PR #14, opened 2016-05-18
Tweaks to produce a functional NPM project page.

jeffposnick:

I had to change a few things with our template and build process to make the NPM project page look like it should.

Here's what the page looks like with these changes in place: https://www.npmjs.com/package/sw-appcache-behavior


addyosmani:
LGTM. Sorry for the late review!

PR #25, opened 2016-07-11
gulpfile updates

### LINKED ISSUE #5: Resolve incompatibility with latest selenium-webdriver
`selenium-webdriver` v2.52.0 was unable to automatically detect the local ChromeDriver instance I had on the `PATH` in my development environment. Rather than debug that incompatibility, I locked down the version to v2.52.0 via a [`npm-shrinkwrap.json`](https://github.com/GoogleChrome/sw-helpers/blob/991a5f3999065ddd540610725736d654c9fa5280/npm-shrinkwrap.json).

Given more time, I should get to the root of the issue and file a bug against `selenium-webdriver` if it turns out to be an issue on their end.

jeffposnick:

Fixes #5

This updates the `gulp test` command to ensure that failures always cause the process to exit with a non-zero error code.

It also resolves an issue that was due to the `gh-pages` module pulling in a dependency that broke the newer `selenium-webdriver` modules.

### DIFF
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -30,7 +30,6 @@ import {globPromise, processPromiseWrapper, taskHarness} from './build-utils.js'
 
 const fsePromise = promisify('fs-extra');
 const tmpPromise = promisify('tmp');
-const ghPagesPromise = promisify('gh-pages');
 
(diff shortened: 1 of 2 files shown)


gauntface:
LGTM

Issue #19, opened 2016-06-25
offline-google-analytics does not work with sendBeacon


igrigorik:
Couple of issues..

1. sendBeacon uses POST to transmit data
1. The URL is not unique ([current code assumes that it is](https://github.com/GoogleChrome/sw-helpers/blob/sw-offline-ga/projects/sw-offline-google-analytics/src/offline-google-analytics-import.js#L42)) and import needs to store and retransmit the POST payload.

To reproduce, enable sendBeacon transport in your GA config:

```
ga("set", "transport", "beacon");
ga("send", "pageview");
```

jeffposnick:
Cool. Thanks for the details; I'll take a look!

jeffposnick:
@philipwalton also let me know about another POST endpoint, [`/batch`](https://developers.google.com/analytics/devguides/collection/protocol/v1/devguide#batch), with similar payload considerations.

jeffposnick:
@igrigorik, I've added in what I hope should support beacon-based transport in https://github.com/GoogleChrome/sw-helpers/commit/5320c8269f3368939bf792c4bc1a47487ca7963f. Feel free to pull in that latest code and try again.

I'm still working on finding an effective way to automate testing at least some of the code, and am waiting on that before releasing this officially.

igrigorik:
@jeffposnick thanks for working on this! Left a comment on the commit.. will take it out for a spin.

jeffposnick:
This was fixed in the version that's been merged.

PR #27, opened 2016-07-18
sw-offline-google-analytics initial release

### LINKED ISSUE #8: Add sw-offline-google-analytics
@addyosmani has some work at https://github.com/GoogleChrome/sw-helpers/tree/offline-analytics but that doesn't use the current monorepo setup. I also have a slightly tweaked version of the original code, and we need things like JSDoc comments.

I'll put together a PR with all that.

jeffposnick:
CC: @philipwalton @igrigorik 

Fixes #8 #19 

This adds the code for a new `sw-offline-google-analytics` project, along with a quick demo and tests. The underlying code is conceptually similar to what's been deployed elsewhere as part of larger projects (like IOWA), and I've gotten some feedback on it from @igrigorik and @philipwalton in the past.

One of the larger changes to that existing codebase was to refactor the imported code into a couple of different modules that handle IndexedDB interactions. I then wrote some tests against those modules. Hopefully this makes up to some extent for not being able to put together a meaningful end-to-end automated test that simulates a lack of network connectivity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,5 @@
 <!-- To make changes, edit templates/README.hbs, not README.md! -->
-[![NPM version][npm-image]][npm-url]
 [![Build Status][travis-image]][travis-url]
-[![devDependency Status](https://david-dm.org/googlechrome/sw-helpers/dev-status.svg)](https://david-dm.org/googlechrome/sw-helpers#info=devDependencies)
 
 # Service Worker Helpers
 
@@ -24,7 +22,7 @@ A service worker implementation of the behavior defined in a page&#x27;s App Cac
                 [API](projects/sw-appcache-behavior#api)
 
 ### sw-offline-google-analytics
-A service worker implementation of the behavior defined in a page&#x27;s App Cache manifest.
+A library to extend a service worker&#x27;s behavior, allowing it to retry failed Google Analytics requests.
 
 **Install**: `npm install --save-dev sw-offline-google-analytics`
 
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ A service worker implementation of the behavior defined in a page&#x27;s App Cac
                 [API](projects/sw-appcache-behavior#api)
 
 ### sw-offline-google-analytics
-A library to extend a service worker&#x27;s behavior, allowing it to retry failed Google Analytics requests.
+A service worker helper library to retry offline Google Analytics requests when a connection is available.
 
 **Install**: `npm install --save-dev sw-offline-google-analytics`
 
--- a/README.md
+++ b/README.md
@@ -23,6 +23,15 @@ A service worker implementation of the behavior defined in a page&#x27;s App Cac
                 [Demo](projects/sw-appcache-behavior#demo) •
                 [API](projects/sw-appcache-behavior#api)
 
+### sw-offline-google-analytics
+A service worker implementation of the behavior defined in a page&#x27;s App Cache manifest.
+
+**Install**: `npm install --save-dev sw-offline-google-analytics`
+
+**Learn More**: [About](projects/sw-offline-google-analytics) •
+                [Demo](projects/sw-offline-google-analytics#demo) •
+                [API](projects/sw-offline-google-analytics#api)
+
 
 ## Development
 
--- a/lib/idb-helper.js
+++ b/lib/idb-helper.js
@@ -99,9 +99,9 @@ IDBHelper.prototype.delete = function(key) {
  */
 IDBHelper.prototype.get = function(key) {
   return this._getDb().then(db => {
-    const tx = db.transaction(this._storeName);
-    const objectStore = tx.objectStore(this._storeName);
-    return objectStore.get(key);
+    return db.transaction(this._storeName)
+      .objectStore(this._storeName)
+      .get(key);
   });
 };
 
@@ -115,9 +115,9 @@ IDBHelper.prototype.get = function(key) {
  */
 IDBHelper.prototype.getAllValues = function() {
   return this._getDb().then(db => {
-    const tx = db.transaction(this._storeName);
-    const objectStore = tx.objectStore(this._storeName);
-    return objectStore.getAll();
+    return db.transaction(this._storeName)
+      .objectStore(this._storeName)
+      .getAll();
   });
 };
 
@@ -132,9 +132,9 @@ IDBHelper.prototype.getAllValues = function() {
  */
 IDBHelper.prototype.getAllKeys = function() {
   return this._getDb().then(db => {
-    const tx = db.transaction(this._storeName);
-    const objectStore = tx.objectStore(this._storeName);
-    return objectStore.getAllKeys();
+    return db.transaction(this._storeName)
+      .objectStore(this._storeName)
+      .getAllKeys();
   });
 };
 
--- a/lib/idb-helper.js
+++ b/lib/idb-helper.js
@@ -1,0 +1,141 @@
+/*
+ Copyright 2016 Google Inc. All Rights Reserved.
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+const idb = require('idb');
+
+/**
+ * A wrapper to store for an IDB connection to a particular ObjectStore.
+ *
+ * @private
+ * @class
+ */
+function IDBHelper(name, version, storeName) {
+  if (name == undefined || version == undefined || storeName == undefined) {
+    throw Error('name, version, storeName must be passed to the constructor.');
+  }
+
+  this._name = name;
+  this._version = version;
+  this._storeName = storeName;
+}
+
+/**
+ * Returns a promise that resolves with an open connection to IndexedDB, either
+ * existing or newly opened.
+ *
+ * @private
+ * @returns {Promise.<DB>}
+ */
+IDBHelper.prototype._getDb = function() {
+  if (this._db) {
+    return Promise.resolve(this._db);
+  }
+
+  return idb.open(this._name, this._version, upgradeDB => {
+    upgradeDB.createObjectStore(this._storeName);
+  }).then(db => {
+    this._db = db;
+    return db;
+  });
+};
+
+/**
+ * Wrapper on top of the idb wrapper, which simplifies saving the key/value
+ * pair to the object store.
+ * Returns a Promise that fulfills when the transaction completes.
+ *
+ * @private
+ * @param {String} key
+ * @param {Object} value
+ * @returns {Promise.<T>}
+ */
+IDBHelper.prototype.put = function(key, value) {
+  return this._getDb().then(db => {
+    const tx = db.transaction(this._storeName, 'readwrite');
+    const objectStore = tx.objectStore(this._storeName);
+    objectStore.put(value, key);
+    return tx.complete;
+  });
+};
+
+/**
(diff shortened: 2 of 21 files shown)


README.md
```diff
@@ -23,6 +23,15 @@ A service worker implementation of the behavior defined in a page&#x27;s App Cac
                 [Demo](projects/sw-appcache-behavior#demo) •
                 [API](projects/sw-appcache-behavior#api)
 
+### sw-offline-google-analytics
+A service worker implementation of the behavior defined in a page&#x27;s App Cache manifest.
```
addyosmani: I assume the text here will be changed :) (currently uses the app-cache manifest text)

lib/idb-helper.js
```diff
@@ -0,0 +1,141 @@
...
+ */
+IDBHelper.prototype.getAllKeys = function() {
+  return this._getDb().then(db => {
+    const tx = db.transaction(this._storeName);
+    const objectStore = tx.objectStore(this._storeName);
+    return objectStore.getAllKeys();
+  });
+};
+
+module.exports = IDBHelper;
```
addyosmani: Minor: I think I've seen a similar IDB helper used in our other projects. Might be useful to generalise this at some point if we find ourselves regularly building the same utils on top of the `idb` module.

README.md
jeffposnick: That looks to be an outdated revision—the current version in the branch should have the correct wording.

projects/sw-offline-google-analytics/README.md
```diff
@@ -15,20 +15,24 @@ Browse sample source code in the [demo directory](https://github.com/GoogleChrom
 
 ### goog.useOfflineGoogleAnalytics
 
-[projects/sw-offline-google-analytics/src/offline-google-analytics-import.js:182-229](https://github.com/GoogleChrome/sw-helpers/blob/213765485d2c3378497e2a80c2b80993a7b09cdf/projects/sw-offline-google-analytics/src/offline-google-analytics-import.js#L182-L229 "Source code on GitHub")
+[projects/sw-offline-google-analytics/src/offline-google-analytics-import.js:45-83](https://github.com/GoogleChrome/sw-helpers/blob/5320c8269f3368939bf792c4bc1a47487ca7963f/projects/sw-offline-google-analytics/src/offline-google-analytics-import.js#L45-L83 "Source code on GitHub")
 
-`goog.useOfflineGoogleAnalytics` is the main entry point to the library
-from within service worker code.
+In order to use the library, call`goog.useOfflineGoogleAnalytics()`.
```
addyosmani: I'm probably being nitpicky :) I found it a little strange that we're initialising offline analytics off the `goog` namespace directly here rather than giving each project it's own dedicated extension to that (e.g `goog.offlineAnalytics.start()`)

projects/sw-offline-google-analytics/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "sw-offline-google-analytics",
   "version": "1.0.0",
-  "description": "A service worker implementation of the behavior defined in a page's App Cache manifest.",
+  "description": "A library to extend a service worker's behavior, allowing it to retry failed Google Analytics requests.",
```
addyosmani: My understanding is the library queues requests while offline and retries them when a network connection is available again. Is that incorrect? If not, it might be worth phrasing this as "A service worker helper library to retry offline Google Analytics requests when a connection is available" (or something better)

lib/idb-helper.js
jeffposnick: This code is a modified version of https://github.com/GoogleChrome/sw-helpers/blob/master/projects/sw-appcache-behavior/src/lib/idb-helpers.js, which I wrote for the `sw-appcache-behavior` project. I moved it up to the project level to make it easier to share within `sw-helpers`. If we think we could use it outside of `sw-helpers`, I'm happy to publish it on NPM.

Assuming this PR is merged, I just filed https://github.com/GoogleChrome/sw-helpers/issues/28 to track moving the existing `sw-appcache-behavior` code to use the project-level library.

projects/sw-offline-google-analytics/package.json
jeffposnick: 👍

projects/sw-offline-google-analytics/README.md
jeffposnick: This matches what's done with `sw-appcache-behavior`, which exposes [`goog.legacyAppCacheBehavior()`](https://www.npmjs.com/package/sw-appcache-behavior#googlegacyappcachebehavior).

I'm open to changing both, and renaming the method exposed in `sw-appcache-behavior` to `goog.appCacheBehavior.legacy()` (or something like that)namespace if you think that's preferable.

Being consistent across projects is my main concern.

addyosmani: Imo it would make sense to switch and be consistent here across projects. Looping in @gauntface and @wibblymat for their take on this just in case I'm a lone voice on this one :)

philipwalton:
Though maybe not required for an initial release, I nice feature to add would be the ability to modify the hit prior to replaying it in the offline case.

Though I'm sure generic modification would be useful in many cases, a specific use-case I can think of is to set a custom dimension value, so at reporting time you could differentiate offline hits from online hits.

In IOWA we listened for the `online` and `offline` events and updated the tracker in the browser, but I don't think that's as fool-proof as doing it at the service worker level.

jeffposnick:
@philipwalton, that sounds like an interesting idea, but my ignorance of the GA API makes me unsure of what to implement. Your feedback is appreciated!

My guess is that it would be one of three possible implementations:

1. Hardcode a `param=value` parameter that gets appended to all replayed requests. This assumes that there's a meaningful combo that could be used in all cases. (Is there?)

2. Hardcode the `param=` portion, and let developers pass in their own `value` when they initialize the library. Like before, this assumes there's a meaningful parameter that could be used in all places, but that the actual value would be specific to each deployment.

3. Just take in a configuration `Object` as a parameter when initializing the library, and serialize whatever gets passed in as `param1=value1&param2=value2`, etc. when replaying. This is obviously the most flexible option, but if you think there's only really one specific `param` that will be useful here, that flexibility comes at the cost of making the best practice less obvious.

What are your thoughts?

philipwalton:
I like option 3.

There is no generalized parameter that's currently reserved to indicate offline/online status, so right now the only option users would have is a custom dimension. And with custom dimensions, the index will be different from property to property.

In my autotrack library, I've standardized on two common options that every plugin that sends hits to GA accepts: [`fieldsObj`](https://github.com/googleanalytics/autotrack/blob/master/docs/common-options.md#fieldsobj) and [`hitFilter`](https://github.com/googleanalytics/autotrack/blob/master/docs/common-options.md#hitfilter).

The former is very similar to what you've described in option 3 (though in the autotrack case it's a defaults object and in the this case it'd probably be an overrides object). The `hitFilter` option is a function that is invoked with the current hit, and whatever it returns becomes the new hit.

I've found that these two options basically handle every possible customization anyone might want to do.

FYI, I'm also happy to send a separate PR for this feature at another time if we just want to get an initial version out quickly.

jeffposnick:
Cool, @philipwalton. I'll get that into this PR. I'll also push a new commit with the additional namespacing described at https://github.com/GoogleChrome/sw-helpers/pull/27#discussion_r71228356

Stay tuned...

philipwalton:
Sounds good. In terms of naming, I think `hitFilter` is good, but I wouldn't use `fieldsObj` since in Measurement Protocol land, these are called [parameters](https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters), not [fields](https://developers.google.com/analytics/devguides/collection/analyticsjs/field-reference). Maybe `parameterOverrides`?

Also, if you implement the `hitFilter` function, for convenience I'd suggest the function being invoked with an object of parameter/value pairs rather than the raw hit payload string, so users don't have to parse that themselves. I assume you're already parsing it, so we may as well expose it parsed.

jeffposnick:
Before I write the code for it, what I was thinking for the interface was

```js
// Within service-worker.js
goog.offlineGoogleAnalytics.initialize({
  dimension1: 'otherValue1',
  dimension2: 'otherValue2'
  // etc.
});
```

Which would cause a hypothetical, queued request for

```
https://www.google-analytics.com/collect?cid=12345.1234&dimension1=value1
```

to be sent as

```
https://www.google-analytics.com/collect?cid=12345.1234&qt=12345&dimension1=otherValue1&dimension2=otherValue2
```

when it was replayed. In other words, the logic would be:

- Take the original request.
- Add in `qt=` with the time offset as a new parameter.
- If a configuration `Object` was passed in to `goog.offlineGoogleAnalytics.initialize()`, then for each `key`/`value`:
  - Replace any instances of `key` in the original URL's parameters with the new `value`.
  - If `key` isn't in the original URL, add in a new `key=value`.

This seems lighter weight than a `hitFilter()` function, but I guess the downside is that you need to provide the values when you start up your service worker instead of being able to calculate them on the fly. That seems like a reasonable tradeoff to me, but maybe @philipwalton objects.

philipwalton:
What if `initialize` accepted a configuration object, and what you've described above were set on the `parameterOverrides` property?

```js
goog.offlineGoogleAnalytics.initialize({
  parameterOverrides: { ... }
});
```

That way it would be easy to add new functionality in the future, e.g. a `hitFilter` function:

```js
goog.offlineGoogleAnalytics.initialize({
  parameterOverrides: { ... },
  hitFilter: (parameters) => { ... }
});
```

jeffposnick:
Thanks again for the feedback, @philipwalton and @addyosmani! Just to summarize the current interface, you now do the following to use the library (this is also reflected in the docs):

```js
// First, import the library into the service worker global scope:
importScripts('path/to/offline-google-analytics-import.js');

// Then, call goog.offlineGoogleAnalytics.initialize():
goog.offlineGoogleAnalytics.initialize({
  parameterOverrides: {
    // Optionally, pass in an Object with additional parameters that will be
    // included in each replayed request.
    dimension1: 'Some Value',
    dimension2: 'Some Other Value'
  }
});
```

Hopefully things look good now, and I'm looking forward to getting 1.0.0 released following a LGTM!

jeffposnick:
(Checking out the broken Travis CI build now...)

projects/sw-offline-google-analytics/README.md
```diff
@@ -32,8 +40,15 @@ and retried the next time the service worker starts up.
...
 
-// Then, call goog.useOfflineGoogleAnalytics() to activate the library.:
-goog.useOfflineGoogleAnalytics();
+// Then, call goog.offlineGoogleAnalytics.initialize():
+goog.offlineGoogleAnalytics.initialize({
+  parameterOverrides: {
+    // Optionally, pass in an Object with additional parameters that will be
+    // included in each replayed request.
+    dimension1: 'Some Value',
+    dimension2: 'Some Other Value'
```
philipwalton: `dimensionXX` is the [`analytics.js` field](https://developers.google.com/analytics/devguides/collection/analyticsjs/field-reference) representation of custom dimension parameters, but since this is the raw hit payload, it should use the [Measurement Protocol parameters](https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters#customs), which are `cd1` and `cd2`.

It should probably also link to the parameter reference:
https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters

projects/sw-offline-google-analytics/demo/service-worker.js
```diff
@@ -8,7 +8,13 @@ importScripts('../build/offline-google-analytics-import.js');
 // First, enable the offline Google Analytics behavior.
 // This will get "first shot" at responding to Google Analytics requests, before
 // our catch-all fetch event listener can handle it.
-goog.useOfflineGoogleAnalytics();
+goog.offlineGoogleAnalytics.initialize({
+  parameterOverrides: {
+    // Add in any additional parameters here, or omit this section to
+    // replay the Google Analytics request without additional parameters.
+    dimension1: 'My Custom Dimension Value'
```
philipwalton: Same comment as above.

projects/sw-offline-google-analytics/README.md
jeffposnick: Changed.

philipwalton:
LGTM

addyosmani:
LGTM2. Let's ship this! 🚢

PR #29, opened 2016-07-22
Use the new shared libs, and rename to goog.appCacheBehavior.fetch

### LINKED ISSUE #28: Use idb-helper.js in sw-appcache-behavior
**Library Affected**:
sw-appcache-behavior

**Browser & Platform**:
all browsers

Bother the page and service worker code for `sw-appcache-behavior` make use of IndexedDB, and it would be good to move that code to use the project-level `idb-helper.js` library that's being introduced as part of the `sw-offline-google-analytics` code.

E.g.: https://github.com/GoogleChrome/sw-helpers/blob/master/projects/sw-appcache-behavior/src/client-runtime.js#L44

jeffposnick:

Fixes #28 

This PR accomplishes two things in `sw-appcache-behavior`:
- It uses the new shared top-level libraries, `idb-helper.js`, `log.js`, and `scope.js`.
- Using `scope.js`, it changes the exposed interface from `goog.legacyAppCacheBehavior` to `goog.appCacheBehavior.fetch`, to match what we're doing with namespaces in `sw-offline-google-analytics`.

This is a breaking change, but given that it's a move from v0.0.4 to v0.0.5, I believe it's in keeping with semver best practices around releases less than v0.1.0.


jeffposnick:
The diff looks larger than it is due to whitespace; https://github.com/GoogleChrome/sw-helpers/pull/29/files?w=1 will give you a more accurate view of what changed.

projects/sw-appcache-behavior/src/appcache-behavior-import.js
```diff
@@ -15,501 +15,511 @@
...
+    return response;
+  });
+}
 
-      return response;
-    });
-  };
-})(self);
+// Add the function to the global service worker scope,
+// as goog.appCacheBehavior.fetch.
```
jeffposnick: I'm open to a different name instead of `goog.appCacheBehavior.fetch`, if anyone has any ideas that they prefer.

jeffposnick:
Gentle **ping**.

wibblymat:
👍

PR #32, opened 2016-07-26
Removes a debugging statement

jeffposnick:

Just some small cruft that was left behind while I was experimenting.


ebidel:
LGTM

PR #34, opened 2016-08-05
Fix the qt param to use offset time

philipwalton:
R: @jeffposnick @addyosmani @wibblymat @gauntface

The [`qt`](https://devsite.googleplex.com/analytics/devguides/collection/protocol/v1/parameters#qt) param for the Measurement Protocol accepts a time offset, not a timestamp. Unfortunately this means hits replayed by the current version of this library are most likely being discarded (sorry for not catching this in my initial review).

### DIFF
--- a/projects/sw-offline-google-analytics/src/lib/replay-queued-requests.js
+++ b/projects/sw-offline-google-analytics/src/lib/replay-queued-requests.js
@@ -38,14 +38,15 @@ module.exports = additionalParameters => {
 
   return idbHelper.getAllKeys().then(urls => {
     return Promise.all(urls.map(url => {
-      return idbHelper.get(url).then(queuedTime => {
+      return idbHelper.get(url).then(hitTime => {
+        const queueTime = Date.now() - hitTime;
         const newUrl = new URL(url);
 
         // URLSearchParams was added in Chrome 49.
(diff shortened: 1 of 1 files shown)


ithinkihaveacat:
Is it possible to record whether a hit occurred online or offline? That would be an interesting stat, and would also help verify that the code is working as expecting.

jeffposnick:
Oops! Thanks for catching that, @philipwalton!

jeffposnick:
(The Travis CI failures are due to the Firefox 48 rollout. I need to move the testing infrastructure to a newer version of `sw-testing-helpers` that supports https://developer.mozilla.org/en-US/docs/Mozilla/QA/Marionette/WebDriver)

jeffposnick:
Actually, the `replay-queued-requests` test failed due to a now-incorrect assumption about the `qt=` parameter would be. (The fact that the Firefox tests are *also* failing made that hard to recognize.)

@philipwalton, I'll update that test logic in a follow-up PR.

jeffposnick:
@ithinkihaveacat, you can do what you suggested via adding in a custom dimension via [`config.parameterOverrides`](https://github.com/GoogleChrome/sw-helpers/tree/master/projects/sw-offline-google-analytics#googofflinegoogleanalyticsinitialize).

PR #39, opened 2016-08-11
Add hitFilter option and better support module loaders

philipwalton:
R: @jeffposnick @addyosmani @wibblymat @gauntface @ebidel

### This pull request adds two primary features:

- **A new `hitFilter` configuration option**
  <br>This options allows developers to programmatically modify the hit params at runtime. This is useful to add/modify params when the `parameterOverrides` option alone isn't sufficient (e.g. when a value needs to be derived from another value).
- **A new index files that exports the `offlineGoogleAnalytics` object without adding it to the `goog` scope**
  <br>In cases where you're using a module bundler to build your service worker, importing the built file can cause unnecessary duplication (especially if you're using `idb` elsewhere). Now users can npm install `sw-helpers` and in their code do the following:

```js
// Imports the module locally.
const offlineGoogleAnalytics = require('sw-helpers/projects/sw-offline-google-analytics/src');

// Initializes the module via the local variable name.
offlineGoogleAnalytics.initialize({...});
```

*__Note:__ `idb` needed to be added to `dependencies` instead of `devDependencies` to make this work.*

### It also updates the logic in two important ways:

- **Replays are now more aggressively retried**
  <br>Since hits between 4-28 hours old may not be precessed, it's important to retry whenever possible. The old logic retried any time the service worker started up. The new logic adds an additional retry any time a successful hit was sent if the previous hit failed.
- **Hits older than 2 days are discarded**
  <br>The previous logic only discarded old hits if they failed after being sent from the service worker. The new logic inspects the age of the hit prior to all replay attempts and only replays hits less than 48 hours old.


gauntface:
minor nit. Is there any way we can publish builds that wouldn't require the src?

require('sw-helpers/projects/sw-offline-google-analytics/src') -> require('sw-helpers/projects/sw-offline-google-analytics')

projects/sw-offline-google-analytics/src/lib/constants.js
```diff
@@ -21,7 +21,7 @@ module.exports = {
     VERSION: 1
   },
   MAX_ANALYTICS_BATCH_SIZE: 20,
-  STOP_RETRYING_AFTER: 86400000, // One day, in milliseconds.
+  STOP_RETRYING_AFTER: 1000 * 60 * 60 * 48, // Two days, in milliseconds.
```
gauntface: Why this increase out of curiosity?

philipwalton:
I think we could, and I agree that would nicer.

In most projects you can just set `main` in `package.json` to whatever entry point you want, but with the mono-repo approach, that isn't possible. We'd have to move the `index.js` file to where we want it.

@jeffposnick and I also talk about making it possible to have this work with just `require('sw-offline-google-analytics')` by adding source files to what gets published to npm, but, again, that won't work here due to the fact that each project depends on stuff in `/lib`.

Depending on how common it ends up being for folks to use module bundlers to build their service worker scripts, we might want to reconsider the mono-repo approach in the future.

projects/sw-offline-google-analytics/src/lib/constants.js
philipwalton: It's a bit complicated to explain, but I'll give it a try.

Google Analytics processes all hits it receives at around 4 a.m. (the next morning) in the timezone of the view. The hit offset time specified via the `qt` param only works if processing for the target day has not occurred yet. Once hit processing is completed for a particular day, no further processing happens (except in the case of fixing errors due to bugs or other unusual circumstances).

This means, in general, the max `qt` time could be as short as 4 hours or as long as 28 hours (assuming normal processing schedules). I extended this to 48 hours just to be safe.

Alternatively we could just always send hits no matter what, and let Google Analytics process what it can.

projects/sw-offline-google-analytics/src/index.js
```diff
@@ -0,0 +1,119 @@
...
+ *     searchParams.set('qt', qt);
+ *   }
+ * });
+ *
+ * // At this point, implement any other service worker caching strategies
+ * // appropriate for your web app.
+ *
+ * @alias goog.offlineGoogleAnalytics.initialize
+ * @param {Object=}   config Optional configuration arguments.
+ * @param {Object=}   config.parameterOverrides Optional
```
jeffposnick: Does it make sense to keep `parameterOverrides` now that there's `hitFilter`?

I'm okay with dropping it, though that breaking change would bump things up to a 2.0.0 release.

If we're going to keep both, I think the docs/sample needs to explain why developers might choose over the other.

projects/sw-offline-google-analytics/src/index.js
```diff
@@ -0,0 +1,119 @@
...
+ * // Then, call goog.offlineGoogleAnalytics.initialize():
+ * goog.offlineGoogleAnalytics.initialize({
+ *   parameterOverrides: {
+ *     // Optionally, pass in an Object with additional parameters that will be
+ *     // included in each replayed request.
+ *     // See https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters
+ *     cd1: 'Some Value',
+ *     cd2: 'Some Other Value'
+ *   },
+ *   hitFilter: (searchParams) =>
```
jeffposnick: We've been leaving out `()` from single-parameter arrow functions (when there aren't any default parameters).

projects/sw-offline-google-analytics/src/index.js
```diff
@@ -0,0 +1,119 @@
...
+ *     // Sets the `qt` param as a custom metric.
+ *     const qt = searchParams.get('qt');
+ *     searchParams.set('qt', qt);
```
jeffposnick: This is effectively a no-op, right? Could you come up with a more compelling example of how developers might use `hitFilter`?

projects/sw-offline-google-analytics/src/lib/replay-queued-requests.js
```diff
@@ -27,46 +26,48 @@ const idbHelper = new IDBHelper(constants.IDB.NAME, constants.IDB.VERSION,
...
-          Object.keys(additionalParameters).sort().forEach(parameter => {
-            newUrl.searchParams.set(parameter, additionalParameters[parameter]);
-          });
-        } else {
-          log('The browser does not support URLSearchParams, ' +
-            'so not setting additional parameters.');
-          newUrl.search += (newUrl.search ? '&' : '') + 'qt=' + queueTime;
-        }
+        // Do not attempt to replay hits that are too old.
+        if (queueTime > constants.STOP_RETRYING_AFTER) return;
```
jeffposnick: My preference would be `return;` on a separate line, surrounded by braces.

(I'm surprised `eslint` doesn't enforce that...)

projects/sw-offline-google-analytics/src/lib/replay-queued-requests.js
```diff
@@ -27,46 +26,48 @@ const idbHelper = new IDBHelper(constants.IDB.NAME, constants.IDB.VERSION,
...
 
-        return fetch(newUrl.toString()).catch(error => {
-          // If this was queued recently, then rethrow the error, to prevent
-          // the entry from being deleted. It will be retried again later.
-          if ((Date.now() - queueTime) < constants.STOP_RETRYING_AFTER) {
-            throw error;
-          }
+        // Do not attempt to replay hits in browsers without URLSearchParams support.
+        if (!('searchParams' in newUrl)) return;
```
jeffposnick: Ditto about `return;` on a newline + braces.

projects/sw-offline-google-analytics/src/index.js
philipwalton: Sorry, this is supposed to be `searchParams.set('cm1', qt);` (as it is in the test file).

This is a real example. It's being used here to track (via a custom metric) what the `qt` param was. This could be used in conjunction with an offline/online change handler to get a sense for how long (on average) users were without connectivity.

Another good example would be if a site had more than one tracker, and the dimension index they wanted to use to track service worker replays were different from each other. In that case they'd have to run logic like this:

```js
switch (searchParams.get('tid')) {
  case 'UA-XXXXX-Y':
    return 1; // or whatever this happens to be for property UA-XXXXX-Y
  case 'UA-XXXXX-Z'
    return 2; // or whatever this happens to be for property UA-XXXXX-Z
  ...
}
```

projects/sw-offline-google-analytics/src/index.js
philipwalton: I'll update for consistency.

projects/sw-offline-google-analytics/src/lib/replay-queued-requests.js
philipwalton: It doesn't enforce it because my way is better :)

But I'll update for consistency. I can also remove this particular condition if we want based on https://github.com/GoogleChrome/sw-helpers/pull/39#discussion_r74503628.

jeffposnick:
I just gave this PR a try locally and am running into issues when building the library:

```sh
$ gulp build --project=sw-offline-google-analytics
[15:23:58] Requiring external module babel-register
[15:23:58] Using gulpfile ~/git/sw-helpers/gulpfile.babel.js
[15:23:59] Starting 'build'...
events.js:154
      throw er; // Unhandled 'error' event
      ^

ReferenceError: Invalid left-hand side in assignment
    at module.exports (/Users/jeffy/git/sw-helpers/node_modules/syntax-error/index.js:13:52)
    at DestroyableTransform._transform (/Users/jeffy/git/sw-helpers/node_modules/browserify/index.js:640:23)
    at DestroyableTransform.Transform._read (/Users/jeffy/git/sw-helpers/node_modules/through2/node_modules/readable-stream/lib/_stream_transform.js:159:10)
    at DestroyableTransform.Transform._write (/Users/jeffy/git/sw-helpers/node_modules/through2/node_modules/readable-stream/lib/_stream_transform.js:147:83)
    at Labeled.Pipeline._write (/Users/jeffy/git/sw-helpers/node_modules/stream-splicer/index.js:70:22)
    at doWrite (/Users/jeffy/git/sw-helpers/node_modules/readable-stream/lib/_stream_writable.js:319:64)
    at writeOrBuffer (/Users/jeffy/git/sw-helpers/node_modules/readable-stream/lib/_stream_writable.js:308:5)
    at Labeled.Writable.write (/Users/jeffy/git/sw-helpers/node_modules/readable-stream/lib/_stream_writable.js:246:11)
    at Labeled.ondata (/Users/jeffy/git/sw-helpers/node_modules/readable-stream/lib/_stream_readable.js:543:20)
    at emitOne (events.js:90:13)
```

It's coming from Browserify, but the underlying code that's making it unhappy is unfortunately not discernible from the stack trace (as far as I can tell). @philipwalton, is this specific to my environment? Are you able to successfully run the command to build the library?

philipwalton:
@jeffposnick yeah, I can run the command. Maybe try a fresh `npm install` and see if it still happens?

jeffposnick:
Okay, it's

```js
module.exports = ({parameterOverrides, hitFilter} = {}) => {
```

in your `replay-queued-requests.js` that's making Browserify unhappy.

Which version of Node are you using? I'm on v5.12.0.

Given that the Gulpfile is run via Babel, I would have thought that somehow would make Browserify happy with all sort of ES2015 weirdness, but maybe not?

philipwalton:
> Which version of Node are you using? I'm on v5.12.0.

I'm on 6.0.0, so I guess that explains it.

> Given that the Gulpfile is run via Babel, I would have thought that somehow would make Browserify happy with all sort of ES2015 weirdness, but maybe not?

Maybe it's because `build-utils` uses `child_process`? Either way, I don't mind removing the destructured params if that makes this more widely available.

Do you have any thoughts on the items Matt and I were discussing?

jeffposnick:
`spawn()` is only called for running `npm publish` and `documentation`. It shouldn't be kicking off the build process. It's probably because the build process uses `browserify` without the `babelify` transform? Which I'd rather avoid adding in, so yeah, if we could expand the compatibility with build environments without sacrificing too much, removing the destructured parameters would be appreciated.

I'm on board with extended the retry window to 2 days.

There were a couple of example-related comments from earlier in the review that still need to be addressed. As for @gauntface's https://github.com/GoogleChrome/sw-helpers/pull/39#issuecomment-239285929, I'd rather leave the approach cleaner for folks who will be consuming via `importScripts()`, which is most likely going to be the most prevalent until we have native ES2015 modules.

projects/sw-offline-google-analytics/src/index.js
philipwalton: `hitFilter` can do everything `parameterOverrides` can do, but it requires a bit more work (as well as a knowledge of the `URLSearchParams` API.

My preference would be to keep it, but I don't have a strong opinion. If you'd prefer to drop it, we'd still have all the features I care about.

philipwalton:
@jeffposnick I believe I've addressed all your review feedback. PTAL.

I've updated the examples in `index.js` with additional  `@examples` tags. I'm not sure if the build supports that or not, so let me know if I need to change it.

jeffposnick:
It looks like the new examples are being picked up, yep. (You can check locally with `gulp documentation:projects --project=sw-offline-google-analytics`, and see how the project's local `README.md` is updated.)

I'm going to merge, and will handle updating the version to 1.1.0, publishing to NPM, and putting together some release notes.

Thanks for the contribution!

PR #46, opened 2016-10-14
Docs

gauntface:

This adds doc building and lint to the sw-routing module (largely to see what the process would be like).


projects/sw-routing/package.json
```diff
@@ -13,6 +13,11 @@
     "email": "jeffy@google.com",
     "url": "https://jeffy.info"
   },
+  "scripts": {
+    "lint": "eslint '.'",
```
jeffposnick: This is already handled in https://github.com/GoogleChrome/sw-helpers/blob/future-of-sw-tooling/gulpfile.js#L46

gauntface:
Please treat this PR as now "Getting Gulp Lint to Pass".

jeffposnick:
👍

PR #49, opened 2016-10-18
Gulp clean

### LINKED ISSUE #50: Make the failing tests pass
It's harder to understand the impact of new changes (and existing issues) with sub-projects when the tests have been failing for a while. Let's see if we can get them passing again before adding in new work. Right now it's hard to understand when the tests last actually went green (maybe 3 months ago? :))

Matt was up for trying to get the tests passing. Locally we're all green now on the appcache behavior. Just need to go green on offline analytics before that work lands.

gauntface:

The tests for appcache behaviour project has some fairly gnarly dependencies on previous tests that I haven't touched in this PR but should be addressed at some point.

Fixes #50 #51


gauntface:
This should be working on the next travis build (Previous issue was passing tests, but selenium-assistant and gulp did not play well).

gauntface:
@jeffposnick @addyosmani 

Can Haz Merge

jeffposnick:
👍

PR #53, opened 2016-10-19
Minify

gauntface:

Minifying :(

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,8 @@
     "babel": "^6.5.2",
     "babel-plugin-external-helpers": "^6.8.0",
     "babel-plugin-transform-async-to-generator": "^6.8.0",
+    "babel-preset-babili": "0.0.5",
+    "babel-preset-es2015": "^6.16.0",
     "blueimp-md5": "^2.3.1",
     "browserify": "^13.1.0",
     "chai": "^3.5.0",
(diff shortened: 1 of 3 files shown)


jeffposnick:
Sorry for the late feedback!

I'd rather avoid going through the transpilation of ES2015 to ES5 just so that we could minify the code using `uglify`.

Since this is code that's run in a service worker, it should be safe to assume that most modern language features are supported without transpilation. (The only thing that I'm currently transpiling is `async`/`await`, which is being converted that to generators.)

https://github.com/babel/babili should (theoretically!) be able to minify the native ES2015 code, without transpilation. I've had some luck [getting it working](https://github.com/jeffposnick/jeffposnick.github.io/blob/work/gulpfile.js#L98) (using CLI; not with Rollup) and I think we should consider it.

gauntface:
I tried babili but the combo would look like this:

rollup -> (Rollup Plugin) rollup-plugin-babel -> babel-preset-babili

Which when I tried it did nothing.

The alternatives are:

1. Drop rollup and just use babel.
2. Try gulp-rollup and see if the streams can be chained together.

How is the async / await stuff being transpiled btw?

How do you expect developers to consume the .mjs files as well? I'm wondering if just offering the source would be enough.

Anyway please revert and propose a better solution - this was a horrible thing to get working thanks to hellish situation the ecosystem is in with rollup + babel + buildsystem.

PR #55, opened 2016-10-23
Changing minification for babili

gauntface:
R: @jeffposnick @addyosmani

Thoughts on this approach over current one?

Not finished but worth a look at this point.


projects/sw-routing/build.js
```diff
@@ -65,6 +70,37 @@ const bundledJSNext = () => {
...
+    .pipe(sourcemaps.write())
+    .pipe(rename(pkg.main))
+    // This is just the UMD model for now.
+    .pipe(gulp.dest(destPath))
+    .on('error', err => {
+      console.error(err);
+      reject(err);
+    })
+    .on('end', () => {
+      console.log('Here ENDED:S');
```
addyosmani: Drop the console.log? :)

projects/sw-routing/build.js
```diff
@@ -65,6 +70,37 @@ const bundledJSNext = () => {
   });
 };
 
-module.exports = () => {
+/** module.exports = () => {
   return Promise.all([minifiedUMD(), bundledJSNext()]);
+};**/
+
+module.exports = () => {
+  console.log('Here :S');
```
addyosmani: Drop the console logs (mentioned later)

projects/sw-routing/build.js
```diff
@@ -65,6 +70,37 @@ const bundledJSNext = () => {
...
+    .pipe(sourcemaps.init())
+    // transform the files here.
+    .pipe(rollup({
+      entry: path.join(__dirname, 'src', 'index.js')
+    }))
+    .pipe(babel({
+      plugins: ['remove-comments'],
+      presets: ['babili']
+    }))
+    .pipe(sourcemaps.write())
```
gauntface: So by default babili doesn't remove the comment so had to add remove-comments to strip out the jsdoc and multiple licence headers.

Will look at the gulp licence task to add it back in.

projects/sw-routing/build.js
gauntface: Yeah - this is not final by a long shot, just wanted to check on opinion of this approach.

projects/sw-routing/build.js
```diff
@@ -13,15 +13,20 @@
...
 
+const gulp = require('gulp');
+const sourcemaps = require('gulp-sourcemaps');
 const path = require('path');
-const rollup = require('rollup').rollup;
-const babel = require('rollup-plugin-babel');
+const rename = require('gulp-rename');
+const rollup = require('gulp-rollup');
+const babel = require('gulp-babel');
+// const babel = require('rollup-plugin-babel');
```
gauntface: It didn't work with the babili preset :(

gauntface:
@addyosmani @jeffposnick should be all good in the hood for a review.

If it's looking OK, would it be worth trying to use this through sw-helpers? One thing I'm slightly dubious of is the transpiling to get Async and Await (i.e. how big is that polyfill) and also how useful the 'js:next' build is.

jeffposnick:
👍  Thanks for taking this on, @gauntface!

Re: standardizing throughout `sw-helpers`, I'd like to when possible, but one wrinkle is that the `sw-appcache-behavior` project has two distinct build targets: one for the client page and one for the service worker. Coming up with a generic build configuration that was flexible enough to handle that might be tough.

We should use this for all the newer modules, though.

The `js:next` build should be very relevant, since we expect folks to consume these as ES2015 modules that get bundled together into their own `service-worker.js`. I think we're doing the right thing in terms of using `js:next` to make them bundle-able via Browserify/Rollup/Webpack/etc.

Regarding transpilation of `async`/`await` to generators, there's not a ton of overhead. Before minification/gzip, it's 3kb, and it looks like it's maybe an extra 500 bytes when both of those are applied. Using `async`/`await` has been SO NICE that I really don't want to give it up.

PR #56, opened 2016-10-24
Lint update

gauntface:
R: @jeffposnick @addyosmani 

This was the worked required to bump the linting of eslint-config-google.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,22 +1,14 @@
 {
-  "extends": "google",
+  "extends": ["eslint:recommended", "google"],
   "env": {
     "serviceworker": true,
-    "browser": true
+    "browser": true,
+    "node": true,
+    "es6": true
   },
   "rules": {
-    "max-nested-callbacks": 0,
-    "valid-jsdoc": [
-      2,
-      {
-        "requireParamDescription": false,
-        "requireReturnDescription": false,
-        "prefer": {
-          "return": "returns"
-        }
-      }
-    ],
-    "no-warning-comments": 0
+    "valid-jsdoc": 1,
+    "require-jsdoc": 1
   },
   "parserOptions": {
     "ecmaVersion": 2017
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 node_modules
 build
 temp
+npm-debug.log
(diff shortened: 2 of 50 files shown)


jeffposnick:
Is it `eslint:recommended` or the update to `google` that mandates the `()` around arrow function parameters? I've never been a fan of that.

If it's made it into the official `google` rules, then I'll relent. Otherwise, if it's part of `eslint:recommended`, I'd argue that we should ignore that rule.

ithinkihaveacat:
@jeffposnick Looks like it [comes from Google](https://github.com/google/eslint-config-google/commit/40e3b15a8d64f8898e53d4d37bf979a6109e76e9#diff-168726dbe96b3ce427e7fedce31bb0bcR282) but I @philipwalton may be able to confirm.

philipwalton:
Google style doesn't strictly require it for single-argument arrow functions, but it does recommend it in all cases to avoid ambiguity. Since the ESLint rule is only granular enough to either always require it or always omit it, I thought it was best to always require it.

Here's the text from the style guide (unfortunately it's still not public or I'd just link to it):

> Tip: It is a good practice to use parentheses even for single-argument arrows, since the code may still parse reasonably (but incorrectly) if the parentheses are forgotten when an additional argument is added.

ithinkihaveacat:
Do the options [`as-needed`](http://eslint.org/docs/rules/arrow-parens#as-needed) and [`requireForBlockBody`](http://eslint.org/docs/rules/arrow-parens#requireforblockbody) help tune it?

philipwalton:
> Do the options as-needed and requireForBlockBody help tune it?

`as-needed` enforces *not* using parens in certain cases. Since the Google style guide recommends using them in all cases, I wouldn't want to add a rules that would prevent someone from following that advice.

philipwalton:
I suppose technically you could argue the rule could be turned off, which would allow folks to use whatever style they want, but it would come at the cost on consistency within a project.

@jeffposnick and @ithinkihaveacat if either of you feel strongly about this, it probably makes sense to move further discussion to the main repo issues: https://github.com/google/eslint-config-google

jeffposnick:
I've made peace with this PR given that the restriction is coming from the latest `google` ruleset. 👍

PR #57, opened 2016-10-25
Cleaner build

### LINKED ISSUE #52: Decide on build artifacts per npm package
**Library Affected**:
All

**Browser & Platform**:
All

Users have asked for us to consider including minified builds in our npm packages (#31). It might be nice to make a call on this so we can consistently apply it across the subprojects.

Fwiw, popular libraries like Lodash include both unminified and minified versions in their npm package.

gauntface:
R: @jeffposnick @addyosmani

Fixes #52

This is the last step I wanted to take for sharing minified code for new modules. I've left the old modules alone for now.

As part of this PR I've juggled the gulp-tasks around so that the individual concerns are seperate (lint, test, build, documentation, publish, serve). Main hope with this is that gulpfile.js is kept minimal and it's clear how tasks relate to each other (for example linting of gulpfile.js and tasks is tucked away in lint.js).


jeffposnick:
👍

PR #62, opened 2016-11-02
Removes predicates, adds ExpressRoute

jeffposnick:
R: @addyosmani @gauntface

Adds the `ExpressRoute` class, matching the functionality you get with `sw-toolbox`'s default routing.

Removes the `predicates` that previously were exposed, since I think we agreed that they ended up being more confusing than helpful. We can cover something similar to `predicates` with examples in our documentation later on.


gauntface:
LGTM from me assuming travis passes.

jeffposnick:
Travis has since been appeased.

PR #65, opened 2016-11-07
Lerna

gauntface:
R: @jeffposnick @addyosmani

It is indeed the release process with lerna.

Was a little bit shakey but still better than manual (hence why we are at 0.0.3)


jeffposnick:
🎊 🎊 🐉

PR #67, opened 2016-11-15
sw-background-queue review

jeffposnick:
Hello @prateekbh!

The easiest way to provide feedback on code is via a GitHub PR review, so I'm creating this one on your behalf.

I understand that you may not be done with all the code quite yet, so feel free to push new updates to your forked repo during the review.

CC: @addyosmani @gauntface


prateekbh:
@jeffposnick much thanks for your efforts
@addyosmani @gauntface
Please feel free to review the core pieces.
I am in process of writing tests and fixing lint.
Integration with sw-goog will follow

packages/sw-background-sync-queue/src/lib/queue.js
```diff
@@ -0,0 +1,143 @@
...
+	 * @memberOf Queue
+	 */
+	getTotalTasks() {
+		return _queue.length;
+	}
+}
+
+const queue = new Queue();
+
+export default queue;
```
jeffposnick: To reinforce my other comments—it would be more in keeping with the other projects if you exported class definitions and then let developers create instances of the classes inside their own code, instead of exporting a single, shared instance of the class that's used globally.

packages/sw-background-sync-queue/src/lib/requestManager.js
```diff
@@ -0,0 +1,136 @@
+import {initializationDefaults} from './constants';
+import queue from './queue.js';
+import responseManager from './responseManager';
+let globalConfig = initializationDefaults;
```
jeffposnick: See other comments about preferring configuration of class instances via variables passed to the constructor and stored in member variables, rather than static configuration used outside of the class structure.

packages/sw-background-sync-queue/src/lib/requestManager.js
```diff
@@ -0,0 +1,136 @@
...
+	 * stores the callbacks object, maintins global config and
+	 * attaches event handler
+	 * @param {Object} config config that gives values like
+	 * maxAge of requests in iDB
+	 * @param {Object} callbacks object that contains function
+	 * ref for various lifecycle callbacks
+	 *
+	 * @memberOf RequestManager
+	 */
+	async initialize(config, callbacks) {
```
jeffposnick: I'd expect there to be a constructor and to make use of the class to store state information.

packages/sw-background-sync-queue/src/lib/requestManager.js
```diff
@@ -0,0 +1,136 @@
...
+	 *
+	 * @param {Request} request request object to transform
+	 * into iDB storable object
+	 * @param {Object} config config object to be
+	 * stored along inthe iDB
+	 * @return {Object} indexable object for iDB
+	 *
+	 * @memberOf RequestManager
+	 */
+	async getQueueableRequest(request, config) {
```
jeffposnick: This is similar to many of the other methods, so I'm not going to repeat the same thing for each one. But it's basically a static method that doesn't really have a connection to the overall `RequestManager` class. It doesn't make use of any class-specific state information.

For functions that will always be static helpers, I'd recommend just putting them inside their own modules not even attempting to wrap them in a class, e.g. https://github.com/GoogleChrome/sw-helpers/blob/030d824d583dbe4fb4bfd59a606caad667740287/packages/sw-broadcast-cache-update/src/lib/broadcast-update.js#L55

packages/sw-background-sync-queue/src/lib/responseManager.js
```diff
@@ -0,0 +1,17 @@
...
+			body: await response.blob(),
+		};
+		idbQHelper.put(hash, idbObject);
+	}
+	getResponse(reqId) {
+	}
+}
+
+const resManager = new ResponseManager();
+export default resManager;
```
jeffposnick: I don't see any reason to wrap this in a `ResponseManager` class. If you have a helper function that doesn't need to keep track of state information, just put it inside a standalone module.

packages/sw-background-sync-queue/src/lib/requestManager.js
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	/**
+	 * function to start playing requests
+	 * in sequence
+	 * @return {Promise} promise that will resolve whenn all requests complete
+	 *
+	 * @memberOf RequestManager
+	 */
+	replayRequests() {
```
jeffposnick: I'd go with one of the canonical approaches to sequential promise invocation, like https://gist.github.com/anvk/5602ec398e4fdc521e2bf9940fd90f84

I'm wary of trying to iterate over them by keeping track of an index the way you're doing it.

packages/sw-background-sync-queue/src/lib/requestManager.js
```diff
@@ -0,0 +1,136 @@
...
+		if(reqData.response) {
+			return this.doFetch(index + 1);
+		}
+
+		let request = new Request(reqData.request.url, {
+			method: reqData.request.method,
+			headers: JSON.parse(reqData.request.headers),
+		});
+
+		if(request.method === 'POST') {
```
jeffposnick: What about `PUT`? Why not just always set `request.body` whenever the original request had a body, instead of explicitly checking for `POST`?

packages/sw-background-sync-queue/src/lib/requestManager.js
```diff
@@ -0,0 +1,136 @@
...
+		}
+
+		// proceed if response do not already exist
+		if(reqData.response) {
+			return this.doFetch(index + 1);
+		}
+
+		let request = new Request(reqData.request.url, {
+			method: reqData.request.method,
+			headers: JSON.parse(reqData.request.headers),
```
jeffposnick: I don't think that this code will work. `Headers` objects aren't JSON-serializable.

packages/sw-background-sync-queue/src/lib/requestManager.js
```diff
@@ -0,0 +1,136 @@
...
+		});
+
+		if(request.method === 'POST') {
+			request.body = reqData.request.body;
```
jeffposnick: This code *might* work if the request body is a string, but I'm fairly sure that both the initial serialization to IndexedDB, and the subsequent deserialization to a local variable here, will fail for binary data.

packages/sw-background-sync-queue/src/lib/requestManager.js
```diff
@@ -0,0 +1,136 @@
...
+		}
+
+		let reqClone = request.clone();
+		return fetch( request )
+			.then( (response) => {
+				responseManager.putResponse(hash, reqData, response.clone());
```
jeffposnick: Are you interested in checking for `response.ok` and explicitly `throw`ing an error to trigger the `.catch()` when it's false? That will allow you to retry when the server responds with, e.g., an HTTP 5xx response. (It also means that you'll end up retrying failures that might be permanent, like HTTP 403 responses.)

packages/sw-background-sync-queue/src/lib/requestManager.js
```diff
@@ -100,14 +102,15 @@ class RequestManager {
...
-	 * @param {any} request
-	 * @param {any} config
-	 * @return
-	 * 
+	 * takes a request and gives back JSON object that is storable in IDB
+	 *
+	 * @param {Request} request request object to transform
+	 * into iDB storable object
+	 * @param {Object} config config object to be
+	 * stored along inthe iDB
```
addyosmani: "inthe" -> "in"

packages/sw-background-sync-queue/src/lib/queue.js
```diff
@@ -6,16 +6,42 @@ import bcmanager from './broadcastManager';
...
+	 *
+	 * @memberOf Queue
+	 */
 	constructor() {
 		this.initialize();
 	}
 
+	/**
+	 * initialize the queue object from the idb
+	 * giving access to any penging queues
```
addyosmani: I would drop "the" and just say "idb". I think there's a typo here for penging -> pending

packages/sw-background-sync-queue/src/lib/queue.js
```diff
@@ -6,16 +6,42 @@ import bcmanager from './broadcastManager';
...
+	 * @param {Object} config config that gives values like maxAge to specify the
+	 *
```
addyosmani: Is the end of the description for `config` missing here?

packages/sw-background-sync-queue/src/lib/queue.js
```diff
@@ -6,16 +6,42 @@ import bcmanager from './broadcastManager';
...
+	 * @memberOf Queue
+	 */
 	async initialize(config) {
 		_config = config;
 		_queue = await idbQHelper.get('queue') || [];
 	}
 
+	/**
+	 * push any request to background sync queue which would be played later
+	 * prefferably when network comes back
```
addyosmani: preferrably -> preferably

packages/sw-background-sync-queue/src/lib/queue.js
```diff
@@ -47,6 +73,12 @@ class Queue {
 		}
 	}
 
+	/**
+	 * clean up the queue, deleting all the tasks who are either damaged or
+	 * whose maxAge has expired
```
addyosmani: Could you expand on what results in a task that is damaged? It may be worth documenting what we mean here. I think maxAge expiration is pretty clear otherwise.

packages/sw-background-sync-queue/demo/sw.js
```diff
@@ -2,14 +2,14 @@ importScripts('../build/background-sync-queue.js')
 
 //initialize bdQ
 goog.backgroundSyncQueue.initialize({},{
-	onRetrySuccess: async (req, res)=>{
+	onRetrySuccess: async (hash, req, res)=>{
 		console.log("success", req, res);
 		var data = await res.json();
```
addyosmani: In the rest of this source you're heavily using `let` and `const`. Is there a reason `var` is used here instead?

packages/sw-background-sync-queue/src/lib/queue.js
```diff
@@ -1,6 +1,7 @@
 import stringHash from 'string-hash';
 import idbQHelper from './idbHelper';
 import requestManager from './requestManager';
+import bcmanager from './broadcastManager'; 
 
 let _queue;
 let _config;
```
addyosmani: I think Jeff may have mentioned this, but I believe you'll want to switch away from module scoped variables here to member variables instead.

packages/sw-background-sync-queue/src/lib/queue.js
```diff
@@ -16,24 +17,39 @@ class Queue {
 	}
 
 	async push(request, config){
-		let localConfig = Object.assign({}, _config, config);
-		let hash = stringHash(request.url + JSON.stringify(request.headers) + await request.text() + Date.now());
-		_queue.push(hash);
+		try{
+			let localConfig = Object.assign({}, _config, config);
+			let hash = stringHash(request.url + JSON.stringify(request.headers) + await request.text() + Date.now());
```
addyosmani: `JSON.stringify(request.headers)` - are you sure this doesn't return an empty object?

projects/sw-background-sync-queue/src/index.js
```diff
@@ -11,7 +11,9 @@
  limitations under the License.
  */
 
-import initialize from './lib/initialization';
+import requestManager from './lib/requestManager';
 import pushIntoQueue from './lib/enqueue';
 
+var initialize = requestManager.initialize;
```
addyosmani: This is another one of those places where you might want to get consistent with the use of `let`/`const` vs `var`

packages/sw-background-sync-queue/demo/index.html
```diff
@@ -0,0 +1,28 @@
...
+	<head></head>
+	<body>
+		<div>Hello dear</div>
+		<button onclick="putReqInQ('1')">send req</button>
+		<button onclick="putReqInQ('2')">send req</button>
+		<button onclick="putReqInQ('3')">send req</button>
+		<button onclick="putReqInQ('4')">send req</button>
+		<button onclick="putReqInQ('5')">send req</button>
+		<script>
+			var _reg;
```
addyosmani: it appears `_reg` gets assigned the response from the Service Worker registration but other than the assignment it doesn't appear to get used. I think it's fine to just `console.log` and skip that part.

packages/sw-background-sync-queue/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "author": {
+    "name": "Prateek Bhatnagar",
+    "email": "prateekbh@google.com"
+  },
+  "license": "Apache-2.0",
+  "repository": "googlechrome/sw-helpers",
+  "bugs": "https://github.com/googlechrome/sw-helpers/issues",
+  "homepage": "https://github.com/GoogleChrome/sw-helpers/tree/master/projects/sw-background-sync-queue",
+  "main": "background-sync-queue.js",
+  "jsnext:main": "background-sync-queue.mjs"
```
addyosmani: 👍 for the `module` property. (Side: we might want to document some of these practices for future subprojects somewhere)

packages/sw-background-sync-queue/src/index.js
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * sw-background-sync-queue Module
+ *
+ * @module sw-background-sync-queue
+ */
+
+import requestManager from './lib/requestManager';
+import queue from './lib/queue';
+
+function initialize(config, callbacks) {
```
addyosmani: If it's just a wrapper, I think Jeff's suggestion makes this a little more terse so 👍

packages/sw-background-sync-queue/src/lib/queue.js
```diff
@@ -0,0 +1,143 @@
+import stringHash from 'string-hash';
+import idbQHelper from './idbHelper';
+import requestManager from './requestManager';
+import bcmanager from './broadcastManager';
+
+let _queue;
```
addyosmani: I had a similar comment in my review. Switching to class-based member variables would be a lot cleaner in this instance.

packages/sw-background-sync-queue/src/lib/queue.js
```diff
@@ -0,0 +1,143 @@
...
+			// broadcast the success
+			bcmanager.postMessage({
+				status: 'failed',
+				request: request,
+			});
+		}
+	}
+
+	/**
+	 * clean up the queue, deleting all the tasks who are either damaged or
```
addyosmani: Could we document what results in a task being damaged? It's pretty clear what leads to a maxAge being expired but it's unclear how a task might get into this other state nor whether there are things that could be done to recover from this state (other than be deleted).

packages/sw-background-sync-queue/src/lib/requestManager.js
```diff
@@ -0,0 +1,136 @@
...
+	 * @param {Object} callbacks object that contains function
+	 * ref for various lifecycle callbacks
+	 *
+	 * @memberOf RequestManager
+	 */
+	async initialize(config, callbacks) {
+		this.attachSyncHandler();
+		globalConfig = Object.assign({}, initializationDefaults, config);
+		globalCallbacks = callbacks;
+		await queue.initialize( globalConfig );
```
addyosmani: There are some calls where objects are wrapped in spaces and others where they aren't. I think this should be initialize(globalConfig).

packages/sw-background-sync-queue/src/lib/requestManager.js
addyosmani: Anne's pattern is also pretty easy on the eye to read. I'd hold preference on that pattern over manually tracking the way it's done here too.

packages/sw-background-sync-queue/src/lib/requestManager.js
addyosmani: A request header of the form `{ 'Content-Type': 'text/html' }` would fail to parse correctly with JSON.parse here.

packages/sw-background-sync-queue/src/lib/requestManager.js
addyosmani: I think you'll either want to do some type checking against the data (is it binary?) or have some sanitization before trying to assign otherwise it will fail.

packages/sw-background-sync-queue/demo/index.html
prateekbh: @jeffposnick @addyosmani i'll be droppping this demo all together to make a more presentable one, this was just to see if the POC was working.
request you to omit the demo portion for now

packages/sw-background-sync-queue/demo/index.html
```diff
@@ -0,0 +1,28 @@
+<html>
+	<head></head>
+	<body>
+		<div>Hello dear</div>
+		<button onclick="putReqInQ('1')">send req</button>
+		<button onclick="putReqInQ('2')">send req</button>
+		<button onclick="putReqInQ('3')">send req</button>
+		<button onclick="putReqInQ('4')">send req</button>
+		<button onclick="putReqInQ('5')">send req</button>
+		<script>
```
prateekbh: please omit everything in demo, as I'll be building two different demos with and without custom fetch handler.
this was just to see if the POC was working

packages/sw-background-sync-queue/src/index.js
```diff
@@ -0,0 +1,31 @@
...
+ limitations under the License.
+ */
+
+/**
+ * sw-background-sync-queue Module
+ *
+ * @module sw-background-sync-queue
+ */
+
+import requestManager from './lib/requestManager';
```
prateekbh: done

packages/sw-background-sync-queue/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "sw-background-sync-queue",
+  "version": "0.0.1",
+  "description": "A service worker implementation of the behavior defined in a page's App Cache manifest.",
+  "keywords": [
+    "service worker",
+    "sw",
+    "offline",
+    "queue"
```
prateekbh: done

packages/sw-background-sync-queue/package.json
```diff
@@ -0,0 +1,21 @@
...
+  ],
+  "author": {
```
prateekbh: done

packages/sw-background-sync-queue/package.json
```diff
@@ -0,0 +1,21 @@
...
+    "name": "Prateek Bhatnagar",
+    "email": "prateekbh@google.com"
+  },
+  "license": "Apache-2.0",
+  "repository": "googlechrome/sw-helpers",
+  "bugs": "https://github.com/googlechrome/sw-helpers/issues",
+  "homepage": "https://github.com/GoogleChrome/sw-helpers/tree/master/projects/sw-background-sync-queue",
```
prateekbh: done

packages/sw-background-sync-queue/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "sw-background-sync-queue",
+  "version": "0.0.1",
+  "description": "A service worker implementation of the behavior defined in a page's App Cache manifest.",
```
prateekbh: done

packages/sw-background-sync-queue/src/lib/constants.js
```diff
@@ -0,0 +1,3 @@
+export const initializationDefaults = {
```
prateekbh: done

packages/sw-background-sync-queue/src/lib/queue.js
```diff
@@ -0,0 +1,143 @@
...
+	 * prefferably when network comes back
+	 *
+	 * @param {Request} request request object to be queued by this
+	 * @param {Object} config optional config to override config params
+	 *
+	 * @memberOf Queue
+	 */
+	async push(request, config) {
+		let localConfig = Object.assign({}, _config, config);
+		let hash = stringHash(request.url + JSON.stringify(request.headers)
```
prateekbh: yeah this sounds better, 
done

packages/sw-background-sync-queue/src/lib/queue.js
```diff
@@ -0,0 +1,143 @@
...
+			+ await request.text() + Date.now());
```
prateekbh: agreed, changed

packages/sw-background-sync-queue/src/index.js
prateekbh: there are some ongoing changes in this, will check

packages/sw-background-sync-queue/src/lib/queue.js
prateekbh: done

packages/sw-background-sync-queue/src/lib/queue.js
```diff
@@ -0,0 +1,143 @@
...
+import requestManager from './requestManager';
+import bcmanager from './broadcastManager';
+
+let _queue;
+let _config;
+
+/**
+ * Core queue class that handles all the enqueue and dequeue
+ * as well as cleanup code for the background sync queue
+ * @class Queue
```
prateekbh: done

packages/sw-background-sync-queue/src/lib/queue.js
```diff
@@ -0,0 +1,143 @@
...
+ */
+class Queue {
+	/**
+	 * Creates an instance of Queue.
+	 *
+	 * @memberOf Queue
+	 */
+	constructor() {
```
prateekbh: done

packages/sw-background-sync-queue/src/lib/queue.js
```diff
@@ -0,0 +1,143 @@
...
+		this.initialize();
```
prateekbh: done, but a question.
I am not declaring a function async. its just i am calling another async function. not awaiting on it.
still a problem?

packages/sw-background-sync-queue/src/lib/queue.js
```diff
@@ -0,0 +1,143 @@
...
+	 * push any request to background sync queue which would be played later
+	 * prefferably when network comes back
+	 *
+	 * @param {Request} request request object to be queued by this
+	 * @param {Object} config optional config to override config params
+	 *
+	 * @memberOf Queue
+	 */
+	async push(request, config) {
+		let localConfig = Object.assign({}, _config, config);
```
prateekbh: actually I think it'll be a good idea to give override to some particular requests.
may be these requests are more critical and i want them to live for a longer time even if they fail. Or may be I want their response to live longer.
Comments?

packages/sw-background-sync-queue/src/lib/queue.js
```diff
@@ -0,0 +1,143 @@
...
+
+	/**
+	 * clean up the queue, deleting all the tasks who are either damaged or
+	 * whose maxAge has expired
+	 *
+	 * @memberOf Queue
+	 */
+	async cleanupQueue() {
+		let currIndex = 0;
+		for(const hash of _queue) {
```
prateekbh: done.. thanks for much cleaner approach

packages/sw-background-sync-queue/src/lib/queue.js
prateekbh: done

packages/sw-background-sync-queue/src/lib/requestManager.js
prateekbh: done

prateekbh: this method now only has initialization stuff, rest all moved to constructor

packages/sw-background-sync-queue/src/lib/requestManager.js
prateekbh: yep changed serialization and de-serialization for headers setting and getting

packages/sw-background-sync-queue/src/lib/requestManager.js
prateekbh: moved it out. Now independent to the class

packages/sw-background-sync-queue/src/lib/requestManager.js
prateekbh: 👍 done

packages/sw-background-sync-queue/src/lib/queue.js
prateekbh: fixed

packages/sw-background-sync-queue/src/lib/queue.js
prateekbh: 😆 gammer checking in comments makes a great idea for VSCode's extension.
thanks for reading it so carefully.
done

packages/sw-background-sync-queue/src/lib/queue.js
prateekbh: yesss indeed

projects/sw-background-sync-queue/src/index.js
prateekbh: changed all together

packages/sw-background-sync-queue/src/lib/requestManager.js
prateekbh: yep done, I thought lint would catch that

packages/sw-background-sync-queue/src/lib/queue.js
prateekbh: done

packages/sw-background-sync-queue/src/lib/broadcastManager.js
```diff
@@ -0,0 +1,2 @@
+let bcmanager = new BroadcastChannel('sw-backgroundsync');
```
prateekbh: done

packages/sw-background-sync-queue/src/lib/idbHelper.js
```diff
@@ -0,0 +1,3 @@
+import IDBHelper from '../../../../lib/idb-helper';
+const idbQHelper = new IDBHelper('bgQueueSyncDB', 1, 'QueueStore');
```
prateekbh: done

packages/sw-background-sync-queue/src/lib/idbHelper.js
```diff
@@ -0,0 +1,3 @@
+import IDBHelper from '../../../../lib/idb-helper';
```
prateekbh: i didn't really understood what u wanted to say here

packages/sw-background-sync-queue/src/lib/queue.js
```diff
@@ -0,0 +1,143 @@
...
+	/**
+	 * push any request to background sync queue which would be played later
+	 * prefferably when network comes back
+	 *
+	 * @param {Request} request request object to be queued by this
+	 * @param {Object} config optional config to override config params
+	 *
+	 * @memberOf Queue
+	 */
+	async push(request, config) {
```
prateekbh: this is breaking for me dont know why... will look into it

packages/sw-background-sync-queue/src/lib/queue.js
prateekbh: done

packages/sw-background-sync-queue/src/lib/queue.js
```diff
@@ -0,0 +1,143 @@
...
+			self.registration.sync.register('bgqueue');
+
+			// broadcast the success
+			bcmanager.postMessage({
+				status: 'success',
+				id: hash,
+				request: queuableRequest,
+			});
+		} catch(e) {
+			// broadcast the success
```
prateekbh: done

packages/sw-background-sync-queue/src/lib/requestManager.js
prateekbh: done

packages/sw-background-sync-queue/src/lib/queue.js
```diff
@@ -0,0 +1,143 @@
...
+			bcmanager.postMessage({
+				status: 'failed',
+				request: request,
```
prateekbh: done, but will explore to give more info to user

packages/sw-background-sync-queue/src/index.js
prateekbh: after removing the singleton nature of the modules, this initialize method no more is just a wrapper

packages/sw-background-sync-queue/src/lib/queue.js
```diff
@@ -0,0 +1,143 @@
...
+
+	/**
+	 * initialize the queue object from the idb
+	 * giving access to any penging queues
+	 * @param {Object} config config that gives values like
+	 * maxAge of requests in iDB
+	 *
+	 * @memberOf Queue
+	 */
+	async initialize(config) {
```
prateekbh: majority code from initialize has been shifted to synchronous constructor

packages/sw-background-sync-queue/src/lib/requestManager.js
prateekbh: removed this.

packages/sw-background-sync-queue/src/lib/responseManager.js
prateekbh: done

jeffposnick:
Just wanted to check in and see whether you're actively writing more code, or whether you're ready for another round of reviews.

prateekbh:
@jeffposnick closing this PR will reopen to master

PR #68, opened 2016-11-16
Moved over to newer JS build process

gauntface:
R: @jeffposnick @addyosmani 

This PR:
- re-adds the sw-appcache-behavior and updates it to work with rollup.
- Adds a check that the project is built before being tested.
- Fixes up some JSDoc warnings.
- Move shared libs to imports.

My only real concern is the change from 'global' to 'self'. I'm guessing this was a browserify quirk / issue, but not 100%. Jeff would appreciate your eyes on that.


jeffposnick:
Thanks, @gauntface! I did some hacky things to force a scoped namespace in the old setup that could be handled by Rollup explicitly now.

I'm going to push a commit to your branch to take care of that.

jeffposnick:
👍  Looks like the demo and tests still work as expected when run against the newly built code.

PR #69, opened 2016-11-17
Adding back in in the analytics suite

gauntface:
R: @jeffposnick @addyosmani

This adds the offline-analytics back in.

The major thing to change was the tests.

Previously everything was run through browserify. I've switched to include appropriate files in the HTML page directly rather than require them in and I've built the tested files at build time rather than in the test itself.

The tested files are built into `build/test/` in sw-offline-google-analytics and I've added that directory to npmignore.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "selenium-assistant": "^3.0.0",
     "serve-index": "^1.8.0",
     "serve-static": "^1.11.1",
+    "sinon": "^1.17.6",
     "sw-testing-helpers": "0.1.4"
   }
 }
--- a/package.json
+++ b/package.json
@@ -31,11 +31,10 @@
     "blueimp-md5": "^2.3.1",
     "browserify": "^13.1.0",
     "chai": "^3.5.0",
-    "chromedriver": "^2.24.1",
+    "chromedriver": "^2.25.1",
     "documentation": "^4.0.0-beta11",
     "eslint-config-google": "^0.7.0",
     "express": "^4.14.0",
-    "fetch-mock": "^5.1.5",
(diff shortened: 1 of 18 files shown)


package.json
```diff
@@ -63,6 +62,7 @@
     "selenium-assistant": "^3.0.0",
     "serve-index": "^1.8.0",
     "serve-static": "^1.11.1",
+    "sinon": "^1.17.6",
```
jeffposnick: Is this being used? I don't see it anywhere else.

packages/sw-offline-google-analytics/test/unit/replay-queued-requests.js
```diff
@@ -0,0 +1,132 @@
...
+  // An arbitrary, but valid, timestamp in milliseconds.
+  const initialTimestamp = 1470405670000;
+  // A 1000 millisecond offset.
+  const timestampOffset = 1000;
+  const originalFetch = window.fetch;
+
+  beforeEach(function() {
+    MockDate.set(initialTimestamp + timestampOffset);
+    fetchedUrls = [];
+    fetch = (requestUrl) => {
```
jeffposnick: Curious about the move from the mocked `fetch` from npm to this brute-force mocking. I'm pretty sure it's fine, but I wonder whether it was motivated by an incompatibility with the new build setup, or something else.

package.json
gauntface: Just used it to fix the fetch thing a little better.

jeffposnick:
👍  There's a test timeout failure in Chrome Canary 56, but it's passing in the other two Chrome releases, so I don't think it should prevent this merge.

PR #70, opened 2016-11-18
Tests

gauntface:
R: @jeffposnick @addyosmani 

Special PR just for you, this should make tests a less flakey.


gauntface:
I'm going to re-look at the appcache behavior tests to see if there is a way to move the selenium driver build call into the test case.

Because it's in 'before()' callback means it won't benefit from retries :(

jeffposnick:
Thanks as always, Matt! I'm 👍  with you merging this, unless you have any more experimentation to do first.

gauntface:
Nah this is good for now. Thanks Jeff.

PR #74, opened 2016-11-21
Higher Level Library - Early Commit

gauntface:
R: @jeffposnick @addyosmani 

This is just the beginning of the higher level library and at the moment it only supports registering routes.

Wanted to gauge what your thoughts are of this PR before moving forward (i.e. catch any sweeping changes / style or approach issues).

Matt


gauntface:
*nudge* @jeffposnick @addyosmani

packages/sw-lib/src/index.js
```diff
@@ -0,0 +1,35 @@
...
+
+/* eslint-env browser */
+/* eslint-disable no-console */
+
+import ErrorFactory from './lib/error-factory';
+import SWLib from './lib/sw-lib';
+import {Route} from '../../sw-routing/src/index.js';
+
+const initialiseLibrary = () => {
+  if (!('ServiceWorkerGlobalScope' in self &&
```
jeffposnick: Consider moving this sort of thing to https://github.com/GoogleChrome/sw-helpers/blob/master/lib/assert.js in case it's useful elsewhere.

It's kind of like a special-case form of the [`isInstance()`](https://github.com/GoogleChrome/sw-helpers/blob/master/lib/assert.js#L32) check.

jeffposnick:
I've got an in-progress review that's pending finalization 😄

packages/sw-lib/src/index.js
```diff
@@ -0,0 +1,35 @@
...
+    self instanceof ServiceWorkerGlobalScope)) {
+    // We are not running in a service worker, print error message
+    throw ErrorFactory.createError('not-in-sw');
+  }
+
+  self.goog = self.goog || {};
+  self.goog.swlib = new SWLib();
```
jeffposnick: You're already getting scoping "for free" via Rollup's [`moduleName` parameter](https://github.com/rollup/rollup/wiki/JavaScript-API#modulename). Can you just rely on that rather than implementing it yourself?

packages/sw-lib/src/index.js
```diff
@@ -0,0 +1,35 @@
...
+  self.goog.swlib.Route = Route;
```
jeffposnick: If we need to create an alias for `Route` under the `goog.swlib` namespace, I think the Rollup-friendly way of doing it would be just to specify it as a named export from this module, and have it inherit the `goog.swlib` namespace from `moduleName`.

packages/sw-lib/src/index.js
```diff
@@ -0,0 +1,35 @@
...
+*/
+
+/* eslint-env browser */
+/* eslint-disable no-console */
+
+import ErrorFactory from './lib/error-factory';
+import SWLib from './lib/sw-lib';
+import {Route} from '../../sw-routing/src/index.js';
+
+const initialiseLibrary = () => {
```
jeffposnick: Modules have an implicit scope, so I think just putting any code that needs to execute unconditionally at the top-level of the module is fine, without wrapping it in an additional function.

That might be slightly controversial, though?

packages/sw-lib/src/lib/error-factory.js
```diff
@@ -0,0 +1,61 @@
+/*
```
jeffposnick: Can you move this code to the top-level [`lib/`](https://github.com/GoogleChrome/sw-helpers/tree/master/lib) directory, which we're using for other modules that are shared throughout the monorepo? I can see it being useful elsewhere as well.

packages/sw-lib/src/lib/error-factory.js
```diff
@@ -0,0 +1,61 @@
...
+    generatedError.message = message;
+    generatedError.stack = stack;
+    return generatedError;
+  }
+
+  /**
+   * Returns the object containing all error names and messages.
+   * @return {Object} Error names and messages.
+   */
+  static getAllErrors() {
```
jeffposnick: I don't see this as being used anywhere, so consider leaving it out.

If it is something you want to keep, the pattern that we're using elsewhere for this sort of thing is to expose constant values as a named export from a `constants` module, e.g. https://github.com/GoogleChrome/sw-helpers/blob/master/packages/sw-broadcast-cache-update/src/lib/constants.js#L31

packages/sw-lib/src/lib/error-factory.js
```diff
@@ -0,0 +1,61 @@
...
+ * A simple class to make errors and to help with testing.
+ */
+class ErrorFactory {
+  /**
+   * @param {string} name The error name to be generated.
+   * @param {Error} [thrownError] The thrown error that resulted in this
+   * message.
+   * @return {Error} The generated error.
+   */
+  static createError(name, thrownError) {
```
jeffposnick: Elsewhere, we're using modules with named exports, which accomplishes the same thing as a Factory class with entirely static methods: https://github.com/GoogleChrome/sw-helpers/blob/master/lib/assert.js

I'm in favor of the module + named exports approach, as selectively importing what you need facilitates tree-shaking, and I think it feels more like "modern" JavaScript (for whatever that's worth).

packages/sw-lib/src/lib/router-wrapper.js
```diff
@@ -0,0 +1,73 @@
...
+   *     1. It can be an Express style route, like: '/example/:anything/route/'
+   *        The only gotcha with this is that it will only capture URL's on your
+   *        origin.
+   *     2. A regex that will be tested against request URL's.
+   *     3. A Route object
+   * @param {function|Handler} handler The handler is ignored if you pass in
+   * a Route, otherwise it's required. The handler will be called when the route
+   * is caught by the capture criteria.
+   */
+  registerRoute(capture, handler) {
```
jeffposnick: We're using anonymous objects with named properties for parameters elsewhere, and it would be good to stay consistent with that approach here.

packages/sw-lib/src/lib/sw-lib.js
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  /**
+   * If there are assets that should be cached on the install step but
+   * aren't revisioned, you can cache them here.
+   * @param {Array<String>} unrevisionedUrls A set of urls to cache when the
+   * service worker is installed.
+   * @return {Promise} Promise resolves if the install could be configured.
+   */
+  warmRuntimeCache(unrevisionedUrls) {
```
jeffposnick: That's a nice way of renaming `sw-toolbox`'s `precache()` method, which I've always found problematic. 👍

packages/sw-lib/src/lib/sw-lib.js
```diff
@@ -0,0 +1,72 @@
...
+
+  /**
+   * If there are assets that are revisioned, they can be cached intelligently
+   * during the install (i.e. old files are cleared from the cache, new files
+   * are added tot he cache and unchanged files are left as is).
+   * @param {Array<String>} revisionedUrls A set of urls to cache when the
+   * service worker is installed.
+   * @return {Promise} Promise resolves if the install could be configured.
+   */
+  cacheRevisionedAssets(revisionedUrls) {
```
jeffposnick: I'm cool with this placeholder for now, with the understanding that it's probably just going to call out to something from `sw-precaching` once that's implemented.

packages/sw-lib/src/lib/router-wrapper.js
gauntface: We had this discussion in the original proposal doc.

The higher level library was going to follow a sw-toolbox style API surface.

packages/sw-lib/src/lib/sw-lib.js
gauntface: yes it will do. This PR is just scaffolding with some implementation of the router side of work.

packages/sw-lib/src/lib/sw-lib.js
gauntface: Done.

packages/sw-lib/src/lib/sw-lib.js
gauntface: Done.

packages/sw-lib/src/lib/router-wrapper.js
gauntface: Done.

packages/sw-lib/src/lib/error-factory.js
gauntface: Done.

gauntface:
@jeffposnick Not sure how to mark this PR as up for review again but it's here.

I can put a demo directory in there if you want but I'd rather avoid it until this is a bit further along.

packages/sw-lib/src/lib/router-wrapper.js
jeffposnick: Right you are.

jeffposnick:
Thanks! It might not be until tomorrow, but I'll take another look.

jeffposnick:
👍 

I don't see any errors in the Travis CI build, but there's something about the build timing out that's logged. I'm assuming that's not a serious issue?

packages/sw-lib/src/lib/sw-lib.js
addyosmani: That's a really nice reframing. I've seen a lot of developers struggle to understand what precaching actually means. Good one, Matt.

packages/sw-lib/test/browser-unit/sw-unit/router.js
```diff
@@ -0,0 +1,126 @@
...
+
+  it('should be able to register a valid express route', function() {
+    const date = Date.now();
+    const expressRoute = `/${date}/:id/test/`;
+    const exampleRoute = `/${date}/1234567890/test/`;
+
+    return new Promise((resolve, reject) => {
+      goog.swlib.router.registerRoute(expressRoute, () => {
+        // TODO: Checking ':id' makes it through to here.
+
```
addyosmani: Are these TODOs for things you'd like to resolve in the future or before this PR lands?

packages/sw-lib/test/utils/dev-server.js
```diff
@@ -0,0 +1,8 @@
+const swTestingHelpers = require('sw-testing-helpers');
+const path = require('path');
+
+const testServer = new swTestingHelpers.TestServer();
+const serverPath = path.join(__dirname, '..', '..', '..', '..');
+testServer.startServer(serverPath, 8080).then((portNumber) => {
+ console.log(`http://localhost:${portNumber}/`);
```
addyosmani: If this is primarily only going to be logged in CI, console message looks fine. I'd otherwise log 'Server started on http://localhost:${portNumber}/' or similar.

gulp-tasks/lint.js
```diff
@@ -39,15 +39,16 @@ const lintPackage = (projectPath) => {
...
-    `./*.js`,
+    `./**/*.js`,
     // Root level JS Files
     `../*.js`,
   ])
   .pipe(eslint())
   .pipe(eslint.format())
   .pipe(eslint.results((results) => {
-    if (results.errorCount > 0 || results.warningCount > 0) {
+    if (results.errorCount > 0) {
```
gauntface: Yup. The top level lib/directory wasn't being caught by the lint task. Once I had fixed that, there are JSDocs missing which is across a number of projects.

Suggest we allow building with warnings until we are at a further along stage and then clean house and enforce the no warnings if we want.

packages/sw-lib/test/browser-unit/sw-unit/router.js
gauntface: Things to resolve in the future.

I have left it for now as it requires some exploration and changes in sw-routing.

packages/sw-lib/test/utils/dev-server.js
gauntface: This is not used beyond manual testing.

Basically the test server used here adds the serviceworker allowed header that you need for local testing - one of the myriad of things I'd like to fix in my totes amaze testing tool ;)

gauntface:
@addyosmani @jeffposnick Happy for this to land? I hopefully fixed the travis tests by swapping 'gulp-mocha' with 'gulp-spawn-mocha' which is meant to fix the problem.

jeffposnick:
👍

PR #75, opened 2016-11-22
Standardize the naming of the build output.

### LINKED ISSUE #73: Consistant naming for build files
cc @jeffposnick @addyosmani 

Just tried using routing in higher level library and writing tests for it and one thing I found a little confusing / non-obvious was the naming of sw-routing's output files.

I expected the name to be either 'sw-routing.js' or 'sw-routing.min.js', instead it was 'routing.umd.min.js'.

Would people agree that main build files should share name of package?

Would people agree that we can drop the umd and just have .min.js?

jeffposnick:
R: @addyosmani @gauntface

Fixes #73


gauntface:
LGTM.

PR #76, opened 2016-11-30
Sw precaching

gauntface:

This is a super duper early cut of the sw-precaching.

It's safe landing in master as the package.json as private set to true. But would be good to get current thoughts.

One thing before landing this: I need to add cache headers.


packages/sw-precaching/src/lib/revisioned-cache-manager.js
```diff
@@ -0,0 +1,66 @@
...
+        parsedAssetHash.path =
+          new URL(parsedAssetHash.path, location.origin).toString();
+      }
+
+      return parsedAssetHash;
+    });
+
+    this._cachedAssets = this._cachedAssets.concat(parsedFileList);
+
+    this._registerEvents();
```
jeffposnick: Since this is something that needs to be called exactly once, how about doing it in the class constructor instead? (See my later comment about the `install` handler.)

packages/sw-precaching/src/lib/revisioned-cache-manager.js
```diff
@@ -0,0 +1,66 @@
...
+  _registerEvents() {
+    if (this._eventsRegistered) {
+      // Only need to register events once.
+      return;
+    }
+
+    this._eventsRegistered = true;
+
+    self.addEventListener('install', (event) => {
+      event.waitUntil(
```
jeffposnick: How about checking that `this._cachedAssets.length > 0` before calling `event.waitUntil()`? That would also play nicely with moving the call to `this._registerEvents()` inside the constructor, since it will be efficient when `this._cachedAssets` is empty.

packages/sw-precaching/src/lib/revisioned-cache-manager.js
```diff
@@ -0,0 +1,66 @@
...
+
+      return parsedAssetHash;
+    });
+
+    this._cachedAssets = this._cachedAssets.concat(parsedFileList);
+
+    this._registerEvents();
+  }
+
+  _registerEvents() {
```
jeffposnick: There's only one event, so `_registerEvent` would be a more appropriate name.

utils/dev-server.js
```diff
@@ -0,0 +1,24 @@
+/*
```
jeffposnick: There's already https://github.com/GoogleChrome/sw-helpers/blob/master/gulp-tasks/serve.js

It would be good to only keep one of them. Is there anything this does that the `gulp serve` task doesn't accomplish? If we keep this, can you modify` gulp serve` so that it starts up this script?

packages/sw-precaching/src/lib/constants.js
```diff
@@ -16,3 +16,6 @@
 export const defaultCacheId = `sw-precaching`;
 export const hashParamName = '_sw-precaching';
 export const version = 'v1';
+export const DB_NAME = 'sw-precaching';
```
gauntface: Good point - I think I was copying sw-appcache-behaviour, but changed here.

packages/sw-precaching/src/lib/install-handler.js
```diff
@@ -13,49 +13,69 @@
...
+const addAssetHashToIndexedDB = (idbHelper, assetAndHash) => {
+  return idbHelper.put(assetAndHash.path, assetAndHash.revision);
+};
+
+const installHandler = ({assetsAndHahes, cacheId} = {}) => {
+  if (!Array.isArray(assetsAndHahes)) {
+    throw ErrorFactory.createError('assets-not-an-array');
+  }
+
+  const cacheName =
```
gauntface: Should we still include a version in the default cache name?

gauntface: One other comment. If the developer adds the library to the window referencing the scope at this level will throw and error.

packages/sw-precaching/src/lib/install-handler.js
```diff
@@ -13,49 +13,69 @@
...
+          if (previousRevisionDetails === assetAndHash.revision) {
+            /* eslint-disable no-console */
+            console.log('    Already Cached? TODO: Need to check cache to ' +
+              'ensure it\'s actually already cached.');
+            /* eslint-enable no-console */
+            return Promise.resolve();
+          }
+        }
+
+        return Promise.all([
```
gauntface: Well there are two problem scenarios:

1.) Cache fails to install.
2.) The indexedDB entry fails to install.

For scenario 1 (Cache failing to install), adding to indexeddb is not a bad scenario since the library should check the cache entry exists before when it finds the manifest has defined both the same asset and revision (i.e. the module should handle this scenario).

For scenario 2, having the cached asset means that the serviceworker can still use that cache. The only difference is that on the next install, the asset will be recached when it wasn't needed (this being due to no revision data being in indexeddb).

I've switched to the cache first then add indexeddb entry since it will probably cause confusion when someone else looks at this, but it does feel like a safe thing to do in parallel.

packages/sw-precaching/src/lib/revisioned-cache-manager.js
```diff
@@ -0,0 +1,66 @@
...
+import installHandler from './install-handler';
+
+class RevisionedCacheManager {
+  constructor() {
+    this._eventsRegistered = false;
+    this._cachedAssets = [];
+  }
+
+  cache({revisionedFiles} = {}) {
+    assert.isInstance({revisionedFiles: revisionedFiles}, Array);
```
gauntface: Done.

packages/sw-precaching/src/lib/revisioned-cache-manager.js
```diff
@@ -0,0 +1,66 @@
...
+
+    const parsedFileList = revisionedFiles.map((revisionedFile) => {
+      let parsedAssetHash = revisionedFile;
+      if (typeof revisionedFile === 'string') {
+        parsedAssetHash = {path: revisionedFile, revision: revisionedFile};
+      }
+
+      // TODO Check revisionedFile.path and revisionedFile.revision
+
+      if (parsedAssetHash.path.indexOf('http') !== 0) {
```
gauntface: Genius. I was hoping that was a better way of doing this. Thank You!!! :)

packages/sw-precaching/src/lib/revisioned-cache-manager.js
gauntface: There is a possibility of cleaning up caches in activate step if the cachename is changed between service worker loads, in which case there would be two events. Happy to change if you would like it for this PR.

packages/sw-precaching/src/lib/revisioned-cache-manager.js
gauntface: There are two options for this, adding the install event regardless of whether anything is cached or not OR adding the install event listener when we know it's going to be used for something.

Sounds like you are opting for the former by putting it in the constructor.

However we can't tell if the developer will actually use this module until they call cache - which is *after* the constructor.

packages/sw-precaching/src/lib/revisioned-cache-manager.js
gauntface: Done.

packages/sw-precaching/src/lib/revisioned-cache-manager.js
gauntface: Just moved to constructor. Going to leave the check in place to ensure it's not registered twice.

utils/dev-server.js
gauntface: Didn't realise that was a thing.

Main thing is to be able to start and use the server from unit tests. Made gulp serve over to using the test server and made test server use the serveStatic and serveIndex modules.

Only other change is to serve from the root of sw-helpers, this is to give the tests access to node_modules (i.e. sinon, mocha etc)

packages/sw-precaching/src/lib/install-handler.js
```diff
@@ -13,49 +13,69 @@
...
+          addAssetHashToIndexedDB(idbHelper, assetAndHash),
+          openCache.add(new Request(assetAndHash.path, {
```
jeffposnick: An option in the manifest that configures cache busting sounds like a good approach.

packages/sw-precaching/src/lib/install-handler.js
jeffposnick: In the back of my head, I was concerned that a mismatch between IDB and the cache could cause problems when trying to setup the routing for precached requests. But I guess IDB won't actually be used for that, in favor of using the parsed manifest as the source of truth. So I'm not as worried about mismatches.

packages/sw-precaching/src/lib/install-handler.js
```diff
@@ -13,49 +13,69 @@
...
+  if (!Array.isArray(assetsAndHahes)) {
+    throw ErrorFactory.createError('assets-not-an-array');
+  }
+
+  const cacheName =
+    `${cacheId || defaultCacheId}-${version}-${self.registration.scope}`;
+  const idbHelper = new IDBHelper(DB_NAME, DB_VERSION, DB_STORENAME);
+
+  return caches.open(cacheName)
+  .then((openCache) => {
```
gauntface: I just tried this - looks like the build process doesn't like doing this WITH the commonjs and resolve plugins for rollup.

Is this just me messing things up? Do you have any fixes for this?

gauntface:
@jeffposnick @addyosmani I've done another pass over this code and sure'd everything up and little and the cache headers are now being set so the tests will fail without the cache busting.

addyosmani:
> I've done another pass over this code and sure'd everything up and little and the cache headers are now being set so the tests will fail without the cache busting.

Will take a look now.

packages/sw-precaching/src/lib/revisioned-cache-manager.js
```diff
@@ -65,6 +65,15 @@ class RevisionedCacheManager {
     this._fileEntriesToCache = this._fileEntriesToCache.concat(parsedFileList);
   }
 
+  /**
+   * This method ensures that the file entry in the file maniest is valid and
```
addyosmani: maniest -> manifest?

packages/sw-precaching/src/lib/revisioned-cache-manager.js
```diff
@@ -92,101 +128,132 @@ class RevisionedCacheManager {
     this._eventsRegistered = true;
 
     self.addEventListener('install', (event) => {
-      if (this._cachedAssets.length === 0) {
+      if (this._fileEntriesToCache.length === 0) {
         return;
```
addyosmani: This does seem like a far more readable name then `cachedAssets`. Nice one.

packages/sw-precaching/src/lib/revisioned-cache-manager.js
```diff
@@ -92,101 +128,132 @@ class RevisionedCacheManager {
...
-              }
-              return false;
-            });
-          } else {
-            return false;
-          }
-        })
-        .then((isAlreadyCached) => {
-          if (isAlreadyCached) {
+    .then((oC) => {
```
addyosmani: Can we expand `oC` to a more readable variable name? I know you need to assign to `openCache` but you could use `openedCache` or something similar.

lib/idb-helper.js
```diff
@@ -19,123 +19,139 @@ import idb from 'idb';
...
  *
  * @private
- * @class
  */
-function IDBHelper(name, version, storeName) {
-  if (name == undefined || version == undefined || storeName == undefined) {
-    throw Error('name, version, storeName must be passed to the constructor.');
+class IDBHelper {
+  constructor(name, version, storeName) {
+    if (name == undefined || version == undefined || storeName == undefined) {
```
addyosmani: Strict `===` are being used elsewhere. Is there a reason the IDB helper is preferring `==`?

lib/idb-helper.js
```diff
@@ -19,123 +19,139 @@ import idb from 'idb';
...
+   * either existing or newly opened.
+   *
+   * @private
+   * @return {Promise<DB>}
+   */
+  _getDb() {
+    if (this._dbPromise) {
+      return this._dbPromise;
+    }
+    console.log('-------------> Opening DB');
```
addyosmani: Should these console messages be limited to a dev mode?

packages/sw-precaching/src/lib/revisioned-cache-manager.js
```diff
@@ -142,10 +97,78 @@ class RevisionedCacheManager {
...
+    const cacheName = cacheId || defaultCacheName;
+
+    return caches.open(cacheName)
+    .then((openCache) => {
+      const cachePromises = assetsAndHahes.map((assetAndHash) => {
+        return this._getRevisionDetails(assetAndHash.path)
+        .then((previousRevisionDetails) => {
+          if (previousRevisionDetails) {
+            if (previousRevisionDetails === assetAndHash.revision) {
+              /* eslint-disable no-console */
```
addyosmani: Disabling disallowing console (http://eslint.org/docs/rules/no-console) here. Was this a debug statement we forgot to drop or intentional? :)

utils/test-server.js
```diff
@@ -33,10 +33,12 @@ app.get('/test/iframe/:random', function(req, res) {
...
 
 app.get('/__echo/filename/:file', function(req, res) {
+  res.setHeader('Cache-Control', 24 * 60 * 60 * 1000);
   res.send(req.params.file);
 });
 
 app.get('/__echo/date/:file', function(req, res) {
+  res.setHeader('Cache-Control', 24 * 60 * 60 * 1000);
   res.send(`${req.params.file}-${Date.now()}`);
 });
```
addyosmani: On the whole the tests here looked fine to me. Very minor comment about console messaging and another on consistent use of equality statements.

packages/sw-precaching/test/browser-unit/sw-unit/revisioned-cache-manager.js
```diff
@@ -114,4 +114,48 @@ describe('Test RevisionedCacheManager', function() {
...
+    null,
+    '',
+    '1234sgdgh',
+    12345,
+    {},
+    [],
+  ];
+
+  badCacheBusts.forEach((badCacheBust) => {
+    it(`should be able to handle bad cache input '${JSON.stringify(badCacheBust)}'`, function() {
```
addyosmani: `it(testLabel), () => {`?

addyosmani:
> 1.) Add more tests to ensure robustness, including:
> 1.1) Cache absolute URLS
> 1.2) Cache absolute URLS for third party origins

Is there further work here on 1.1 and 1.2 you need to still land?

lib/idb-helper.js
jeffposnick: That's my (somewhat old) code.

It's `==` to catch `null` as well as `undefined`, since they both would be treated the same way. Not sure how likely it is for someone to explicitly pass in `null`...

jeffposnick:
Thanks to the GitHub UI I can't find the inline comment about `async`/`await` and the build process, so I'll respond here: if you run the `rollupBabel` plugin first, and then follow it with `resolve` and then `commonjs`, you should be able to transpile `async`/`await` and still pull in CommonJS modules. You can take a look at a working configuration in my PR for the cache expiration code:

https://github.com/GoogleChrome/sw-helpers/blob/f68cac2a9eeef570039f2f87157aca732d55d57c/packages/sw-cache-expiration/build.js#L31

I found that using `async`/`await` made a big difference in readability, so it's could be worth switching.

gauntface:
Ok. I've moved over to async and await (Thanks Jeff for the gulpfile) and I've added checks for absolute urls, both same origin and foreign origin.

There is a question as to what we do in terms of CORs support, but that should probably be discussed in the fizzle meeting on Monday.

PR #77, opened 2016-11-30
goog.cacheExpiration.Behavior code

jeffposnick:
R: @addyosmani @gauntface

This includes working code for the cache expiration behavior, along with a quick demo.

Still to come is better documentation (elaborating on the basic JSDocs that are present) and tests.


lib/assert.js
```diff
@@ -60,11 +60,21 @@ function isSWEnv() {
     self instanceof ServiceWorkerGlobalScope);
 }
 
+function isValue(object, expectedValue) {
+  const parameter = Object.keys(object).pop();
```
addyosmani: If `object` were a different type (e.g `Object.keys([]).pop()`), this would result in `actualValue` = object[undefined] if we weren't doing type validation before consuming the `object`. 

It might be worth checking your type  before using it. That said, there's prior art in this file as we aren't checking this in the `isType` function either. Your call on whether to change the pattern :)

packages/sw-cache-expiration/README.md
```diff
@@ -17,79 +17,92 @@ Browse sample source code in the [demo directory](https://github.com/GoogleChrom
...
 
-[packages/sw-cache-expiration/src/lib/behavior.js:25-156](https://github.com/GoogleChrome/sw-helpers/blob/a6f471056b52998c278dcb22dfc96169ea99932a/packages/sw-cache-expiration/src/lib/behavior.js#L25-L156 "Source code on GitHub")
+[packages/sw-cache-expiration/src/lib/behavior.js:48-263](https://github.com/GoogleChrome/sw-helpers/blob/43dea0fe7c5c2dca3500b988667de8b9ead46a3a/packages/sw-cache-expiration/src/lib/behavior.js#L48-L263 "Source code on GitHub")
 
-TODO Behavior Description
+**Examples**
 
-#### cacheName
+```javascript
+// Used as an automatically invoked as "behavior" by a RequestWrapper:
```
addyosmani: Suggestion: Add a H4 or bolded title for the example so it's easy to catch what each snippet is doing before needing to read through source or comments.

packages/sw-cache-expiration/README.md
```diff
@@ -17,79 +17,92 @@ Browse sample source code in the [demo directory](https://github.com/GoogleChrom
...
 
-Get the cache name
+// Set up a route to match any requests made against the example.com domain.
+// The requests will be handled with a stale-while-revalidate policy, and the
+// cache size will be capped at 10 entries.
+const route = new goog.routing.RegExpRoute({
+  match: ({url}) => url.domain === 'example.com',
+  handler: new goog.runtimeCaching.StaleWhileRevalidate({requestWrapper})
+});
+```
```
addyosmani: 👍

addyosmani:
You've already noted that more docs are on the way, but looking at the current cache expiration behavior most of the implementation makes sense in it's current form. I've only got very minor comments on the first commit.

packages/sw-cache-expiration/demo/service-worker.js
```diff
@@ -1,26 +1,33 @@
...
+      cacheName,
+      maxEntries: 2,
+      maxAgeSeconds: 10,
+    }),
+  ],
+});
+
+const route = new goog.routing.RegExpRoute({
+  regExp: /\.txt$/,
+  handler: new goog.runtimeCaching.StaleWhileRevalidate({requestWrapper}),
```
gauntface: This level of indirection makes it hard to reason with what is going on.

StaleWhileRevalidate > RequestWrapper > CacheExpirationBehavior.

Nothing to action on here, but reading this code, I have no idea why the RequestWrapper exists and how it works with the StaleWhileRevalidate and then how that plays with Behavior :(

packages/sw-cache-expiration/src/lib/behavior.js
```diff
@@ -14,144 +14,257 @@
 */
 
 import idb from 'idb';
-import Configuration from './configuration';
 import assert from '../../../../lib/assert';
-import {idbName, idbVersion} from './constants';
+import {idbName, idbVersion, urlProperty, timestampProperty} from './constants';
```
gauntface: Nit: Could you change these to urlPropertyName and timestampPropertyName?

I'm also finding it weird have normal variables lingering around and using like this rather than all caps variable names.

packages/sw-cache-expiration/src/lib/behavior.js
```diff
@@ -14,144 +14,257 @@
...
+   *        will be expired using a LRU policy once the cache reaches this size.
+   * @param {Number} [input.maxAgeSeconds] The maximum age for fresh entries.
    */
-  constructor({configuration}) {
-    if (typeof configuration === 'undefined') {
-      configuration = {};
-    }
+  constructor({cacheName, maxEntries, maxAgeSeconds} = {}) {
+    assert.isType({cacheName}, 'string');
+    assert.atLeastOne({maxEntries, maxAgeSeconds});
```
gauntface: Should check these are numbers as well.

packages/sw-cache-expiration/src/lib/behavior.js
```diff
@@ -14,144 +14,257 @@
...
+   * Returns a promise for the IndexedDB database used to keep track of state
+   * for `this.cacheName`.
+   *
+   * @private
+   * @return {DB} An open DB instance.
    */
-  get cacheName() {
-    return this.configuration.cacheName;
-  }
+  async getDB() {
```
gauntface: Am I right in assuming you can't use indexedDBHelper because you need to index on a different property?

packages/sw-cache-expiration/src/lib/behavior.js
```diff
@@ -14,144 +14,257 @@
...
+    const oldEntries = this.maxAgeSeconds ?
+      await this.findOldEntries({now}) :
+      [];
+
+    // Once that's done, check for the maximum size.
+    const extraEntries = this.maxEntries ?
+      await this.findExtraEntries() :
+      [];
+
+    const urls = [...new Set(oldEntries.concat(extraEntries))];
```
gauntface: What is this code doing?

1. Concat the two arrays into a new array.
2. Make a Set from the new array.
3. Spread the set into an array.

Are the entries not just lists of URLs? If they are, can't step 2 and 3 just be skipped?

packages/sw-cache-expiration/src/lib/behavior.js
```diff
@@ -14,144 +14,257 @@
...
     const expireOlderThan = now - (this.maxAgeSeconds * 1000);
     const urls = [];
-    return this.db.then((db) => {
-      const tx = db.transaction(this.cacheName, 'readwrite');
-      const store = tx.objectStore(this.cacheName);
-      store.iterateCursor((cursor) => {
+    const db = await this.getDB();
+    const tx = db.transaction(this.cacheName, 'readonly');
+    const store = tx.objectStore(this.cacheName);
+    const index = store.index(timestampProperty);
```
gauntface: Nit: Worth renaming index to timestampIndex?

packages/sw-cache-expiration/src/lib/behavior.js
```diff
@@ -14,144 +14,257 @@
...
+
+    index.iterateCursor(async (cursor) => {
```
gauntface: What does adding `async` do here?

This is probably me just being new to async and await but I always assumed async meant that you'd need to then use await somewhere in the function for it to serve a purpose.

packages/sw-cache-expiration/src/lib/behavior.js
```diff
@@ -14,144 +14,257 @@
...
+   *
+   * @private
+   * @return {Array<string>} A list of the URLs that were expired.
+   */
+  async findExtraEntries() {
+    const urls = [];
+    const db = await this.getDB();
+    const tx = db.transaction(this.cacheName, 'readonly');
+    const store = tx.objectStore(this.cacheName);
+    const index = store.index(timestampProperty);
```
gauntface: Nit: Worth renaming index to timestampIndex?

lib/assert.js
gauntface: I've had one or two situations where I've avoid the assert library because the object wrapping resulted in some weird issues :(

packages/sw-cache-expiration/README.md
jeffposnick: That's unfortunately not easy to control; it relies on the templates used by the [`documentation`](https://www.npmjs.com/package/documentation) package.

Longer term we may not use `documentation`, and perhaps what we move to will have a better story in terms of controlling the output.

lib/assert.js
jeffposnick: Wrapping in objects is more work for us when we write the library code, but I think it's worth the benefit offered to folks consuming our libraries. E.g., the pattern allows for a standard way of reporting incorrect library usage, like the following:

Assume the following assertions in our library code:

```js
assert.isType({cacheName}, 'string');
assert.atLeastOne({maxEntries, maxAgeSeconds});
```

Assume a third party developer has the following buggy usage:
```js
new goog.cacheExpiration.Behavior({
  cacheName,
});
```

That leads to `Please set at least one of the following parameters: 'maxEntries', 'maxAgeSeconds'` being logged.

Another example of buggy library usage:
```js
new goog.cacheExpiration.Behavior({
  maxEntries: 2,
  maxAgeSeconds: 10,
});
```

leads to `The 'cacheName' parameter has the wrong type. (Expected: string, actual: undefined)` being logged.

Wrapping the parameters passed to the assertions in an object is the easiest way I could think of to get the actual parameter names passed along to the assertion library, and third party developers get much more useful error messages as a result.

@gauntface, I'd love to fix whatever bugs were stopping you from using the assertions.

packages/sw-cache-expiration/demo/service-worker.js
jeffposnick: I know.

I'm going to add some comments to the demo as requested by @jpmedley, but I think we're just going to have to do a good job of documenting how this pattern is used throughout the framework, and try to paper over it with the higher-level wrapper for those who would rather use that.

packages/sw-cache-expiration/src/lib/behavior.js
jeffposnick: Changed.

packages/sw-cache-expiration/src/lib/behavior.js
jeffposnick: Done.

packages/sw-cache-expiration/src/lib/behavior.js
jeffposnick: Yup, basically.

Longer term, it might make sense to move the other libraries from `indexedDBHelper` (which dates back to the Appcache helper library) to `idb` or `idb-keyval` directly, given the additional flexibility they offer.

packages/sw-cache-expiration/src/lib/behavior.js
jeffposnick: Oops, this was just leftover from a previous version in which I was doing `IndexedDB` operations inside the loop. I'll remove it.

packages/sw-cache-expiration/src/lib/behavior.js
jeffposnick: Done.

packages/sw-cache-expiration/src/lib/behavior.js
jeffposnick: Done.

PR #80, opened 2016-12-05
sw-lib precache

gauntface:
Adds cacheRevisionedAssets to swlib.

### DIFF
--- a/packages/sw-lib/src/lib/error-factory.js
+++ b/packages/sw-lib/src/lib/error-factory.js
@@ -21,6 +21,9 @@ const errors = {
     'string, a Regex to capture request URLs or a Route instance.',
   'empty-express-string': 'The Express style route string must have some ' +
     'characters, an empty string is invalid.',
+  'bad-revisioned-cache-list': `The 'cacheRevisionedAssets()' method expects` +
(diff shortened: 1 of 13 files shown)


gauntface:
I may have tweaked the readme template to including dependency badges AND bumped the dependencies so we are all up to date :)

packages/sw-lib/src/lib/error-factory.js
```diff
@@ -21,6 +21,9 @@ const errors = {
     'string, a Regex to capture request URLs or a Route instance.',
   'empty-express-string': 'The Express style route string must have some ' +
     'characters, an empty string is invalid.',
+  'bad-revisioned-cache-list': `The 'cacheRevisionedAssets()' method expects` +
```
jeffposnick: This is fine for now, but longer term, we should think about including URLs with links to published documentation about the expected input, rather than inline examples of the type of input we expect. That way if we end up changing the expected format we just need to adjust the definitive JSDocs and don't have to worry about updating error strings.

gauntface: Agreed. I think it make sense before any notion of 1.0 release to have sweep of user niceties (i.e. ensure docs are available and flow well + good error messages). Main reason I opted for ErrorFactory is to somewhat make this easier.

PR #81, opened 2016-12-05
Adding the start of a CLI tool. No logic at the moment

gauntface:
R: @jeffposnick @addyosmani

Super early and small PR for the CLI / Module piece. Has no logic other than showing help / version text. But comments on general structure and testing approach would be appreciated.

### DIFF
--- a/packages/sw-cli/src/cli/cli-help.txt
+++ b/packages/sw-cli/src/cli/cli-help.txt
@@ -5,10 +5,10 @@ Usage:
 
 Commands:
 
-    generate-sw  [<path of project>]
-    build-file-manifest [<path of project>]
+    generate-sw
+    build-file-manifest
 
 Options
 
-    -h --help       Show this help text.
-    -v --version    Show current version of sw-cli
+    -h --help               Show this help text.
+    -v --version            Show current version of sw-cli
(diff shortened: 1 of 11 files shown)


packages/sw-cli/src/cli/cli-help.txt
```diff
@@ -0,0 +1,14 @@
+sw-cli
```
jeffposnick: Not a reflection on your code at all, but it's pretty disappointing that there isn't a better solution for `node` CLI help than an external text file. I was taking a look at what `yeoman` does and I guess it uses the same approach: https://github.com/yeoman/yo/blob/master/lib/usage.txt

😦

packages/sw-cli/src/cli/index.js
```diff
@@ -0,0 +1,140 @@
...
+  /**
+   * If a command is given in the command line args, this method will handle
+   * the appropriate action.
+   * @param {string} command The command name.
+   * @param {object} args The arguments given to this command.
+   * @param {object} flags The flags supplied with the command line.
+   * @return {Promise} A promise for the provided task.
+   */
+  handleCommand(command, args, flags) {
+    switch (command) {
```
jeffposnick: I think you're missing `break`s and I'm not sure why there a `{}` around one of the clauses in this `switch`.

gauntface: They are returns and I'm removed the braces/

Issue #78, opened 2016-12-05
Revamp the defaults around opaque responses


gauntface:
cc @addyosmani @jeffposnick 

**Library Affected**: sw-precache (Possibly some of the behavior modules as well)

**Browser & Platform**: All

**Issue or Feature Request Description**: How should we handle CORs / Bad requests?

This has been a common request / issue with developers using sw-toolbox (perhaps sw-precache as well). Arguments for and against which requests are cached and which aren't. Previously there was an option / config to define which status codes would be treated as valid status codes.

    // A regular expression to apply to HTTP response codes. Codes that match
    // will be considered successes, while others will not, and will not be
    // cached.
    successResponses: /^0|([123]\d\d)|(40[14567])|410$/

This breaks down to:

## 0 Opaque responses

**Good**: Works with CDN content from other domains.
**Bad**: Response may be bad and this means we could be serving up bad responses when a network 
response is available.

## 1XX Informational

> This class of status code indicates a provisional response, consisting only of the Status-Line and optional headers, and is terminated by an empty line. Since HTTP/1.0 did not define any 1xx status codes, servers must not send a 1xx response to an HTTP/1.0 client except under experimental conditions.

**Good**: ...
**Bad**: ...

Is this a common status code to use? When is this useful to cache?

## 2XX Success

> This class of status codes indicates the action requested by the client was received, understood, accepted and processed successfully.

**Good**: Extremely common and expected to be cached.
**Bad**: ...

## 3XX Redirection

> This class of status code indicates that further action needs to be taken by the user agent in order to fulfil the request. The action required may be carried out by the user agent without interaction with the user if and only if the method used in the second request is GET or HEAD. A user agent should not automatically redirect a request more than five times, since such redirections usually indicate an infinite loop.

**Good**: Can save a network trip from the server.
**Bad**: Redirect for auth may cause issues if cached and returned. Doesn't the `fetch` API handle redirects? Meaning that this should cache the final response rather than a 3XX redirect response?

## 401, 404, 405, 406, 407, 410 Client Error

> The 4xx class of status code is intended for cases in which the client seems to have erred. Except when responding to a HEAD request, the server should include an entity containing an explanation of the error situation, and whether it is a temporary or permanent condition. These status codes are applicable to any request method. User agents should display any included entity to the user.

401: Unauthorizaed
404: Not found
405: Method Not Allowed
406: Not Acceptable
407: Proxy Auth Required
410: Gone

[...]

PR #83, opened 2016-12-06
Extending tests for precaching module to ensure different scenarios a…

gauntface:
R: @jeffposnick @addyosmani 

This adds tests to cover the following scenarios:

- Duplicate entries are only cached onces
- Duplicate entries with the same path but different revisions throw an error
- 404 URLs aren't cached by default (Should this be extended to a range of status codes?
- CORs and Non-CORs requests are cached by default.

This is somewhat in prep for #78 where we'll introduce a way to override the default behavior. For now its response.ok || response.type === 'opaque' only.

### DIFF
--- a/packages/sw-precaching/src/lib/revisioned-cache-manager.js
+++ b/packages/sw-precaching/src/lib/revisioned-cache-manager.js
@@ -32,7 +32,7 @@ class RevisionedCacheManager {
    */
   constructor() {
     this._eventsRegistered = false;
-    this._fileEntriesToCache = [];
+    this._fileEntriesToCache = {};
     this._idbHelper = new IDBHelper(dbName, dbVersion, dbStorename);
 
     this._registerEvents();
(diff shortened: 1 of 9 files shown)


packages/sw-precaching/src/lib/revisioned-cache-manager.js
```diff
@@ -143,6 +156,14 @@ class RevisionedCacheManager {
 
       event.waitUntil(
         this._performInstallStep()
+        .then(() => {
+          // Closed indexedDB now that we are done with the install step
+          this._close();
+        }, (err) => {
```
jeffposnick: Would you mind switching to `.catch()`? I find the two-argument `.then()` to be much harder to follow.

packages/sw-precaching/src/lib/revisioned-cache-manager.js
```diff
@@ -158,23 +179,43 @@ class RevisionedCacheManager {
...
       let requestUrl = this._cacheBustUrl(fileEntry);
-      const response = await fetch(requestUrl, {
+      let fetchOptions = {
         credentials: 'same-origin',
-      });
-      await openCache.put(fileEntry.path, response);
-      await this._putRevisionDetails(fileEntry.path, fileEntry.revision);
+      };
+
+      if (requestUrl.indexOf(location.origin) !== 0) {
```
jeffposnick: As per the earlier conversation, I'd avoid doing this implicit `no-cors` fallback.

packages/sw-precaching/src/lib/revisioned-cache-manager.js
```diff
@@ -158,23 +179,43 @@ class RevisionedCacheManager {
...
+        await fetch(requestUrl, {
+          method: 'OPTIONS',
+        });
+      } catch (err) {
+        // If this error is received, it means pre-flight failed.
+        fetchOptions.mode = 'no-cors';
+      }
+      }
+
+      const response = await fetch(requestUrl, fetchOptions);
```
jeffposnick: I believe this is the bit that's going to be moved to using `RequestWrapper`.

packages/sw-precaching/src/lib/revisioned-cache-manager.js
gauntface: Done.

packages/sw-precaching/src/lib/revisioned-cache-manager.js
gauntface: Done.

Issue #48, opened 2016-10-18
Improve project health [meta bug]


addyosmani:
There are a number of project health improvements we could make to the current setup:
- [x] **[Make the failing tests pass](https://github.com/GoogleChrome/sw-helpers/issues/50)** - it's harder to understand the impact of new changes (and existing issues) with sub-projects when the tests have been failing for a while. Let's see if we can get them passing again before adding in new work. Right now it's hard to understand when the tests last actually went green (maybe [3 months ago](https://travis-ci.org/GoogleChrome/sw-helpers/builds/147627696)? :))
- [x] **[Consider scoping tests to supported browsers](https://github.com/GoogleChrome/sw-helpers/issues/51)** - when running `--project=sw-offline-google-analytics` I see a lot of tests failing in Safari (which doesn't support Service Worker). Selenium assistant can find any and all browsers available, we just need to check them against a browserId.
- [x] **Provide more insight into build health per project** - I believe we currently only display a single build status for our CI for the overall repo. I'd like to suggest we also include badges for individual projects in the README so folks can instantly tell what is green and what is not.
- [x] **[Decide on build artifacts](https://github.com/GoogleChrome/sw-helpers/issues/52)** - users have asked for us to consider including minified builds in our npm packages (#31). It would be good to decide on this so we can consistently apply it across the subprojects.
- [x] **Get linting to pass** - This has [just](https://github.com/GoogleChrome/sw-helpers/pull/46) been done as part of the work on docs.

gauntface:
@addyosmani I'm going to close this now that we have shields on the homepage for each project.

PR #84, opened 2016-12-08
Adding matrix to travis

gauntface:
R: @jeffposnick @addyosmani

This is the first step in fixing the remaining issues on: #48

> I believe we currently only display a single build status for our CI for the overall repo. I'd like to suggest we also include badges for individual projects in the README so folks can instantly tell what is green and what is not.

This PR doesn't include the individual shields, but I will do that tomorrow in a follow up PR once I get the server accessible somewhere.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,18 @@ language: node_js
 node_js:
   - 'stable'
 
+env:
+  - PROJECT=sw-appcache-behavior
+  - PROJECT=sw-background-sync-queue
+  - PROJECT=sw-broadcast-cache-update
+  - PROJECT=sw-cache-expiration
+  - PROJECT=sw-cli
+  - PROJECT=sw-lib
+  - PROJECT=sw-offline-google-analytics
+  - PROJECT=sw-precaching
+  - PROJECT=sw-routing
+  - PROJECT=sw-runtime-caching
+
 # Read more here: https://docs.travis-ci.com/user/gui-and-headless-browsers/#Using-xvfb-to-Run-Tests-That-Require-a-GUI
 before_script:
   - "export DISPLAY=:99.0"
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -5,13 +5,27 @@ language: node_js
 node_js:
   - 'stable'
 
+env:
+  # Keep star at top so it runs first - it'll be the longest task
+  - PROJECT="*"
```
jeffposnick: Is there still a reason to leave in `*`? As long as all the individual projects are also listed here, I'm not sure that it adds any value, and we'll be running everything twice.

gauntface: It is only covering the case someone fails to add their project to travis.yml which I can very much see happening. But beyond there is very little use to this granted. Shall I remove it?

gauntface:
@jeffposnick happy for this to land?

jeffposnick:
I just took a look at the Travis runs and it seems like we're spending ~1 minute per project on the `npm install` step, even though the same `package.json` is used for each of the runs. How about configuring:

```yaml
cache:
  directories:
    - node_modules
```

as per https://blog.travis-ci.com/2013-12-05-speed-up-your-builds-cache-your-dependencies?

gauntface:
Just added this in. Will look at doing the same for the selenium browsers later today as well (once this + shields is in).

gauntface:
So something weird is occuring with Travis since switching on node module caching (although I think that's just a coincidence. Node failed to install Node 7 (i.e. stable) and now it's failing to find modules that were originally installed.

Hopefully a new build soon will pass without these errors.

jeffposnick:
Hmm, each of the projects are still taking ~1 minute on `npm install` and installing their own copies of everything.

Maybe the next full build will actually take advantage of the cache?

gauntface:
Yeah I cleared the cache to get the tests to pass (I think the first attempt was broken and continued breaking.

gauntface:
Well the caches page on travis is showing '309.03MB' begin cached for the branch and PR, so I think it's working.

@jeffposnick are you happy to take this PR or do you want me to push some "fake changes" to see if the caching is working?

PR #85, opened 2016-12-09
Precaching Rejig

gauntface:
R: @jeffposnick @addyosmani

This is a slight change on the previous precaching PR (previous comments have been taken into account here).

- This has tests for caching relative, absolute, CORs and Local requests. The entries in the file manifest are now parsed by a CacheEntry model (this handles defaults and checking the input is valid).

- Rejigged some of the code so that the RevisionedCacheManager is a little leaner and the code is easier to read.

- Changed {path: '', revision: ''} -> {url: '', revision: ''}

### DIFF
--- a/packages/sw-precaching/src/lib/models/revisioned-cache-entry.js
+++ b/packages/sw-precaching/src/lib/models/revisioned-cache-entry.js
@@ -1,8 +1,33 @@
+import {cacheBustParamName} from '../constants';
+
 class RevisionedCacheEntry {
-  constructor({revisionID, revision, request}) {
+  constructor({revisionID, revision, request, cacheBust}) {
     this.revisionID = revisionID;
     this.revision = revision;
     this.request = request;
+    this.cacheBust = cacheBust;
+  }
+
+  get cacheBustRequest() {
+    return new Request(this._cacheBustUrl(this.request.url, this.revision));
+  }
+
+  /**
+   * This method takes a file entry and if the `cacheBust` parameter is set to
+   * true, the cacheBust parameter will be added to the URL before making the
+   * request. The response will be cached with the absolute URL without
+   * the cache busting search param.
+   * @param {String} requestURL This is an object with `url`, `revision` and
+   * `cacheBust` parameters.
+   * @param {String} revision Revision to use in the cache bust.
+   * @return {String} The final URL to make the request to then cache.
+   */
+  _cacheBustUrl(requestURL, revision) {
+    const parsedURL = new URL(requestURL, location.origin);
+    parsedURL.search += (parsedURL.search ? '&' : '') +
+      encodeURIComponent(cacheBustParamName) + '=' +
+      encodeURIComponent(revision);
+    return parsedURL.toString();
   }
 }
 
--- a/packages/sw-precaching/src/lib/models/revisioned-cache-entry.js
+++ b/packages/sw-precaching/src/lib/models/revisioned-cache-entry.js
@@ -1,0 +1,9 @@
+class RevisionedCacheEntry {
+  constructor({revisionID, revision, request}) {
+    this.revisionID = revisionID;
+    this.revision = revision;
+    this.request = request;
+  }
+}
+
+export default RevisionedCacheEntry;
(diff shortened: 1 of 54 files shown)


packages/sw-precaching/src/lib/models/revisioned-cache-entry.js
```diff
@@ -0,0 +1,34 @@
...
+   * true, the cacheBust parameter will be added to the URL before making the
+   * request. The response will be cached with the absolute URL without
+   * the cache busting search param.
+   * @param {String} requestURL This is an object with `url`, `revision` and
+   * `cacheBust` parameters.
+   * @param {String} revision Revision to use in the cache bust.
+   * @return {String} The final URL to make the request to then cache.
+   */
+  _cacheBustUrl(requestURL, revision) {
+    const parsedURL = new URL(requestURL, location.origin);
```
jeffposnick: Optional: You could follow the model at https://github.com/GoogleChrome/samples/blob/gh-pages/service-worker/custom-offline-page/service-worker.js#L24 where you first try to set `{cache: 'reload'}` and if that works, avoid using a URL parameter.

`{cache: 'reload'}` is only supported in Firefox and I haven't seen any progress towards getting it supported in Chrome, though.

packages/sw-precaching/src/lib/revisioned-cache-manager.js
```diff
@@ -15,8 +15,10 @@
 
 import assert from '../../../../lib/assert';
 import ErrorFactory from './error-factory';
-import IDBHelper from '../../../../lib/idb-helper.js';
-import {defaultCacheName, dbName, dbVersion, dbStorename, cacheBustParamName}
+import StringCacheEntry from './models/string-cache-entry.js';
+import DefaultsCacheEntry from './models/defaults-cache-entry.js';
+import RevisionDetailsModel from './models/revision-details-model.js';
+import {defaultCacheName}
   from './constants';
```
jeffposnick: Probably put this back on the previous line.

packages/sw-precaching/src/lib/models/revisioned-cache-entry.js
gauntface: I've added this in with a different feature support check `'cache' in Request.prototype`

jeffposnick:
I'm seeing

```
  1) sw-precaching Browser Tests should pass all browser based unit tests in  Google Chrome Dev:
     Error: Issues in Google Chrome Dev.
Google Chrome Dev had 1 test failures.
------------------------------------------------
[Failed Test 1]
    - sw-precaching Test Revisioned Caching > should cache and fetch files
        Cannot read property 'should' of undefined
------------------------------------------------
      at assistantDriver.getSeleniumDriver.then.then.then (packages/sw-precaching/test/automated-test-suite.js:74:17)
      at ManagedPromise.invokeCallback_ (node_modules/selenium-webdriver/lib/promise.js:1366:14)
      at TaskQueue.execute_ (node_modules/selenium-webdriver/lib/promise.js:2970:14)
      at TaskQueue.executeNext_ (node_modules/selenium-webdriver/lib/promise.js:2953:27)
      at asyncRun (node_modules/selenium-webdriver/lib/promise.js:2813:27)
      at node_modules/selenium-webdriver/lib/promise.js:676:7
      at process._tickCallback (internal/process/next_tick.js:103:7)
```

It's not related to this PR. I don't know if you want to merge this now and then fix that, or fix the underlying test in a separate PR that you rebase into this.

gauntface:
I'll fix the test before merging this. Feel free to merge your PR though if you are happy with it.

Bad builds make for an unhappy gauntface.

gauntface:
@jeffposnick I realised the problem I was hitting.

selenium-assistant was updating Chrome dev where my local machine wasn't up-to-date (i.e. tests were on 57, locally I was testing on 56).

In Chrome 57 a change was introduced so that the cachedResponse's `.url` value included the search params, where as 56 and below, it was just the URL used to cache the response with.

this lead to issues in the way the tests were checking the validity of the cached responses.

This PR now has some extensive logging for all Browser and Service Worker tests WITH stack traces relevant to failing tests, so in the future tests should be easier to spot the specific failing test and the specific location of the error.

Finally. I've moved to a default port number for the test server when nothing is defined. This should prevent the case where node attempts to pick a free port number to then realise that it's in use (it's a rare scenario but is happening frequently across Travis tests.

jeffposnick:
Thanks for dealing with all the nitty-gritty! 👍

PR #87, opened 2016-12-12
Refactor the build process; build minified + non-minified libraries

### LINKED ISSUE #31: Feature Request: Include minified offline-google-analytics.js in npm install
**Library Affected**:
sw-google-offline-analytics

**Browser & Platform**:
All browsers

It would be great if the npm install also included a minified version of the offline-google-analytics.js file 👍

jeffposnick:
R: @addyosmani @gauntface
CC: @deanhume

This PR refactors much of the common build boilerplate into a helper method. That helper method also takes care of generating both minified and non-minified output, which fixes #31.

There are some packages that still require slightly different build setups, either due to their testing requirements or due to their legacy packaging requirements. In those cases I've switched over the part of the build that could be abstracted out into the general configuration, and left behind the things that needed to stay separate.

### DIFF
--- a/build-utils.js
+++ b/build-utils.js
@@ -15,17 +15,20 @@
 
 /* eslint-disable no-console, valid-jsdoc */
 
+const babel = require('gulp-babel');
 const childProcess = require('child_process');
-const path = require('path');
-const promisify = require('promisify-node');
+const commonjs = require('rollup-plugin-commonjs');
 const fs = require('fs');
 const gulp = require('gulp');
 const gulpif = require('gulp-if');
-const sourcemaps = require('gulp-sourcemaps');
-const rename = require('gulp-rename');
-const rollup = require('gulp-rollup');
-const babel = require('gulp-babel');
 const header = require('gulp-header');
+const path = require('path');
+const promisify = require('promisify-node');
+const rename = require('gulp-rename');
+const resolve = require('rollup-plugin-node-resolve');
+const rollup = require('gulp-rollup');
+const rollupBabel = require('rollup-plugin-babel');
+const sourcemaps = require('gulp-sourcemaps');
 
 const globPromise = promisify('glob');
 
@@ -105,6 +108,7 @@ function buildJSBundle(options) {
     const sources = [
       `${options.projectDir}/src/**/*.js`,
       '{lib,packages}/**/*.js',
+      // Explicitly avoid matching node_modules/.bin/*.js
       'node_modules/*/**/*.js',
     ];
 
@@ -133,10 +137,22 @@ function buildJSBundle(options) {
  *        ('umd', 'es', etc.) to the path to use for the output.
  * @param {String} projectDir The path of the project directory.
  * @param {String} moduleName The name of the module, used in the UMD output.
- * @param {Array.<Object>} [plugins] A list of Rollup plugins to use.
  * @returns {Array.<Object>}
  */
-function generateBuildConfigs(formatToPath, projectDir, moduleName, plugins) {
+function generateBuildConfigs(formatToPath, projectDir, moduleName) {
+  const plugins = [
+    rollupBabel({
+      plugins: ['transform-async-to-generator', 'external-helpers'],
+      exclude: 'node_modules/**',
+    }),
+    resolve({
+      jsnext: true,
+      main: true,
+      browser: true,
+    }),
+    commonjs(),
+  ];
+
   // This is shared throughout the full permutation of build configs.
   const baseConfig = {
     rollupConfig: {
@@ -167,7 +183,9 @@ function generateBuildConfigs(formatToPath, projectDir, moduleName, plugins) {
       return fullConfig;
     });
   }).reduce((previous, current) => {
-    // Flatten the two-dimensional array.
+    // Flatten the two-dimensional array. E.g.:
+    // Input is [[umd-minified, umd-unminified], [es-minified, es-unminified]]
+    // Output is [umd-minified, umd-unminified, es-minified, es-unminified]
     return previous.concat(current);
   }, []);
 }
--- a/build-utils.js
+++ b/build-utils.js
@@ -20,6 +20,7 @@ const path = require('path');
 const promisify = require('promisify-node');
 const fs = require('fs');
 const gulp = require('gulp');
+const gulpif = require('gulp-if');
 const sourcemaps = require('gulp-sourcemaps');
 const rename = require('gulp-rename');
 const rollup = require('gulp-rollup');
@@ -27,6 +28,8 @@ const babel = require('gulp-babel');
 const header = require('gulp-header');
 
 const globPromise = promisify('glob');
+
+const LICENSE_HEADER = fs.readFileSync('LICENSE-HEADER', 'utf8');
 
 /**
  * Wrapper on top of childProcess.spawn() that returns a promise which rejects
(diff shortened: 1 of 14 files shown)


build-utils.js
```diff
@@ -94,41 +97,85 @@ function taskHarness(task, projectOrStar, ...args) {
...
  */
 function buildJSBundle(options) {
-  const destPath = path.join(options.projectDir, 'build');
-  const licenseHeader = fs.readFileSync('LICENSE-HEADER', 'utf8');
+  return new Promise((resolve, reject) => {
+    const outputName = options.buildPath.replace(/^build\//, '');
+    const sources = [
+      `${options.projectDir}/src/**/*.js`,
+      '{lib,packages}/**/*.js',
+      'node_modules/*/**/*.js',
```
gauntface: Nit: could this not be node_modules/**/*.js?

build-utils.js
```diff
@@ -94,41 +97,85 @@ function taskHarness(task, projectOrStar, ...args) {
...
+
+      // Remove the '.min.' from the file name if the output isn't minified.
+      if (!fullConfig.minify) {
+        fullConfig.buildPath = fullConfig.buildPath.replace('.min.', '.');
+      }
+
+      return fullConfig;
     });
-  });
+  }).reduce((previous, current) => {
```
gauntface: This is probably due to the chaining of maps and reduce. But what is the input and output of this? I thought I was following this but apparently not.

build-utils.js
```diff
@@ -94,41 +97,85 @@ function taskHarness(task, projectOrStar, ...args) {
...
+      'node_modules/*/**/*.js',
+    ];
 
-  if (options.outputName.indexOf('build/') !== 0) {
-    throw new Error('Expected options.ouputName to start with \'build/\'');
-  }
+    gulp.src(sources)
+      .pipe(sourcemaps.init())
+      .pipe(rollup(options.rollupConfig))
+      .pipe(gulpif(options.minify, babel({
```
gauntface: Nit: Won't this result in non-transpiled versions of the code as well? If so could we rename minify to minifyTranspile or go for a `buildtype = 'production'` style variable.

packages/sw-appcache-behavior/build.js
```diff
@@ -32,7 +32,7 @@ module.exports = () => {
           commonjs(),
         ],
       },
-      outputName: 'build/client-runtime.js',
+      buildPath: 'build/client-runtime.js',
```
gauntface: NIT: should these be swapped out with a value in package.json or does that not apply in this package?

packages/sw-routing/build.js
```diff
@@ -15,38 +15,20 @@
 
 const commonjs = require('rollup-plugin-commonjs');
 const nodeResolve = require('rollup-plugin-node-resolve');
-const path = require('path');
 const pkg = require('./package.json');
-const {buildJSBundle} = require('../../build-utils');
+const {buildJSBundle, generateBuildConfigs} = require('../../build-utils');
 
-const pluginsConfig = [
+const plugins = [
```
gauntface: Is there a reason this doesn't have the babel plugin?

packages/sw-runtime-caching/build.js
```diff
@@ -13,41 +13,18 @@
...
 */
 
-const rollupBabel = require('rollup-plugin-babel');
-const path = require('path');
-const {buildJSBundle} = require('../../build-utils');
 const pkg = require('./package.json');
+const rollupBabel = require('rollup-plugin-babel');
+const {buildJSBundle, generateBuildConfigs} = require('../../build-utils');
+
+const plugins = [rollupBabel({
```
gauntface: If it doesn't hurt to have the commonjs and noderesolve plugins, could we just collapse these modules to just use the same set of plugins in a shared way?

build-utils.js
jeffposnick: Maybe? I think you were the one who used that glob pattern to begin with: https://github.com/GoogleChrome/sw-helpers/blame/master/build-utils.js#L112 and I was just keeping it consistent.

I'm happy to change it and see if there's an issue.

build-utils.js
jeffposnick: There's an optional Babel transpilation that happens depending on the `plugins` definition in `options.rollupConfig`.

And then, independent of that and after all the Rollup transformations happen, there's another Babel invocation, but with it just configured to minify via Babili.

It's kind of confusing that Babel in the second case is basically just a harness of kick off minification, but that's how it goes.

build-utils.js
jeffposnick: I'll add a comment. It's just a pattern for flattening arrays.

packages/sw-appcache-behavior/build.js
jeffposnick: This library is messy in that there are two different bundled outputs, one for consumption by the SW and the other for the browser. I didn't know how to cleanly list the metadata in the `package.json`'s `main`, since it wasn't clean whether it should be the path to the SW or browser script. So I decided to leave out the `package.json` metadata entirely.

Longer term, maybe splitting up the pieces into two different modules, each with their own `package.json` would make the most sense? Or some other refactoring that's outside the scope of this PR.

packages/sw-routing/build.js
jeffposnick: There's no transpilation needed. The only thing we're transpiling elsewhere is `async`/`await` ➡️  generators, and there's none of that in this bundle.

jeffposnick: But, as per the next comment, I'll just use a common set of plugins.

packages/sw-runtime-caching/build.js
jeffposnick: I was initially hesitant to add things to the build process that weren't needed for a given library, but I just did a quick comparison of before/after when using all the plugins, and there's either a negligible or no change in build time and output file size.

So I'll just standardize.

build-utils.js
jeffposnick: Just tried changing it, and you get a

```
[10:59:25] Starting 'build'...

events.js:160
      throw er; // Unhandled 'error' event
      ^
Error: Input files must be buffered!
```

if you don't explicitly use the original glob pattern. I think without that original pattern, `gulp.src` picks up the symlinks in `node_modules/.bin` and gets confused?

Anyway, I'm assuming that's why you had it set up that way. I'll add a comment.

jeffposnick:
PTAL.

There are probably still Travis CI failures, as they seemed to be unrelated to this PR.

packages/sw-appcache-behavior/build.js
gauntface: Ignore if that's the case. Just wanted to make sure that was the case.

I'm going to hit the same problem with some push modules I intend to write next quarter (i.e. lib in window and lib in SW). Just leave as is for now.

gauntface:
Thanks for making those changes. If the travis builds fail, could you open an issue with links to the overall builds and I'll dig into them.

I'm in the middle of putting together a PR to make logging on Browser and SW tests more visible and useful at top level, so hopefully things will be a bit easier to diagnose soon.

jeffposnick:
I've filed https://github.com/GoogleChrome/sw-helpers/issues/88 to track the build failures.

gauntface:
Thanks, feel free to merge this if you are happy with this. Slow Travis builds are killing my productivity.

PR #89, opened 2016-12-15
Adding root dir question to cli

gauntface:
R: @jeffposnick 

First step towards #82

Some minor changes to the flow. The CLI detects directories in the current working directory and offers them in a list, otherwise the user can manually enter in a directory name.

This is the same code as previous PR - I just re-did to avoid weird merge conflict / output from Github.


packages/sw-cli/src/cli/errors.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
```
jeffposnick: It's more in keeping with the other packages to put this in a `lib/` directory.

packages/sw-cli/src/cli/index.js
```diff
@@ -121,8 +123,113 @@ class SWCli {
...
+        resolve(directoryContents);
+      });
+    })
+    .then((directoryContents) => {
+      return directoryContents.filter((directoryContent) => {
+        return fs.statSync(directoryContent).isDirectory();
+      });
+    })
+    .then((subdirectories) => {
+      const blacklistDirectoryNames = [
```
jeffposnick: Consider pulling this out into a `constants.js` module.

packages/sw-cli/src/cli/errors.js
gauntface: Done.

packages/sw-cli/src/cli/index.js
gauntface: Done.

PR #90, opened 2016-12-15
sw-cacheable-response-behavior implementation

jeffposnick:
Related to https://github.com/GoogleChrome/sw-helpers/issues/78, this implements a behavior with the `cacheWillUpdate` callback. It effectively lets you pass in one or more statuses and one or more headers that are used to determine whether a response is cacheable or not. It also includes changes to the `RequestWrapper` in the runtime caching module so that it will automatically call `cacheWillUpdate` prior to adding a `Response` to the cache. If there are no `cacheWillUpdate` callbacks defined, `RequestWrapper` will just check the `.ok` property to determine whether a `Response` can be cached.


packages/sw-runtime-caching/src/lib/request-wrapper.js
```diff
@@ -140,7 +145,23 @@ class RequestWrapper {
     assert.atLeastOne({request});
 
     const response = await this.fetch({request});
-    if (response.ok || response.type === 'opaque') {
+
+    // .ok is true if the response status is 2xx. That's the default condition.
+    let cacheable = response.ok;
+    // If we have any cacheWillUpdate callbacks defined...
+    if (this.callbacks.cacheWillUpdate) {
```
gauntface: Nit: Could this be changed to behaviorCallbacks. Just grokking this code it's not obvious that callbacks relates to behaviors.

packages/sw-runtime-caching/src/lib/request-wrapper.js
```diff
@@ -140,7 +145,23 @@ class RequestWrapper {
...
+      // ...run throuh the callbacks one at a time, using reduce() to keep track
+      // of the previous result. We explicitly use reduce() here rather than
+      // some() since we want to make sure to run all of the callbacks, even
+      // if we already have a true result from one.
+      // This allows cacheWillUpdate callbacks to have side effects other, than
+      // just returning true or false.
+      cacheable = this.callbacks.cacheWillUpdate.reduce((previous, cb) => {
```
gauntface: When we discussed this we were going to enforce a single will cache behavior.

Having multiple just feels error prone.

packages/sw-cacheable-response-behavior/src/lib/behavior.js
```diff
@@ -0,0 +1,121 @@
...
+   * If multiple criteria are present (e.g. both `statuses` and `headers`), then
+   * the `Response` needs to meet all of the criteria to be cacheable.
+   *
+   * @param {Object} input The input object to this function.
+   * @param {Array<Number>} [input.statuses] The status codes that are
+   *        checked when determining whether a `Response` is cacheable.
+   * @param {Object<String,String>} [input.headers] The header values that are
+   *        checked when determining whether a `Response` is cacheable.
+   */
+  constructor({statuses, headers} = {}) {
```
gauntface: Was there a reason this only takes statuses and headers?

It feels like a callback with the response would cater for a lot more scenarios than enforcing only these options.

For example, if the developer wanted to only cache responses with a certain set of headers '.some' wouldn't be correct here. I'm also wondering if the developer would ever want to restrict caching based on the request as well as the response (i.e. only cache is the request header and the response header match....Edge case I know)

packages/sw-runtime-caching/src/lib/request-wrapper.js
jeffposnick: Changed.

packages/sw-cacheable-response-behavior/src/lib/behavior.js
jeffposnick: I think this is the difference between defining a lifecycle callback and then providing a behavior module that implements that callback.

The `cacheWillUpdate` lifecycle callback should be defined as broadly as possible, and it's a good idea to pass in the original `Request` to it. I've modified `RequestWrapper` to take that into account.

With `goog.cacheableResponse.Behavior`, we have a wrapper to make implementing `cacheWillUpdate` for some common scenarios (header values, status values) easier for developers. Folks don't have to use `goog.cacheableResponse.Behavior` if that's too limiting, though—they can pass in any object that implements `cacheWillUpdate` and returns a `boolean`.

I think that gives the flexibility to folks who need it.

I'm also happy to tweak the `goog.cacheableResponse.Behavior` implementation if it makes sense, but I think we already have a way to provide full flexibility for folks who need it.

jeffposnick:
Actually, I think I need to adjust this to allow opaque responses by
default, and then we can impose stricter requirements in the contexts when
it's needed, instead of the other way around.

gauntface:
I thought we agreed to make opaque responses invalid by default and require developer to opt in?

jeffposnick:
That's entirely possible regarding the opaque response decision. I trust your memory better than mine. I'll leave as-is.

packages/sw-runtime-caching/src/lib/request-wrapper.js
```diff
@@ -62,23 +62,30 @@ class RequestWrapper {
...
+            this.behaviorCallbacks[callbackName].push(
               behavior[callbackName].bind(behavior));
           }
         }
       });
     }
+
+    if (this.behaviorCallbacks.cacheWillUpdate) {
+      if (this.behaviorCallbacks.cacheWillUpdate.length !== 1) {
+        throw Error('You cannot register more than one behavior that ' +
```
gauntface: Could you move this to use an ErrorFactory?

gauntface:
@jeffposnick All in I'm happy with this only comment is to move to the ErrorFactory if possible.

packages/sw-runtime-caching/src/lib/request-wrapper.js
jeffposnick: Sure thing. Still getting used to that pattern.

jeffposnick:
As per https://github.com/GoogleChrome/sw-helpers/pull/90#issuecomment-267649270, we should be good to merge.

PR #91, opened 2016-12-17
Warm runtime

gauntface:
R: @jeffposnick @addyosmani 

This delightful little wonder introduces caching of un-revisioned caching. To make use of a big chunk of the code from the "revisioned" caching I had to move quite a few things around but the core logic is the same.

@jeffposnick I tried to move to the RequestWrapper, but due to caching being done async from the function being completed it's cause issues with my tests.


addyosmani:
>This delightful little wonder introduces caching of un-revisioned caching

Caching of unrevisioned assets? :)

packages/sw-precaching/src/lib/controllers/base-manager.js
```diff
@@ -0,0 +1,171 @@
...
+class BaseCacheManager {
+  constructor() {
+    this._entriesToCache = {};
+  }
+
+  cache(rawEntries) {
+    rawEntries.forEach((rawEntry) => {
+      const precacheEntry = this._parseEntry(rawEntry);
+      this._addEntryToInstallList(precacheEntry);
+    });
```
addyosmani: Is there any reason to not directly pass `this._parseEntry(rawEntry)` to `_addEntryToInstallList` here?

packages/sw-precaching/src/lib/controllers/base-manager.js
```diff
@@ -0,0 +1,171 @@
...
+      })
+    );
+  }
+
+  _onDuplicateEntryFound(newEntry, previous) {
+    throw new Error('This should be overriden by extending class.');
+  }
+
+  _getCache() {
+    throw new Error('This should be overriden by extending class.');
```
addyosmani: overridden

packages/sw-precaching/src/lib/controllers/base-manager.js
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  _isAlreadyCached(precacheEntry, openCache) {
+    throw new Error('This should be overriden by extending class.');
+  }
+
+  _onEntryCached(precacheEntry) {
+    // NOOP
+  }
```
addyosmani: Is there a reason why we're nooping here?

packages/sw-precaching/src/lib/precache-manager.js
```diff
@@ -0,0 +1,94 @@
...
+      // Only need to register events once.
+      return;
+    }
+
+    this._eventsRegistered = true;
+
+    self.addEventListener('install', (event) => {
+      const promiseChain = Promise.all([
+        this._revisionedManager._performInstallStep(),
+        // this._unrevisionedManager._performInstallStep(),
```
addyosmani: Drop commented out line?

packages/sw-precaching/src/lib/controllers/base-manager.js
gauntface: This can throw and require both extending classes implement this (one of which will be a noop and one will be doing same as it is currently) or we can NOOP here and leave one class implementing and the other not.

Happy to change if you feel strongly.

packages/sw-precaching/src/lib/controllers/unrevisioned-manager.js
```diff
@@ -0,0 +1,62 @@
...
+      throw ErrorFactory.createError('invalid-unrevisioned-entry',
+        new Error('Invalid file entry: ' +
+          JSON.stringify(precacheEntry)));
+    }
+
+    return precacheEntry;
+  }
+
+  _onDuplicateEntryFound(newEntry, previous) {
+    // NOOP. Just ignore duplication entries.
```
jeffposnick: `duplication` ➡️  `duplicate`

packages/sw-precaching/src/lib/controllers/unrevisioned-manager.js
```diff
@@ -0,0 +1,62 @@
...
+   * the cache entry exists for the `fileEntry.path` this method returns true.
+   * False otherwise.
+   * @param {Object} fileEntry A file entry with `path` and `revision`
+   * parameters.
+   * @param {Cache} openCache The cache to look for the asset in.
+   * @return {Promise<Boolean>} Returns true is the fileEntry is already
+   * cached, false otherwise.
+   */
+  async _isAlreadyCached(fileEntry, openCache) {
+    return false;
```
jeffposnick: I'm assuming this is just a placeholder—can you put in a `TODO:` comment or something along those lines making that clear?

packages/sw-precaching/src/lib/models/precache-entries/base-precache-entry.js
```diff
@@ -1,14 +1,14 @@
-import {cacheBustParamName} from '../constants';
+import {cacheBustParamName} from '../../constants';
 
 class RevisionedCacheEntry {
-  constructor({revisionID, revision, request, cacheBust}) {
-    this.revisionID = revisionID;
+  constructor({entryID, revision, request, cacheBust}) {
+    this.entryID = entryID;
```
jeffposnick: It could be helpful to use the `assert` library to validate the parameters if there are any that are required and/or need to be of a specific type.

packages/sw-precaching/src/lib/models/precache-entries/object-precache-entry.js
```diff
@@ -1,43 +1,43 @@
-import ErrorFactory from '../error-factory';
-import RevisionedCacheEntry from './revisioned-cache-entry.js';
+import ErrorFactory from '../../error-factory';
+import RevisionedCacheEntry from './base-precache-entry.js';
 
 class DefaultsCacheEntry extends RevisionedCacheEntry {
-  constructor({revisionID, revision, url, cacheBust}) {
+  constructor({entryID, revision, url, cacheBust}) {
```
jeffposnick: Could I convince you to move to the `assert` library for these checks?

I'm happy enough with explicitly using `ErrorFactory` for more complex error reporting, but `assert` was created and is used elsewhere for basic parameter validation.

In addition to being consistent with the other modules, using `assert` would make it easier to toggle on this sort of validation during development and disable it in a production deployment. (Assuming you think that's a good idea—chime in on https://github.com/GoogleChrome/sw-helpers/issues/93)

packages/sw-precaching/src/lib/models/precache-entries/object-precache-entry.js
```diff
@@ -1,43 +1,43 @@
...
-    // Check revisionID type
-    if (typeof revisionID === 'undefined') {
-      revisionID = new URL(url, location.origin).toString();
-    } else if (typeof revisionID !== 'string' || revisionID.length === 0) {
-      throw ErrorFactory.createError('invalid-file-manifest-entry',
-        new Error('Bad revisionID Parameter. It should be a string with at ' +
-          'least one character: ' + JSON.stringify(revisionID)));
+    // Check entryID type
+    if (typeof entryID === 'undefined') {
+      entryID = new URL(url, location.origin).toString();
```
jeffposnick: This shouldn't be`location.origin`, which would cause all relative URLs to be resolved with `/` as the base.

Using `location` would match what `sw-precache` [currently does](https://github.com/GoogleChrome/sw-precache/blob/master/service-worker.tmpl#L56), i.e. use the SW's script location as the base.

There's an argument to be made that using `registration.scope` as the base would make more sense than using the SW script location. I could see that as being more intuitive. (`self.registration.scope` wasn't implemented when `sw-precache` first came out, so it's a historical accident that it's not used.) But it could make it harder to transition from using `sw-precache` to using the new library for developers who also set custom scopes.

Thoughts?

jeffposnick:
Happy to chat off-thread about what you're running into that prevents the use of `RequestWrapper` here.

packages/sw-precaching/src/lib/controllers/base-manager.js
```diff
@@ -0,0 +1,145 @@
...
+   * cached.
+   */
+  async _performInstallStep(cacheName) {
+    if (Object.keys(this._entriesToCache).length === 0) {
+      return;
+    }
+
+    let openCache = await this._getCache();
+    const entriesToCache = Object.values(this._entriesToCache);
+    const cachePromises = entriesToCache.map(async (precacheEntry) => {
```
gauntface: Fixed.

packages/sw-precaching/src/lib/controllers/base-manager.js
```diff
@@ -0,0 +1,145 @@
...
+   * to be deleted from the cache, this method will compare the URL's
+   * and figure out which assets are no longer required to be precached.
+   */
+  async _cleanUpOldEntries() {
+    if (!await caches.has(this._cacheName)) {
+      // Cache doesn't exist, so nothing to delete
+      return;
+    }
+
+    const requestsCachedOnInstall = Object.values(this._entriesToCache)
```
gauntface: Moved to Map. Thanks for the info.

packages/sw-precaching/src/lib/controllers/base-manager.js
```diff
@@ -0,0 +1,145 @@
...
+        return openCache.delete(cachedRequest);
+      })
+    );
+  }
+
+  /**
+   * A simple helper method to get the cache used for precaching assets.
+   * @return {Cache} The cache to be used for precaching.
+   */
+  _getCache() {
```
gauntface: Done.

packages/sw-precaching/src/lib/controllers/revisioned-manager.js
```diff
@@ -0,0 +1,102 @@
...
+   */
+  _parseEntry(input) {
+    if (typeof input === 'undefined' || input === null) {
+      throw ErrorFactory.createError('invalid-revisioned-entry',
+        new Error('Invalid file entry: ' + JSON.stringify(input)));
+    }
+
+    let precacheEntry;
+    switch(typeof input) {
+      case 'string': {
```
gauntface: Done.

packages/sw-precaching/src/lib/controllers/unrevisioned-manager.js
gauntface: Done.

packages/sw-precaching/src/lib/controllers/unrevisioned-manager.js
gauntface: My thinking for the unrevisioned manager is that we have no way of knowing if the current entry is up to date and as a result caching it on a new service worker install would make sense (I believe this is was sw-toolbox does).

If you think thats a bad idea and unrevisioned assets should only be re-cached at run time then I can change this, but my gut says that it'll cause more issues than it'll solve.

packages/sw-precaching/src/lib/models/precache-entries/object-precache-entry.js
gauntface: Can you give me an example of where location.origin vs location breaks relative URLs? (Just so I can add a test to capture this edge case)

Scope feels like it would be very error prone, but being relative to location of service worker makes most sense to me. (i.e. the scope could be different to the location of the service worker).

jeffposnick: Assume `location` is `https://example.com/path/to/sw.js`.
Assume `url` is `assets/image.png`.

If you use `location.origin` as the base, then `entryID` will end up being set to `https://example.com/assets/image.png`. That's *probably* not what the developer wants when they specify a relative URL.

If you use `location` as the base, then `entryID` will end up being set to `https://example.com/path/to/assets/image.png`. That's presumably what the developer wants.

I'm cool with using `location` as the base and not the SW scope, for consistency with `sw-precache`.

packages/sw-precaching/src/lib/controllers/unrevisioned-manager.js
jeffposnick: Okay, I was confused by the comments JSDoc and the fact that the method signature takes in parameters that are never used. Can you just update those to make it clear that it's just a method that ignores the input and always returns `false`?

packages/sw-precaching/src/lib/controllers/base-manager.js
```diff
@@ -0,0 +1,145 @@
...
+    return caches.open(this._cacheName);
+  }
+
+  _parseEntry(input) {
+    throw new Error('_parseEntry should be overriden by extending class.');
```
gauntface: Done.

gauntface: Done

packages/sw-precaching/src/lib/models/precache-entries/base-precache-entry.js
jeffposnick: That's fine if the checks are done elsewhere.

packages/sw-precaching/src/lib/models/precache-entries/object-precache-entry.js
gauntface: awesome. I've added in tests for this. But my understanding is that relative path mean it'll request assets by the service worker scope, but rather request them based on the service workers location.

i.e. register('/sw.js', {scope: '/news/})

Would be asset 'image.png' would be '/image.png' and not '/news/image.png'.

Is that right or is this some edge case behavior?

gauntface:
@jeffposnick @addyosmani All of the raised issues have been fixed, could you take another look.

packages/sw-precaching/src/lib/models/precache-entries/object-precache-entry.js
jeffposnick: You're correct in summing out what the behavior would be.

It's really a toss-up as to whether the SW location or the scope is a better base to use when there's a relative URL.

This is only an issue when a user explicitly changes the default scope (which usually isn't needed), and can be remediated by using absolute URLs for the assets, so I'm not too worried about it.

jeffposnick:
LGTM. I don't know if @addyosmani has any more feedback, but otherwise, feel free to merge.

We should still chat about what changes could be made to `RequestWrapper` to get it usable here, but that doesn't need to block this PR.

PR #94, opened 2017-01-03
Adding tests for file extension selection

### LINKED ISSUE #82: SW-CLI Usage
@jeffposnick @addyosmani 

Sorry I haven't covered this topic sooner - it's only lightly covered in the sw-goog doc.

I've started to consider the scenarios of the CLI usage. If we can cover the primary use cases in the CLI, I'm confident the right API can be exposed as a module for gulp and grunt etc.

In the doc we outline the main use cases as generating a service worker and building a file manifest.

## Generating a Service Worker

The user journeys of this are likely to be:

1. First time use of the CLI (either first use by the user or first time using for a new project).
2. Repeat use of the CLI (manually running to fix or make changes).
3. Automated running of CLI on a CI like Travis.

### First Use - `sw-cli generate-sw`

For the first use we can use an interactive CLI that guides the developer somewhat, asking them for:
1. Ask for the root directory.
1. Ask them to confirm the types of files they wish to cache (i.e. *.html, *.png, *.jpg) etc - this will be based on the files found in the directory.
1. Ask them what the manifest file should be called (default: precache-manifest.json).
1. Ask them what the service worker file should be called (default: sw.js).
1. Ask if we should save this to a config file sw-cli-config.json (Y / n).

The result of this is a service worker and file manifest. If the user selects to generate a config file, it'll also be generated and the config will include a more extensive set of config options (i.e. options that are excluded from the interactive CLI)..

### Second Use - `sw-cli generate-sw`

The second flow would go through the same flow:

1.) If a config file is found with name sw-cli-config.json then we'll ask the developer if they wish to use it.
2.) If not found or used, the above flow will be taken
3.) We will ask the developer if they wish to override the service worker file and / or config file if required.

### CI Use - `sw-cli generate-sw --non-interactive`

This mode will require the default config file to exist on the current working directory OR expect the '--config' flag to be set. Overwriting tan existing service worker file will be blocked by default but --force can be used to force it (I'd expect developers to clean the directory before overwriting).

## Build File Manifest

The manifest can use the same config file from the previous step and could follow the same interactive flow:

1. Ask for the root directory.
1. Ask them to confirm the types of files they wish to cache (i.e. *.html, *.png, *.jpg) etc - this will be based on the files found in the directory.
1. Ask them what the manifest file should be called (default: precache-manifest.json).
1. Ask if we should save this to a config file sw-cli-config.json (Y / n).

The second use and CI use would follow the same pattern as above. The only difference is any questions regarding SW are skipped.

Does this seem like a reasonable first approach?

gauntface:
R: @jeffposnick @addyosmani

Fixes #82

This adds in the ability to select file extensions that the developer would wish to cache under the current directory.

This may need changed if developers aren't a fan of it, but I think it matches precache in terms of just working (everything is selected by default) but adds a nice level of flexibility to cut out unexpected files.

### DIFF
--- a/packages/sw-cli/src/cli/index.js
+++ b/packages/sw-cli/src/cli/index.js
@@ -26,6 +26,8 @@ const logHelper = require('../lib/log-helper');
 const pkg = require('../../package.json');
 const constants = require('../lib/constants.js');
 
+const DEBUG = false;
+
 /**
  * This class is a wrapper to make test easier. This is used by
  * ./bin/index.js to pass in the args when the CLI is used.
@@ -126,17 +128,12 @@ class SWCli {
   generateSW() {
     return this._getRootOfWebApp()
     .then((rootDirectory) => {
+      return this._getFileExtensionsToCache(rootDirectory);
+    })
+    .then((fileExtensionsToCache) => {
 
     });
     /** return inquirer.prompt([
-      {
-        name: 'rootDir',
-        message: 'What is the root of your web app?',
-        validate: () => {
-          // TODO: Validate user input
-          return true;
-        },
-      },
       {
         name: 'fileExtensions',
         message: 'Which file types would you like to cache?',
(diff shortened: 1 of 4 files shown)


packages/sw-cli/src/cli/index.js
```diff
@@ -230,6 +227,91 @@ class SWCli {
...
+      const promises = directoryContents.map((directoryContent) => {
+        const fullPath = path.join(directory, directoryContent);
+        if (fs.statSync(fullPath).isDirectory()) {
+          if (!constants.blacklistDirectoryNames.includes(directoryContent)) {
+            return this._getFileContents(fullPath);
+          } else {
+            return [];
+          }
+        } else {
+          return Promise.resolve(fullPath);
```
jeffposnick: Nit: I don't think the `Promise.resolve()` wrapper is needed.

packages/sw-cli/src/cli/index.js
```diff
@@ -230,6 +227,91 @@ class SWCli {
...
+    })
+    .then((fileResults) => {
+      return fileResults.reduce((collapsedFiles, fileResult) => {
+        return collapsedFiles.concat(fileResult);
+      }, []);
+    });
+  }
+
+  _getFileExtensions(files) {
+    const fileExtensions = {};
```
jeffposnick: What do you think about this approach, using ES2015 collections?

```
const fileExtensions = new Set(files.map(path.extname));

if (DEBUG && fileExtensions.has('')) {
  // log error
}

return fileExtensions;
// If you need to return an actual array, then use
// return [...fileExtensions]
```

gauntface: The debug log isn't a thrown error, it's just a warning log message and if I do the set I'll love the information regarding which file would be excluded from the preache (it's mostly files like .eslintrc or LICENSE).

Personally I can see the log being useful, but if you disgree I'm happy to use Set and then remove the empty string from the set.

jeffposnick: I've got no objections to the logging. The potential to use `Set` (and then removing the `''` entry) was the main point I was trying to get across.

gauntface: I get that, but rather than "just lose '' " With this approach I keep the relationship between a file and the result extension '', meaning the log can say - "unable to cache  file 'LICENSE'" which while not useful for that example, there may be scenarios where that log can save some head scratching.

jeffposnick: Ah, gotcha. Then feel free to go ahead with the iterating-over-each-item approach, though you might still benefit from using a `Set` instead of an `Object` with dummy `true` values.

It's not a big deal either way.

gauntface: Good point. Moved over to Set for that. Happy to merge if tests are all ok?

PR #95, opened 2017-01-03
Explicitly set `redirect: follow` when precaching

jeffposnick:
R: @addyosmani @gauntface

See https://github.com/GoogleChrome/sw-precache/pull/233

The same thing should apply here.

### DIFF
--- a/packages/sw-precaching/src/lib/revisioned-cache-manager.js
+++ b/packages/sw-precaching/src/lib/revisioned-cache-manager.js
@@ -256,6 +256,7 @@ class RevisionedCacheManager {
 
     let response = await fetch(fileEntry.cacheBustRequest, {
         credentials: 'same-origin',
+        redirect: 'follow',
       });
 
     if (response.ok) {


gauntface:
@jeffposnick Would you be willing to write a test to confirm this behaviour doesn't revert in the future?

gauntface:
Going to merge and write one as I have some extra time :)

jeffposnick:
I am going to write a test one of these days. I swear!

srosset81:
@jeffposnick for requests other than precaching, I guess we need to manually add the `{ redirect: 'follow' }` option in order to avoid the "network protocol violation". We have a few users reporting this error and I think it must be related to this problem.

jeffposnick:
Hello @srosset81—For requests other than precaching, you can set the fetch options when configuring the strategy, as needed.

E.g. (using the v3 syntax):

```js
  new RegExp('/path/prefix'),
  workbox.strategies.networkFirst({
    cacheName: 'my-cache',
    fetchOptions: {
      redirect: 'follow',
    }
  })
);
```

Though as per https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch#Parameters, the default should be `'follow'`.

If you're running into trouble, and need help debugging, please open a fresh issue with reproduction details.

PR #96, opened 2017-01-04
Adding test for redirects in sw-precaching

gauntface:
R: @jeffposnick @addyosmani 

This adds redirects for testing the sw-precache.

### DIFF
--- a/packages/sw-precaching/test/browser-unit/cache-testing.js
+++ b/packages/sw-precaching/test/browser-unit/cache-testing.js
@@ -319,7 +319,7 @@ describe('sw-precaching Test Revisioned Caching', function() {
   it('should manage redirected revisioned requests', function() {
     return window.goog.swUtils.activateSW('data/response-types/redirect-revisioned-sw.js')
     .then((iframe) => {
-      goog.__TEST_DATA['redirect'].map((redirectPath) => {
+      const promises = goog.__TEST_DATA['redirect'].map((redirectPath) => {
         const sections = redirectPath.split('/').filter((section) => {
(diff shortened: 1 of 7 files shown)


gauntface:
Note: The `redirect: 'follow'` doesn't seem to add anything in terms of browser support for browsers today. Still worth having for older browsers, but it would be good to confirm it actually has an affect on a browser somewhere.

packages/sw-precaching/test/browser-unit/cache-testing.js
```diff
@@ -306,13 +306,61 @@ describe('sw-precaching Test Revisioned Caching', function() {
...
     }, (err) => {
       // NOOP - The error is not to do with the error throw in SW, so nothing
       // to check.
     });
   });
+
+  it('should manage redirected revisioned requests', function() {
+    return window.goog.swUtils.activateSW('data/response-types/redirect-revisioned-sw.js')
+    .then((iframe) => {
+      goog.__TEST_DATA['redirect'].map((redirectPath) => {
```
jeffposnick: It looks like you're missing a `return`, since there isn't an implicit `return` when an arrow function has a block as its body.

Assuming that this wasn't intentional, could you check other tests that might also follow a similar pattern and add `returns` there as well? I see that line 346 in this file, at least, also seems to be missing it.

gauntface: Good catch - thank you.

Fixed in the latest commit.

gauntface:
@jeffposnick happy to merge this?

PR #101, opened 2017-01-05
Docs for sw-precaching.

gauntface:
R: @jeffposnick @addyosmani 

I did some tidy up in the process, but all minor tweaks.


jeffposnick:
It's a little tough to find all the new bits in GitHub's web UI, since some files were moved around, but from what I can tell, this is 👍

PR #105, opened 2017-01-06
Adds a waitOnCache parameter to fetchAndCache()

### LINKED ISSUE #99: Add a flag to RequestWrapper to wait on cache operations
This would make it possible for @gauntface to use `RequestWrapper` inside of the precaching code, as well as make it easier to write tests that depend on entries being written to the cache.

jeffposnick:
R: @addyosmani @gauntface

Fixes #99

This adds a new `waitOnCache` option that can be passed to `RequestWrapper#fetchAndCache()`. I *think* that's enough to allow @gauntface to start using `RequestWrapper` inside the precaching code, which would be a nice thing to do to ensure that all fetches/caching goes through a common code path.

If there's anything else that's needed to make `RequestWrapper` usable, @gauntface, let me know and I'll adjust this PR.

This also fixes up some paths in the demo code, which is unfortunately how I'm testing the changes. (Real tests are forthcoming!)

### DIFF
--- a/packages/sw-runtime-caching/demo/service-worker.js
+++ b/packages/sw-runtime-caching/demo/service-worker.js
@@ -2,9 +2,9 @@
 /* global goog */
 
 importScripts(
-  '../build/routing.min.js',
-  '../../sw-runtime-caching/build/runtime-caching.min.js',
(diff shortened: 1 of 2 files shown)


gauntface:
LGTM from me. This should mean I can switch over.

PR #106, opened 2017-01-06
Add file manifest creation.

gauntface:
R: @jeffposnick @addyosmani 

This PR is the start of building the file manifest.

At the moment it's missing quite a few bits / features that sw-precache has, but this PR was getting pretty big so I figured it's worth getting eyes on it as is and will follow your guidance on how to proceed (i.e. accept this PR as is or re-open with more full implementation.


packages/sw-cli/src/cli/index.js
```diff
@@ -309,7 +322,12 @@ class SWCli {
         );
       }
     });
-    return [...fileExtensions];
+    return [...fileExtensions].map((fileExtension) => {
```
jeffposnick: Optional, slightly fancier:

```js
return [...fileExtensions].map((fileExtension) => fileExtension.replace(/^\./, '');
```

packages/sw-cli/src/cli/index.js
```diff
@@ -370,14 +388,164 @@ class SWCli {
...
+    return inquirer.prompt([
+      {
+        name: 'saveConfig',
+        message: 'Last Question - Would you like to save these settings to ' +
+          'a config file?',
+        type: 'confirm',
+        default: true,
+      },
+    ])
+    .then((results) => {
```
jeffposnick: Again, slightly fancier, to take advantage of the implicit single-line arrow function return:

```js
.then((results) => results.saveConfig)
```

packages/sw-cli/src/cli/index.js
```diff
@@ -370,14 +388,164 @@ class SWCli {
...
+        errors['unable-to-get-save-config'],
+        err
+      );
+      throw err;
+    });
+  }
+
+  _buildFileManifestFromGlobs(manifestFilePath, globs) {
+    let globbedFiles = [];
+    globs.forEach((globPattern) => {
```
jeffposnick: I sometimes hesitate to use `reduce()`, since they can lead to inscrutable code, but I'm just throwing this out there because this is the ideal use case:

```js
const globbedFiles = globs.reduce((accumulated, globPattern) => {
  const files = this._getFileManifestDetails(globPattern);
  return accumulated.concat(files);
}, []);
```

Or, this might work too, though I'm not 100% sure it will flatten the arrays in the way you want it:

```js
const globbedFiles = [...globs.map((globPattern) => this._getFileManifestDetails(globPattern))];
```

Or, you know, keep your `forEach()` if you'd prefer.

packages/sw-cli/src/cli/index.js
```diff
@@ -370,14 +388,164 @@ class SWCli {
...
+    try {
+      mkdirp.sync(path.dirname(manifestFilePath));
+    } catch (err) {
+      logHelper.error(errors['unable-to-make-manifest-directory'], err);
+      return Promise.reject(err);
+    }
+
+    return new Promise((resolve, reject) => {
+      fs.readFile(
+        path.join(__dirname, '../lib/templates/file-manifest.js.tmpl'), 'utf8',
```
jeffposnick: You should pass in each of the components of the path to `path.join()` so that it has the intended effect:

```js
path.join(__dirname, '..', 'lib', 'templates', 'file-manifest.js.tmpl')
```

packages/sw-cli/src/cli/index.js
```diff
@@ -370,14 +388,164 @@ class SWCli {
...
+        return template(templateString)({
+          manifestEntries: manifestEntries,
+        });
+      } catch (err) {
+        logHelper.error(errors['populating-manifest-tmpl-failed'], err);
+        throw err;
+      }
+    })
+    .then((populatedTemplate) => {
+      return new Promise((resolve, reject) => {
```
jeffposnick: Since you're already inside a promise chain, I think you could get away without this `Promise` wrapper. Just `return Promise.reject(err)` if needed inside the block.

packages/sw-cli/src/cli/index.js
```diff
@@ -370,14 +388,164 @@ class SWCli {
...
+        });
+      });
+    });
+  }
+
+  _filterFiles(files) {
+    // Filter oversize files.
+    files = files.filter((fileDetails) => {
+      if (fileDetails.size > constants.maximumFileSize) {
+        logHelper.warn(`Skipping file '${fileDetails.fil}' due to size. ` +
```
jeffposnick: `fileDetails.fil` ➡️  `fileDetails.file`

packages/sw-cli/src/cli/index.js
```diff
@@ -370,14 +388,164 @@ class SWCli {
...
+
+      const fileHash = this._getFileHash(file);
+      return {
+        file,
+        hash: fileHash,
+        size: fileSize,
+      };
+    });
+
+    return fileDetails.filter((details) => {
```
jeffposnick: Another case where collapsing the block into a single-line arrow function would be more concise.

packages/sw-cli/src/lib/templates/file-manifest.js.tmpl
```diff
@@ -0,0 +1,6 @@
+self.__file_manifest = [<% manifestEntries.forEach((manifestEntry) => { %>
```
jeffposnick: Can this be simplified by using `JSON.stringify(manifestEntries)`?

packages/sw-cli/src/cli/index.js
```diff
@@ -146,22 +151,30 @@ class SWCli {
...
+      logHelper.warn('File Extensions to Cache: ' + fileExtentionsToCache);
+      logHelper.warn('File Manifest: ' + fileManifestName);
+      logHelper.warn('Service Worker: ' + serviceWorkerName);
+      logHelper.warn('Save to Config File: ' + saveConfig);
+      logHelper.warn('');
+
+      const relativePath = path.relative(process.cwd(), rootDirectory);
+
+      const globs = [
+        path.join(relativePath, '**', '*') +
```
gauntface: I'm reluctant to drop the relativePath from the glob pattern and because of that I would stick with path.join BUT I've swapped it with your `.replace(path.sep, '/')` approach in sw-precache.

packages/sw-cli/src/cli/index.js
gauntface: Noice.

packages/sw-cli/src/cli/index.js
gauntface: Done.

packages/sw-cli/src/cli/index.js
gauntface: Gone for your first proposal ;)

packages/sw-cli/src/cli/index.js
gauntface: Done.

packages/sw-cli/src/cli/index.js
gauntface: fs.writeFile is async, so the outer promise is needed to block the promise chain until the file is written.

packages/sw-cli/src/cli/index.js
gauntface: ta

packages/sw-cli/src/cli/index.js
gauntface: done.

packages/sw-cli/src/lib/templates/file-manifest.js.tmpl
gauntface: why yes it can.

gauntface:
@jeffposnick happy for this to land?

PR #108, opened 2017-01-06
Sw lib caching strategies

gauntface:
R: @jeffposnick @addyosmani 

This adds caching strategies to sw-lib as well as adding some docs.


packages/sw-lib/src/index.js
```diff
@@ -26,6 +26,33 @@ if (!assert.isSWEnv()) {
...
+/**
+ * The sw-lib module is a high-level library that makes it easier to
+ * configure routes with caching strategies as well as manage precaching
+ * of assets during the install step of a service worker.
+ *
+ * @example
+ * importScripts('/<Path to Module>/build/sw-lib.min.js');
+ *
+ * goog.swlib.cacheRevisionedAssets([
+ *     '/styles/main.1234.css',
```
addyosmani: I think what you're trying to show here is that `cacheRevisionedAssets` accepts both filename paths with revisions in there or an object with the raw path and a separate revision number. Is that the case? 

It might be worth splitting these into two examples (one showing caching a few files with one approach (e.g `/styles/main.1234.css', '/js/main.1234.js' etc) and then the object variation. Might make it more easy to read through.

packages/sw-lib/src/index.js
```diff
@@ -26,6 +26,33 @@ if (!assert.isSWEnv()) {
...
+ *       url: '/index.html',
+ *       revision: '1234'
+ *     }
+ * ]);
+ *
+ * goog.swlib.warmRuntimeCache([
+ *     '/scripts/main.js',
+ *     new Request('/images/logo.png')
+ * ]);
+ *
```
addyosmani: Same comment as above.

packages/sw-lib/test/browser-unit/sw-unit/caching-strategies.js
```diff
@@ -0,0 +1,44 @@
...
+
+const expect = self.chai.expect;
+self.chai.should();
+mocha.setup({
+  ui: 'bdd',
+  reporter: null,
+});
+
+describe('Test swlib.cacheOnly', function() {
+  it('should be accessible goog.swlib.cacheOnly', function() {
```
addyosmani: We talked about why these functions aren't just using arrow syntax offline. I think your reason about Mocha breaking decorated `it` bindings etc if arrows are used was fine. I did see https://github.com/skozin/arrow-mocha in case we wanted to explore how to work around this in future.

packages/sw-lib/src/index.js
gauntface: Done.

packages/sw-lib/src/index.js
gauntface: Done.

packages/sw-lib/test/browser-unit/sw-unit/caching-strategies.js
gauntface: I'm probably going to stick with the function name for now. Ideally Mocha would change to have a more explicit approach of declaring the timeout and retries of a unit test.

gauntface:
@addyosmani happy for this to land?

jeffposnick:
I'm taking a look right now, too.

packages/sw-lib/src/index.js
```diff
@@ -26,6 +26,54 @@ if (!assert.isSWEnv()) {
...
+ *       revision: '1234'
+ *     },
+ *     {
+ *       url: '/about.html',
+ *       revision: 'abcd'
+ *     }
+ * ]);
+ *
+ * // If you have assets that aren't revisioned, you can cache them
+ * // during the installation of you service work using warmRuntimeCache()
```
jeffposnick: "service work" ➡️  "service worker"

packages/sw-lib/src/index.js
```diff
@@ -26,6 +26,54 @@ if (!assert.isSWEnv()) {
...
+ *       url: '/index.html',
+ *       revision: '1234'
+ *     },
+ *     {
+ *       url: '/about.html',
+ *       revision: 'abcd'
+ *     }
+ * ]);
+ *
+ * // If you have assets that aren't revisioned, you can cache them
```
jeffposnick: Suggested rephrasing:

> If you have assets that aren't revisioned, but which you plan to reference using a [runtime caching strategy](https://developers.google.com/web/fundamentals/instant-and-offline/offline-cookbook/#serving-suggestions), you can add them to your runtime cache

packages/sw-lib/src/lib/sw-lib.js
```diff
@@ -17,36 +17,51 @@
...
   /**
-   * Initialises the classes router wrapper.
+   * Initialises an instance of SWLib. An instance of this class is
+   * accessible when the module is imported into a service worker as
+   * `self.goog.swlib`.
    */
   constructor() {
     this._router = new RouterWrapper();
     this._precacheManager = new PrecacheManager();
+    this._strategies = {
```
jeffposnick: When we talked about this last week, I misunderstood and thought that your plan was to export the class definitions—basically, aliasing them under the `sw-lib` module namespace. I didn't realize that your suggestion was to create instances of each class and include those in each `SWLib` instance.

These strategies are most useful when you have some control over their behavior via the `RequestWrapper` class, but there's no way to configure these instances. The default behavior will work, but in the back of my head I worry that people won't realize that they can do things like change the cache name or configure cache expiration if they're only familiar with using these unconfigured instances.

Do you all share those concerns?

packages/sw-lib/src/lib/sw-lib.js
```diff
@@ -60,10 +75,18 @@ class SWLib {
   }
 
   /**
-   * If there are assets that should be cached on the install step but
-   * aren't revisioned, you can cache them here.
-   * @param {Array<String>} unrevisionedFiles A set of urls to cache when the
-   * service worker is installed.
+   * Any assets you wish to cache which can't be revisioned should be
```
jeffposnick: How about:

> Any assets you wish to cache **ahead of time** which can't be revisioned should be...

Since in many of the cases for unversioned resources, just populating the caches via a runtime caching strategy is desirable.

packages/sw-lib/src/index.js
gauntface: Fixed.

packages/sw-lib/src/lib/sw-lib.js
gauntface: Yes and no.

This API matches that of `sw-toolbox`, that's all I was aiming for with this as it does cover a simple use case.

There are three 1 options:

1. This approach and expect developers to use the smaller modules is desired (not ideal).
2. This approach + expose the classes as well, i.e. allow `new goog.swlib.strategies.<Name of Strategy>()`. Simple use case plus allows configuration.
3. Just include the class strategies. Only con is that is requires some boilerplate for basic use.

packages/sw-lib/src/lib/sw-lib.js
gauntface: done.

packages/sw-lib/src/index.js
gauntface: At the moment the precaching module doesn't configure fetch routes, so technically both require "run time caching strategies".

packages/sw-lib/src/lib/sw-lib.js
jeffposnick: The `sw-toolbox`-y way of doing this was to pass in the configuration as the third parameter.

I guess we could take inspiration from that and allow developers to pass in a `RequestWrapper` as an optional third parameter.

For that approach to work, the second parameter should be the an uninstantiated reference to a `sw-runtime-caching` class, and then `SWLib` would construct the class at runtime, passing along the `RequestWrapper` to the constructor if it's provided.

gauntface: In toolbox I would always go:

    toolbox.router.all('/', toolbox.fastest);

The options generally would be global rather than per route.

Would it be worth me removing the strategies from this PR, landing the doc changes and then discussing the strategies in an issue?

jeffposnick: Yep, yanking them out of this PR and discussing the approach off-thread would make sense.

There are lots of common use cases for per-route configs that we'd need to support, like one route that used a dedicated `image` cache with a specific cache expiration policy, and a different route that used a dedicated `api-response` cache with a different policy.

gauntface:
@jeffposnick the strategies are now removed - any further changes you'd like to comments?

jeffposnick:
There's a few extra bits to remove to make the linting happy, but once Travis is green, feel free to merge.

```
/home/travis/build/GoogleChrome/sw-helpers/packages/sw-lib/src/lib/sw-lib.js
  22:3   error  'CacheFirst' is defined but never used            no-unused-vars
  22:15  error  'CacheOnly' is defined but never used             no-unused-vars
  22:26  error  'NetworkFirst' is defined but never used          no-unused-vars
  22:40  error  'NetworkOnly' is defined but never used           no-unused-vars
  22:53  error  'StaleWhileRevalidate' is defined but never used  no-unused-vars
```

PR #114, opened 2017-01-10
Clean up the parameters that are passed to the handler

### LINKED ISSUE #107: RegexpRoute and ExpressRoute Improving Params Output.
**Library Affected**: *sw-routing*

**Browser & Platform**: All

At the moment the express and regex routes will pass in an object into the handle method with a 'params' value.

With a regex route of `const regexRoute = /\/1234567890\/(\w+)\//;` I'd expect to be able to access the capturing group via params, which we can do, but it's not labelled clearly.

The above would result in:

    (args) => {
      args.params[0] -----> "/1234567890/test/"
      args.params[1] -----> "test"
      args.params.index -----> 21
      args.params.input -----> "http://localhost:3000/1234567890/test/"
    }

Given that the `args` variable also includes `args.url`, the params.input value is not useful and `params[0]` is not exactly useful either.

With express, this follows the same pattern.

For express route of `/:date/:id/test/`I would expect params to have a data and id value, but it's just index values with '0' being route and 1 being the date value and 2 being the id value.

    (args) => {
      args.params[0] -----> "/1234567890/some-id/test/"
      args.params[1] -----> 1234567890
      args.params[2] -----> "some-id"
      args.params.index -----> XX
      args.params.input -----> "/1234567890/some-id/test/"
    }

I'd expect:

    (args) => {
      args.params['date'] -----> "1234567890"
      args.params['id'] -----> "some-id"
    }

Also worth nothing that the 'date' value in this example is parsed as a number rather than a string. Not sure if we should be consistent with that or not.

jeffposnick:
R: @addyosmani @gauntface

Fixes #107 

This leads to less cruft being passed along to the handler's `params` when using the `RegExpRoute` or `ExpressRoute` classes along with capture groups/named parameters.

I've got some tests for this module in flight at https://github.com/GoogleChrome/sw-helpers/tree/sw-routing-tests and I'll add tests for this modified functionality as part of a PR for that branch.

### DIFF
--- a/packages/sw-routing/src/lib/express-route.js
+++ b/packages/sw-routing/src/lib/express-route.js
@@ -31,14 +31,42 @@ import pathToRegExp from 'path-to-regexp';
 class ExpressRoute extends Route {
   /**
    * @param {string} path The path to use for routing.
+   *        If the path contains [named parameters](https://github.com/pillarjs/path-to-regexp#named-parameters),
+   *        then an Object mapping parameter names to the corresponding value
+   *        will be passed to the handler via `params`.
    * @param {function} handler The handler to manage the response.
+   * @param {string} [method] Only match requests that use this
+   *        HTTP method. Defaults to `'GET'` if not specified.
    */
   constructor({path, handler, method}) {
     assert.isType({path}, 'string');
     assert.hasMethod({handler}, 'handle');
 
-    const regExp = pathToRegExp(path);
-    const match = ({url}) => url.pathname.match(regExp);
+    let keys = [];
+    // keys is populated as a side effect of pathToRegExp. This isn't the nicest
+    // API, but so it goes.
+    // https://github.com/pillarjs/path-to-regexp#usage
+    const regExp = pathToRegExp(path, keys);
+    const match = ({url}) => {
+      const regexpMatches = url.pathname.match(regExp);
+
+      // Return null immediately if this route doesn't match.
+      if (!regexpMatches) {
+        return null;
+      }
+
+      // If the route does match, then collect values for all the named
+      // parameters that were returned in keys.
+      // If there are no named parameters then this will end up returning {},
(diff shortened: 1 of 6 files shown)


packages/sw-routing/src/lib/express-route.js
```diff
@@ -31,14 +31,42 @@ import pathToRegExp from 'path-to-regexp';
...
+
+      // Return null immediately if this route doesn't match.
+      if (!regexpMatches) {
+        return null;
+      }
+
+      // If the route does match, then collect values for all the named
+      // parameters that were returned in keys.
+      // If there are no named parameters then this will end up returning {},
+      // which is truthy, and therefore a sufficient return value.
```
gauntface: What happens to the params value when there is no capture groups? Previously it was `true` which didn't feel like the right value given what it represents.

jeffposnick: As per the inline comment in the code:

> If there are no named parameters then this will end up returning {}...

That will result in `params` being set to `{}`, which seemed reasonable to me. Do you think a different value would make more sense? It needs to be something truthy.

gauntface: Personally I'd expect params to be undefined if the route defines no routes. The match method needs to be truthy, but the callback to the developer facing API could have this filtered out if the value is not an object, meaning you could return true and it would get parsed out.

jeffposnick: Okay, I've updated the code to behave the way you describe.

jeffposnick:
Travis is failing because the new `sw-lib` tests have landed in `master` while this PR has been in flight, and they need to be adjusted to account for the revised `params` behavior. I'll add a commit to this PR that cleans those up.

jeffposnick:
Merging now that Travis CI is happy.

It's always a good sign when tests break when interfaces change 😄

PR #116, opened 2017-01-10
Moving lint and build into test dependencies.

### LINKED ISSUE #115: gulp test should build project.
Running gulp test alone doesn't build the project, this can lead to old builds being tested :(

gauntface:
R: @jeffposnick @addyosmani 

Fixes #115

One problem I hit is that gulp-eslint seems to exit the gulp process early. I stumbled across a hack to make it work by adding gulp-debug to the pipe chain, but I'm not sure what the problem is.

But this change does mean testing will always be run against a new build and linting is now part of the test task.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,5 @@ before_script:
   - sleep 3 # give xvfb some time to start
 
 script:
-  - gulp lint --project "${PROJECT}"
-  - gulp build --project "${PROJECT}"
+  # Linting and building is handled in gulp test.
   - gulp test --project "${PROJECT}"
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,5 +28,6 @@ before_script:
   - sleep 3 # give xvfb some time to start
 
 script:
-  # Linting and building is handled in gulp test.
+  - gulp lint --project "${PROJECT}"
+  # Building is handled in gulp test.
   - gulp test --project "${PROJECT}"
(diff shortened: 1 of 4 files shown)


jeffposnick:
Regarding the `gulp-debug` bit, I'm diving into what's going on to try to isolate the source of breakage independent of this PR.

I have a minimal reproduction of the aborted `gulp` process when just running a `gulp lint project='sw-lib'` and `gulp lint project='sw-precaching'` against the current `master`, with all the other projects seeming to complete without issue.

I'm going to poke around a bit more to see what those two projects have in common w.r.t. linting. (I feel like this might be why, when I run `gulp lint` locally, it doesn't pick up some of the things that Travis CI picks up—`gulp lint` is also aborting early, while Travis CI now does per-project linting.)

I'd like to hold off on merging this PR until we can clean up whatever that issue is, and then we can hopefully omit the `gulp-debug` hack.

gauntface:
Would you accept this PR if I remove lint from the test task?

I specifically wanted build and test to work together, linting was a nice to have.

jeffposnick:
Yes, if you could remove the linting and the hacky bits, I'd be fine with this PR. I'll file a separate issue to investigate what's going on with the linting failures.

jeffposnick:
(See https://github.com/GoogleChrome/sw-helpers/issues/119)

jeffposnick:
👍

PR #117, opened 2017-01-10
Manifest e2e

gauntface:
R: @jeffposnick @addyosmani 

Fixes path to be relative to the project directory in the manifest and tests that the manifest is valid JavaScript.


packages/sw-cli/test/end-to-end.js
```diff
@@ -0,0 +1,105 @@
...
+      },
+    });
+
+    const cli = new SWCli();
+    return cli.handleCommand('generate-sw')
+    .then(() => {
+      const injectedSelf = {};
+      const manifestContent =
+        fs.readFileSync(path.join(exampleProject, manifestName));
+      vm.runInNewContext(manifestContent, {
```
jeffposnick: Could you provide a comment explaining why this is necessary? I'm not familiar with using the `vm` module and some more context (`NewContext`, if you will...) would be useful.

jeffposnick:
Just the one suggestion about an explanatory comment, but otherwise it's 👍

PR #121, opened 2017-01-11
Use local documentation

jeffposnick:
R: @addyosmani @gauntface @prateekbh 

I still can't shake the feeling that there were some provisions in place already to use the local `documentation` binary, but I must be confused with another project.

In any case, this is a small change to launch it from `node_modules/.bin/documentation`, meaning that it doesn't need to be installed globally.

It might be irrelevant if we switch to another documentation generator, but it shouldn't hurt.


gauntface:
LGTM

PR #125, opened 2017-01-12
Remove cacheName from sw-lib cachingStrategy.

### LINKED ISSUE #122: Cache Expiration Behavior + Cache Name + RequestWrapper
@jeffposnick I'm getting a little lost with how these pieces fit together.

A RequestWrapper can take a `cacheName` but if one isn't supplied, a default `cacheName` is used.

The RequestWrapper also takes `behaviors` as an options.

The cache expiration behavior requires a `cacheName`.

One of the follow needs to be done:

1. Cache expiration behavior should get the cache name from the request wrapper, rather than from it's constructor. Pros: Simple and single place for cache name. Cons: Tightly couples it with RequestWrapper
2. The cache expiration behavior should use the same default as request wrapper. Pros: Simple. Cons: Could be flakey with changes to either class and / or default values.
3. Request Wrapper shouldn't have a cache name as an option and behaviors should explicitly require cache names where appropriate.
4. There is a way to get a common default cache from a library and manually pass that in where relevant. Pros: developer control. Cons: Not automatic, could be frustrating sharing same value everywhere.

I'm probably in favor of option 1 or 2.

gauntface:
R: @jeffposnick @addyosmani 

Fixes #122

Just removes a now unnecessary cacheName parameter.

### DIFF
--- a/packages/sw-lib/test/browser-unit/sw-unit/caching-strategies.js
+++ b/packages/sw-lib/test/browser-unit/sw-unit/caching-strategies.js
@@ -69,7 +69,6 @@ describe('Test caching strategies.', function() {
       // between the request wrapper and the cache expiration behavior.
       const CACHE_NAME = 'hello-world-' + Date.now();
       const CACHE_EXPIRATION = {
-        cacheName: CACHE_NAME,
         maxEntries: 10,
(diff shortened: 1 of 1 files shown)


jeffposnick:
👍

PR #127, opened 2017-01-12
Cache browser downloads

jeffposnick:
R: @gauntface

We're currently spending ~40 seconds for each entry in `packages/` downloading browsers, leading to much longer Travis CI build times than are necessary.

As per my conversation with @gauntface, this may be enough to get Travis to persist the downloads across builds. Let's what the actual Travis CI run for this PR looks like, but I guess it might take a re-run to actually trigger the use of the cache?


addyosmani:
In general this approach appears sound to me. Appears the Travis run is still running.

gauntface:
Looking at this test: https://travis-ci.org/GoogleChrome/sw-helpers/jobs/191411257 line 328

    [21:25:56] Finished 'download-browsers' after 506 ms

PR #129, opened 2017-01-12
Appveyor

### LINKED ISSUE #118: AppVeyor / CLI Testing on Windows
One thing that is evident in sw-precache is that windows can introduce some "interesting" quirks with globs and file paths.

In an attempt to make sure I don't mess this up I'd like to add AppVeyor to our set up but only to run sw-cli tests.

@jeffposnick @addyosmani any objections to this?

We can configure it to either run if specific message is in the commit I.e. the phrase "cli" comes up, to try and save overall time and only use it when needed. What do you think?

gauntface:
R: @jeffposnick @addyosmani 

I'm hoping this will bring in a work appveyor test which will mean windows testing for CLI moving forward.

Fixes #118

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,3 +1,7 @@
+only_commits:
+  files:
+    - packages/sw-cli/
+
 install:
   - ps: Install-Product node Stable
   - npm install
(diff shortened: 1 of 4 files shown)


jeffposnick:
🎆 👍

jeffposnick:
It's great how it led to you identifying an issue with an existing test!

PR #148, opened 2017-01-13
Better JSDoc examples for sw-runtime-caching

jeffposnick:
R: @addyosmani @gauntface

This improves the JSDocs for the members of the `sw-runtime-caching` package.


packages/sw-runtime-caching/src/lib/constants.js
```diff
@@ -35,5 +35,4 @@ export const behaviorCallbacks = [
   'cacheDidUpdate',
   'cacheWillUpdate',
   'fetchDidFail',
-  'requestWillFetch',
```
jeffposnick: Though this might be added in at some point, `RequestWrapper` doesn't currently support `requestWillFetch`, so I'm removing it from here and from the JSDoc.

packages/sw-runtime-caching/src/lib/cache-first.js
```diff
@@ -18,7 +18,18 @@ import assert from '../../../../lib/assert';
 
 /**
  * @memberof module:sw-runtime-caching
- * @extends module:sw-runtime-caching.Handler
+ * @extends Handler
```
gauntface: Does this link in the docs correctly?

packages/sw-runtime-caching/src/lib/network-only.js
```diff
@@ -18,7 +18,18 @@ import assert from '../../../../lib/assert';
...
  * @memberof module:sw-runtime-caching
- * @extends module:sw-runtime-caching.Handler
+ * @extends Handler
+ *
+ * @example
+ * // Set up a route to match any requests made for URLs that end in .txt
+ * // The requests are handled with a network-only strategy.
+ * const route = new goog.routing.RegExpRoute({
+ *   regExp: /\.txt$/,
+ *   handler: new goog.runtimeCaching.CacheFirst(),
```
gauntface: 'CacheFirst' needs changing here.

jeffposnick: Done.

packages/sw-runtime-caching/src/lib/cache-first.js
jeffposnick: The link doesn't working with the current `documentation`-based generation, no. But it didn't work either with the `module:` prefix, and at least the output is less verbose with that removed.

gauntface: gotcha. Looks like gulp documentation is replacing the version of docs I had up. But I believe that has linked up the extending classes correctly. May be a jsdocs feature.

PR #156, opened 2017-01-17
Demos to Docs

gauntface:
R: @jeffposnick @addyosmani 

This moves the demos into the docs directories so they'll be published with other markdown files.

Also switched from builds to node_modules, this is just to seperate out the master code from the published demo code. It might be good or bad for us in the long run, but I think it'll largely sanitize keeping the demos working moving forward.

Apologies the PR is so big (it's largely just deleting and adding files).

Main changes are to the Readme templates and adjusting paths in each demo to use node_modules rather than build.


jeffposnick:
I'm having a hard time picturing how this all fits together. I've got a local copy of this branch; are there specific commands that I should run in order to simulate what the process of publishing the demos will be like (`gulp documentation`)?

Also, I'm confused by checking in things `docs/demos/node_modules/` to git, as `node_modules` tends to be managed out of source control via `npm`/`yarn`. What's the thinking behind that?

jeffposnick:
I'm supportive of this general idea, but I have a few things that I'm not sure about.

Are the `dependencies` version numbers in `docs/demos/package.json` managed automatically via a script, or will we have to bump them explicitly? (I feel like this might be automated by `npm-publish-scripts`?)

What's the process for testing out the demos against local copies of the libraries, instead of against a version released on `npm`? There are lots of times I will make a local change to one or more libraries and want to confirm behavior in an existing demo. This workflow is partly due to the abysmal test coverage for my code, but at the same time, playing around with code changes in Chrome's DevTools is often a really useful way of confirming new behavior.

gauntface:
Package.json would require a manual update.

For that kind of flow I would recommend I just change back to a gulp build and update cycle. This might break the demos and / or result in demos showing different behavior than what is released, but I'm not too fussed about this tbh.

Shall I do that?

jeffposnick:
> For that kind of flow I would recommend I just change back to a gulp build and update cycle. This might break the demos and / or result in demos showing different behavior than what is released, but I'm not too fussed about this tbh.

Does that mean that instead of the demos going against `sw-*` libraries installed via `npm`, the demos would always go against a locally obtained copy of the library bundles? If so, yes, I think that change would be worthwhile.

gauntface:
I'm going to abandon this. There isn't going to be a clean way to build from gulp that is going to work for your flow of building and testing (given the lack of tests).

I'm going to redo this PR with less changes and just drop the demos from the readme's until we can find a clean solution.

PR #157, opened 2017-01-18
CLI - Generate sw

gauntface:
R: @jeffposnick @addyosmani 

This is a super basic, bare bones implementation of service worker generation.

### DIFF
--- a/packages/sw-cli/package.json
+++ b/packages/sw-cli/package.json
@@ -38,7 +38,7 @@
     "sw-lib": "0.0.7",
     "update-notifier": "^1.0.3"
   },
-  "devdependencies": {
+  "devDependencies": {
     "fs-extra": "^2.0.0"
   }
 }
--- a/packages/sw-cli/package.json
+++ b/packages/sw-cli/package.json
@@ -35,6 +35,10 @@
     "lodash.template": "^4.4.0",
     "minimist": "^1.2.0",
     "mkdirp": "^0.5.1",
+    "sw-lib": "0.0.7",
     "update-notifier": "^1.0.3"
+  },
+  "devdependencies": {
+    "fs-extra": "^2.0.0"
   }
 }
--- a/packages/sw-cli/src/lib/templates/sw.js.tmpl
+++ b/packages/sw-cli/src/lib/templates/sw.js.tmpl
@@ -1,6 +1,22 @@
-importScripts(
-  "<%= manifestPath %>",
-  "<%= swlibPath %>"
-);
+importScripts('<%= swlibPath %>');
 
-self.goog.swlib.cacheRevisionedAssets(self.__file_manifest);
+/**
+ * DO NOT EDIT THE FILE MANIFEST ENTRY
+ *
+ * The method cacheRevisionedAssets() does the following:
+ * 1. Cache URLs in the manifest to a local cache.
+ * 2. When a network request is made for any of these URLs the response
+ *    will ALWAYS comes from the cache, NEVER the network.
+ * 3. When the service worker changes ONLY assets with a revision change are
+ *    updated, old cache entries are left as is.
+ *
+ * By changing the file manifest manually, your users may end up not receiving
+ * new versions of files because the revision hasn't changed.
+ *
+ * Please use sw-cli or some other tool / approach to generate the file
+ * manifest which accounts for changes to local files and update the revision
+ * accordingly.
+ */
+const fileManifest = <%= JSON.stringify(manifestEntries) %>;
+
+self.goog.swlib.cacheRevisionedAssets(fileManifest);
--- a/packages/sw-cli/src/lib/templates/sw.js.tmpl
+++ b/packages/sw-cli/src/lib/templates/sw.js.tmpl
@@ -1,0 +1,6 @@
+importScripts(
+  "<%= manifestPath %>",
+  "<%= swlibPath %>"
+);
+
+self.goog.swlib.cacheRevisionedAssets(self.__file_manifest);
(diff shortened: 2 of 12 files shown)


gauntface:
FYI: I'm going to use Saucelabs for the windows testing just to make sure I haven't / don't mess up any paths on Windows.....I'm pretty sure I have / will.

packages/sw-cli/package.json
```diff
@@ -30,6 +30,10 @@
     "inquirer": "^2.0.0",
     "lodash.template": "^4.4.0",
     "minimist": "^1.2.0",
-    "mkdirp": "^0.5.1"
+    "mkdirp": "^0.5.1",
+    "sw-lib": "0.0.7"
+  },
+  "devdependencies": {
```
jeffposnick: This should be `devDependencies`.

packages/sw-cli/src/lib/templates/sw.js.tmpl
```diff
@@ -0,0 +1,6 @@
+importScripts(
+  "<%= manifestPath %>",
```
jeffposnick: Changes to files listed in `importScripts` won't trigger service worker updates yet. (The [tracking bug](https://bugs.chromium.org/p/chromium/issues/detail?id=648295) says that it's targeted for Chrome 57, but I think the branch for that happened already, so I assume Chrome 58 is more likely.)

Therefore, we need to inline the manifest file if we want changes to the manifest to trigger the `install` events.

packages/sw-cli/package.json
gauntface: Fixed.

packages/sw-cli/src/lib/templates/sw.js.tmpl
gauntface: Yeah I haven't tackled this problem yet.

I'm still in favor of adding a timestamp in the top of the service worker, this is generic and safe and results in a new service worker.

Secondly, if we follow this approach, using the CLI to rebuild the manifest can then update the service worker timestamp, throwing an error (or simply not updating the manifest file) if there is any issue in doing that (i.e. sw file can't be found, no existing timestamp is found or .....).

If we inline the contents we can ONLY rewrite the service worker file which prevents people from being able to edit the service worker directly which is :-1: for me.

What do you think?

gauntface:
What is the benefit of inlining files? That makes the service worker much harder to read and understand what is going on and will prevent developers from being able to alter and grow it's use beyond the CLI / tooling we provide out of the box.

packages/sw-cli/src/lib/templates/sw.js.tmpl
jeffposnick: Using a timestamp to trigger a SW update is overly aggressive. We want SW updates only when something our SW cares about changes, not whenever there's a build performed.

Granted, that doesn't introduce very much overhead, assuming our `install`/`activate` handlers are lightweight.

But more to your point, I understand that you want to have a model in which folks can edit the top-level service worker file, and I support that. I am confused as to *when* you think the top-level file should be edited, though. My understanding was that folks would edit the file prior to running it through the Rollup bundling step, not edit the file afterwards.

In your model, are folks taking the file that's been spit out by this build process, including the timestamp you propose, and then editing that? Won't their edits be lost the next time the file is generated?

jeffposnick:
My inline comment covers what my thoughts were about editing the top-level service worker. I agree that being able to alter that file is an important goal.

One benefit of inlining files is being able to reliably trigger the SW update flow only when something we care about changes (until `importScripts` can do that).

Another benefit is that it's easier to deploy a single, bundled SW file than it is to deploy three separate files, all of which have to be present at runtime in order for the SW to function. That was one of the reasons I had for inlining `sw-toolbox` code in the final output of `sw-precache`. You also only have to worry about getting the HTTP caching headers right (normally, `max-age: 0`) for a single file and not three of them if everything is bundled.

gauntface:
The latest version of this inlines the file manifest - seems to be the most natural thing for all developers and I've added a note to above the manifest explaining why developers should not edit it manually.

The sw-lib file is still in there as an importScript but it is now named: sw-lib.vX.X.X.min.js. This means that cache expiration etc is not a problem when developers are using sw-cli (i.e. if they update sw-cli, sw-cli will update with a new sw-lib version which will result in a new file, with a new name and a change in the SW.

I know we will still have plenty to discuss tomorrow, but just wanted to post an update on current changes.

addyosmani:
Now that we have an inlined manifest and a comment discouraging developers from directly editing the generated Service Worker, it looks like we're heading in a good direction. Thanks for keeping the train chugging along on this folks 🚋 💨 

Jeff's feedback on inlining sw-lib _would_ improve the deployment story and I can't disagree that it simplifies having to reason about the HTTP caching headers for fewer files, but I also wonder how much of an effort deploying a few additional files are (given a /dist already contains a plethora these days). I would be more inclined to think the reason to inline would be reliably triggering the SW update flow, but perhaps this is a good topic for us to discuss tomorrow.

I'd love for us to think about moving forward and opening up an issue for deciding on whether to inline or not separately so Matt isn't too blocked on moving this work forward.

jeffposnick:
The approach in this PR is on par with what you could get with `sw-precache`, so from the perspective of getting something working and useful checked in for now, I'm 👍

My thoughts about an end state that I'd consider ideal and an imrpovement over the `sw-precache`-era functionality are probably best summarized at https://github.com/GoogleChrome/sw-helpers/issues/82#issuecomment-274214709. Given the controversy about that approach, we can discuss potential changes without blocking this PR.

PR #158, opened 2017-01-18
Add in same-origin check for ExpressRoute

### LINKED ISSUE #151: Make sure that ExpressRoute only matches same-origin requests
Right now the `goog.routing.ExpessRoute` just looks at `.pathname` of the request URL, and compares it to the path used to the configure the route.

This means that a request for `https://cross-domain.com/path/to/file` would match a route that was configured with `new ExpressRoute({path: '/path/to/file', hander})`, even if the service worker were served from `https://example.com`.

I *think* the decision was made not to support that additional `origin` configuration option that `sw-toolbox` supported, and which was pretty confusing. (@gauntface @addyosmani, can you confirm that's what we decided?)

I'll point out in the docs that `ExpressRoute` should only be used for same-origin routing, and if cross-origin routes are needed, something like `RegExpRoute` is a better choice.

jeffposnick:
R: @addyosmani @gauntface

Fixes #151

### DIFF
--- a/lib/error-factory.js
+++ b/lib/error-factory.js
@@ -37,7 +37,7 @@ class ErrorFactory {
       throw new Error(`Unable to generate error '${name}'.`);
     }
 
-    let message = this._errors[name];
+    let message = this._errors[name].replace(/\s+/g, ' ');
     let stack = null;
     if (thrownError) {
       message += ` [${thrownError.message}]`;
--- a/packages/sw-routing/src/lib/error-factory.js
+++ b/packages/sw-routing/src/lib/error-factory.js
@@ -18,7 +18,7 @@ import ErrorFactory from '../../../../lib/error-factory';
 const errors = {
   'express-route-requires-absolute-path': `When using ExpressRoute, you must
     provide a path that starts with a '/' character. You can only match
-    same-origin requests. For more flexibility, use RegExpRoute.`
+    same-origin requests. For more flexibility, use RegExpRoute.`,
 };
 
 export default new ErrorFactory(errors);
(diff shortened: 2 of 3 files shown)


gauntface:
What happens when a user **does** pass in a different origin?

jeffposnick:
I added in an explicit check to make sure that the `path` starts with `'/'`, and I'm using `ErrorFactory` to throw a custom error if it doesn't.

I also made a slight tweak to the base `ErrorFactory` class to have it strip out repeated whitespace in the error message before `throw`ing. This makes it easier to define multi-line error messages using ES2015 string literal backticks, while still respecting our line length linting, and not having extra newlines and whitespace make it through to the message itself. See https://github.com/GoogleChrome/sw-helpers/pull/158/files#diff-cad42d520bcc18461d07406f91d64442R19 for an example of what I mean about the nicer syntax.

If you'd rather I not tweak `ErrorFactory` like that, I can back it out of this PR.

gauntface:
ErrorFactory change looks good.

With the slash, is there any chance someone would go: `new ExpressRoute('site-section/:sectionId')` I could picture this being intended as a path relative to the service worker location.

jeffposnick:
I poked around the [`path-to-regexp`](https://github.com/pillarjs/path-to-regexp#usage) library that's used under the hood. It would be okay with paths that don't start with '/'. Moreover, `path-to-regexp` would be okay with full URLs that include origins, and would convert them into a reasonable-looking `RegExp`.

The reason not to support anything other than paths that start with '/' is that we're billing this library as being "Express"-style routing, and from what I can tell from the [Express docs](https://expressjs.com/en/guide/routing.html), it always requires routes to start with '/'.

So it's a somewhat artificial requirement meant to enforce Express's restrictions. Of course, if we didn't behave like Express then this module would undergo an existential crisis, so I'm in favor of maintaining that requirement.

gauntface:
I'm a tad confused why your build is unhappy given no changes in this should have upset - trying to fix travis....

jeffposnick:
Looks like it's cleared up now, but yes, that is odd...

I'm going to merge.

PR #160, opened 2017-01-18
Address comment in #82 - update user of new versions

gauntface:
R: @jeffposnick @addyosmani 

Notify user of new CLI versions.


jeffposnick:
👍

PR #161, opened 2017-01-18
Checks the freshness of a cached response before using it

### LINKED ISSUE #120: Take Date: header into account for sw-cache-expiration
The change in https://github.com/GoogleChrome/sw-toolbox/pull/206 should also be taken into account in `sw-cache-expiration`, i.e. the `Date:` header value should factor in to deciding whether a response is fresh.

jeffposnick:
R: @addyosmani @gauntface

Fixes #120 

This adds in support for a new lifecycle callback, `cacheWillMatch`, which gives behaviors a chance to intercept cache matches before they're returned to the handler, to check whether the matched value should be used.

It also modifies the `sw-cache-expiration` behavior to take advantage of this new callback, by examining the `Response` object that was retrieved from the cache, making sure that it's fresh, via a comparison of its `Date:` header to the configured `maxAgeSeconds` expiration.

This matches the behavior that was recently implemented in `sw-toolbox`: https://github.com/GoogleChrome/sw-toolbox/pull/206

### DIFF
--- a/packages/sw-cache-expiration/src/lib/behavior.js
+++ b/packages/sw-cache-expiration/src/lib/behavior.js
@@ -121,18 +121,18 @@ class Behavior {
    * `goog.runtimeCaching` handlers when a `Response` is about to be returned
    * from a [Cache](https://developer.mozilla.org/en-US/docs/Web/API/Cache) to
    * the handler. It allows the `Response` to be inspected for freshness and
-   * prevents it from being used if the `Response`'s `Date` header value is
+   * prevents it from being used if the `Response`'s `Date:` header value is
    * older than the configured `maxAgeSeconds`.
    *
    * Developers who are not using `goog.runtimeCaching` would normally not call
-   * this method directly; instead, use [`isResponseFresh`](#isResponseFresh)
-   * to perform the same freshness check.
+   * this method directly; instead, [`isResponseFresh`](#isResponseFresh) can be
+   * used to perform the same freshness check.
    *
    * @private
    * @param {Object} input The input object to this function.
    * @param {Response} input.cachedResponse The `Response` object that's been
    *        read from a cache and whose freshness should be checked.
-   * @return {Response|null} Either the `cachedResponse`, if it's fresh, or
+   * @returns {Response|null} Either the `cachedResponse`, if it's fresh, or
    *          `null` if the `Response` is older than `maxAgeSeconds`.
    */
   cacheWillMatch({cachedResponse} = {}) {
@@ -144,16 +144,16 @@ class Behavior {
   }
 
   /**
-   * This checks whether a `Response` is fresh, based on the `Response`'s
-   * `Date` header and the configured `maxAgeSeconds`.
+   * Checks whether a `Response` is fresh, based on the `Response`'s
+   * `Date:` header and the configured `maxAgeSeconds`.
    *
-   * If `maxAgeSeconds` or the `Date` header is not set then it will
+   * If `maxAgeSeconds` or the `Date:` header is not set then it will
    * default to returning `true`.
    *
    * @param {Object} input The input object to this function.
    * @param {Response} input.cachedResponse The `Response` object that's been
    *        read from a cache and whose freshness should be checked.
-   * @return {boolean} Either the `true`, if it's fresh, or `false` if the
+   * @returns {boolean} Either the `true`, if it's fresh, or `false` if the
    *          `Response` is older than `maxAgeSeconds`.
    */
   isResponseFresh({cachedResponse} = {}) {
@@ -166,7 +166,7 @@ class Behavior {
       if (dateHeader) {
         const now = Date.now();
         const parsedDate = new Date(dateHeader);
-        // If the Date header was invalid for some reason, parsedDate.getTime()
+        // If the Date: header was invalid for some reason, parsedDate.getTime()
         // will return NaN, and the comparison will always be false. That means
         // that an invalid date will be treated as if the response is fresh.
         if ((parsedDate.getTime() + (this.maxAgeSeconds * 1000)) < now) {
--- a/packages/sw-cache-expiration/src/lib/behavior.js
+++ b/packages/sw-cache-expiration/src/lib/behavior.js
@@ -121,18 +121,18 @@ class Behavior {
    * `goog.runtimeCaching` handlers when a `Response` is about to be returned
    * from a [Cache](https://developer.mozilla.org/en-US/docs/Web/API/Cache) to
    * the handler. It allows the `Response` to be inspected for freshness and
-   * prevents it from being used if the `Response`'s `Date:` header value is
+   * prevents it from being used if the `Response`'s `Date` header value is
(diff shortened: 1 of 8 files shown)


packages/sw-cache-expiration/src/lib/behavior.js
```diff
@@ -118,15 +118,78 @@ class Behavior {
 
   /**
    * A "lifecycle" callback that will be triggered automatically by the
-   * goog.runtimeCaching handlers when an entry is added to a cache.
+   * `goog.runtimeCaching` handlers when a `Response` is about to be returned
+   * from a [Cache](https://developer.mozilla.org/en-US/docs/Web/API/Cache) to
+   * the handler. It allows the `Response` to be inspected for freshness and
+   * prevents it from being used if the `Response`'s `Date:` header value is
```
jpmedley: Please remove colon from header names. Check throughout.

packages/sw-cache-expiration/src/lib/behavior.js
```diff
@@ -118,15 +118,78 @@ class Behavior {
...
+   * older than the configured `maxAgeSeconds`.
+   *
+   * Developers who are not using `goog.runtimeCaching` would normally not call
+   * this method directly; instead, [`isResponseFresh`](#isResponseFresh) can be
+   * used to perform the same freshness check.
```
jpmedley: I won't consider this blocking. In the future, please try to use active voice instead of passive. For example, "instead, use `isResponseFresh` to perform the same freshness check.

packages/sw-cache-expiration/src/lib/behavior.js
```diff
@@ -118,15 +118,78 @@ class Behavior {
...
+   * This checks whether a `Response` is fresh, based on the `Response`'s
+   * `Date:` header and the configured `maxAgeSeconds`.
+   *
+   * If `maxAgeSeconds` or the `Date:` header is not set then it will
+   * default to returning `true`.
+   *
+   * @param {Object} input The input object to this function.
+   * @param {Response} input.cachedResponse The `Response` object that's been
+   *        read from a cache and whose freshness should be checked.
+   * @returns {boolean} Either the `true`, if it's fresh, or `false` if the
```
jpmedley: Didn't you say in the meeting that it was @return instead of @returns?

packages/sw-runtime-caching/src/lib/request-wrapper.js
```diff
@@ -117,7 +120,16 @@ class RequestWrapper {
     assert.atLeastOne({request});
 
     const cache = await this.getCache();
-    return await cache.match(request, this.matchOptions);
+    const response = await cache.match(request, this.matchOptions);
```
gauntface: Nit: Change response to cachedResponse.

packages/sw-runtime-caching/src/lib/request-wrapper.js
```diff
@@ -117,7 +120,16 @@ class RequestWrapper {
...
+
+    // Give the cacheWillMatch callbacks a chance to modify the cached response.
+    // TODO: Right now, we run all the callbacks sequentially, and return the
+    // final result. Instead, it might make sense to enforce a limit of a single
+    // cacheWillMatch callback, and return an error if there are multiple ones.
+    const callbacks = this.behaviorCallbacks.cacheWillMatch || [];
+    return callbacks.reduce((cachedResponse, callback) => {
```
gauntface: Change cachedResponse to previousResponseValue

packages/sw-runtime-caching/src/lib/request-wrapper.js
```diff
@@ -117,7 +120,16 @@ class RequestWrapper {
...
+      return callback({cachedResponse});
```
gauntface: This feels like a really weird loop and I think it's the naming.

An array of 'cacheWillMatch' functions, sounds like an array of functions that will return true or false if the cache will match, but it's not it's a function that filters the cache response.

Then looping over some callbacks will get the actual response, others will get null from the previous cached response.

If the previousResponse is null, there is no point calling the callback, it'll have no information to look at. If you agree with this, then this should be a for loop that loops over until it gets a null and then just return early.

packages/sw-runtime-caching/src/lib/request-wrapper.js
```diff
@@ -117,7 +120,16 @@ class RequestWrapper {
     assert.atLeastOne({request});
 
     const cache = await this.getCache();
-    return await cache.match(request, this.matchOptions);
+    const response = await cache.match(request, this.matchOptions);
+
+    // Give the cacheWillMatch callbacks a chance to modify the cached response.
```
jpmedley: Need backticks on `cacheWillMatch`.

packages/sw-runtime-caching/src/lib/request-wrapper.js
```diff
@@ -117,7 +120,16 @@ class RequestWrapper {
...
+    // TODO: Right now, we run all the callbacks sequentially, and return the
+    // final result. Instead, it might make sense to enforce a limit of a single
+    // cacheWillMatch callback, and return an error if there are multiple ones.
```
jpmedley: Need backticks on `cacheWillMatch`.

packages/sw-runtime-caching/src/lib/request-wrapper.js
jeffposnick: Done.

packages/sw-runtime-caching/src/lib/request-wrapper.js
jeffposnick: Done.

packages/sw-runtime-caching/src/lib/request-wrapper.js
jeffposnick: These are inline comments, not JSDoc. They don't get run through the Markdown processor or included as part of the published documentation, so Markdown syntax seems unneeded.

packages/sw-runtime-caching/src/lib/request-wrapper.js
jeffposnick: These are inline comments, not JSDoc. They don't get run through the Markdown processor or included as part of the published documentation, so Markdown syntax seems unneeded.

packages/sw-runtime-caching/src/lib/request-wrapper.js
jeffposnick: Yeah, it is weird. I struggled with this, as mentioned in the "TODO:" comment about whether it made sense to support multiple callbacks.

My takeaway from your comment is that it would be simpler if we just enforced a limit of 1 callback (similar to what we do for `cacheWillUpdate`), and that would end up being much less of a surprise to the developer who accidentally supplies multiple callbacks and doesn't understand how they all interact with each other.

jeffposnick:
I've changed the approach to prevent multiple `cacheWillMatch` callbacks, and hopefully that makes things clearer and provides less opportunity for developer confusion. PTAL.

PR #162, opened 2017-01-18
Adding contribution guidelines

### LINKED ISSUE #132: README: Add note about npm install/yarn root
Minor one. When you clone sw-helpers, it's not clear right now what the expectation would be when you `npm install` or `yarn` root. 

One might guess this will recursively try installing all the dependencies for the sub-packages in the mono-repo. Or just the deps needed to hack on sw-helpers, but since we don't appear to note this in the readme it's unclear.

gauntface:
R: @jeffposnick @addyosmani

Fixes #132

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -1,0 +1,75 @@
+---
+layout: index
+title: Contributing
+navigation_weight: 1
+---
+# Contributing
+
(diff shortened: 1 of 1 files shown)


jeffposnick:
👍

PR #163, opened 2017-01-19
Adding ability to define routes for precached assets

### LINKED ISSUE #136: sw-precaching and Routing
sw-precaching at the moment ONLY caches assets, it doesn't define any routes.

There are a few things to consider:

1. If it sets up the routes inside the module, that blocks off any possibility of external routes doing the caching.
1. If we want external routes to be defined to do this, there is no exposure of the cache name and be default I think caching strategies look in a custom cache name.

Solutions:
1. Add the routes in sw-precaching.
1. Add the routes in sw-precaching with opt out in options.
1. Add a way for routes to match against any cache (i.e. caches.matches() )
1. Add a way to get the cache name from sw-precaching and configure route for that.

gauntface:
R: @jeffposnick @addyosmani

Fixes #136

This makes the cache name and cached assets available so Routes can be easily added for precaching.

### DIFF
--- a/packages/sw-precaching/package.json
+++ b/packages/sw-precaching/package.json
@@ -21,5 +21,9 @@
   "homepage": "https://github.com/GoogleChrome/sw-helpers/tree/master/packages/sw-precaching",
   "main": "build/sw-precaching.min.js",
   "module": "build/sw-precaching.min.mjs",
-  "jsnext:main": "build/sw-precaching.min.mjs"
+  "jsnext:main": "build/sw-precaching.min.mjs",
+  "devDependencies": {
+    "sw-routing": "0.0.7",
+    "sw-runtime-caching": "0.0.7"
+  }
 }
--- a/packages/sw-precaching/src/lib/controllers/base-cache-manager.js
+++ b/packages/sw-precaching/src/lib/controllers/base-cache-manager.js
@@ -17,7 +17,7 @@ class BaseCacheManager {
   }
 
   /**
-   * This method will add the entries to the install list.
+   * Adds entries to the install list.
    * This will manage duplicate entries and perform the caching during
    * the install step.
    * @param {Array<String|Request|Object>} rawEntries A raw entry that can be
@@ -32,7 +32,7 @@ class BaseCacheManager {
   }
 
   /**
-   *This method gives access to the cache used by thie caching manager.
+   * Gives access to the cache name used by thie caching manager.
    * @return {String} The cache name used for this manager.
    */
   getCacheName() {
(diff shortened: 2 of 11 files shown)


packages/sw-precaching/package.json
```diff
@@ -21,5 +21,9 @@
   "homepage": "https://github.com/GoogleChrome/sw-helpers/tree/master/packages/sw-precaching",
   "main": "build/sw-precaching.min.js",
   "module": "build/sw-precaching.min.mjs",
-  "jsnext:main": "build/sw-precaching.min.mjs"
+  "jsnext:main": "build/sw-precaching.min.mjs",
+  "devDependencies": {
```
jeffposnick: Just checking: is this for the benefit of https://github.com/GoogleChrome/sw-helpers/pull/163/files#diff-0dba01216e2ffb2625e2115ac8158c95R1, or is it used elsewhere as well?

I remember @prateekbh ran into similar issues when he needed to use cross-module dependencies in his tests, and I think he worked around that by adding in a bunch of additional bundling steps into his build process.

I'm not sure which approach is preferable, but if you or @prateekbh have strong feelings, it would be useful to weigh in and go with a consistent approach across all of the projects.

packages/sw-precaching/src/lib/controllers/base-cache-manager.js
```diff
@@ -32,6 +32,28 @@ class BaseCacheManager {
...
   /**
+   *This method gives access to the cache used by thie caching manager.
+   * @return {String} The cache name used for this manager.
+   */
+  getCacheName() {
+    return this._cacheName;
+  }
+
+  /**
+   * This method returns an array of URL's that will be cached by this
```
jeffposnick: Could you adjust this comment to make it clear that you're returning fully qualified URLs (even though the manifest contains relative URLs)?

packages/sw-precaching/src/lib/controllers/base-cache-manager.js
```diff
@@ -32,6 +32,28 @@ class BaseCacheManager {
...
+   * cache manager.
+   * @return {Array<String>} An array of URLs that will be cached.
+   */
+  getCachedUrls() {
+    const urls = [];
```
jeffposnick: Please ignore if you find this less readable, but you could rewrite as:

```
return [...this._entriesToCache.keys()]
  .map(url => new URL(url, location).href)
```

or equivalently:

```
return Array.from(this._entriesToCache.keys())
  .map(url => new URL(url, location).href)
```

packages/sw-precaching/src/lib/precache-manager.js
```diff
@@ -148,6 +149,29 @@ class PrecacheManager {
   }
 
   /**
+   * Get the name of the cache used for revisioned assets.
```
jeffposnick: I think this JSDoc is incorrect.

packages/sw-precaching/src/lib/precache-manager.js
```diff
@@ -148,6 +149,29 @@ class PrecacheManager {
...
+   *
+   * @example
+   * const revisionedMngr = precacheMAnager.getRevisionedCacheManager();
+   */
+  getRevisionedCacheManager() {
+    return this._revisionedManager;
+  }
+
+  /**
+   * Get the name of the cache used for unrevisioned assets.
```
jeffposnick: I think this JSDoc is incorrect.

packages/sw-precaching/src/lib/controllers/base-cache-manager.js
```diff
@@ -32,6 +32,28 @@ class BaseCacheManager {
   }
 
   /**
+   *This method gives access to the cache used by thie caching manager.
```
jpmedley: "This method gives" to "Gives".

In reference materials, definitions start with a verb, even though it's not a complete sentence. Please check throughout.

packages/sw-precaching/src/lib/controllers/base-cache-manager.js
jpmedley: "This method returns" to "Returns".

packages/sw-precaching/package.json
gauntface: This test is a smoke test and is just to ensure the precaching route is feasible. I really didn't want to add a build step for routing as it's likely to get messy and un-maintainable since it won't be the same build step as sw-routing. #twocents

packages/sw-precaching/src/lib/controllers/base-cache-manager.js
gauntface: Done.

packages/sw-precaching/src/lib/controllers/base-cache-manager.js
gauntface: Done.

packages/sw-precaching/src/lib/precache-manager.js
gauntface: correct and fixed.

packages/sw-precaching/src/lib/precache-manager.js
gauntface: correct and fixed.

packages/sw-precaching/src/lib/controllers/base-cache-manager.js
gauntface: done.

packages/sw-precaching/src/lib/controllers/base-cache-manager.js
gauntface: done.

Issue #164, opened 2017-01-20
Every Module Should Have a Demo.


gauntface:
This is a checklist of things missing from current docs.

- [ ] Every module has a demo.
    - [ ] Add some styles to demos.
    - [ ] Make it more evident as to what the demo is showing off.
- [x] Every module has a description for the module in JSDocs.
- [x] Every module has some JSDoc OR we have some markdown file introducing each project.
- [x] Add examples to every method.
- [x] Add Example of public constructor.
- [x] Add fuller examples on each index.

gauntface:
cc @jpmedley I'll go through this with you sometime next week to get your thoughts / input on this as well as how to use [npm-publish-scripts](https://googlechrome.github.io/npm-publish-scripts/) (if you are interested).

jpmedley:
"how to use npm-publish-scripts (if you are interested)." Never ask if I'm interested. Ask when I have the time.

prateekbh:
@gauntface current JSDocs also $0 thingy, after every parameter. May be we can cover to remove them as well
e.g. https://github.com/GoogleChrome/sw-helpers/tree/master/packages/sw-background-sync-queue#constructor

gauntface:
@prateekbh what is the $0 thingy you are referring to?

addyosmani:
I have an issue for migration docs tracked at #109 - should this be folded in here or best to keep that separate?

gauntface:
@addyosmani Keep it seperate for now - I've change the title for this to just cover Demos.

gauntface:
Going to close this - demo's still need some love, but can be done if and when we feel it's important

PR #165, opened 2017-01-20
Minor doc change

gauntface:
R: @jeffposnick @addyosmani 

This is an extremely reduced PR from previous doc PR.

Reference docs on github pages are now referenced for a project rather than the invidual Readme's for projects and the hosted demo links have been removed for now (links to the demo code is still there).

I've opened an issue to cover a number of gaps in this approach (#164) but I think need more time to properly fix.


gauntface:
@jpmedley just to let you know that **all** the README's in this project are auto generated from `gulp documentation` task. This PR fixes the documentation task so it'll actually run and no longer relies on building all the modules.

All that being said, your commit needs to be made to the text that is used to generate the README rather than change directly on the readme. In this case it was the package.json (See my last commit). This change then filters to all references for that string (think it was 2 or 3 files instead of just the one).

PR #176, opened 2017-01-24
Glob pattern

### LINKED ISSUE #166: sw-cli Handle Single File Type Selection
The glob 'lib/**/*.{js}' is invalid.

gauntface:
R: @jeffposnick @addyosmani

Fixes #166 

Handles the case where there is only one file extension to add to glob.


jeffposnick:
👍

Issue #145, opened 2017-01-13
sw-precaching ability to set cache name


gauntface:
At the moment the precaching and runtime caching modules have got their own caches. Without really trying to cache in particular places I've ended up with 3 caches and it just looks / feels messy.

Is this ok? Should we consolidate to a single default cache? Is there any issues with that?

jeffposnick:
Different caches for different purposes doesn't strike me as a problem. (I mean, the `sw-precache`'s old behavior was a mess, I know, but we're not talking about that...)

It's nice to be able to look at the top-level list of caches and make sure that you have multiple ones to confirm that both your precaching and runtime caching are behaving properly.

It's also a nice distinction to know that all the entries in a specific cache are managed via the same expiration policy, which wouldn't be the case if we mixed up precached and runtime cache entries.

gauntface:
Moving to RequestWrapper should fix this.

gauntface:
@jeffposnick @addyosmani 

I was considering how best to implement custom cachenames for the precaching.

Out of the box all of this works but does assume that precaching + warmRuntime cache are fine being in only two caches (i.e. assumes developers won't want to do something like):

```javascript
revisionedCacheManager.cache({
  revisionedUrls: [....]
  cacheName: 'images'
})

revisionedCacheManager.cache({
  revisionedUrls: [....]
  cacheName: 'styles'
})

revisionedCacheManager.cache({
  revisionedUrls: [....]
  cacheName: 'scripts'
})
```

This isn't infeasible to implement, just the current API prevents it.

There are a few options to support this:

1. Remove the precache-manager completely. All this class does is trigger the install and activate steps. It could be argued that this is useful in sw-lib, but less so for this low level module. Only loss from the current design is that developers of sw-precaching would need to implement the install and activate events on their own.
2. Expose the Revisioned and Unrevisioned CacheManagers as well as the PrecacheManager. Offers all the current APIs + exposes the raw managers. Downside is that this introduces questions as to when use the raw managers vs the precache manager.
2a. (Slight variant) Add a new method `addManager` to precacheManager, this will then loop through the managers and trigger the appropriate install and activate calls. (Offers little to the developer, but helps to prevent any bad practices and makes the raw managers feel natural to use).
[...]

PR #177, opened 2017-01-24
Moves sw-precaching over to using requestWrapper

gauntface:
R: @jeffposnick @addyosmani

This is a step towards #145 although I need to spend a little more time thinking about best approach for cache names

### DIFF
--- a/packages/sw-precaching/src/lib/controllers/base-cache-manager.js
+++ b/packages/sw-precaching/src/lib/controllers/base-cache-manager.js
@@ -121,13 +121,7 @@ class BaseCacheManager {
 
     let response = await this._requestWrapper.fetch({
       request: precacheEntry.getNetworkRequest(),
-      waitOnCache: true,
     });
-
-    /** let response = await fetch(precacheEntry.getNetworkRequest(), {
-      credentials: 'same-origin',
-      redirect: 'follow',
-    });**/
     if (response.ok) {
       const openCache = await this._getCache();
       await openCache.put(precacheEntry.request, response);
--- a/packages/sw-precaching/src/lib/controllers/base-cache-manager.js
+++ b/packages/sw-precaching/src/lib/controllers/base-cache-manager.js
@@ -1,5 +1,5 @@
 import ErrorFactory from '../error-factory';
-
+import {RequestWrapper} from '../../../../sw-runtime-caching/src/index';
 /**
  * This class handles the shared logic for caching revisioned and unrevisioned
  * assets.
@@ -14,6 +14,7 @@ class BaseCacheManager {
   constructor(cacheName) {
     this._entriesToCache = new Map();
     this._cacheName = cacheName;
+    this._requestWrapper = new RequestWrapper({cacheName: this._cacheName});
   }
 
   /**
@@ -118,10 +119,15 @@ class BaseCacheManager {
       return;
     }
 
-    let response = await fetch(precacheEntry.getNetworkRequest(), {
+    let response = await this._requestWrapper.fetch({
+      request: precacheEntry.getNetworkRequest(),
+      waitOnCache: true,
+    });
+
+    /** let response = await fetch(precacheEntry.getNetworkRequest(), {
       credentials: 'same-origin',
       redirect: 'follow',
-    });
+    });**/
     if (response.ok) {
       const openCache = await this._getCache();
       await openCache.put(precacheEntry.request, response);
--- a/packages/sw-precaching/src/lib/controllers/base-cache-manager.js
+++ b/packages/sw-precaching/src/lib/controllers/base-cache-manager.js
@@ -13,8 +13,8 @@ class BaseCacheManager {
    */
   constructor(cacheName) {
     this._entriesToCache = new Map();
-    this._cacheName = cacheName;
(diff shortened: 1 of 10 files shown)


packages/sw-precaching/src/lib/controllers/base-cache-manager.js
```diff
@@ -118,10 +119,15 @@ class BaseCacheManager {
       return;
     }
 
-    let response = await fetch(precacheEntry.getNetworkRequest(), {
+    let response = await this._requestWrapper.fetch({
```
jeffposnick: Have you considered using [`RequestWrapper.fetchAndCache`](https://github.com/GoogleChrome/sw-helpers/blob/master/packages/sw-runtime-caching/src/lib/request-wrapper.js#L177) here instead of `RequestWrapper.fetch`? All the "value add" stuff is in `RequestWrapper.fetchAndCache`.

packages/sw-precaching/src/lib/controllers/base-cache-manager.js
```diff
@@ -14,6 +14,7 @@ class BaseCacheManager {
   constructor(cacheName) {
     this._entriesToCache = new Map();
     this._cacheName = cacheName;
+    this._requestWrapper = new RequestWrapper({cacheName: this._cacheName});
```
jeffposnick: Rather than always constructing your own `RequestWrapper`, how about allowing developers to optionally pass in their own?

Developers might want to do that to allow them to assign a `BroadcastCacheUpdate` behavior to their `RequestWrapper`, so that they could announce to controlled pages when there's an update to a precached resource. It could also be useful to override the `CacheableResponse` behavior if they have custom criteria for determining whether a precaching response is valid.

One problem I see is figuring out what would happen if the pass in both a `cacheName` and a `RequestWrapper` that already has a differently named cache.

packages/sw-precaching/src/lib/controllers/base-cache-manager.js
gauntface: There is one problem with that. The current precaching code will fetch request with the cache busting and then cache it against the origin URL to make future cache.matches possible with the original URL. fetchAndCache will just cache the URL with the cache busting.

gauntface:
@jeffposnick In case you the comment gets lost:

1. I can use fetchAndCache and the current code in precache will cache the reponse against the origin request URL (i.e. the URL without the cache busting search param)
2. I've actually moved unrevisioned assets to the default cache from requestwrapper.

I still need to make the cachenames customisable by a developer, but I'll tackle that in a seperate PR.

packages/sw-precaching/src/lib/controllers/base-cache-manager.js
addyosmani: I hadn't considered the use-case of someone wanting to construct/pass their own Requestwrapper to control pages when there's updates made. Given @gauntface might need to spend time designing how best to handle passing both cacheName and a RequestWrapper that already has a different cache name, perhaps this can be deferred to another PR?

addyosmani:
> I still need to make the cachenames customisable by a developer, but I'll tackle that in a seperate PR.

This SGTM. I left a comment on the first of your commits which may be showing up as outdated, but it looks like you're going to tackle that work in an iteration PR anyway. Full steam ahead!

packages/sw-precaching/src/lib/controllers/base-cache-manager.js
```diff
@@ -13,7 +13,8 @@ class BaseCacheManager {
    */
   constructor(cacheName) {
     this._entriesToCache = new Map();
-    this._cacheName = cacheName;
+    this._requestWrapper = new RequestWrapper({cacheName});
```
jeffposnick:
```js
this._requestWrapper = new RequestWrapper({
  cacheName,
  fetchOptions: {
    credentials: 'same-origin',
  },
});
```

gives you the same [`fetch` options](https://github.com/GoogleChrome/sw-helpers/blob/master/packages/sw-runtime-caching/src/lib/request-wrapper.js#L50) that the previous code (and like `sw-precache`) uses.

jeffposnick:
Just the one additional comment about `fetchOptions`.

Would you like me to modify `RequestWrapper.fetchAndCache` so that it takes in both the request and a cache key URL to support the use case you need? That seems like a straightforward change, and I've created https://github.com/GoogleChrome/sw-helpers/issues/178 to track that if you'd in favor.

It doesn't need to block this PR.

packages/sw-precaching/src/lib/controllers/base-cache-manager.js
gauntface: Possible stupid question, is there a reason for this not to be the default option?

jeffposnick: Definitely not a stupid question. It's not the default for the underlying `fetch()`, so matching `fetch()` behavior by default unless otherwise required seemed like the least-surprising thing to do.

We ran into a scenario with IOWA in which we specifically needed this set, which is how it ended up in `sw-precache` initially.

I'm open to setting different values by default in `RequestWrapper` directly if you'd like to open an issue to discuss.

jeffposnick:
👍

PR #181, opened 2017-01-26
Sw lib docs

gauntface:
R: @jeffposnick @jpmedley 

This is a big set tweaks to docs for sw-lib but will hopefully be a bit easier to follow for readers.

### DIFF
--- a/packages/sw-lib/src/index.js
+++ b/packages/sw-lib/src/index.js
@@ -35,7 +35,7 @@ if (!assert.isSWEnv()) {
  *
  * console.log(self.goog.swlib);
  *
- * @example <caption>Cache some assets and register routes.</caption>
+ * @example <caption>Caching assets and registering routes.</caption>
  *
  * goog.swlib.cacheRevisionedAssets([
  *   {
--- a/packages/sw-lib/src/index.js
+++ b/packages/sw-lib/src/index.js
@@ -18,7 +18,6 @@
 
 import ErrorFactory from './lib/error-factory';
 import SWLib from './lib/sw-lib';
-import {Route} from '../../sw-routing/src/index.js';
 import assert from '../../../lib/assert.js';
 
 if (!assert.isSWEnv()) {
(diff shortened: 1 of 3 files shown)


packages/sw-lib/src/lib/router.js
```diff
@@ -87,10 +64,11 @@ class RouterWrapper {
...
    * 1. A regex that will be tested against request URL's.
-   * 1. A [Route]{@link module:sw-routing.Route} instance.
-   * @param {function|Handler} handler The handler is ignored if you pass in
-   * a Route object, otherwise it's required. The handler will be called when
-   * the route is caught by the capture criteria.
+   * 1. A [Route]{@link module:sw-lib.SWLib#Route} instance.
+   * @param {function|Handler} handler The handler argument is ignored if
+   * you pass in a Route object, otherwise it's required. The handler
+   * will be called when the route is caught by the capture criteria.
+   * If required, provide a function or a runtime caching strategy.
```
jpmedley: Always start with the description. In this case start with "Called when the...".

packages/sw-lib/src/lib/router.js
```diff
@@ -38,46 +38,23 @@ import ErrorFactory from './error-factory.js';
...
-class RouterWrapper {
+class Router {
   /**
-   * Constructs a new RouterWrapper.
+   * An instance of this call can be accessed via `goog.swlib.router`. You
+   * should not instantiate this class yourself.
    */
   constructor() {
-    this._router = new Router();
+    this._router = new SWRoutingRouter();
```
jeffposnick: Just checking: developers will never interact with `SWRoutingRouter`, right? It's just used internally as an alias so that it doesn't conflict with `Router`?

Because that's a confusing name to use for something that would be developer-facing.

jeffposnick:
👍  with the one comment.

packages/sw-lib/src/index.js
```diff
@@ -27,53 +26,37 @@ if (!assert.isSWEnv()) {
...
- * ]);
+ * console.log(self.goog.swlib);
  *
- * // If you have assets that aren't revisioned, you can cache them
- * // during the installation of you service worker using warmRuntimeCache()
- * goog.swlib.warmRuntimeCache([
- *     '/scripts/main.js',
- *     '/images/default-avater.png'
- * ]);
+ * @example <caption>Cache some assets and register routes.</caption>
```
jpmedley: Use participle phrases for captions. >> "Caching assets and registering routes."

packages/sw-lib/src/lib/router.js
gauntface: No developer will see it, that name is internal only. Plus, the docs reference that SWLib.Router is a wrapper on top of sw-routing.Router.

gauntface:
@jpmedley all your changes are in.

PR #182, opened 2017-01-26
Fetch and cache

### LINKED ISSUE #178: Support alternative cache key URLs in RequestWrapper.fetchAndCache
CC: @gauntface 

**Library Affected**:
`sw-runtime-caching`

It could be useful to support passing in an additional optional parameter to [`RequestWrapper.fetchAndCache`](https://github.com/GoogleChrome/sw-helpers/blob/81fedc4a1183c6137d19c07800d806090df17ce3/packages/sw-runtime-caching/src/lib/request-wrapper.js#L177) allowing developers to provide an alternative cache key URL, if they don't want the `request.url` used.

The use case is when the `request` includes cache-busting URL parameters, which ideally wouldn't be used in the cache key.

The revised method signature could be:

```js
async fetchAndCache({request, waitOnCache, cacheKey}) {
  // ...
}
```

gauntface:
R: @jeffposnick @addyosmani 

Fixes #178 

The one behavior I'm unsure of is how to highlight that the cache of the response didn't occur because it was an invalid response.

For now I've implemented a check for (!cacheable && waitOnCache) => i.e. the developer wanted to wait on the cache.put() that is never going to happen so throw. Not the nicest solution, but not sure how else to approach this use case.

### DIFF
--- a/packages/sw-precaching/src/lib/controllers/base-cache-manager.js
+++ b/packages/sw-precaching/src/lib/controllers/base-cache-manager.js
@@ -124,18 +124,18 @@ class BaseCacheManager {
       return;
     }
 
-    let response = await this._requestWrapper.fetch({
-      request: precacheEntry.getNetworkRequest(),
-    });
-    if (response.ok) {
-      const openCache = await this._getCache();
-      await openCache.put(precacheEntry.request, response);
+    try {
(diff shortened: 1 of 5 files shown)


jeffposnick:
👍

PR #183, opened 2017-01-26
Adding Titles of Modules and sw-routing docs

gauntface:
R: @jeffposnick @jpmedley 

Adding sw-routing docs.

### DIFF
--- a/packages/sw-background-sync-queue/src/index.js
+++ b/packages/sw-background-sync-queue/src/index.js
@@ -14,8 +14,8 @@
 /**
  * # sw-background-sync-queue
  *
- * A service worker implementation of the a queue which is triggered by
- * the background sync event.
+ * Queues failed requests and uses the Background Sync API to replay those
+ * requests at a later time when the network state has changed.
  *
  * @module sw-background-sync-queue
  */
--- a/packages/sw-background-sync-queue/src/index.js
+++ b/packages/sw-background-sync-queue/src/index.js
@@ -12,7 +12,10 @@
  */
 
 /**
- * sw-background-sync-queue Module
+ * # sw-background-sync-queue
+ *
+ * A service worker implementation of the a queue which is triggered by
+ * the background sync event.
  *
  * @module sw-background-sync-queue
  */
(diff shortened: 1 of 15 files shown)


packages/sw-background-sync-queue/src/index.js
```diff
@@ -12,7 +12,10 @@
  */
 
 /**
- * sw-background-sync-queue Module
+ * # sw-background-sync-queue
+ *
+ * A service worker implementation of the a queue which is triggered by
```
jpmedley: 'a queue' or 'the queue'?

packages/sw-routing/src/lib/express-route.js
```diff
@@ -50,16 +47,22 @@ import pathToRegExp from 'path-to-regexp';
...
+   * @param {Object} input
+   * @param {string} input.path The path to use for routing.
    *        If the path contains [named parameters](https://github.com/pillarjs/path-to-regexp#named-parameters),
    *        then an Object mapping parameter names to the corresponding value
    *        will be passed to the handler via `params`.
-   * @param {Object} handler - An Object with a `handle` method. That method
-   *        will be used to respond to matching requests.
-   * @param {string} [method] Only match requests that use this
+   * @param {Object} input.handler - An Object with a `handle` method. That
+   *        method will be used to respond to matching requests.
```
jpmedley: An Object with a `handle` method that will be used to respond to matching requests.

Combine the pieces into one.

packages/sw-background-sync-queue/src/index.js
gauntface: This is copied straight out of package.json - not my words. @prateekbh ?

packages/sw-routing/src/lib/express-route.js
gauntface: Done.

packages/sw-background-sync-queue/src/index.js
gauntface: Reading this description I'm not convinced it's the best way to describe the module.

Would something along the lines of:

"Queues failed requests and uses the Background Sync API to replay those requests at a later time when the network state has changed."

jpmedley: That sounds good.

gauntface:
@jpmedley that should be the last change in.

PR #184, opened 2017-01-26
Initial attempt at an e2e test for RegExpRoute

jeffposnick:
This is a bit of a throwaway PR, so that I can solicit feedback from @gauntface


packages/sw-routing/test/automated-test-suite.js
```diff
@@ -49,7 +49,7 @@ describe(`sw-routing Browser Tests`, function() {
         .then(() => swTestingHelpers.mochaUtils.startWebDriverMochaTests(
           assistantDriver.getPrettyName(),
           globalDriverBrowser,
-          `${baseTestUrl}/test/browser-unit/`
+          `${baseTestUrl}/test/browser/`
```
gauntface: Pro this, but could we rename sw-unit to just sw rather than 'unit'?

My rationale for this originally was that some tests are unit tests in the browser, some are unit tests in sw and some service workers are essentially registered as assets to then test resulting behavior.

I've not be particularly standardised so would be good to get some consensus and strive to make all tests follow that pattern.

packages/sw-routing/test/browser/end-to-end/regexp-route.js
```diff
@@ -0,0 +1,18 @@
+/* global goog, expect */
+
+describe('End to End Test of RegExpRoute', () => {
+  it('should work properly when there are multiple routes', () => goog.swUtils.activateSW('end-to-end/sw/regexp-route.js')
```
gauntface: Two things here:

1. Arrow functions will not work with Mocha binding - not a problem until you want to use this.timeout() or this.retries so generally I just avoid fat arrows for these callbacks to avoid future confusion.
2. Nit* I'm not a fan of arrow function straight into promise chain.

packages/sw-routing/test/browser/end-to-end/sw/regexp-route.js
```diff
@@ -0,0 +1,25 @@
...
+  regExp: new RegExp('static$'),
+  handler: {
+    handle: () => Promise.resolve(new Response('static response')),
+  },
+}));
+
+routes.push(new goog.routing.RegExpRoute({
+  regExp: new RegExp('echo/(\\w+)$'),
+  handler: {
+    handle: ({params}) => Promise.resolve(new Response(params[0])),
```
gauntface: Could you make this add two or three params and return it as JSON? Minor tweak but would be a slightly strong test.

packages/sw-routing/test/browser/register-unit-tests.js
```diff
@@ -1,11 +1,11 @@
-describe('Service Worker Unit Tests', function() {
+describe('Service Worker Unit Test Registration', function() {
   beforeEach(() => window.goog.swUtils.cleanState());
 
   const swUnitTests = [
-    'sw-unit/namespace.js',
-    'sw-unit/route.js',
-    'sw-unit/express-route.js',
-    'sw-unit/regexp-route.js',
+    'unit/namespace.js',
```
gauntface: Pro the renaming (See above) would like to just standardise and implement it across the tests (Mine are due a refactor).

gauntface:
Generally +1 for these, some minor comments on style / mocha quirks, others are more on standard approach to file structure. Other than that, one small request for test change.

gauntface:
@jeffposnick not wanting to derail this (honest) but would you be open to having the following format:

- test/*.js (Node unit tests / Test Orchestrating with Selenium).
- test/browser/*.{html|js} (Unit tests that should run in the browser).
- test/sw/*.{js} (Unit tests that should run in the service worker).
- test/static/*.{js|txt|...} (Any files that are used by tests but aren't the actual test cases).

Looking at my tests they are in need of a refactoring and would be good to try and get something that is semi-self explanatory in terms of file structure.

jeffposnick:
All that sounds good. Let me know if the modified directory structure looks okay to you, and I've moved away from arrow functions for the top-level `describe()`/`it()` blocks.

If these changes look okay then I guess we can merge it in and I'll write additional e2e tests as part of the next PR.

PR #185, opened 2017-01-26
Updating sw-runtime docs

gauntface:
R: @jeffposnick @jpmedley 

Nearly done with a rough pass over all the docs.

JEFFFFFF Please note that I changes an 'await fetch().catch()' to a try {} catch (err) {} statement. Can you double check this is correct or am I missing a feature of await that means .catch will still work / acts differently to try and catch?

### DIFF
--- a/packages/sw-runtime-caching/src/lib/cache-only.js
+++ b/packages/sw-runtime-caching/src/lib/cache-only.js
@@ -20,7 +20,7 @@ import assert from '../../../../lib/assert';
  * An implementation of a [cache-only](https://developers.google.com/web/fundamentals/instant-and-offline/offline-cookbook/#cache-only)
  * request strategy.
  *
- * The advantage to using this vs. directly calling `caches.match()` is that
+ * The advantage to using this versus directly calling `caches.match()` is that
  * it will use the cache configuration and trigger the behaviors defined in
  * the underlying `RequestWrapper`.
  *
--- a/packages/sw-runtime-caching/src/lib/cache-only.js
+++ b/packages/sw-runtime-caching/src/lib/cache-only.js
@@ -17,8 +17,12 @@ import Handler from './handler';
 import assert from '../../../../lib/assert';
 
 /**
- * @memberof module:sw-runtime-caching
- * @extends Handler
+ * An implementation of a [cache-only](https://developers.google.com/web/fundamentals/instant-and-offline/offline-cookbook/#cache-only)
+ * request strategy.
+ *
+ * The advantage to using this vs. directly calling `caches.match()` is that
+ * it will use the cache configuration and trigger the behaviors defined in
+ * the underlying `RequestWrapper`.
  *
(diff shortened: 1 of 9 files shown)


packages/sw-runtime-caching/src/lib/cache-only.js
```diff
@@ -17,8 +17,12 @@ import Handler from './handler';
 import assert from '../../../../lib/assert';
 
 /**
- * @memberof module:sw-runtime-caching
- * @extends Handler
+ * An implementation of a [cache-only](https://developers.google.com/web/fundamentals/instant-and-offline/offline-cookbook/#cache-only)
+ * request strategy.
+ *
+ * The advantage to using this vs. directly calling `caches.match()` is that
```
jpmedley: Spell versus instead of abbreviating.

packages/sw-runtime-caching/src/lib/handler.js
```diff
@@ -45,11 +46,11 @@ class Handler {
...
-   *        in to the method. If used in conjunction with the `Route` class,
-   *        then the return value from the `match` function in `Route` will
-   *        be passed in via this parameter.
-   * @return {Promise.<Response>} A response, obtained from whichever strategy
-   *         is implemented.
+   *        in to the method. If used in conjunction with the
+   *        {@link module:sw-routing.Route|Route} class, then the return value
+   *        from the `match` function in the Route constructor
+   *        will be passed in as the `params` value.
+   * @return {Promise.<Response>} A promise solving with a response.
```
jpmedley: resolving, not solving.

packages/sw-runtime-caching/src/lib/network-only.js
```diff
@@ -17,8 +17,12 @@ import Handler from './handler';
 import assert from '../../../../lib/assert';
 
 /**
- * @memberof module:sw-runtime-caching
- * @extends Handler
+ * An implementation of a [network-only](https://developers.google.com/web/fundamentals/instant-and-offline/offline-cookbook/#network-only)
+ * request strategy.
+ *
+ * The advantage to using this vs. directly calling `fetch()` is that it will
```
jpmedley: vs. again.

packages/sw-runtime-caching/src/lib/request-wrapper.js
```diff
@@ -104,7 +105,16 @@ class RequestWrapper {
   }
 
   /**
-   * @return {Cache} An open `Cache` instance based on the configured
+   * This method will open a cache and maintain a reference to that cache
```
jpmedley: Change to "Opens a cache and maintains a reference ..."

gauntface:
@jpmedley all done ;)

jeffposnick:
Re: @gauntface's original question about moving to `try/catch`, I think that looks okay. I'll write some tests to confirm 😄

PR #186, opened 2017-01-27
Adding a default precaching route

gauntface:
R: @jeffposnick @addyosmani 

This will add a precaching route for the precached assets.

The warmRuntimeCache() is still not used in anyway (assumption being the developer would add routes to take advantage of the cached assets).

Let me know what you think.


packages/sw-lib/src/lib/sw-lib.js
```diff
@@ -351,6 +352,25 @@ class SWLib {
...
+   */
+  _registerDefaultRoutes() {
+    const revisionedManager = this._precacheManager.getRevisionedCacheManager();
+    const cacheFirstHandler = this.cacheFirst({
+      cacheName: revisionedManager.getCacheName(),
+    });
+
+    const route = new this.Route({
+      match: ({url, event}) => {
+        const cachedUrls = revisionedManager.getCachedUrls();
```
jeffposnick: Calling `getCachedUrls()` [is non-trivial](https://github.com/GoogleChrome/sw-helpers/blob/a1451310054b75e14882703a4cdb06dd944fdcca/packages/sw-precaching/src/lib/controllers/base-cache-manager.js#L53). How about you initialize `cachedUrls` outside of the `match()` function and then reuse the value each time `match()` is called?

gauntface: I can't call it outside of the function as the precache entries won't be defined until *after* this route is defined.

What I could do however is make the function call trivial by generating the array in the install event when the list of URL's is "finalised" / can't change. What that be ok with you?

jeffposnick: Using the `install` event sounds fine.

An alternative might be to lazy-generate the list via a getter, and then return the same list subsequent times its called? Something like

https://github.com/GoogleChrome/sw-helpers/blob/beada35eca632b46459cf1a5a336f37f988de115/packages/sw-broadcast-cache-update/src/lib/behavior.js#L103

If there are usage scenarios in which we don't need the `getCachedUrls()` value then that approach could be slightly more efficient than doing it in the `install` handler unconditionally. But we're not talking about anything too significant, so feel free to implement it however you think is cleanest.

gauntface: I am  more than happy to do that, only edge case (and it is an edge case) is to reset that variable is more entries are added, but I'll go with this approach because I think it's a little safer / easier to follow.

gauntface:
@jeffposnick the last commit has the change we discussed.

packages/sw-lib/src/lib/sw-lib.js
jeffposnick: Yeah, there are a couple of places in the various modules where we assume that properties will never change, even though they're not actually protected via setters or immutable or anything like that.

I've kind of just been hoping that it won't be a problem, since developers would normally have to go out of their way to explicitly change the internal state. We should revisit if we come across API surfaces that developers might accidentally use and which have those potential side effects.

PR #190, opened 2017-01-31
Unit tests for sw-broadcast-cache-update

jeffposnick:
R: @addyosmani @gauntface

End to end tests will follow.

### DIFF
--- a/packages/sw-broadcast-cache-update/src/index.js
+++ b/packages/sw-broadcast-cache-update/src/index.js
@@ -60,11 +60,11 @@
 import Behavior from './lib/behavior';
 import broadcastUpdate from './lib/broadcast-update';
 import {cacheUpdatedMessageType} from './lib/constants';
-import sameResponses from './lib/responses-are-same';
+import responsesAreSame from './lib/responses-are-same';
 
 export {
   Behavior,
   broadcastUpdate,
   cacheUpdatedMessageType,
-  sameResponses,
+  responsesAreSame,
 };
--- a/packages/sw-broadcast-cache-update/test/sw/behavior.js
+++ b/packages/sw-broadcast-cache-update/test/sw/behavior.js
@@ -17,7 +17,14 @@ describe('Test of the Behavior class', function() {
   const source = 'test-source';
 
   it(`should throw when Behavior() is called without any parameters`, function() {
-    expect(() => new goog.broadcastCacheUpdate.Behavior()).to.throw();
+    let thrownError = null;
+    try {
+      new goog.broadcastCacheUpdate.Behavior();
+    } catch(err) {
+      thrownError = err;
+    }
+    expect(thrownError).to.exist;
+    expect(thrownError.name).to.equal('channel-name-required');
   });
 
   it(`should use the channelName from the constructor`, function() {
@@ -49,6 +56,9 @@ describe('Test of the Behavior class', function() {
     const behavior = new goog.broadcastCacheUpdate.Behavior({channelName});
     const broadcastChannel = behavior.channel;
     expect(broadcastChannel).to.be.instanceof(BroadcastChannel);
+    // behavior.channel is a getter that create a BroadcastChannel the first
+    // time it's called, and this test confirms that it returns the same
+    // BroadcastChannel object when called twice.
     expect(broadcastChannel).to.eql(behavior.channel);
     expect(broadcastChannel.name).to.equal(channelName);
   });
--- a/packages/sw-broadcast-cache-update/test/sw/behavior.js
+++ b/packages/sw-broadcast-cache-update/test/sw/behavior.js
@@ -1,0 +1,55 @@
+importScripts(
+  '/node_modules/mocha/mocha.js',
+  '/node_modules/chai/chai.js',
+  '/node_modules/sw-testing-helpers/build/browser/mocha-utils.js',
+  '/packages/sw-broadcast-cache-update/build/sw-broadcast-cache-update.min.js'
+);
+
+const expect = self.chai.expect;
+mocha.setup({
+  ui: 'bdd',
+  reporter: null,
+});
+
+describe('Test of the Behavior class', function() {
+  const channelName = 'test-channel';
+  const headersToCheck = ['one', 'two'];
+  const source = 'test-source';
+
+  it(`should throw when Behavior() is called without any parameters`, function() {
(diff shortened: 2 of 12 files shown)


packages/sw-broadcast-cache-update/src/index.js
```diff
@@ -60,11 +60,11 @@
...
 import {cacheUpdatedMessageType} from './lib/constants';
-import sameResponses from './lib/responses-are-same';
+import responsesAreSame from './lib/responses-are-same';
 
 export {
   Behavior,
   broadcastUpdate,
   cacheUpdatedMessageType,
-  sameResponses,
+  responsesAreSame,
```
jeffposnick: I realized when writing the tests that this was being exported with the wrong name...

packages/sw-broadcast-cache-update/test/sw/behavior.js
```diff
@@ -0,0 +1,55 @@
...
+  reporter: null,
+});
+
+describe('Test of the Behavior class', function() {
+  const channelName = 'test-channel';
+  const headersToCheck = ['one', 'two'];
+  const source = 'test-source';
+
+  it(`should throw when Behavior() is called without any parameters`, function() {
+    expect(() => new goog.broadcastCacheUpdate.Behavior()).to.throw();
```
gauntface: Could you define an error that is expected so that we know it's not some other error, for example browserCacheUpdate is undefined causing the error rather that the empty constructor causing the error..

packages/sw-broadcast-cache-update/test/sw/behavior.js
```diff
@@ -0,0 +1,55 @@
...
+  it(`should use a default value for source when one isn't provided`, function() {
+    const behavior = new goog.broadcastCacheUpdate.Behavior({channelName});
+    expect(behavior.source).to.not.be.empty;
+  });
+
+  it(`should create and reuse a BroadcastChannel based on channelName`, function() {
+    const behavior = new goog.broadcastCacheUpdate.Behavior({channelName});
+    const broadcastChannel = behavior.channel;
+    expect(broadcastChannel).to.be.instanceof(BroadcastChannel);
+    expect(broadcastChannel).to.eql(behavior.channel);
```
gauntface: This check isn't needed because the broadcastChannel was just pulled form behavior.channel.

packages/sw-broadcast-cache-update/test/sw/broadcast-update.js
```diff
@@ -0,0 +1,40 @@
...
+  const source = 'test-source';
+
+  it(`should throw when broadcastUpdate() is called without any parameters`, function() {
+    expect(() => goog.broadcastCacheUpdate.broadcastUpdate()).to.throw();
+  });
+
+  it(`should trigger the appropriate message event on a BroadcastChannel with the same channel name`, function(done) {
+    const secondChannel = new BroadcastChannel(channelName);
+    goog.broadcastCacheUpdate.broadcastUpdate({channel, cacheName, source, url});
+    secondChannel.addEventListener('message', (event) => {
```
gauntface: Would it be safer to add the event listener before broadcasting the update?

packages/sw-broadcast-cache-update/test/sw/responses-are-same.js
```diff
@@ -0,0 +1,58 @@
...
+  reporter: null,
+});
+
+describe('Test of the responsesAreSame function', function() {
+  const firstHeaderName = 'x-first-header';
+  const secondHeaderName = 'x-second-header';
+  const headersToCheck = [firstHeaderName, secondHeaderName];
+
+  it(`should throw when responsesAreSame() is called without any parameters`, function() {
+    expect(() => goog.broadcastCacheUpdate.responsesAreSame()).to.throw();
```
gauntface: Could you define the error to be thrown here.

packages/sw-broadcast-cache-update/test/sw/responses-are-same.js
```diff
@@ -0,0 +1,58 @@
...
+      [firstHeaderName]: 'same', [secondHeaderName]: 'same'}});
+    expect(goog.broadcastCacheUpdate.responsesAreSame({
+      first, second, headersToCheck})).to.be.true;
+  });
+
+  it(`should return true when one header matches and the other is missing`, function() {
+    const first = new Response('', {headers: {
+      [firstHeaderName]: 'same'}});
+    const second = new Response('', {headers: {
+      [firstHeaderName]: 'same'}});
```
gauntface: I think this is meant to have a second header that exist in the first response

packages/sw-broadcast-cache-update/test/sw/behavior.js
jeffposnick: Sure thing. I've been thinking more about https://github.com/GoogleChrome/sw-helpers/issues/98 and I think we can end up in a place where we have non-generic, contextual `Error`s without all the overhead of configuring an instance of an `ErrorFactory`. But for now I'll go that route.

packages/sw-broadcast-cache-update/test/sw/behavior.js
jeffposnick: [`.channel` is a getter](https://github.com/GoogleChrome/sw-helpers/blob/e97bebda4b7e50873935a18d35b68bfb809479ab/packages/sw-broadcast-cache-update/src/lib/behavior.js#L98), and this is specifically checking to make sure that when it's called twice in succession, it returns the same `BroadcastChannel`.

That's completely non-obvious from the test, though, so I'll add a comment.

packages/sw-broadcast-cache-update/test/sw/broadcast-update.js
jeffposnick: Yes. Switching.

packages/sw-broadcast-cache-update/test/sw/responses-are-same.js
jeffposnick: Will do.

packages/sw-broadcast-cache-update/test/sw/responses-are-same.js
jeffposnick: It's not; this is testing a specific condition in which both `Response`s are missing the same header, and that should be `true`. I'll update the description of the test to make that clear.

jeffposnick:
PTAL, @gauntface.

PR #194, opened 2017-02-04
Test harness template, along with migration of sw-broadcast-cache-upd…

jeffposnick:
We discussed moving some of the HTML test harness logic to a template. Here's one approach—let me know what you think, @gauntface. Happy to make change or just abandon if you're not in favor. It does simplify setting up new package tests at the expense of extra "magic" that might make it more inaccessible to new developers.

I've just moved over one package's tests to use the template, but would theoretically move over the rest if we want to go with this approach.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "gulp-rollup": "^2.5.1",
     "gulp-sourcemaps": "^2.4.0",
     "gulp-spawn-mocha": "^3.1.0",
+    "handlebars": "^4.0.6",
     "jsdoc-strip-async-await": "^0.1.0",
     "lerna": "2.0.0-beta.30",
     "minimist": "^1.2.0",
--- a/packages/sw-broadcast-cache-update/test/automated-test-suite.js
+++ b/packages/sw-broadcast-cache-update/test/automated-test-suite.js
@@ -34,7 +34,7 @@ describe(`${packageName} Browser Tests`, function() {
 
   // Set up the web server before running any tests in this suite.
   before(() => testServer.start('.').then((portNumber) => {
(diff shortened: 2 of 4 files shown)


gauntface:
Generally I'm all for this.

One convention I've tried to stick to with the server is to start generated requests with '__', so could we change the endpoint to something similar to: '/__test/mocha/:pkg/' just so it's clear there is no file and it's something served up by our test server.

jeffposnick:
I was considering the `__` prefix, too, but the tests that need to pull in additional subresources end up being cleaner if you can use relative URLs and assume that the `<pkg>/test/browser/` is always going to be the base, instead of using some synthetic `__` URL as a the base.

Of course, we could work around that by setting `<base href="/packages/<pkg>/test/browser/>` in the generated HTML. The downside is that it's even more "magic" going on. But it would be a cleaner approach. Want me to do that?

gauntface:
That's tough but I think I would favor the base trick. Basically if there is no __test directory, hopefully new users will hunt for where it's defined, we could just add a comment in service instance explaining the base trick.

The only other consideration I had is that at the moment I tend to use the serve-index to navigate files to get to a projects test index.html page, which I won't be able to do, so maybe we could add an index page that links to each projects test page? (This is an aside to this PR, just wanted to float the idea)

jeffposnick:
I can add in a 301 redirect in the Express code from `/packages/:pkg/test/browser/` to `/__test/mocha/:pkg`.

gauntface:
Let's do it. If it gets in the way of something else we can address it then.

Thanks for looking into all this Jeff.

jeffposnick:
The CI is still finishing, but PTAL and merge if it looks good. I'll do a follow-up PR to migrate over the other existing tests.

PR #199, opened 2017-02-07
Refactoring sw-cli - node module + CLI usage

gauntface:
R: @jeffposnick @addyosmani

Apologies that this is such a big PR.

There isn't much that has changed but the individual steps that existed to generate the SW / build manifest are now seperate files - advantage is that they are self contained.

Logic for service worker generation is pulled out from CLI and can be used as a node module as well with this PR.

Once this PR is ok to land I'll write some docs for how to this use as a node module (And write some tests to ensure the surface actually works).


jeffposnick:
I want to give the CLI a test drive before commenting more, but I'm running into:

```sh
[~/git/sw-helpers/packages/sw-cli/build/cli]$ node bin.js generate-sw
? What is the root of your web app? (Use arrow keys)
  ──────────────
❯ Manually Enter Path Unable to get the root directory of your web app.
TypeError: Path must be a string. Received undefined
    at assertPath (path.js:7:11)
    at Object.join (path.js:1211:7)
    at Promise.then.then.then.then (/Users/jeffy/git/sw-helpers/packages/sw-cli/build/cli/index.js:236:19)
```

That error pops up immediately, before I have the chance to enter anything.

jeffposnick:
The error turned out to be due to an old version of `inquisitor` that was being used by the script due to me not having installed the `sw-cli`-specific `node_modules`.

I don't think I'll have time to thoroughly test this out tonight, but will give some feedback tomorrow.

jeffposnick:
I was able to run the script successfully and generated a the manifest, SW file, and config file in my current directory. I then re-ran it to see if it would regenerate the manifest, and got the following:

```sh
[~/git/sw-helpers/packages/sw-routing/demo]$ node ~/git/sw-helpers/packages/sw-cli/build/cli/bin.js build-file-manifest
/Users/jeffy/git/sw-helpers/packages/sw-cli/build/cli/index.js:121
        return this.buildFileManifest();
                    ^

TypeError: this.buildFileManifest is not a function
    at SWCli.handleCommand (/Users/jeffy/git/sw-helpers/packages/sw-cli/build/cli/index.js:121:21)
    at SWCli.argv (/Users/jeffy/git/sw-helpers/packages/sw-cli/build/cli/index.js:53:19)
    at Object.<anonymous> (/Users/jeffy/git/sw-helpers/packages/sw-cli/build/cli/bin.js:21:11)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Module.runMain (module.js:605:10)
    at run (bootstrap_node.js:418:7)
```

packages/sw-cli/src/lib/utils/generate-glob-pattern.js
```diff
@@ -0,0 +1,14 @@
...
+module.exports = (relativePath, fileExtentionsToCache) => {
+  // Glob patterns only work with forward slash
+  // https://github.com/isaacs/node-glob#windows
+  const globPath = path.join(relativePath, '**', '*').replace(path.sep, '/');
+  if (fileExtentionsToCache.length > 1) {
+    // Return pattern '**/*.{txt,md}'
+    return globPath + `.{${fileExtentionsToCache.join(',')}}`;
+  } else {
+    // Return pattern '**/*.txt'
+    return globPath + `.${fileExtentionsToCache[0]}`;
```
gauntface: Very good point - it wasn't an issue for the CLI because I had a validation check, but for node module - yes it could be. (I'll add tests for these in a seperate PR because chances are there will be more edge cases like that)

gauntface:
@jeffposnick I just added some changes that you requested.

The build manifest was never really implemented and tested, so I've just altered it to print an error message and exit with a bad status code.

My plan is to implement the following PRs:

1. An API for generating SW when used as a node-module.
1. An API for generating a manifest used as a node-module.
1. Add in support for creating and reading in the config file for CLI and node module.
1. Add in support for building manifest from CLI (either as Questions or config)

Happy to re-order that list if there is anything you think is more or less important.

jeffposnick:
Okay, the incremental approach sounds fine.

PR #203, opened 2017-02-09
Rename "[Bb]ehavior" to "[Pp]lugin"

### LINKED ISSUE #196: Rename "behavior" to "plugin"
Feedback from @devnook: the term "behavior" is a lot less common than "plugin". We should consider just adopting "plugin" throughout the project.

This has the advantage of kinda-sorta looking like how [Webpack supports extensibility](https://webpack.github.io/docs/using-plugins.html#built-in-plugins).

jeffposnick:
R: @addyosmani @gauntface

Fixes #196 

Based on feedback from @devnook


jeffposnick:
Fixed the test, and merging.

PR #204, opened 2017-02-10
requestWillFetch event

### LINKED ISSUE #201: Add a new requestWillFetch lifecycle callback
**Library Affected**:
*sw-runtime-caching*

There's a use case for plugins (*née* behaviors) that modify an outgoing `Request` before it's requested from the network, via `fetch()`. The request URL might be rewritten, a header might be added, etc.

We should add in a new lifecycle callback that is invoked by `RequestWrapper` [right before the `fetch()` is performed](https://github.com/GoogleChrome/sw-helpers/blob/c26d5b81f1ec4c8417d4e657712e3518d3e43826/packages/sw-runtime-caching/src/lib/request-wrapper.js#L182). A plugin that implemented the callback would expose a method like:

```js
requestWillFetch({request}) {
  // do something synchronously, and return a Request object
}
```

and then the `Request` object that the plugin's `requestWillFetch()` returns will be either be forwarded on to the next registered callback (if there are additional callbacks) or used as a parameter to the actual `fetch()` (if it's the last callback).

@mullens, does that sound sufficient for your use case?

Assigning to @prateekbh as he expressed an interest in implementing the changes to `RequestWrapper` to support this.

prateekbh:

Fixes #201

Adding code to support new event requestWillFetch.
Plugins will be able to make changes to request before they are actually being fetched.


packages/sw-runtime-caching/src/lib/request-wrapper.js
```diff
@@ -178,6 +178,18 @@ class RequestWrapper {
   async fetch({request}) {
     assert.atLeastOne({request});
 
+    if (this.behaviorCallbacks.requestWillFetch) {
```
jeffposnick: `this.behaviorCallbacks` was renamed to `this.pluginCallbacks` (see https://github.com/GoogleChrome/sw-helpers/pull/203).

Sorry!

packages/sw-runtime-caching/src/lib/request-wrapper.js
```diff
@@ -178,6 +178,18 @@ class RequestWrapper {
...
+      for (let callback of this.behaviorCallbacks.requestWillFetch) {
```
jeffposnick: `this.behaviorCallbacks` was renamed to `this.pluginCallbacks` (see https://github.com/GoogleChrome/sw-helpers/pull/203).

Sorry!

packages/sw-runtime-caching/src/lib/request-wrapper.js
```diff
@@ -178,6 +178,18 @@ class RequestWrapper {
...
+        const returnedRequest = callback({request});
+        if (returnedRequest && returnedRequest.then) {
```
jeffposnick: I'd rather we just stick to only supporting synchronous callbacks initially, unless we have an actual need for `async` callbacks or callbacks that return `Promise`s. It will make the documentation cleaner and simpler.

gauntface: +1 Jeff's comments here.

prateekbh: doesn't webP plugin already had one? saving device's pixel ratio/DPR in IDB and using it while modifying it in IDB?

jeffposnick: I wasn't aware of that detail.

In that case, I'd say we should just always expect the callback to return a `Promise<Request>`, since it's easy enough to for a synchronous callback to use `Promise.resolve(Request)`.

gauntface: +1 Jeff's comment and if we role with this please change the check for an explicit Promise rather than a check for the "then" parameter

packages/sw-runtime-caching/src/lib/request-wrapper.js
prateekbh: No worries! I personally like plugins better than Behavours

packages/sw-runtime-caching/src/lib/request-wrapper.js
prateekbh: Done

packages/sw-runtime-caching/src/lib/request-wrapper.js
prateekbh: makes sense 👍 
changed code to use `assert` and now only expect the callback to return a promise

prateekbh:
@gauntface @jeffposnick changes made :)

PR #209, opened 2017-02-14
Generate-sw from node module

gauntface:
R: @jeffposnick @addyosmani 

This PR is just a re-jig of the utils and making the generateSW method work as a node module (with an ok API).

The next PR I'll get the build manifest working (Should enable anyone to play with adding it to their build process).


packages/sw-cli/src/lib/generate-sw.js
```diff
@@ -1,33 +1,70 @@
...
+ *   rootDirectory: './build/',
+ *   globPatterns: ['**\/*.{html,js,css}'],
+ *   globIgnores: ['admin.html']
+ *   serviceWorkerName: 'sw.js'
+ * })
+ * .then(() => {
+ *   console.log('Service worker generated.');
+ * });
+ *
+ * This method will generate a working Service Worker with an inlined
```
jpmedley: The phrase 'service worker` should be lower case. Please check throughout.

packages/sw-cli/src/lib/generate-sw.js
```diff
@@ -1,33 +1,70 @@
...
- * @return {Promise} The promise returned here will be used to exit the
- * node process cleanly or not.
+ * @param {Object} input
+ * @param {String} input.rootDirectory The root of the files you wish to
+ * be cached. This will also be the directory the service worker and library
+ * files are written to.
+ * @param {Array<String>} input.globPatterns Patterns to glob for when
+ * generating the build manifest.
+ * @param {String|Array<String>} input.globIgnores Add glob patterns that
+ * should exclude files from the build manifest.
```
jpmedley: Always use parallel structures for descriptions of parameters. Specifically, this one should match `input.globPatterns` so:

'Patterns to exclude when generating the build manifest.'

jeffposnick:
👍

PR #211, opened 2017-02-15
sw-cache-expiration unit tests

jeffposnick:
R: @addyosmani @gauntface

Unit tests for `sw-cache-expiration`. There's also a bug fix for an issue I found in the underlying library (yay testing!) and I'm exposing a few additional symbols that could be useful.

The unit tests are fairly in depth, so I'm not sure that I'm going to prioritize writing end-to-end tests. The in-depth-ness also means that they're a little gnarly, so please feel free to push back if it's too unclear what's going on.

### DIFF
--- a/packages/sw-cache-expiration/src/lib/error-factory.js
+++ b/packages/sw-cache-expiration/src/lib/error-factory.js
@@ -1,0 +1,27 @@
+/*
+ Copyright 2016 Google Inc. All Rights Reserved.
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+import ErrorFactory from '../../../../lib/error-factory';
+
+const errors = {
+  'max-entries-or-age-required': `Either the maxEntries or maxAgeSeconds
+    parameters (or both) are required when constructing Plugin.`,
+  'max-entries-must-be-number': `The maxEntries parameter to the Plugin
+    constructor must either be a number or undefined.`,
+  'max-age-seconds-must-be-number': `The maxAgeSeconds parameter to the Plugin
+    constructor must either be a number or undefined.`,
+};
+
+export default new ErrorFactory(errors);
--- a/packages/sw-cache-expiration/src/lib/plugin.js
+++ b/packages/sw-cache-expiration/src/lib/plugin.js
@@ -75,18 +75,21 @@ class Plugin {
    * @param {string} input.cacheName Name of the cache the Responses belong to.
    * @return {DB} An open DB instance.
    */
-  async getDB({cacheName}) {
-    if (!this._dbs.has(cacheName)) {
-      const openDb = await idb.open(idbName, idbVersion, (upgradeDB) => {
+  async getDB({cacheName} = {}) {
+    assert.isType({cacheName}, 'string');
+
+    const idbId = `${idbName}-${cacheName}`;
+    if (!this._dbs.has(idbId)) {
+      const openDb = await idb.open(idbId, idbVersion, (upgradeDB) => {
         const objectStore = upgradeDB.createObjectStore(cacheName,
           {keyPath: urlPropertyName});
         objectStore.createIndex(timestampPropertyName, timestampPropertyName,
           {unique: false});
       });
-      this._dbs.set(cacheName, openDb);
+      this._dbs.set(idbId, openDb);
     }
 
-    return this._dbs.get(cacheName);
+    return this._dbs.get(idbId);
   }
 
   /**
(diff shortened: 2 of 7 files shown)


packages/sw-cache-expiration/src/lib/plugin.js
```diff
@@ -303,7 +326,9 @@ class Plugin {
    * @param {string} input.cacheName Name of the cache the Responses belong to.
    * @return {Array<string>} A list of the URLs that were expired.
    */
-  async findExtraEntries({cacheName}) {
+  async findExtraEntries({cacheName} = {}) {
```
gauntface: Nit: But this name seems like the opposite of the description: "findExtraEntries" > "Expires entries base on the the maximum cache size."

packages/sw-cache-expiration/test/sw/plugin.js
```diff
@@ -0,0 +1,226 @@
...
+    return plugin.getCache({cacheName}).then((firstCache) => {
+      return plugin.getCache({cacheName}).then((secondCache) => {
+        expect(firstCache).to.eql(secondCache);
+      });
+    });
+  });
+
+  it(`should return true when there's no cachedResponse passed to isResponseFresh()`, function() {
+    const plugin = new Plugin({maxAgeSeconds});
+    expect(plugin.isResponseFresh()).to.be.true;
```
gauntface: This feels a bit of a weird test in that I would have expected `inResponseFresh()` to throw if there was not response to check for freshness.

Is there a use case for this returning true?

packages/sw-cache-expiration/test/sw/plugin.js
```diff
@@ -0,0 +1,226 @@
...
+    const plugin = new Plugin({maxAgeSeconds});
+    const date = new Date(now).toUTCString();
+    const cachedResponse = new Response('', {headers: {date}});
+    expect(plugin.isResponseFresh({cachedResponse, now})).to.be.true;
+  });
+
+  it(`should return false when isResponseFresh() is called and the Date: header in the cachedResponse is not recent`, function() {
+    const plugin = new Plugin({maxAgeSeconds});
+    // This will construct a date that is 1 second past the expiration.
+    const date = new Date(now - ((1 + maxAgeSeconds) * 1000)).toUTCString();
```
gauntface: Feel free to ignore this but worth considering:

You could condense some of these tests down into a single unit test (i.e. fresh with time as normal, then advance time and test again, still fresh, then advance past the maximum and ensure not fresh).

Using sinon fake timers you can control the time of the environment: (Example of using them here: https://github.com/gauntface/route/blob/master/packages/device-clock/test/node/device-clock.js#L17 and advance with https://github.com/gauntface/route/blob/master/packages/device-clock/test/node/device-clock.js#L46) Was thoroughly impressed how easy it was to use - no worries if you are happy leaving as is though.

packages/sw-cache-expiration/test/sw/plugin.js
```diff
@@ -0,0 +1,226 @@
...
+    const cachedResponse = new Response('', {headers: {date}});
+    expect(plugin.isResponseFresh({cachedResponse, now})).to.be.false;
+  });
+
+  it(`should return cachedResponse when cacheWillMatch() is called and isResponseFresh() is true`, function() {
+    const plugin = new Plugin({maxAgeSeconds});
+    const date = new Date(now).toUTCString();
+    const cachedResponse = new Response('', {headers: {date}});
+    expect(plugin.cacheWillMatch({cachedResponse, now})).to.eql(cachedResponse);
```
gauntface: Have you added now as an argument to accommodate testing or is it useful as part of the API?

packages/sw-cache-expiration/src/lib/plugin.js
jeffposnick: Thanks; they JSDocs were wrong.

packages/sw-cache-expiration/test/sw/plugin.js
jeffposnick: The underlying method gets called as part of the `RequestWrapper` lifecycle callback `cacheWillMatch`. `cacheWillMatch` might be called with an undefined `Response` passed in if there isn't yet an entry in the cache for a given URL, and we just want to return `true` in that case to work with the semantics that `RequestWrapper` expects.

packages/sw-cache-expiration/test/sw/plugin.js
jeffposnick: I could imagine someone who's using the underlying module outside of the larger framework wanting that level of control.

packages/sw-cache-expiration/src/lib/error-factory.js
```diff
@@ -0,0 +1,27 @@
+/*
+ Copyright 2016 Google Inc. All Rights Reserved.
```
jpmedley: Should the copyright year be updated?

packages/sw-cache-expiration/test/sw/plugin.js
jeffposnick: I was thinking of testing that sort of scenario as part of end-to-end tests, but my goal right now is to get some sort of test coverage for all the code, and then go back and cover additional scenarios. So I'll defer on that for now.

packages/sw-cache-expiration/src/lib/error-factory.js
jeffposnick: We could, but let's just do that in a separate PR that covers the entire set of modules if we care.

PR #213, opened 2017-02-15
Get file manifest entries

gauntface:
R: @jeffposnick @addyosmani

This will allow developers to get a JavaScript Array of Objects containing a URL and revision string. When this is ok to land, I'll add the method that will generate the file manifest and write to a file (Basically take this an produce the JS file for it).


packages/sw-cli/src/lib/get-file-manifest-entries.js
```diff
@@ -0,0 +1,49 @@
...
+ * To get a list of files and revision details that can be used to ultimately
+ * precache assets in a service worker.
+ *
+ * @param {Object} input
+ * @param {Array<String>} input.globPatterns  Patterns used to select files to
+ * include in the file entries.
+ * @param {Array<String>} input.globIgnores  Patterns used to exclude files
+ * from the file entries.
+ * @param {String} input.rootDirectory The directory run the glob patterns over.
+ * @return {Array<Object>} An array of ManifestEntries will include
```
jeffposnick: "ManifestEntries" is used here as if it's a class name, which I don't think is the case. Could you actually include a sample of what each entry in the return array will look like? Either here in the `@return` or in the main JSDoc for this function.

PR #214, opened 2017-02-15
Build manifest file

gauntface:
R: @jeffposnick @addyosmani

This adds the ability to build manifest file. I'm certain there will be plenty of things to change but this should be a useful first step.


packages/sw-cli/src/lib/generate-file-manifest.js
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * @memberof module:sw-cli
+ */
+const generateFileManifest = (input) => {
+  if (!input || typeof input !== 'object' || input instanceof Array) {
+    return Promise.reject(
+      new Error(errors['invalid-generate-file-manifest-arg']));
+  }
+
+  const manifestFilePath = input.manifestFilePath;
```
jeffposnick: I'm not sold on `manifestFilePath` as the parameter name here. I feel like something with `output` or `dest` would be more explicit.

(I'm looking at what `webpack` currently does, since folks might be familiar with that, and they [use](https://webpack.js.org/guides/get-started/#using-webpack-with-a-config) `output: {filename, path}`, though breaking it up into separate properties for the file and path is overkill for us.)

gauntface:
@jeffposnick changed the name to dest and new builds seem happy :)

packages/sw-cli/src/lib/generate-file-manifest.js
```diff
@@ -0,0 +1,53 @@
+const getFileManifestEntries = require('./get-file-manifest-entries');
+const writeFileManifest = require('./utils/write-file-manifest');
+const errors = require('./errors');
+
+/**
+ * @example <caption>Generate a service worker for a project.</caption>
+ * const swCLI = require('sw-cli');
+ *
+ * swCLI.generateFileManifest({
+ *   manifestFilePath: './build/manifest.js'
```
jeffposnick: This should be updated to `dest`.

packages/sw-cli/src/index.js
```diff
@@ -33,6 +34,19 @@ const getFileManifestEntries = require('./lib/get-file-manifest-entries');
  *   console.log('Service worker generated.');
  * });
  *
+ * @example <caption>Generate a build manifest file.</caption>
+ * const swCLI = require('sw-cli');
+ *
+ * swCLI.generateFileManifest({
+ *   manifestFilePath: './build/scripts/manifest.js',
```
jeffposnick: This should be updated to `dest`.

PR #221, opened 2017-02-21
Add ES2015 .mjs module output to the sw-lib build process

jeffposnick:
R: @gauntface

Adding in `.mjs` output seems to work fine, and it's needed to consume `sw-lib` as an ES module.

Hopefully this was just an oversight and there isn't a technical reason why we can't output an ES module here.


gauntface:
It was simply that I wasn't sure if this would work given that it's written with requires - if it works and you've tested then LGTM.

jeffposnick:
Yep, I've tried the change locally and it works when pulled in via Rollup.

PR #225, opened 2017-02-21
Adds a format option to writeFileManifest(), defaulting to 'iife'

jeffposnick:
R: @addyosmani @gauntface

Fixes #221 

I've added an extra test to confirm that the exception we expect is thrown when a bad format is passed in. I didn't add an extra test to actually confirm that the output of the `writeFileManifest({format: 'es'})` command is a valid ES2015 file with the expected entries, though. I didn't see where the equivalent test was to confirm that the previous output was valid—if you could point me in the right direction, @gauntface, I'll add that test to this PR.

### DIFF
--- a/packages/sw-build/src/lib/generate-file-manifest.js
+++ b/packages/sw-build/src/lib/generate-file-manifest.js
@@ -30,10 +30,8 @@ const errors = require('./errors');
  * generating the build manifest.
  * @param {String|Array<String>} input.globIgnores Patterns to exclude when
  * generating the build manifest.
- * @param {String} [input.format] Either [`'iife'`](https://en.wikipedia.org/wiki/Immediately-invoked_function_expression)
- * (the default) or `'es'`. If you plan on reading in the manifest from your
- * service worker via [`importScripts()`](), then use `'iife'`. If you plan on
- * reading in the manifest as an imported ES2015 module, use `'es'`.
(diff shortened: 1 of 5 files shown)


packages/sw-build/src/lib/generate-file-manifest.js
```diff
@@ -28,6 +30,10 @@ const errors = require('./errors');
  * generating the build manifest.
  * @param {String|Array<String>} input.globIgnores Patterns to exclude when
  * generating the build manifest.
+ * @param {String} [input.format] Either [`'iife'`](https://en.wikipedia.org/wiki/Immediately-invoked_function_expression)
```
gauntface: This is a little wordy for reference doc, could we reword to something along the lines of:

Default format is an [`'iife'`](https://en.wikipedia.org/wiki/Immediately-invoked_function_expression), but also accepts `'es'`, which outputs an ES2015 module. 

Then open an issue to document how to use sw-build with a build process and the options available and link from this comment to those docs (once written)?

gauntface:
I don't think there is a good test for the validity of the manifest file.

Because it's inlined in the generated sw (Which is where all the tests were originally written for) it's assumed it works, but you are right, this should have additional tests. I'd likely take the approach of running the JS file through vm.runInNewContext(JSfileContent) and examine the values the file defines.

Maybe raise and issue to implement the tests and not block on this PR.

PR #226, opened 2017-02-21
Manifest gen

gauntface:
R: @jeffposnick @addyosmani 

This will add the command to the CLI to generate the manifest only.

    $ sw-cli generate-manifest

It needs a question for the format to publish but can add that in a separate PR or after the other open PR has landed.


packages/sw-cli/src/index.js
```diff
@@ -119,9 +120,8 @@ class SWCli {
     switch (command) {
       case 'generate-sw':
         return this._generateSW();
-      case 'build-file-manifest':
-        logHelper.error(`TODO: Implement.`);
-        return Promise.reject();
+      case 'generate-manifest':
```
addyosmani: Sorry for the late comment. For the initial MVP of sw-cli, I think opting for the generate-* style commands makes sense. Longer term we might want to consider whether we'd benefit from grouping these the way we do in Yeoman: e.g `sw-cli generate:sw` `sw-cli generate:manifest` etc. 

We might also want to think about whether we'll ever need to support method chaining for these commands, but that can come later.

PR #230, opened 2017-02-23
Make the Errors generated by assert more useful.

### LINKED ISSUE #98: Normalize ErrorFactory and assertion libraries
Placeholder bug to remind myself that there's work to be done to normalize how we're using `Error` objects in the `assert` library and the `ErrorFactory` stuff.

More details to come.

jeffposnick:
R: @addyosmani @gauntface

Fixes #98 

This might be a little controversial, so feel free to push back.

The changes here don't end up using `ErrorFactory` in the `assert` module. I think they add functionality to `assert` that brings it on par with what you'd get with `ErrorFactory`, without the overhead of explicitly maintaining a large mapping of error identifiers. Folks are obviously free to continue using `ErrorFactory` where appropriate, e.g. in contexts that assertions don't make sense.

With this change in place, failing assertions that were defined like:

```js
class Router {
  registerRoutes({routes} = {}) {
    assert.isInstance({routes}, Array);
    // ...
  }
}
```

produce `Error` objects that have fields like:

```json
{
  "message": "Invalid call to Router.registerRoutes() — The 'routes' parameter must be an instance of 'Array'",
  "name": "assert.isInstance"
}
```

The `message` is set to something meaningful, including the underlying method that triggered the assertion failure, and featuring unmangled method/class names.

The `name` is set to the specific type of assertion that failed—`assert.isInstance`, `assert.atLeastOne`, etc.—meaning that you can use the unit testing pattern of checking `ex.name` to ensure that an expected error is thrown (rather than an unexpected or generic error).

We can theoretically extend this code to automatically include a URL pointing to the specific documentation for the method that failed—i.e. map `Router.registerRoutes()` to https://googlechrome.github.io/sw-helpers/reference-docs/stable/v0.0.8/module-sw-routing.Router.html#registerRoutes, but I haven't tackled that yet.

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -15,10 +15,12 @@
 
 /* eslint-disable require-jsdoc */
 
+import ErrorStackParser from 'error-stack-parser';
+
 function atLeastOne(object) {
   const parameters = Object.keys(object);
   if (!parameters.some((parameter) => object[parameter] !== undefined)) {
-    throw Error('Please set at least one of the following parameters: ' +
+    throwError('Please set at least one of the following parameters: ' +
       parameters.map((p) => `'${p}'`).join(', '));
   }
 }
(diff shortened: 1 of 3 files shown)


gauntface:
Big +1 on this. The stack info with function name + variable name should be more than enough to help developers and ultimately saves on custom errors. Nice work.

PR #235, opened 2017-02-23
Adding cacheable response plugin

gauntface:
R: @jeffposnick @addyosmani 

This adds `sw-cachable-response` to sw-lib, just to make it friendly.

This is part of the writing up a doc on request support (https://github.com/GoogleChrome/sw-helpers/issues/78) and noticed it could be easier

### DIFF
--- a/packages/sw-lib/test/browser-unit/library-namespace.js
+++ b/packages/sw-lib/test/browser-unit/library-namespace.js
@@ -16,6 +16,32 @@
 describe('Test Behaviors of Loading the Script', function() {
   this.timeout(5 * 60 * 1000);
 
+  const deleteIndexedDB = () => {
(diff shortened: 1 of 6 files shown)


packages/sw-lib/test/browser-unit/library-namespace.js
```diff
@@ -16,6 +16,32 @@
 describe('Test Behaviors of Loading the Script', function() {
   this.timeout(5 * 60 * 1000);
 
+  const deleteIndexedDB = () => {
+    return new Promise((resolve, reject) => {
+      // TODO: Move to constants
+      const req = indexedDB.deleteDatabase('sw-precaching');
```
addyosmani: Can we file issues (where there's time) for some of the TODOs?

PR #236, opened 2017-02-23
Docs on Caching CDN Requests

### LINKED ISSUE #78: Revamp the defaults around opaque responses

gauntface:

Fixes #78

This is to close this issue, kick start some sort of doc content and I'm assuming someone on the hack day will hit this problem.


docs/advanced/index.md
```diff
@@ -0,0 +1,59 @@
...
+In sw-lib, you can set up some additional options with a route that will allow
+support for additional status codes and / or based on specific headers.
+
+```javascript
+const cdnCacheStrategy = goog.swlib.staleWhileRevalidate({
+  cacheableResponse: {
+    statuses: [0],
+  },
+});
+goog.swlib.router.registerRoute(/http://cdn.mysite.com/styles/.*/, cdnCacheStrategy);
```
jeffposnick: Could you switch this example to use `new RegExp('^https://cdn.mysite.com/styles/')`

- It uses `https` instead of `http`.
- It uses a `RegExp(string)`-style constructor, avoiding the need to escape the `/` characters inside of a RegExp literal.
- It ensures that the match is done against the beginning of the URL string, which is where we expect the `https://...` to appear.
- It omits the `.*`, as it's not necessary. It would only be needed if there was a trailing `$` character in the RegExp. As long as the RegExp matches, even if it doesn't match the entire URL string, it will trigger the route.

(This tells me that the docs around RegExp-based routes should be improved. I had a similar conversation with @jpmedley earlier.)

docs/advanced/index.md
```diff
@@ -0,0 +1,59 @@
...
+```
+
+This same approach can be used with all of th caching strategies supported by
+sw-lib including `cacheFirst()`, `cacheOnly()`, `networkFirst()`,
+`networkOnly()` and `StaleWhileRevalidate()`.
```
jeffposnick: `StaleWhileRevalidate` ➡️  `staleWhileRevalidate`

docs/advanced/index.md
```diff
@@ -0,0 +1,59 @@
...
+title: Advanced
+navigation_weight: 1
+---
+
+# Caching CDN Requests or Non-200 Responses
+
+If an attempt is made to cache an asset that is on a different origin or
+the asset returns a non-200 response, these modules will throw an error. The
+reason for this default behavior is to ensure that only good and local
+responses are cached.
```
jpmedley: 'default behavior' implies a way to override it. Either remove 'default' or describe the override.

gauntface:
@jeffposnick @jpmedley  Could I request a second pass on this please.

docs/advanced/index.md
```diff
@@ -0,0 +1,110 @@
...
+
+Depending on how you are using these libraries will depend on how you can add
+support for external resources.
+
```
jpmedley: Having depending/depend is consfusing. Try something like: 'Use of these libraries depends on how you support external resources.'.

docs/advanced/index.md
```diff
@@ -0,0 +1,110 @@
...
+## sw-lib
+
+In sw-lib, you can set up some additional options with a route that will allow
+support for additional status codes and / or based on specific headers.
```
jpmedley: I rewrote this yesterday because I found it confusing. I think it's the parsing of the and/or against two different verbs. How about, "In sw-lib you can set up additional options with routes that support additional status codes, specific headers, or both."?

docs/advanced/index.md
```diff
@@ -0,0 +1,110 @@
...
+```javascript
+const cdnCacheStrategy = goog.swlib.staleWhileRevalidate({
+  cacheableResponse: {
+    statuses: [0],
+  },
+});
+goog.swlib.router.registerRoute(new RegExp('^https://cdn.mysite.com/styles/'), cdnCacheStrategy);
+```
+
+This same approach can be used with all of th caching strategies supported by
```
jpmedley: Fix 'th' > 'the'.

While you're editing, please use active voice. In other words, "use this approach" instead of "this approach can be used."

docs/advanced/index.md
```diff
@@ -0,0 +1,110 @@
...
+sw-lib including `cacheFirst()`, `cacheOnly()`, `networkFirst()`,
+`networkOnly()` and `staleWhileRevalidate()`.
+
+## Lower Level Modules
+
+To cache a response with a status code other than 2XX status code when using
+the lower level modules use the `sw-cacheable-response` plugin with a
```
jpmedley: I also have trouble parsing this sentence. This would be a little easier if the qualification were the opening clause. "When using lower level modules, use the `s-c-r` plugin with a `RW` to cache" etc..

docs/advanced/index.md
```diff
@@ -0,0 +1,110 @@
...
+const route = new goog.routing.RegExpRoute({
+  match: ({url}) => url.domain === 'example.com',
+  handler: new goog.runtimeCaching.StaleWhileRevalidate({requestWrapper})
+});
+```
+
+# Fine Grained Request Caching
+
+If you need more than the ability to define static status and header values
+you can create your own "plugin" which needs to implement the `cacheWillUpdate`
```
jpmedley: Method names need to end in (). Please check throughout.

docs/advanced/index.md
```diff
@@ -0,0 +1,110 @@
...
+method. Wherever this plugin is used, it will be passed a
+[Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) object
+and returning a boolean of true will result in the caching of the response,
+other you prevent caching by returning false.
```
jpmedley: Did you mean 'otherwise'?

docs/advanced/index.md
```diff
@@ -0,0 +1,110 @@
...
+});
+```
+
+# Fine Grained Request Caching
+
+If you need more than the ability to define static status and header values
+you can create your own "plugin" which needs to implement the `cacheWillUpdate`
+method. Wherever this plugin is used, it will be passed a
+[Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) object
+and returning a boolean of true will result in the caching of the response,
```
jpmedley: Independent clauses. Need a comma after 'and'.

gauntface:
@jpmedley @jeffposnick Should be good for another review - thanks for input so far.

addyosmani:
These changes LGTM. Thanks for working on the docs, Matt!

addyosmani:
This looks like it's good to land now. Ships away!

PR #239, opened 2017-02-23
Fixes Readme Appveyor Badge

gauntface:
R: @jeffposnick @addyosmani


addyosmani:
Just to check on AppVeyor support: I recall when we were adding this to other libraries (in the past) it was useful making sure the whole team was added as admins. Have we done this? :) (just in case someone gets run over by a 🚌 or something) :)

PR #240, opened 2017-02-23
Make sw-cli public

gauntface:
R: @jeffposnick @addyosmani 

I can haz?

### DIFF
--- a/packages/sw-cli/package.json
+++ b/packages/sw-cli/package.json
@@ -1,7 +1,6 @@
 {
   "name": "sw-cli",
   "version": "0.0.8",
-  "private": true,
   "description": "A CLI tool to generate a service worker and a file manifest making use of the sw-lib module.",
   "keywords": [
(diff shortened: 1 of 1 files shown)


packages/sw-cli/package.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "sw-cli",
   "version": "0.0.8",
-  "private": true,
```
addyosmani: 👍 Noticed this is also in https://www.npmjs.com/package/sw-cli and we've been keeping the package updated. We might want to chat sometime about whether we version up all of the packages in the repo at the same time (e.g like Angular does) or keep them individual.

gauntface: Lerna will make them all publish with same version if they change and when they do - they jump to the latest version number. I think this has been good so far and makes sense to avoid needless churn for updates for projects depending on us.

Issue #242, opened 2017-02-27
Debug failing test in Chrome Beta/Canary


jeffposnick:
**Library Affected**:
`sw-routing`

One of our tests is failing in non-production versions of Chrome:

```sh
[Failed] End to End Test of RegExpRoute > should work properly when there are multiple routes
    expected 'Cannot GET /static\n' to equal 'static response'
    [Stack Trace]
    AssertionError: expected 'Cannot GET /static\n' to equal 'static response'
        at Promise.all.iframe.contentWindow.fetch.then.then (end-to-end.js:7:43)
        at <anonymous>
```

gauntface:
@jeffposnick what does "non-production versions of Chrome" mean?

jeffposnick:
Beta + Dev/Canary.

jeffposnick:
tl;dr it looks like there's a race condition in the `goog.swUtils.activateSW()` promise that, the first time it's called, causes it to resolve before the service worker has taken control of the iframe. @gauntface, this might be something for you to dig into further.

Here's the full test that's failing in Chrome M57+, but works fine in M56. It's from https://github.com/GoogleChrome/sw-helpers/blob/master/packages/sw-routing/test/browser/end-to-end.js:

```
describe('End to End Test of RegExpRoute', function() {
  it('should work properly when there are multiple routes', function() {
    return goog.swUtils.activateSW('../static/regexp-route.js')
      .then((iframe) => Promise.all([
        iframe.contentWindow.fetch('/static')
          .then((response) => response.text())
          .then((text) => expect(text).to.equal('static response')),

        iframe.contentWindow.fetch('/echo3/1st/abc/2nd/def/3rd/ghi')
          .then((response) => response.json())
          .then((json) => expect(json).to.eql(['abc', 'def', 'ghi'])),

        iframe.contentWindow.fetch('/will-not-match')
          .then((response) => expect(response.status).to.equal(404)),
      ]));
  });
});
```

Some observations:
- Those tests all rely on `goog.swUtils.activateSW()` to register and then wait for the SW to activate, at which point they make requests that test the SW's functionality.
- If I reorder the tests in that file, which ever one I list first will fail. E.g. moving `'End to End Test of Router'` to the first position will cause that to fail. The rest of the tests will all succeed, so the failure isn't due to any specific test.
[...]

PR #248, opened 2017-02-28
Save config with sw-cli

gauntface:
R: @jeffposnick @addyosmani 

This is the last major piece that I wanted to land in sw-cli. This adds the config file that is picked up before generating the SW - this allows skipping of the questions.

I will hook this up with the build manifest in a seperate PR - just to reduce the size of PRs.


gauntface:
Latest PR Fixes #242

PR #249, opened 2017-02-28
Adds doc pages for older modules

gauntface:

This will tick off:

> Every module has some JSDoc OR we have some markdown file introducing each project.

on #164


jeffposnick:
👍

PR #254, opened 2017-02-28
Fixes #223

### LINKED ISSUE #223: Add in custom error message for missing globPatterns
**Library Affected**:
`sw-build`

There currently isn't a check to make sure the `globPatterns` parameter is properly set to an array when configuring `sw-build`. If you mistakenly leave it out, you get a runtime error along the lines of 

```sh
TypeError: Cannot read property 'reduce' of undefined
    at Object.getFileManifestEntries (sw-build/build/lib/get-file-manifest-entries.js:44:35)
```

instead of a meaningful error indicating that the parameter is required.

gauntface:
R: @jeffposnick @addyosmani 

Simple check and run for globPatterns input.


gauntface:
I just tried using assert - at the moment sw-build and sw-cli are plain old node modules (i.e. aren't run through rollup etc). As a result using assert introduces a world of hurt because it uses es2015 import for error-stack-parser.

PR #258, opened 2017-03-01
Rollup-based build recipe

jeffposnick:
R: @jpmedley @addyosmani @gauntface

Here's a recipe for how to structure your service worker using ES2015 imports, and then bundle that source file into a final service worker using `rollup`. It has examples for kicking off that process as both part of an `npm scripts` chain or as `gulp` tasks.

I don't know whether I need to do anything more than drop this `.md` file into a directory, so please let me know if there are other things I should do to make sure this gets picked up.


docs/recipes/build/rollup.md
```diff
@@ -0,0 +1,135 @@
+# [Rollup](https://github.com/rollup/rollup) Bundling
+
+## Dependencies
```
gauntface: There needs to be a title and intro that explains what this doc is for / going to achieve.

docs/recipes/build/rollup.md
```diff
@@ -0,0 +1,135 @@
...
+
+- A working installation of `node` (4.0+) & `npm`.
+- `npm install --save-dev sw-build sw-lib rollup rollup-plugin-node-resolve`
+
+## Service Worker Code
+
+This approach assumes that your `src/sw.js` file is using [ES2015 module
```
gauntface: This could be part of the intro - i.e. explanation of if you are using Rollup and would like to use ES2015 modules in your service worker - here is how you can get the manifest revision details.

docs/recipes/build/rollup.md
```diff
@@ -0,0 +1,135 @@
...
+// import other libraries, e.g. idb-keyval, Firebase Messaging, etc.
+
+swLib.cacheRevisionedAssets(manifest);
+
+// Additional runtime service worker logic goes here.
+```
+
+## Option 1: Vanilla node Build Script
+
+The following is a basic `node` build script that can be run as part of a larger
```
gauntface: I don't understand this paragraph - what is the goal?

Why is it referencing npm scripts?

docs/recipes/build/rollup.md
```diff
@@ -0,0 +1,135 @@
...
+build process, e.g. in `package.json` at the end of an
+[`npm scripts`](https://docs.npmjs.com/misc/scripts) `"build"` chain as
+`&& node build-sw.js`.
```
gauntface: Is it explained after this section?

docs/recipes/build/rollup.md
```diff
@@ -0,0 +1,135 @@
+# [Rollup](https://github.com/rollup/rollup) Bundling
+
+## Dependencies
+
+- A working installation of `node` (4.0+) & `npm`.
```
jpmedley: 'A working installation of' to 'Working installations of'.

docs/recipes/build/rollup.md
```diff
@@ -0,0 +1,135 @@
...
+
+const bundleConfig = {
+  format: 'iife',
+  dest: `${BUILD_DIR}/sw.js`,
+};
+
+// manifest generation ->
+//   rollup configuration ->
+//   write bundle to disk
+swBuild.generateFileManifest(swBuildConfig)
```
gauntface: I'd inline config here - It's weird go sw-build config, rollup config, sw-build, then rollup promise.

docs/recipes/build/rollup.md
```diff
@@ -0,0 +1,135 @@
...
+  .then(() => rollup(rollupConfig))
+  .then((bundle) => bundle.write(bundleConfig));
+
+```
+
+## Option 2: [gulp](http://gulpjs.com/) Tasks
```
gauntface: Maybe refrain from using the term "Options" and keep to recipes - i.e. using sw-build with Rollup and Gulp

docs/recipes/build/rollup.md
```diff
@@ -0,0 +1,135 @@
...
+
+The same basic code can be split up into two separate `gulp` tasks, if you
+already are using that for your build process.
+
+**Note:** As with before, make sure that you run the `write-sw` task at the very
```
gauntface: Maybe bring this out in a note in the intro and why it's important and then no need to mention again.

docs/recipes/build/rollup.md
```diff
@@ -0,0 +1,135 @@
...
+    globPatterns: [
+      `./${BUILD_DIR}/{css,images,js}/**/*`,
+      `./${BUILD_DIR}/index.html`,
+    ],
+    rootDirectory: BUILD_DIR,
+    format: 'es',
+  });
+});
+
+gulp.task('write-sw', ['write-manifest'], () => {
```
gauntface: real nit - but would bundle-sw.js be a better name? I'm just thinking that write-manifest is generating out of thin air and sw.js will be taking an existing file and spitting it out the other side.

jeffposnick:
Thanks for the feedback, and PTAL.

.gitignore
```diff
@@ -4,3 +4,4 @@ temp
 npm-debug.log
 packages/**/LICENSE
 .DS_Store
+!docs/recipes/build
```
gauntface: Where is this from? Just curious - not blocking.

docs/recipes/build/rollup.md
```diff
@@ -0,0 +1,144 @@
...
+
+These recipes cover the following `sw-lib` use case:
+- You want to write and maintain your own `sw.js` file that uses precaching via
+`sw-lib.cacheRevisionedAssets()` alongside other service worker logic.
+- You'd like to use [ES2015 import syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import)
+to include `sw-lib` and other modules into your service worker.
+- You are comfortable using [`Rollup`](https://github.com/rollup/rollup) to
+handle transforming the `sw.js` file you maintain into a production-ready,
+self contained service worker file, containing the `sw-lib` library code along
+with your custom logic.
```
gauntface: I'm not going to block because this going to be personal, but I kind of prefer the second half of this intro and think you could go along the lines of:

"If you are using Rollup and ES2015 imports, you can use sw-lib as an import in your service worker and sw-build can output a precache manifest that can be imported as well.

This recipe illustrates how you can use ES2015 imports in your service worker with sw-lib and sw-build."

docs/recipes/build/rollup.md
```diff
@@ -0,0 +1,144 @@
...
+
+// This will precache the files in your manifest, and keep them up to date.
+swLib.cacheRevisionedAssets(manifest);
+
+// Any additional runtime service worker logic can go here.
+```
+
+## Your Build Process
+
+Here are two alternatives that lead to an equivalent, ready-to-deploy service
```
gauntface: This feels like a wordy way to say:

"You have two options, you can build your service worker using Gulp or npm scripts"

docs/recipes/build/rollup.md
```diff
@@ -0,0 +1,144 @@
...
+`npm scripts` or `gulp` as your task runner.
+
+**Note:** Regardless of which task runner you're using, make sure that the
+generation of the service worker file happens *last* in the chain of your build
+tasks. This ensures that the service worker looks at the final contents of your
+`build/` directory when determining which files to precache and keep up to date.
+
+### `npm scripts`
+
+The following is a basic `node` script that can be run as part of a larger build
```
gauntface: Nit: I'd add an example `"scripts": {"build": "node ./build.js"}` as it's not clear where / what this Javascript / should do.

docs/recipes/build/rollup.md
```diff
@@ -0,0 +1,144 @@
...
+## Service Worker Code
+
+This approach assumes that you have a `src/sw.js` that uses [ES2015 module
+import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import)
+syntax, and is structured like:
+
+```js
+// Contents of src/sw.js:
+
+import manifest from '/tmp/manifest.js';
```
gauntface: Is it worth highlighting after this code sample that in the npm run script / gulp process you'll write the manifest file using sw-build before running the service worker file through rollup?

.gitignore
jeffposnick: We've got an entry a few lines up that ignores `build` (in the context of the build output for a given module).

This is just to negate that so that I'm able to call the directory that hosts the recipes `build` and still check it in to git.

docs/recipes/build/rollup.md
jeffposnick: I've added an inline comment to that effect.

docs/recipes/build/rollup.md
jeffposnick: There are more than two options, but I'll rephrase to make it less wordy 😄

docs/recipes/build/rollup.md
jeffposnick: If you're following this recipe, it's not like to look like `"scripts": {"build": "node ./build.js"}`

More likely is `"scripts": {"build": "node build.js && node build-sw.js"}` (chaining the additional build process outlined here to the end of your existing build process).

But I can include that sample.

docs/recipes/build/rollup.md
jeffposnick: I think I'll keep as-is. This will probably get tweaked more based on hackathon feedback.

jeffposnick:
I've addressed the feedback, hopefully in a satisfactory fashion.

Tweaking prose in a PR vs. in a Google Doc is always a pain, so this might not be the perfect final version, but I'm happy to address any changes in a follow-up PR.

GitHub won't let me merge until @jpmedley approves this, though.

Issue #259, opened 2017-03-02
Send cloned request in fetchDidFail


prateekbh:
**Library Affected**:
*sw-runtime-caching*

**Browser & Platform**:
*All*

**Issue Description**:
In this piece of code
https://github.com/GoogleChrome/sw-helpers/blob/5e7b7b9434c4af2b398ef823aa84257704398d85/packages/sw-runtime-caching/src/lib/request-wrapper.js#L179-L203
you'll see that we fetch the `request` and then send the same `request` to plugins in event: `fetchDidFail`.
Imagine some plugin trying to read request.body or every trying to clone with request.clone will get error of 'Body already used'.

*When reporting bugs, please include relevant JavaScript Console logs and links to public URLs at which the issue could be reproduced.*

PR #260, opened 2017-03-02
fixes bug #259, cloned request for every plugin

prateekbh:

Fixes #*259*

This sends a clone of request to plugins in `fetchDidFail` event

### DIFF
--- a/packages/sw-runtime-caching/src/lib/request-wrapper.js
+++ b/packages/sw-runtime-caching/src/lib/request-wrapper.js
@@ -178,7 +178,7 @@ class RequestWrapper {
    */
   async fetch({request}) {
     assert.atLeastOne({request});
-
+    const clonedRequest = request.clone();
     if (this.pluginCallbacks.requestWillFetch) {
       for (let callback of this.pluginCallbacks.requestWillFetch) {
         const returnedPromise = callback({request});
@@ -194,7 +194,9 @@ class RequestWrapper {
     } catch (err) {
       if (this.pluginCallbacks.fetchDidFail) {
         for (let callback of this.pluginCallbacks.fetchDidFail) {
-          callback({request});
+          // New clone for every plugin, so taht one plugin's read/clone
+          // does not break other plugin
+          callback({request: clonedRequest.clone()});
         }
       }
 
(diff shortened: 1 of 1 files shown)


packages/sw-runtime-caching/src/lib/request-wrapper.js
```diff
@@ -178,7 +178,7 @@ class RequestWrapper {
    */
   async fetch({request}) {
     assert.atLeastOne({request});
-
+    const clonedRequest = request.clone();
```
jeffposnick: Can you check `this.pluginCallbacks.fetchDidFail` here and only clone the request if there's one or more callbacks? Something like:

```js
const clonedRequest = this.pluginCallbacks.fetchDidFail ? request.clone() : null;
```

packages/sw-runtime-caching/src/lib/request-wrapper.js
```diff
@@ -194,7 +194,7 @@ class RequestWrapper {
     } catch (err) {
       if (this.pluginCallbacks.fetchDidFail) {
         for (let callback of this.pluginCallbacks.fetchDidFail) {
-          callback({request});
+          callback({request: clonedRequest});
```
jeffposnick: You'll run into the same problem here if there are multiple `fetchDidFail` callbacks and the first one consumes the `clonedRequest` body. You can either change this to

```js
callback({request: clonedRequest.clone()});
```

which has the drawback of always creating 1 extra clone beyond what's needed, or alternatively, do something where you check the length of `this.pluginCallbacks.fetchDidFail`, create an array containing enough extra clones to match that length, and then pop clones off that array so that you always use the exact number you need.

prateekbh: yeah i had this exact idea if you see one of my commits in this PR, then was conflicted..
but yeah `callback({request: clonedRequest.clone()});` makes perfect sense
will do

prateekbh: done

packages/sw-runtime-caching/src/lib/request-wrapper.js
prateekbh: done

prateekbh:
Umm... I cannot me this anymore. Can someone go ahead and do it

PR #261, opened 2017-03-03
Support for a function as a Route's handler

### LINKED ISSUE #231: Direct support for functions in Route({handler})
**Library Affected**:
`sw-routing`

We currently require that when a `Route` is instantiated, the `{handler}` parameter passed to the constructor is an object that exposes a `handle` function, and that `handle` function is what's responsible for returning a `Promise<Response>`.

That works okay when you're using one of the handlers from `sw-runtime-caching` (like `StaleWhileRevalidate`, etc.), but when you just want to do something simple and provide your own function, it leads to awkwardly wrapping the `{handler}` value in an anonymous object with a `handle`, like so:

```js
const route = new goog.routing.RegExpRoute({
  regExp: new RegExp('^/article/'),
  handler: {handle: () => caches.match('app-shell.html')},
});
```

My initial motivation for this was to enforce consistency and reduce overloaded behavior (since the `sw-runtime-caching` classes all are objects which expose `handle` functions), but I'm starting to change my mind about whether it's a good idea.

How do folks feel about relaxing that restriction, and making it so that you could just pass in a function directly as the `{handler}` value? It would lead to code that looks like:

```js
const route = new goog.routing.RegExpRoute({
  regExp: new RegExp('^/article/'),
  handler: () => caches.match('app-shell.html'),
});
```

Right now we have an assertion that checks whether the `{handler}` value is an object that exposes a `handle` property, so we would also relax that and allow the value to be a `function` or an object with `handle` (to ensure we still support `sw-runtime-caching`).

jeffposnick:
R: @addyosmani @gauntface

Fixes #231

### DIFF
--- a/packages/sw-routing/src/lib/route.js
+++ b/packages/sw-routing/src/lib/route.js
@@ -49,11 +49,9 @@ import {defaultMethod, validMethods} from './constants';
  *     return event.request.mode === 'navigation' &&
  *            url.pathname.startsWith('/path/to/');
  *   },
- *   handler: {
- *     handle: ({event}) => {
- *       // Do something that returns a Promise.<Response>, like:
- *       return caches.match(event.request);
- *     },
+ *   handler: ({event}) => {
+ *     // Do something that returns a Promise.<Response>, like:
+ *     return caches.match(event.request);
  *   },
  * });
  *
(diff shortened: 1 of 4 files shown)


packages/sw-routing/src/lib/route.js
```diff
@@ -72,20 +70,27 @@ class Route {
    *        and `event`, which is a [FetchEvent](https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent).
    *        `match` should return a truthy value when the route applies, and
    *        that value is passed on to the handle function.
-   * @param {Object} input.handler An Object with a `handle` method. That
-   *        function is passed an object with the same `url` and `event`
-   *        properties as `match` received, along with an additional property,
-   *        `params`, set to the truthy value that `match` returned.
+   * @param {Object|function} input.handler An Object with a `handle` method, or
```
gauntface: Nit: Could you swap this around to function|Object and the comment.

Basically I expect most developers will never care about this because they'll use a default caching strategy, but if they do use this I imagine function will be the preferred way.

packages/sw-routing/test/static/route.js
```diff
@@ -9,38 +9,29 @@ const routes = [];
...
-      new Response(JSON.stringify(params.slice(1)), {
-        headers: {'content-type': 'application/json'},
-      })
-    ),
-  },
+  handler: ({params}) => Promise.resolve(new Response(
+    JSON.stringify(params.slice(1)), {
+      headers: {'content-type': 'application/json'},
+    })
+  ),
```
gauntface: Could you tests the to relevant places to make sure the object way is supported as well.

jeffposnick: It's [covered in the unit tests](https://github.com/GoogleChrome/sw-helpers/pull/261/files#diff-9e4ba02317a767b390dca5110a56ed26R38), so I figured it was safe to stick with one approach instead of covering all variations within the end-to-end tests.

packages/sw-routing/src/lib/route.js
jeffposnick: Done.

jeffposnick:
Sorry for the slow turnaround—PTAL.

packages/sw-routing/test/static/route.js
gauntface: I'd like just one - it's largely to protect from accidental breaking in the future.

jeffposnick: Cool. Added.

PR #274, opened 2017-03-07
Add demos

gauntface:

Fixes ##272


packages/sw-precaching/demo/sw.js
```diff
@@ -0,0 +1,44 @@
+/* global goog */
+
+importScripts('./sw-precaching.v0.0.13.min.js');
```
jeffposnick: I've been pulling in libraries as `../build/my-lib.js` for the other demos. For production `importScript()` uses we want to encourage pulling in versioned libraries (or using a module bundler), but I wonder if it can lead to more trouble than it's worth to try to make sure that the right version of the library ends up in the `demo/` directory, and that we update it each time we push out a new version.

Also... I don't see `sw-precaching.v0.0.13.min.js` included in this PR. So maybe this isn't meant to work to begin with?

packages/sw-lib/demo/sw.js
```diff
@@ -0,0 +1,37 @@
...
+
+// A regular expression to catch a range of strings.
+// These only need to match part of the URL.
+goog.swlib.router.registerRoute(
+  /\/images\/(.*\/)?.*\.(png|jpg|jpeg|gif)/,
+  goog.swlib.cacheFirst()
+);
+
+// Use express style routes to capture URLs
+goog.swlib.router.registerRoute('/styles/:filename', goog.swlib.cacheFirst());
```
jeffposnick: Showing off URL capture when the handler doesn't make use of it could be confusing.

I never really came up with a good use case for URL capture, though...

Probably more common and worth illustrating would be using `/styles/(.*)` as a wildcard?

packages/sw-lib/demo/sw.js
```diff
@@ -0,0 +1,37 @@
...
+
+goog.swlib.cacheRevisionedAssets(self.__file_manifest);
+
+// Specific URL defined with a string.
+goog.swlib.router.registerRoute('/example/', goog.swlib.staleWhileRevalidate());
+
+// A regular expression to catch a range of strings.
+// These only need to match part of the URL.
+goog.swlib.router.registerRoute(
+  /\/images\/(.*\/)?.*\.(png|jpg|jpeg|gif)/,
```
jeffposnick: `new RegExp('/images/.+(?:png|jpg|gif)')` could be a cleaner example.

packages/sw-lib/demo/sw.js
```diff
@@ -0,0 +1,37 @@
...
+  goog.swlib.staleWhileRevalidate({
+    cacheName: 'example-cache',
+    cacheExpiration: {
+      maxEntries: 10,
+      maxAgeSeconds: 7 * 24 * 60 * 60,
+    },
+  }));
+
+// To cache third party content that doesn't support CORS, allow '0' status code
+goog.swlib.router.registerRoute('/example-third-party-caching/',
```
jeffposnick: Express-style routes only match same-origin requests 😨  To match third-party URLs, you're going to need to use a RegExp.

We do currently fail with an error message you if you try to pass in a full URL, like `https://cors-domain.com/path/to`, in as an Express-style route. But in this case, we would just assume that `'/example-third-party-caching/'` was a same-origin route, and it wouldn't end up matching anything.

I don't know what we could do to try to guess developer's intent here and warn them...

.travis.yml
```diff
@@ -39,3 +39,4 @@ before_script:
 script:
   # Building is handled in gulp test.
   - gulp test --project "${PROJECT}"
+  - if [[ "$TRAVIS_BRANCH" = "master" && "$TRAVIS_PULL_REQUEST" = "false" ]]; then npm install -g npm-publish-scripts && npm-publish-scripts publish-docs --non-interactive; fi
```
jeffposnick: This is included in https://github.com/GoogleChrome/sw-helpers/pull/273 as well. I guess if that other PR gets merged, this line just disappear from the diff?

gauntface:
@jeffposnick changed the demos slightly - still wrinkles IMO in terms of demonstrating some of the use cases but better than what it was.

PR #278, opened 2017-03-07
Fixes #93

### LINKED ISSUE #93: Toggle assertions/debug logging via environment setting
We should come up with a standard toggle for enabling/disabling assertions and additional logging via a central framework-level environment value (`'production'` or `'development'`).

The default value for that value, when not explicitly set, could be `'development'` if we detect that we're being served from `localhost`, and `'production'` otherwise.

gauntface:
R: @jeffposnick @addyosmani 

Fixes #93

Adds logging levels. Only thing that you may want changed is how the logging level is set (Please see reference docs in the code for example of how it *could* be done).


gauntface:
to be honest I would like this for the hackathon in case we can get some modules logging added - but we should circle back and see if there is a better approach.

I'll raise and issue for post hackathon and will merge after.

jeffposnick:
Yup, that's what I meant to indicate in my review as well 😄

PR #284, opened 2017-03-08
Enhance the README.

jpmedley:

Try this on for size.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,57 @@
 
 # SW Helpers
 
+## Contents
+
+- [Ways to Use the Libraries](#ways-to-use-the-libraries)
+- [Setting Up](#setting-up)
+- [The Libraries](#the-libraries)
+  - [sw-appcache-behavior](#sw-appcache-behavior)
+  - [sw-broadcast-cache-update](#sw-broadcast-cache-update)
+  - [sw-build](#sw-build)
+  - [sw-cache-expiration](#sw-cache-expiration)
+  - [sw-cli](#sw-cli)
+  - [sw-lib](#sw-lib)
+  - [sw-offline-google-analytics](#sw-offline-google-analytics)
+  - [sw-precaching](#sw-precaching)
+  - [sw-routing](#sw-routing)
+  - [sw-runtime-caching](#sw-runtime-caching)
+- [External Contributions](#external-contributions)
+- [License](#license)
+
+The sw-helpers library is ...
+
+## Ways to Use the Libraries
+
+Despite the number of modules in this package, there are a number of approaches to using these tools which many be used separately or together in any number of combinations.
+
+* **Precaching, runtime caching and routing**: Use [sw-lib](.) to quickly
+  implement a [caching strategies](.) or a route.
+* **Build process**: Generate a service worker or manifest in your build process
+  using [sw-build](.).
+* **Command line**: Generate a service worker or manifest from the command line.
+* **Go beyond the basics**: Implement more advanced capabilities and more
+  sophisticated use cases using any of the other libraries, which are [described
+  below](#the-libraries).
+
+## Setting Up
+
+Each module is installed separately using the command line [as listed below](#the-libraries). To use a particular module:
+
+1. Install the module. For example:
+
+   `npm install --save-dev sw-lib`
+2. Copy the module's JavaScript and map files to your serving directory. To work properly these must be in the same directory as your service worker. For example:
+
+   `cp node_modules/sw-lib/build/* app/`
+3. Import the modules to your service worker file. For example:
+
+   `importScripts('sw-lib.min.js');`
+
 ## The Libraries
+
+All of the modules are described below. Each has an _About_ page with basic usage instructions and a _Demo_ directory with an example. The main page of the documentation is [here](https://googlechrome.github.io/sw-helpers/#main).
+
 ### sw-appcache-behavior
 
 [![Build Status](https://travis-shields.appspot.com/shield/GoogleChrome/sw-helpers/master/PROJECT%3D%22sw-appcache-behavior%22)][travis-url]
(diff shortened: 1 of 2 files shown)


addyosmani:
This will need a rebase.

README.md
```diff
@@ -6,6 +6,56 @@
 
 # SW Helpers
 
+## Contents
+
+- [Ways to Use the Libraries](#ways-to-use-the-libraries)
+- [Setting Up](#setting-up)
```
addyosmani: Nit: it's common for libraries to use the `Setup` verbiage. Will leave to your discretion.

README.md
```diff
@@ -6,6 +6,56 @@
...
+
+Each module is installed separately using the command line [as listed below](#the-libraries). To use a particular module:
+
+1. Install the module. For example:
+   `npm install --save-dev sw-lib`
+2. Copy the module and map files to your serving directory. To work properly these must be in the same directory as your service worker. For example:
+   `cp node_modules/sw-lib/build/* app/`
+3. Import the modules to your service worker file. For example:
+   `importScripts('sw-lib.min.js');`
+
```
addyosmani: @gauntface @jeffposnick To what extent do we want to start modernizing our installation guidelines to include `yarn` as well as `npm`? It's certainly not a blocker here. Just suggesting as a discussion point.

README.md
```diff
@@ -6,6 +6,56 @@
...
+# Using
+
+As listed under [The Libraries](#the-libraries), each module has an _About_ page with basic usage instructions and a _Demo_ directory with an example. The main page of the documentation is [here](https://googlechrome.github.io/sw-helpers/#main).
```
addyosmani: Does each module have a demo? I was under the impression most did, but it would be good to sanity check in case we're setting up expectations incorrectly.

README.md
```diff
@@ -6,6 +6,56 @@
...
+  - [sw-cache-expiration](#sw-cache-expiration)
+  - [sw-cli](#sw-cli)
+  - [sw-lib](#sw-lib)
+  - [sw-offline-google-analytics](#sw-offline-google-analytics)
+  - [sw-precaching](#sw-precaching)
+  - [sw-routing](#sw-routing)
+  - [sw-runtime-caching](#sw-runtime-caching)
+- [External Contributions](#external-contributions)
+- [License](#license)
+
```
addyosmani: I like the introduction of the TOC.

README.md
jpmedley: Three forms generally used in titles IT documents:
* Noun ("Setup")
* Gerund ("Setting Up")
* Infinitive ("To Set Up")
What we try for is consistency within a single doc set. I'll see what we do most in this set and make them consistent.

gauntface:
Is this done in the template/readme.md or in the root readme.md?

If the later - please move these changes into the template and then run gulp documentation.

README.md
gauntface: My only concern here is that there is no heirachy - i.e. all the libraries look equal.

README.md
```diff
@@ -6,7 +6,58 @@
...
+The sw-helpers library is ...
+
+## Ways to Use the Libraries
+
+Despite the number of modules in this package, there are a number of approaches to using these tools which many be used separately or together in any number of combinations.
+
+* **Basic caching strategies**: Use [sw-lib](.) to quickly implement one of the
```
gauntface: sw-lib is precaching, runtime caching and routing. Less emphasis on the caching strategies IMO.

README.md
```diff
@@ -6,7 +6,58 @@
...
+  [standard caching strategies](.).
+* **Build process**: Incorporate sw-helpers into your build process using
+  [sw-build](.).
```
gauntface: build process is largely for generating service worker or build a file manaifest 'Incorporate sw-helpers' feels vague and not representative.

README.md
```diff
@@ -6,7 +6,58 @@
...
+* **Command line**: Use a command line interface to generate a basic service
+  worker and play around with the features of the library.
```
gauntface: It's a service worker or manifest generator - nothing available for "playing around with the features of the library".

README.md
```diff
@@ -6,7 +6,58 @@
...
+* **Go beyond the basics**: Implement more advanced capabilities and more
+  sophisticated use cases using any of the other libraries, which are described
+  below.
```
gauntface: I'd link straight to the section - "described below." confused me as the next step of setting up.

README.md
```diff
@@ -6,7 +6,58 @@
...
+
+## Setting Up
+
+Each module is installed separately using the command line [as listed below](#the-libraries). To use a particular module:
+
+1. Install the module. For example:
+
+   `npm install --save-dev sw-lib`
+2. Copy the module and map files to your serving directory. To work properly these must be in the same directory as your service worker. For example:
```
gauntface: I'd say "copy the modules build files" rather then "the module and map files"

gauntface: Also they don't need to be in the same directory as your service worker - they can be anywhere and just imported: i.e:

```
importScripts('/hello/world/blurgh/sw/stuff/sw-lib.min.js);
```

jpmedley: Better check that. I originally had mine in a `js/` folder and found that not everything worked. @jeffposnick

README.md
jeffposnick: I passed along an (internal) document I put together that categorized the libraries into four buckets. You might want to reuse that categorization here.

README.md
jeffposnick: Your resources pulled in via `importScripts()` can be located anywhere, as long as the correct URL is provided.

The only time that the location matters if for the top-level service worker script.

README.md
jpmedley: If the {{#each projects}} macro can't print them in the order of that doc, then I can't change the toc. Having a toc in a different order from the actual contents is a anti-pattern.

Issue #282, opened 2017-03-08
Refactor Plugin classes


jeffposnick:
**Library Affected**:
many

For many of the packages (e.g., `sw-cache-expiration`), the `Plugin` class that exposes `RequestWrapper` lifecycle callbacks (e.g. `cacheDidUpdate`) is currently the same place that implements the core functionality, which is reusable outside of the `RequestWrapper` context.

It would be cleaner to break that up so that there's a base class which implements the core functionality, and then a subclass that exposes the addition lifecycle callbacks.

Additionally, it would be better to use a more descriptive name than `Plugin` for each of those classes, since our assertion libraries will log the class name if there are bad parameters passed to the constructor. So `Plugin` inside of the `sw-cache-expiration` project would be better named `CacheExpirationPlugin`.

gauntface:
I think I get what you are saying and big plus one for this.

Seperating the plugin callbacks from methods should make it easier to use as standalone module.

jeffposnick:
Modules to refactor:
- [ ] `sw-background-sync-queue`
- [ ] `sw-broad-cache-update`
- [ ] `sw-cache-expiration`
- [ ] `sw-cacheable-response`

jeffposnick:
Three of the items from https://github.com/GoogleChrome/sw-helpers/issues/282#issuecomment-285520914 are complete, with `sw-background-sync-queue` not yet updated.

@prateekbh, are you okay with me assigning this issue to you? You can follow the model used in the other PRs. I'd do it myself, but I know there's a bit of an ongoing discussion about the exported interface and namespace for the `sw-background-sync-queue` module, and that should probably be addressed prior to/at the same time as the plugin refactoring is performed.

prateekbh:
> For many of the packages (e.g., sw-cache-expiration), the Plugin class that exposes RequestWrapper lifecycle callbacks (e.g. cacheDidUpdate) is currently the same place that implements the core functionality

@jeffposnick  in background-sync-queue all the core functionality is already abstracted behind a proxy class, which just calls the core methods. This was done so that this lib can be used with or without `sw-helpers`.

https://github.com/GoogleChrome/sw-helpers/blob/master/packages/sw-background-sync-queue/src/lib/background-sync-queue.js#L119-L121

If I read your description right, this shouldn't be a problem

jeffposnick:
[...]

PR #287, opened 2017-03-09
Refactoring of the Plugin class in BroadcastCacheUpdate

jeffposnick:
R: @addyosmani @gauntface

It ended up touching many more files/docs/tests than I thought it would, so I've limited this PR to just handle it for the `BroadcastCacheUpdate` class. I'll do follow-up PRs for the other libraries, and use #282 to track all of them.

### DIFF
--- a/packages/sw-broadcast-cache-update/src/index.js
+++ b/packages/sw-broadcast-cache-update/src/index.js
@@ -19,22 +19,23 @@
  * A helper library that uses the Broadcast Channel API to announce when
  * two Response objects differ.
  *
- * The main use of this module will be instantiating a new `Plugin` and
- * passing it to a
- * {@link  module:sw-runtime-caching.RequestWrapper|RequestWrapper},
+ * The main use of this module will be instantiating a new
+ * `BroadcastCacheUpdatePlugin` and passing it to a
+ * {@link module:sw-runtime-caching.RequestWrapper|RequestWrapper},
  * as shown in the first example below.
  *
  * You can listen for updates from your web app by adding an event listener on
- * a browser channel with the same channel name as the Plugin,
- * which 'cache-updates' in the second example below.
+ * a `BroadcastChannel` within a page, using the same channel name as
+ * what's used within the service worker, as shown in the second example below.
  *
- * @example <caption>Using the broadcastCacheUpdate.Plugin class in a
+ * @example <caption>Using the BroadcastCacheUpdatePlugin class in a
  * service worker.</caption>
  *
  * const requestWrapper = new goog.runtimeCaching.RequestWrapper({
  *   cacheName: 'text-files',
  *   plugins: [
- *     new goog.broadcastCacheUpdate.Plugin({channelName: 'cache-updates'}),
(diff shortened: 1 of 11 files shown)


packages/sw-broadcast-cache-update/src/index.js
```diff
@@ -57,13 +58,15 @@
...
+import BroadcastCacheUpdate from './lib/broadcast-cache-update';
+import BroadcastCacheUpdatePlugin from './lib/broadcast-cache-update-plugin';
 import broadcastUpdate from './lib/broadcast-update';
 import {cacheUpdatedMessageType} from './lib/constants';
 import responsesAreSame from './lib/responses-are-same';
 
 export {
-  Plugin,
+  BroadcastCacheUpdate,
+  BroadcastCacheUpdatePlugin,
```
gauntface: Nothing to block on but I'm not crazy about this naming `goog.broadcastUpdate.BroadcastCacheUpdatePlugin` but I can honestly see why it is that way.

If anyway has any alternatives - please suggest

packages/sw-broadcast-cache-update/src/lib/broadcast-cache-update.js
```diff
@@ -27,47 +27,34 @@ import {defaultHeadersToCheck, defaultSource} from './constants';
  * For efficiency's sake, the underlying response bodies are not compared;
  * only specific response headers are checked.
  *
- * @example <caption>Added as a "plugin" to a `RequestWrapper` to
- * automatically dispatch messages on a cache update</caption>
+ * This class is meant to be agnostic about the service worker environment it's
```
gauntface: Nit: Could you rephrase this to focus on "This class can be used to <insert use here> and can be used in your service worker without any of the other sw-helper modules."

packages/sw-broadcast-cache-update/src/lib/broadcast-cache-update.js
```diff
@@ -27,47 +27,34 @@ import {defaultHeadersToCheck, defaultSource} from './constants';
...
- *     first: oldResponse,
- *     second: newResponse,
- *     cacheName
- *   });
- * }
+ * Promise.all([
+ *   caches.open(cacheName).then((cache) => cache.match(url)),
+ *   fetch(url),
+ * ]).then(([first, second]) => {
+ *   if (cacheResponse) {
```
gauntface: I dont think cacheResponse is defined

packages/sw-broadcast-cache-update/src/index.js
jeffposnick: +1. I didn't know what else to go with, though.

I'll open an issue to discuss possible changes.

packages/sw-broadcast-cache-update/src/lib/broadcast-cache-update.js
jeffposnick: Done.

packages/sw-broadcast-cache-update/src/lib/broadcast-cache-update.js
jeffposnick: Righto.

PR #291, opened 2017-03-10
Fix a typo.

jpmedley:


packages/sw-broadcast-cache-update/src/lib/broadcast-cache-update-plugin.js
```diff
@@ -55,7 +55,7 @@ class BroadcastCacheUpdatePlugin extends BroadcastCacheUpdate {
    *
    * @private
    * @param {Object} input The input object to this function.
-   * @param {string} input.cacheName Name of the cache the responses belong to.
+   * @param {string} input.cacheName Name of the cache the responses belongs to.
```
jeffposnick: "responses" is plural, so "belong" is correct. (There are two responses, and they both belong to the same cache.)

PR #294, opened 2017-03-12
Adds a networkTimeoutSeconds option to the NetworkFirst strategy.

### LINKED ISSUE #246: Add networkTimeoutSeconds
See: https://github.com/GoogleChrome/sw-toolbox/issues/221 for details

jeffposnick:
R: @addyosmani @gauntface

Fixes #246

This is code fundamentally the same as [`sw-toolbox`'s implementation](https://github.com/GoogleChrome/sw-toolbox/blob/master/lib/strategies/networkFirst.js#L35).

I feel bad about not having a test for the new functionality, but then I feel worse about not having written tests for any of the `sw-runtime-caching` classes yet. So it balances out?

(But I will write tests for them in a future PR.)

### DIFF
--- a/packages/sw-runtime-caching/demo/service-worker.js
+++ b/packages/sw-runtime-caching/demo/service-worker.js
@@ -8,17 +8,14 @@ importScripts(
 );
 
 // Have the service worker take control as soon as possible.
-self.addEventListener('install', (event) => {
-  event.waitUntil(self.skipWaiting());
-});
-self.addEventListener('activate', (event) => {
-  event.waitUntil(self.clients.claim());
-});
+self.addEventListener('install', () => self.skipWaiting());
+self.addEventListener('activate', () => self.clients.claim());
 
 const requestWrapper = new goog.runtimeCaching.RequestWrapper({
   cacheName: 'text-files',
   plugins: [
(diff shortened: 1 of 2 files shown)


packages/sw-runtime-caching/demo/service-worker.js
```diff
@@ -29,4 +26,5 @@ const route = new goog.routing.RegExpRoute({
 
 const router = new goog.routing.Router();
 router.registerRoute({route});
-router.setDefaultHandler({handler: new goog.runtimeCaching.NetworkFirst()});
+router.setDefaultHandler({handler: new goog.runtimeCaching.NetworkFirst(
+  {networkTimeoutSeconds: 10})});
```
gauntface: Does this make sense for the default route?

gauntface:
If you can raise an issue for the tests - LGTM

packages/sw-runtime-caching/demo/service-worker.js
jeffposnick: Just in terms of demonstrating usage, it doesn't hurt. We may tweak when we revisit/normalize the demos.

jeffposnick:
https://github.com/GoogleChrome/sw-helpers/issues/301 filed to track the lack of tests.

PR #295, opened 2017-03-12
sw-build support for templatedUrls

### LINKED ISSUE #263: dynamicUrlToDependencies equivalent
**Library Affected**:
*sw-build*

We should implement the equivalent of `sw-precache`'s [`dynamicUrlToDependencies`](https://github.com/GoogleChrome/sw-precache#dynamicurltodependencies-objectstringarraystring) in `sw-build`. I realize that I need it to port over my [iFixit PWA Demo](https://github.com/GoogleChrome/sw-precache/tree/master/app-shell-demo), as it uses a server-rendered `/shell` URL that doesn't correspond directly to a single file on disk.

As per https://github.com/GoogleChrome/sw-precache/issues/164#issue-174292972, I think that name was confusing, and we do have the opportunity to pick something different here. I was thinking `serverRenderedUrls`, but am open to alternatives.

jeffposnick:
R: @addyosmani @gauntface

Fixes #263 

This implements what's basically `sw-precache`'s `dynamicUrlToDependencies`, but with a name that @gauntface [agreed on](https://github.com/GoogleChrome/sw-precache/issues/164#issuecomment-284524338). It also has support for both strings and globs as the value that uniquely determines the URL's contents, so there should be parity with @HenrikJoreteg's changes made in https://github.com/GoogleChrome/sw-precache/pull/262

I'm not 100% sure that I've made the changes in all the right places (do I need to touch something in `sw-cli` too?), as I'm not as familiar with this portion of the codebase. Feedback is appreciated.


packages/sw-build/src/lib/get-file-manifest-entries.js
```diff
@@ -2,7 +2,7 @@ const errors = require('./errors');
 const filterFiles = require('./utils/filter-files');
 const getCompositeDetails = require('./utils/get-composite-details');
 const getFileDetails = require('./utils/get-file-details');
-const getStringDetails = require('./utils/get-string-hash');
+const getStringDetails = require('./utils/get-string-details');
```
addyosmani: This was changed from get-string-hash -> get-string-details. I found the original name more specific description wise. Although you're generating hash -> updating -> returning digest. Small nit to consider.

gauntface:
LGTM from my side - only concern was accepting strings - not sure it makes sense vs just adding a custom {url: '', revision: ''} object in the service - for sw-precache this wasn't an option so the linked PR was required. 

Happy to accept with or without for now - just raise an issue to discuss if you land as is :)

Otherwise code +1

packages/sw-build/src/lib/get-file-manifest-entries.js
jeffposnick: `get-string-details` and `get-string-hash` are two distinct modules, and the reference to `get-string-hash` in the previous commit was a copy/paste error 😦

jeffposnick:
Actually, since this is already reviewed, I'll merge as-is, and opened https://github.com/GoogleChrome/sw-helpers/issues/300 to discuss whether to remove string support.

PR #322, opened 2017-03-14
Fix example formatting

arthurevans:


jeffposnick:
Thanks, @arthurevans!

PR #375, opened 2017-03-21
Logging

gauntface:
R: @jeffposnick @addyosmani 

This introduces the following possible ways of logging:

```javascript

logHelper.log('Hello World 1');
logHelper.warn('Hello World 1');
logHelper.debug('Hello World 1');
logHelper.error('Hello World 1');


logHelper.log({
    message: 'Hello World 2'
});
// Same for .debug, .warn, .error

logHelper.log({
    message: 'Hello World 3',
    data: {
        example: 'sw-helpers 4 life',
        this: 'Will print out as key value pairs under Extra Information'
    }
});
// Same for .debug, .warn, .error

logHelper.log({
    message: 'Hello World 4',
    error: new Error('This is printed at top level')
});
// Same for .debug, .warn, .error
```

chrome logs
firefox logs

### DIFF
--- a/lib/log-group.js
+++ b/lib/log-group.js
@@ -1,6 +1,14 @@
 /* eslint-disable no-console */
 
+/**
+ * A simple helper to manage the print of a set of logs
+ */
 class LogGroup {
+  /**
+   * @param {object} input
+   * @param {string} input.title
+   * @param {boolean} input.isPrimary
+   */
   constructor({title, isPrimary}) {
     this._isPrimary = isPrimary || false;
     this._groupTitle = title || '';
@@ -8,10 +16,16 @@ class LogGroup {
     this._childGroups = [];
   }
 
-  addLog(logFunc) {
-    this._logs.push(logFunc);
+  /**
+   *@param {object} logDetails
+   */
+  addLog(logDetails) {
+    this._logs.push(logDetails);
   }
 
+  /**
+   * @param {object} group
+   */
   addChildGroup(group) {
     if (group._logs.length === 0) {
       return;
@@ -20,6 +34,9 @@ class LogGroup {
     this._childGroups.push(group);
   }
 
+  /**
+   * prints out this log group to the console.
+   */
   print() {
     let isFirefox = false;
     if (/Firefox\/\d*\.\d*/.exec(navigator.userAgent)) {
@@ -39,6 +56,10 @@ class LogGroup {
     this._closeGroup(isFirefox);
   }
 
+  /**
+   * Prints the specific logDetails object.
+   * @param {object} logDetails
+   */
   _printLogDetails(logDetails) {
     const logFunc = logDetails.logFunc ? logDetails.logFunc : console.log;
     let allArguments = [logDetails.message];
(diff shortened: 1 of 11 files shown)


lib/log-group.js
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * A simple helper to manage the print of a set of logs
+ */
+class LogGroup {
+  /**
+   * @param {object} input
+   * @param {string} input.title
+   * @param {boolean} input.isPrimary
+   */
+  constructor({title, isPrimary}) {
```
jeffposnick: Using a default parameter makes this a little nicer, since otherwise the object destructuring will fail when you forget to pass in an empty object:

```js
constructor({title, isPrimary} = {}) {
}
```

lib/log-group.js
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    this._childGroups.push(group);
+  }
+
+  /**
+   * prints out this log group to the console.
+   */
+  print() {
+    let isFirefox = false;
```
jeffposnick: How about storing this in a class variable and reusing it, since it will never change?

lib/log-group.js
```diff
@@ -0,0 +1,114 @@
...
+   * @param {boolean} isFirefox
+   */
+  _openGroup(isFirefox) {
+    if (this._isPrimary) {
+      // Only start a group is there are child groups
+      if (this._childGroups.length === 0) {
+        return;
+      }
+
+      const logDetails = this._logs.splice(0, 1)[0];
```
jeffposnick: This is clearer as

```js
const logDetails = this._logs.shift();
```

lib/log-group.js
```diff
@@ -0,0 +1,114 @@
...
+      if (isFirefox) {
+        // Firefox doesn't support colors lgos in console.group.
```
jeffposnick: lgos => logs

lib/log-helper.js
```diff
@@ -41,58 +50,186 @@ class LogHelper {
...
+      logLevel < minValidLogLevel) {
+      return false;
+    }
+
+    return true;
   }
 }
 
-export default new LogHelper();
+self.goog.temp = new LogHelper();
```
jeffposnick: Is the `temp` naming here intentional?

lib/log-group.js
```diff
@@ -0,0 +1,112 @@
+/* eslint-disable no-console */
+
+/**
+ * A simple helper to manage the print of a set of logs
+ */
```
jpmedley: This looks public-facing. Can you beef it up? I can't quite infer what I'm supposed to do with this.

gauntface: It's not public.

addyosmani:
I have some tweaks in mind for the styling of the logs, but will file later. This initially LGTM to land as is.

gauntface:
@jeffposnick @addyosmani 

In light of incoming change requests from Addy, I've added in "test" at the top level + `gulp build:shared`. This is for top level code that is shared.

gauntface:
Edge doesn't support colors so will fix that up and reopen the PR.

PR #377, opened 2017-03-22
Logging

gauntface:
R: @jeffposnick @addyosmani 

This introduces the following possible ways of logging:

```javascript

logHelper.log('Hello World 1');
logHelper.warn('Hello World 1');
logHelper.debug('Hello World 1');
logHelper.error('Hello World 1');


logHelper.log({
    message: 'Hello World 2'
});
// Same for .debug, .warn, .error

logHelper.log({
    message: 'Hello World 3',
    data: {
        example: 'sw-helpers 4 life',
        this: 'Will print out as key value pairs under Extra Information'
    }
});
// Same for .debug, .warn, .error

logHelper.log({
    message: 'Hello World 4',
    error: new Error('This is printed at top level')
});
// Same for .debug, .warn, .error
```

chrome logs
firefox logs

Edge doesn't support color in console AND it's group support is broken (everything ends up out of sync / out of order - so indentation is a mess).

I've dropped the grouping on edge and fallen back to just plain logs.

However, this looks really spammy. Compare Chrome and Edge for logs with (Message, Error and Data).

error in chrome

edge error

### DIFF
--- a/gulp-tasks/build.js
+++ b/gulp-tasks/build.js
@@ -16,8 +16,6 @@ const path = require('path');
 const gulp = require('gulp');
 const promisify = require('promisify-node');
 
-const resolve = require('rollup-plugin-node-resolve');
-const commonjs = require('rollup-plugin-commonjs');
 const {taskHarness, buildJSBundle} = require('../utils/build');
 
 const fsePromise = promisify('fs-extra');
--- a/gulp-tasks/build.js
+++ b/gulp-tasks/build.js
@@ -16,7 +16,9 @@ const path = require('path');
 const gulp = require('gulp');
 const promisify = require('promisify-node');
 
-const {taskHarness} = require('../utils/build');
+const resolve = require('rollup-plugin-node-resolve');
+const commonjs = require('rollup-plugin-commonjs');
+const {taskHarness, buildJSBundle} = require('../utils/build');
 
 const fsePromise = promisify('fs-extra');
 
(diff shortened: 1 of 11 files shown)


gulp-tasks/build.js
```diff
@@ -43,6 +45,28 @@ const buildPackage = (projectPath) => {
...
 };
 
+gulp.task('build:shared', () => {
+  return Promise.all([
+    buildJSBundle({
+      rollupConfig: {
+        entry: path.join(__dirname, '..', 'lib', 'log-helper.js'),
+        format: 'umd',
+        moduleName: 'goog.logHelper',
+        plugins: [
```
jeffposnick: Are these plugins needed? I don't see any external `require()`s in `log-helper.js`.

gulp-tasks/build.js
```diff
@@ -43,6 +45,28 @@ const buildPackage = (projectPath) => {
     });
 };
 
+gulp.task('build:shared', () => {
+  return Promise.all([
```
jeffposnick: There's only one promise, so you can ditch the `Promise.all()`.

gulp-tasks/build.js
```diff
@@ -43,6 +45,28 @@ const buildPackage = (projectPath) => {
     });
 };
 
+gulp.task('build:shared', () => {
```
jeffposnick: Just to confirm: this `umd` bundle is being consumed by the test, and not anything else, right?

gauntface: yes just the test

PR #380, opened 2017-03-23
Normalize RegExpRoute & ExpressRoute w.r.t. cross-origin requests

### LINKED ISSUE #237: Clarify how RegExp routing matches
**Library Affected**:
`sw-routing`, `sw-lib`

There's been some confusion about how the RegExp is applied when matching against URLs for routing. The docs should be explicit that the match doesn't have to apply to the whole URL, and that the `new RegExp(string)` syntax tends to be cleaner when including URL paths, since otherwise the `/` characters need to be escaped.

E.g. if we wanted to match all requests made for any resource on https://example.com, like https://example.com/page1, https://example.com/page2, a good RegExp to use would be `new RegExp('^https://example.com/')`. This ensures that the match is made at the start of the URL (via the `^` character) and it's not necessary to include any sort of `.*` wildcard, since that RegExp will match both 'https://example.com/page1' and 'https://example.com/page2' as-is.

### LINKED ISSUE #281: Relax same-origin requirement for Express-style routes?
**Library Affected**:
*sw-routing*

In reviewing examples/docs from @gauntface and @jpmedley, I've seen a couple of uses of Express-style routes to attempt to match cross-origin requests. As of https://github.com/GoogleChrome/sw-helpers/issues/151, we will only match same-origin requests if someone provides an Express-style route, and we will throw a runtime error if you pass in a string that includes a full URL, begining with `'http'`.

Seeing the natural inclination of both of those folks to try to use the library in that way, though, makes me want to at least revisit that decision. Here are the options that I see:

1. Leave things as-is. Documentation along with the runtime assertions will hopefully make sure that folks know how to use Express-style routes, and they'll fall back to RegExp routes for cross-origin requests.
2. Open things up so that you can pass in a full URL as an Express-style route, but overload the routing logic—when you pass in a full URL, you perform matches against the entire incoming URL for both same- and cross-origin requests, but when you pass in a partial URL you only perform matches against same-origin requests. (This is similar to the situation we ended up with in `sw-toolbox`, except we triggered cross-origin matching based on the presence of a separate `{origin}` option.)
3. Open things up to accept full URLs, but don't use overloaded routing logic. Always match all routes, whether they were specified as partial or full URLs, against both same- and cross-origin requests.

The distinction between 2. and 3. is illustrated by considering the following route:

```js
swLib.router.registerRoute('/path/to/file', staleWhileRevalidate);
```

What if there were an incoming request for `https://cross-origin-domain.com/path/to/file`. Should that route match?

It wouldn't match if we went with option 2., and that's probably what the developer intends, but then we need to explain clearly how we use change our cross-domain matching restrictions depending on whether the input was a full or partial URL.

It would match if we went with option 3., but matching in this case is probably *not* what the developer intended. I'm mostly worried about these false positives when a developer specifies a route that has a very broad wildcard, like `/(.*)`, which would end up matching all requests, both same- and cross-origin.

jeffposnick:
R: @addyosmani @gauntface

Fixes #281 and fixes #237 

I *think* this accurately reflects what we agreed upon as a resolution to #281. I may be incorrect about what the takeaway from that was, so feel free to correct me and I'll correct the logic in this PR!

This change opens up `ExpressRoute` so that if you pass in a `path` value starting with `'https://'`, it will match cross-origin requests that go against the origin that you specify.

If you pass in a `path` value starting with `'/'` then it will continue to only match same-origin requests, which is similar to the old behavior.


packages/sw-lib/src/lib/router.js
```diff
@@ -54,14 +54,13 @@ import ErrorFactory from './error-factory.js';
...
 class Router extends SWRoutingRouter {
-
   /**
    * @param {String|Regex|Route} capture The capture for a route can be one
    * of three types.
-   * 1. It can be an Express style route, like: '/example/:anything/route/'
-   *    The only gotcha with this is that it will only capture URL's on your
-   *    origin.
-   * 1. A regex that will be tested against request URL's.
+   * 1. It can be an Express style route, like '/wildcard/prefix/(.*)' for
```
gauntface: Could you keep the naming here ':someName' just to probably seperate express from regex and could you include an express name in the cross-origin route.

packages/sw-routing/src/lib/error-factory.js
```diff
@@ -16,9 +16,9 @@
 import ErrorFactory from '../../../../lib/error-factory';
 
 const errors = {
-  'express-route-requires-absolute-path': `When using ExpressRoute, you must
-    provide a path that starts with a '/' character. You can only match
-    same-origin requests. For more flexibility, use RegExpRoute.`,
+  'express-route-invalid-path': `When using ExpressRoute, you must
+    provide a path that starts with a '/' character (to match same-origin
+    requests) or that starts with 'https://' (to match cross-origin requests)`,
```
gauntface: It can also start with 'http://' and what is this error thrown? If it's when the route can't be passed, it may be worth just saying the route must be a valid URL starting with a '/' or 'http://' and leave out the same-origin / cross-origin request details (I'm just thinking that this message is overly verbose).

packages/sw-routing/src/lib/express-route.js
```diff
@@ -25,29 +25,37 @@ import pathToRegExp from 'path-to-regexp';
  * library to transform the `path` parameter into a regular expression, which is
  * then matched against the URL's path.
  *
- * Please note that `ExpressRoute` can only match requests for URLs that are on
- * the same-origin as the current page. If you need to match cross-origin
- * requests, you can use either a generic [`Route`]{@link Route} or a
- * [`RegExpRoute`]{@link RegExpRoute}.
+ * Please note that `ExpressRoute` can match either same-origin or cross-origin
+ * requests. To match only same-origin requests, use a `path` value that begins
+ * with `'/'`, e.g. `'/path/to/(.*)'`. To match cross-origin requests, use
```
gauntface: Same as above - could we use an express name

packages/sw-routing/src/lib/express-route.js
```diff
@@ -66,8 +74,8 @@ class ExpressRoute extends Route {
    *        HTTP method. Defaults to `'GET'` if not specified.
    */
   constructor({path, handler, method}) {
-    if (path.substring(0, 1) !== '/') {
-      throw ErrorFactory.createError('express-route-requires-absolute-path');
+    if (!(path.startsWith('/') || path.startsWith('https://'))) {
```
gauntface: why no http:// ? Can't images use HTTP without causing the insecure content issue?

I'm also not 100% sw-helpers should be enforcing this - but instead leave the browser to enforce this.

Also is this checked needed? This was only added to block cross origin urls.

packages/sw-routing/src/lib/error-factory.js
jeffposnick: I've updated the message to reflect the fact that we're going to support `http:` URLs.

This is thrown when you try to instantiate an `ExpressRoute` using an invalid `path`. I am going to keep it verbose because the information is useful in this context—if someone passed in a bad `path`, I want to give them as much relevant information as I can so that they can figure out what they should pass in instead.

packages/sw-lib/src/lib/router.js
jeffposnick: Okay, if you prefer that format for the example I'll switch back.

packages/sw-routing/src/lib/express-route.js
jeffposnick: If you prefer that, I'll switch back. (Those wildcards are valid in Express routes, though, so I figured including some examples using them wouldn't hurt.)

packages/sw-routing/src/lib/express-route.js
jeffposnick: I'll open it up to `http` if we want to support that use case.

I do think that this check is needed, yes. If they don't pass in something that starts with `'/'` or `http` then it's not a valid `path`, so shouldn't we tell them that? The alternative would be to allow them to create an `ExpressRoute` that will never match anything, and that doesn't seem preferable.

jeffposnick:
PTAL.

I was not planning on changing the behavior of `RegExpRoute` and was assuming that it would continue to match against the full URL unconditionally, regardless of whether the request is same- or cross-origin. https://github.com/GoogleChrome/sw-helpers/issues/237 tracks providing a better explanation in the docs/samples about `RegExpRoute`'s behavior.

jeffposnick:
Okay! So this is a fun topic.

I think the latest round of changes leads to an acceptable behavior, and takes into account the feedback about `RegExpRoute` route as well. But please look over the PR again and let me know if I got anything wrong.

Possibly controversial: for cross-origin requests, the match has to occur at the start of the string. @gauntface suggested that we should require the the match has to apply to the entire string, but since same-origin requests play nicely with partial-string matches, I wanted to keep the same semantics with cross-origin requests.

This means that `new RegExp('https://cross-origin\.com')` can match `'https://cross-origin.com/path/to/file'`, which is what I think developers would expect.

If I were to require the the match applies to the entire string, then something like `new RegExp('https://cross-origin\.com/.*')` would be required.

packages/sw-routing/src/lib/express-route.js
```diff
@@ -66,8 +74,8 @@ class ExpressRoute extends Route {
    *        HTTP method. Defaults to `'GET'` if not specified.
    */
   constructor({path, handler, method}) {
-    if (path.substring(0, 1) !== '/') {
-      throw ErrorFactory.createError('express-route-requires-absolute-path');
+    if (!(path.startsWith('/') || path.startsWith('http'))) {
+      throw ErrorFactory.createError('express-route-invalid-path');
     }
```
jeffposnick: I switched things over to check for a prefix of `'http'` rather than `'https://'`. @gauntface's comment was on an earlier incarnation of this PR. Does this look okay now?

jeffposnick:
> I'm slightly confused. The previous comment noted that cross-origin requests require matches to occur at the start of the string but we later say that we want to preserve the same partial-string match semantics of same-origin requests. Don't both require the match to occur at the start of the string? I may be reading this incorrectly.

Assuming a cross-origin URL of `'https://cross-origin.com/path/to/file'`:
1. `/https:\/\/cross-origin\.com/` leads to a partial-string match that begins at character 0. Under this PR, it would trigger the route.
2. `/https:\/\/cross-origin\.com.*/` leads to a full-string match that begins at character 0. Under this PR, it would also trigger the route.
3. `/file/` leads to a partial-string match that begins at character 30(ish). Under this PR, it would not trigger the route. (It would trigger the route if it were a same-origin URL, though.)

I'm of the opinion that developers would expect 1. to match. If we required that the regex match the full string for cross-origin requests, then developers would have to remember to put in that trailing `.*` wildcard, *à la* 2.

If @addyosmani and @gauntface feel like 1. **shouldn't** match, then I'd need to change the logic in this PR.

(And if you feel like the partial-string match in 3. shouldn't trigger the route for same-origin requests, then that would be even more of a change.)

gauntface:
I'd vote for option 1 **not** matching because it's only partial. @addyosmani what do you think?

jeffposnick:
If 1. doesn't match, what do you think about 3. matching for same-origin URLs? I'd think it should either be both or neither. (That's what I meant about preserving the same semantics.)

To illustrate why 3. is problematic:

If the request URL is `'https://sameorigin.com/path/to/file.jpg'`, should `/jpg$/` match that? Or is a wildcard like`/.*jpg$/` required? And then if we require `/.*jpg$/`, how do you prevent `/.*jpg$/` from matching a cross-origin URL like `'https://cross-origin.com/path/to/file.jpg'`?

gauntface:
I understand what you are asking now.

I'd argue that both your example regex's are dangerous, but I like the flexibilty, maybe we should just allow them and see what developer feedback is.

Is there any chance we can capture some of this is in tests that are a little easier to follow / parse? I'm thinking something like:

```
const testCases = [
  {
    requestUrl: 'http://thirdparty.origin.com/',
    regex: /\//,
    express: '/'
    shouldMatch: false,
    reason: 'Because third party origin must match partially on the full url.'
  },
  {
    requestUrl: 'http://current-site.origin.com/',
    regex: /\//,
    express: '/'
    shouldMatch: true,
    reason: 'Because the sites origin only has to match partially on the pathname.'
  },
  {
    requestUrl: 'http://thirdparty.origin.com/',
    regex: /http:\/\/thirdparty.origin.com\//,
    express: 'http://thirdparty.origin.com/'
    shouldMatch: true,
    reason: 'Because third party origin will match the full url.'
  },
  {
    requestUrl: 'http://current-site.origin.com/',
    regex: /http:\/\/thirdparty.origin.com\//,
    express: 'http://thirdparty.origin.com/'
    shouldMatch: false,
    reason: 'Because the origins do not match.'
  },
];
```

jeffposnick:
Some of those scenarios are covered by tests in this PR, but sure, I'm happy to add in a commit to this PR with a dedicated set of tests using a consistent pattern of inputs.

But I'm not sure what the takeaway from this is:

> I'd argue that both your example regex's are dangerous, but I like the flexibilty, maybe we should just allow them and see what developer feedback is.

Are you okay with the logic as-is in this PR?

gauntface:
Yes - fine with the logic in this PR.

addyosmani:
> I'd argue that both your example regex's are dangerous, but I like the flexibilty, maybe we should just allow them and see what developer feedback is.

As long as we're open to reconsidering if we find developer feedback requires a change, this approach seems fine to me.

addyosmani:
I'm happy to approve with the current logic and assume we're going to land the commit with additional test coverage before merging this in. Feel free to re-ping if you need a review on those changes.

jeffposnick:
Will do.

jeffposnick:
Merging, now that there are additional tests.

PR #381, opened 2017-03-23
Use a specific cache name when finding the NavigationRoute response

### LINKED ISSUE #321: registerNavigationRoute() should default to looking in the precaching cache
**Library Affected**:
*sw-lib*

https://github.com/GoogleChrome/sw-helpers/pull/307 added support for `swLib.registerNavigationRoute()`, and it uses `caches.match(url)` under the hood to return the response. It would make sense to change that to default to `caches.match(url, {cacheName})`, where `cacheName` defaults to `this._revisionedCacheManager.getCacheName()` but can be changed via an option if the developer wants something else.

It's not crucial for the hackathon, but it can work around an issue when upgrading from `sw-precache`, when there might still be an older entry with the same URL in one of the caches that `sw-precache` had created.

jeffposnick:
R: @addyosmani @gauntface

Fixes #321

### DIFF
--- a/packages/sw-lib/src/lib/router.js
+++ b/packages/sw-lib/src/lib/router.js
@@ -54,6 +54,15 @@ import ErrorFactory from './error-factory.js';
  * @memberof module:sw-lib
  */
 class Router extends SWRoutingRouter {
+  /**
+   * Constructs a light wrapper on top of the underlying `Router`.
+   * @param {String} revisionedCacheName The cache name used for entries cached
+   *        via cacheRevisionedAssets().
+   */
+  constructor(revisionedCacheName) {
+    super();
+    this._revisionedCacheName = revisionedCacheName;
+  }
 
   /**
    * @param {String|Regex|Route} capture The capture for a route can be one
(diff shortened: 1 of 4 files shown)


packages/sw-lib/src/lib/router.js
```diff
@@ -112,14 +121,23 @@ class Router extends SWRoutingRouter {
...
+   *        used by cacheRevisionedAssets().
    */
   registerNavigationRoute(url, options = {}) {
     if (typeof url !== 'string') {
       throw ErrorFactory.createError('navigation-route-url-string');
     }
 
+    // Allow folks to explicitly pass in a null/undefined cacheName option if
+    // they want that behavior.
+    const cacheName = 'cacheName' in options ?
```
addyosmani: What's the use-case for specifying a null or undefined cacheName? (I may be missing something obvious). Is it for the case where folks want to treat options.cacheName as optional and ignore setting a name themselves?

jeffposnick: Passing in `null`/`undefined` will default to the `caches.match()` behavior that was in place prior to this PR, i.e. return the first match for the URL, regardless of which cache it's in. Not necessarily the ideal use case (hence this PR), but I figured I'd provide a backdoor in case people really wanted that behavior?

jeffposnick:
I'm going to add in some additional logging statements just to clarify things, since @gauntface's `log-helper` library is 🔥

packages/sw-routing/src/lib/navigation-route.js
```diff
@@ -69,9 +70,32 @@ class NavigationRoute extends Route {
...
-      return event.request.mode === 'navigate' &&
-        whitelist.some((regExp) => regExp.test(url.pathname)) &&
-        !blacklist.some((regExp) => regExp.test(url.pathname));
+      let matched = false;
+      let message;
+
+      if (event.request.mode === 'navigate') {
+        if (whitelist.some((regExp) => regExp.test(url.pathname))) {
+          if (blacklist.some((regExp) => regExp.test(url.pathname))) {
+            message = `The navigation route is not being used, since while ` +
```
gauntface: Could we shorten this message?

`The navigation request is in both the whitelist and blacklist of the NavigationRoute.`

jeffposnick: The implication is ambiguous then—the developer won't necessarily know that the route isn't being used. I'll shorten the second clause of the sentence, but will keep the first clause.

packages/sw-routing/src/lib/navigation-route.js
```diff
@@ -69,9 +70,32 @@ class NavigationRoute extends Route {
...
+
+        logHelper.debug({
+          that: this,
+          message,
+          data: {'request-url': url.href, whitelist, blacklist, handler},
+        });
+      }
+
+      return matched;
     };
```
addyosmani: I'm glad we switched to using the log-library for this 👍

PR #382, opened 2017-03-24
Use a single fetch handler per Router

### LINKED ISSUE #354: Use a single fetch event handler within Router
**Library Affected**:
*sw-routing*

Right now, `Router` creates multiple `fetch` event handlers, one for each call to `registerRoutes()`. There probably should be a single global `fetch` handler that it registers, and it evaluates the list of routes at runtime.

jeffposnick:
R: @addyosmani @gauntface

The diff is pretty busy, but this is basically just shifting the `fetch` handler into the `Router` constructor, and storing the list of routes in a private member variable.

Fixes #354

### DIFF
--- a/packages/sw-routing/src/lib/router.js
+++ b/packages/sw-routing/src/lib/router.js
@@ -53,7 +53,7 @@ class Router {
    * Start with an empty array of routes, and set up the fetch handler.
    */
   constructor() {
-    this._routes = [];
+    this._routes = new Map();
 
     self.addEventListener('fetch', (event) => {
       const url = new URL(event.request.url);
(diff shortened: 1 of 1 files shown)


packages/sw-routing/src/lib/router.js
```diff
@@ -122,7 +71,7 @@ class Router {
 
       let responsePromise;
       let matchingRoute;
-      for (let route of (routes || [])) {
+      for (let route of this._routes) {
         if (route.method !== event.request.method) {
```
gauntface: Did we drop support for 'all' like sw-toolbox had?

If so, should we group routes together by method - reducing this loop for each request?

jeffposnick: Yeah, we don't have support for `all`. I can switch to keeping track of routes per-method to optimize the loop.

jeffposnick:
PTAL at the revised approach to confirm you're still 👍

PR #386, opened 2017-03-28
add replay tag in bg sync queue

### LINKED ISSUE #384: Background Sync queue needs way to manual replay
**Library Affected**:
sw-background-sync-queue

**Browser & Platform**:
all browsers

There is a need for a manual way to replay all queued requests.
This could be also used in case of sw-offline-analytics where all the requests can be manually fired up on the `document ready` may be...

prateekbh:

Fixes #384

Adds another constant tag for replaying the queue

### DIFF
--- a/packages/sw-background-sync-queue/src/lib/constants.js
+++ b/packages/sw-background-sync-queue/src/lib/constants.js
@@ -6,6 +6,7 @@ const defaultQueueName = 'DEFAULT_QUEUE';
 const tagNamePrefix = 'SW_BACKGROUND_QUEUE_TAG_';
 const broadcastMeta = 'SW_BACKGROUND_SYNC_QUEUE';
 const allQueuesPlaceholder = 'QUEUES';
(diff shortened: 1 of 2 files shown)


jeffposnick:
The code change is 👍 

Are there any tests that can be updated to ensure that this behaves as expected?

prateekbh:
@jeffposnick @addyosmani current tests are mostly unit tests for sub pieces, there is no test on service worker behaviour as such.
Can you guys point me to any other tests which cover service worker functionality in the repo(e.g activate/fetch event etc)?

jeffposnick:
https://github.com/GoogleChrome/sw-helpers/blob/master/packages/sw-routing/test/browser/end-to-end.js is an example of an end-to-end test.

If there aren't any existing tests that would need to be updated to reflect this new functionality, then I'd be fine with moving ahead with this PR and then filing a separate issue to track the creation of new end-to-end tests that covered it.

prateekbh:
yeah there are no end to end tests right now, i'll open a separate PR for these

PR #388, opened 2017-03-29
Fixes #355

### LINKED ISSUE #355: Clarify that developers shouldn't hardcode the manifest
**Library Affected**:
*sw-lib*

The https://googlechrome.github.io/sw-helpers/reference-docs/stable/v0.0.13/module-sw-lib.html#main docs page has the following example:

```js
// Caching assets and registering routes.
goog.swlib.cacheRevisionedAssets([
  {
    url: '/',
    revision: '1234'
  },
  {
    url: '/styles/main.css',
    revision: 'abcd'
  }
]);
```

Reading this page gave @kaycebasques the impression that you should hardcode the manifest array and update the `revision` details by hand, when in fact we want people to use `sw-build` to generate that information. We should update the docs to make it clear that the manifest array should be generated as part of a build process.

gauntface:
R: @jeffposnick @addyosmani 

Fixes #355

### DIFF
--- a/packages/sw-lib/src/index.js
+++ b/packages/sw-lib/src/index.js
@@ -41,11 +41,11 @@ if (!assert.isSWEnv()) {
  * goog.swlib.cacheRevisionedAssets([
  *   {
  *     url: '/',
- *     revision: '1234'
+ *     revision: '613e6c7332dd83e848a8b00c403827ed'
  *   },
  *   {
  *     url: '/styles/main.css',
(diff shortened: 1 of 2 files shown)


jeffposnick:
How about taking it one step further and including values that look like actual MD5 hashes rather than `1234` and `abcd` in the samples? That would hammer things home.

gauntface:
Done.

PR #391, opened 2017-03-30
Refactor the CacheableResponsePlugin to a separate class

jeffposnick:
R: @addyosmani @gauntface

This is part of #282, in which the Plugin classes are refactored to extend a base class, and expose the lifecycle methods.

In addition to refactoring, it also adds tests, as this module was previously untested.

### DIFF
--- a/packages/sw-cacheable-response/test/automated-test-suite.js
+++ b/packages/sw-cacheable-response/test/automated-test-suite.js
@@ -1,0 +1,69 @@
+/*
+ Copyright 2017 Google Inc. All Rights Reserved.
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+const seleniumAssistant = require('selenium-assistant');
+const swTestingHelpers = require('sw-testing-helpers');
+const testServer = require('../../../utils/test-server.js');
+
+const RETRIES = 3;
+const TIMEOUT = 10 * 1000;
+const packageName = 'sw-cacheable-response';
+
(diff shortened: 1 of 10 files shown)


packages/sw-cacheable-response/test/automated-test-suite.js
```diff
@@ -0,0 +1,69 @@
...
+ limitations under the License.
+ */
+
+const seleniumAssistant = require('selenium-assistant');
+const swTestingHelpers = require('sw-testing-helpers');
+const testServer = require('../../../utils/test-server.js');
+
+const RETRIES = 3;
+const TIMEOUT = 10 * 1000;
+const packageName = 'sw-cacheable-response';
```
gauntface: Could you make this upper case - sorry I was reading this and thought it was some sort of automated script for each package.

packages/sw-cacheable-response/test/sw/cacheable-response.js
```diff
@@ -0,0 +1,108 @@
...
+);
+
+const expect = self.chai.expect;
+mocha.setup({
+  ui: 'bdd',
+  reporter: null,
+});
+
+describe('Test of the CacheableResponse class', function() {
+  const validStatus = 418;
```
gauntface: Can you make all these upper case since they dont change.

packages/sw-cacheable-response/test/automated-test-suite.js
jeffposnick: Updated.

packages/sw-cacheable-response/test/sw/cacheable-response.js
jeffposnick: I'd prefer to leave these names as-is, since the `const` does make it clear that they don't change, and they're not in the top-level scope.

(Okay, the real reason that I'd like to leave them as-is is because it makes it easier to use them later on via ES2015 object literal shorthand, e.g. `new Response('', {headers});` vs. `new Response('', {headers: HEADERS});`.)

Let me know if that's unreasonable and I'll switch 😈

jeffposnick:
I've changed one of the names, and I also realized that adding some logging would be a good idea.

Let me know your thoughts about the other variable names.

packages/sw-cacheable-response/test/sw/cacheable-response.js
gauntface: I'd rather you changed to VALID_HEADERS - just makes reading and following the tests easier

PR #398, opened 2017-03-31
Static file globs

gauntface:
R: @jeffposnick @addyosmani 

First PR on moving over to sw-precache.

globPatterns => staticFileGlobs

Will add the CLI flag for this once sw-build has the rename in place - just for consistency.


jeffposnick:
The code changes look fine. From a naming perspective, do you prefer `globPatterns` to `staticFileGlobs`? Because if so, instead of abandoning `globPatterns` everywhere, we could just accept `staticFileGlobs` as an alias to `globPatterns`, but encourage new developers to use `globPatterns`.

addyosmani:
Imo, both `globPatterns` and `staticFileGlobs` are both equally understandable but open to aliasing depending on what Matt feels works best here.

packages/sw-cli/src/index.js
```diff
@@ -223,7 +223,7 @@ class SWCli {
         rootDirPath, fileExtentionsToCache);
       return swBuild.generateFileManifest({
         rootDirectory: rootDirPath,
-        globPatterns: [globPattern],
+        staticFileGlobs: [globPattern],
         globIgnores: [
           path.join(rootDirPath, fileManifestName),
         ],
```
addyosmani: All 👍

gauntface:
Naming I don't care about at this point to be honest.

gauntface:
I'm going to fix the Windwos build as I have a hunch on where the issue but not 100%.

jeffposnick:
👍

addyosmani:
All 💚

PR #399, opened 2017-03-31
Revamps the defaults around caching & logging opaque responses

### LINKED ISSUE #78: Revamp the defaults around opaque responses

jeffposnick:
R: @addyosmani @gauntface

Fixes #78 

This uses the logging layer implemented in `sw-cacheable-response` to let the developer know when there's a response that would otherwise be cached, but isn't due to the response code.

It changes the defaults for `NetworkFirst` and `StaleWhileRevalidate` to allow for opaque responses with status of `0` to be cached and used. The default for everything else is to require a response status of `200`.

One thing that it does not do is add any tests for the various behaviors, since `sw-runtime-caching` is currently untested 😨  I've got some WIP tests at https://github.com/GoogleChrome/sw-helpers/tree/runtime-caching-test and am fine with holding off on merging this PR until those tests are checked in, and then I'll modify this PR to include additional tests for response codes. Or you could just merge it sooner with the understanding that I'll cover the response code tests in the separate testing PR.

In any case, feel free to review for general feedback in the meantime.


packages/sw-cacheable-response/src/lib/cacheable-response.js
```diff
@@ -65,14 +65,19 @@ class CacheableResponse {
...
    * @return {boolean} `true` if the `Response` is cacheable, based on the
    *          configuration of this object, and `false` otherwise.
    */
-  isResponseCacheable({response} = {}) {
+  isResponseCacheable({request, response} = {}) {
     assert.isInstance({response}, Response);
+    if (request) {
+      assert.isInstance({request}, Request);
+    }
+
```
addyosmani: Not a blocker: I wonder if it would be more clear for us to be rejecting a promise for isResponseCacheable style interfaces rather than completely throwing, but I guess we would still need to return from the assertion so perhaps this is fine after-all.

packages/sw-runtime-caching/src/lib/network-first.js
```diff
@@ -13,13 +13,19 @@
...
 
+import {CacheableResponse} from '../../../sw-cacheable-response/src/index';
 import Handler from './handler';
 import assert from '../../../../lib/assert';
 
 /**
  * An implementation of a [network first](https://developers.google.com/web/fundamentals/instant-and-offline/offline-cookbook/#network-falling-back-to-cache)
  * request strategy.
  *
+ * By default, [opaque responses](http://stackoverflow.com/questions/39109789/what-limitations-apply-to-opaque-responses)
```
addyosmani: I wonder if there's value using the shorter http://stackoverflow.com/q/39109789 form for links that are going to appear in source but will defer to your judgement call.

packages/sw-runtime-caching/src/lib/request-wrapper.js
```diff
@@ -96,6 +96,16 @@ class RequestWrapper {
       if (this.pluginCallbacks.cacheWillUpdate.length !== 1) {
         throw ErrorFactory.createError('multiple-cache-will-update-plugins');
       }
+      this.cacheableResponseCheck = this.pluginCallbacks.cacheWillUpdate[0];
+    } else {
+      // If the was no plugin implementing cacheWillUpdate passed in, then
```
addyosmani: If there was no plugin?

packages/sw-runtime-caching/src/lib/request-wrapper.js
```diff
@@ -279,17 +306,10 @@ class RequestWrapper {
...
-        responses with a status of 200 are cached. You can configure the
-        cacheableResponse plugin to change this default.`.replace(/\s+/g, ' '));
-
-      if (waitOnCache) {
-        // If the developer request to wait on the cache but the response
-        // isn't cacheable, throw an error.
-        throw ErrorFactory.createError('invalid-reponse-for-caching');
-      }
+    } else if (!cacheable && waitOnCache) {
+      // If the developer request to wait on the cache but the response
```
addyosmani: If the developer requested

addyosmani:
> It changes the defaults for NetworkFirst and StaleWhileRevalidate to allow for opaque responses with status of 0 to be cached and used. The default for everything else is to require a response status of 200.

This seems reasonable to me. 

> One thing that it does not do is add any tests for the various behaviors, since sw-runtime-caching is currently untested 😨 I've got some WIP tests at https://github.com/GoogleChrome/sw-helpers/tree/runtime-caching-test and am fine with holding off on merging this PR until those tests are checked in, and then I'll modify this PR to include additional tests for response codes

Once feedback is addressed, I'm fine with us merging this earlier. Do we have an issue for tracking tests missing there?

jeffposnick:
The lack of tests is tracked in https://github.com/GoogleChrome/sw-helpers/issues/301

packages/sw-cacheable-response/src/lib/cacheable-response.js
jeffposnick: `isResponseCacheable()` synchronously returns a `boolean` when it's used correctly, so `throw`ing when it's used incorrectly, as opposed to returning a rejected promise, seems like the right thing to do.

gauntface: Why are we throwing on this if it's only used for debugging?

packages/sw-runtime-caching/src/lib/network-first.js
```diff
@@ -13,13 +13,19 @@
...
 
+import {CacheableResponse} from '../../../sw-cacheable-response/src/index';
 import Handler from './handler';
 import assert from '../../../../lib/assert';
 
 /**
  * An implementation of a [network first](https://developers.google.com/web/fundamentals/instant-and-offline/offline-cookbook/#network-falling-back-to-cache)
  * request strategy.
  *
+ * By default, [opaque responses](http://stackoverflow.com/q/39109789)
```
gauntface: Can we not link to a stack overflow comment and just explain here or some other place on the docs.

gauntface: Also worth say "By default, Network First will cache ...." default is a tad generic.

packages/sw-runtime-caching/src/lib/network-first.js
```diff
@@ -53,6 +59,10 @@ class NetworkFirst extends Handler {
   constructor(input = {}) {
     super(input);
 
+    const cacheableResponse = new CacheableResponse({statuses: [0, 200]});
+    this.requestWrapper.cacheableResponseCheck =
```
gauntface: This line of code  . . . . . wah?

Can this be simplified, abstracted to something more sane, something?

gauntface: Could this not be a plugin?

packages/sw-runtime-caching/src/lib/network-first.js
```diff
@@ -53,6 +59,10 @@ class NetworkFirst extends Handler {
...
+      cacheableResponse.isResponseCacheable.bind(cacheableResponse);
+
```
addyosmani: It seems a little strange that you need to bind a cacheableResponse's entry to...the parent? Could you explain what's going on here?

packages/sw-runtime-caching/src/lib/request-wrapper.js
```diff
@@ -96,6 +96,16 @@ class RequestWrapper {
       if (this.pluginCallbacks.cacheWillUpdate.length !== 1) {
         throw ErrorFactory.createError('multiple-cache-will-update-plugins');
       }
+      this.cacheableResponseCheck = this.pluginCallbacks.cacheWillUpdate[0];
```
gauntface: Sorry but I don't follow the differences between this.cacheableResponseCheck and this._defaultCacheableResponseCheck where default isn't static and IF it's set it's another weird binding thing.

packages/sw-runtime-caching/src/lib/request-wrapper.js
```diff
@@ -106,6 +116,29 @@ class RequestWrapper {
...
 
   /**
+   * Provides a method of setting the cacheableResponseCheck after the
+   * `RequestWrapper` has already been constructed, but only if it wasn't set
+   * inside the constructor.
+   *
+   * @param {function} callback The function used to determine whether a
+   *        response is cacheable.
+   */
+  set cacheableResponseCheck(callback) {
```
gauntface: This makes it even hard to reason with the above.......

packages/sw-runtime-caching/src/lib/request-wrapper.js
addyosmani: What is this entry in cacheWillUpdate supported to contain at this point?

packages/sw-runtime-caching/src/lib/request-wrapper.js
```diff
@@ -96,6 +96,16 @@ class RequestWrapper {
...
+    } else {
+      // If the was no plugin implementing cacheWillUpdate passed in, then
+      // construct a new one which just checks to make sure the response's
+      // status is 200. This matches the default behavior of cache.add().
+      // The default check might be overridden later on if someone calls the
+      // cacheableResponseCheck setter.
+      const cacheableResponse = new CacheableResponse({statuses: [200]});
+      this._defaultCacheableResponseCheck =
+        cacheableResponse.isResponseCacheable.bind(cacheableResponse);
     }
```
addyosmani: Similar question to earlier on logic.

packages/sw-runtime-caching/src/lib/request-wrapper.js
```diff
@@ -106,6 +116,29 @@ class RequestWrapper {
...
+    }
+  }
+
+  /**
+   * @private
+   * @return {function} The function used to determine whether a response is
+   *         cacheable.
+   */
+  get cacheableResponseCheck() {
+    return this._cacheableResponseCheck || this._defaultCacheableResponseCheck;
```
gauntface: Can this not be just set in the constructor? i.e. when will this dynamically be set / changed? And when it does, could we actually go with functions for this instead of magic here.

setCacheableResponseCheck() and performCacheableResponseCheck()

packages/sw-runtime-caching/src/lib/stale-while-revalidate.js
```diff
@@ -39,6 +45,23 @@ import assert from '../../../../lib/assert';
...
+   * @param {RequestWrapper} [input.requestWrapper] An optional `RequestWrapper`
+   *        that is used to configure the cache name and request plugins. If
+   *        not provided, a new `RequestWrapper` using the
+   *        [default cache name](#defaultCacheName) will be used.
+   */
+  constructor(input = {}) {
+    super(input);
+
+    const cacheableResponse = new CacheableResponse({statuses: [0, 200]});
+    this.requestWrapper.cacheableResponseCheck =
```
gauntface: Again this feels really weird and bad and likely to trip someone up sooner or later.

gauntface:
What happens with NetworkOnly? I assume there is no messing with the status code - but wanted to confirm.

jeffposnick:
My takeaway from the review feedback is that enforcing the "plugin" behavior for the cacheable response check has led to unnecessarily confusing code. I'm going to rethink that approach, letting developers continue to initialize it as a "plugin" when they explicitly create their own `RequestWrapper`, but providing a saner way to allow loosely-coupled classes within `sw-runtime-caching` to override the default behavior when needed.

gauntface:
I'm probably missing a lot or prior context - which is probably a sign that we made some decisions and now it's too magical that we can't decypher it from first glance.

Was it a rule that there could only be one of these plugins which the developer could overwrite with a custom implementation of that plugin?

Was there a reason for moving away from using a plugin?

jeffposnick:
> Was it a rule that there could only be one of these plugins which the developer could overwrite with a custom implementation of that plugin?

That's a limitation that's enforced for a certain subset of the lifecycle events, like `cacheWillUpdate`, yes. There's a runtime error if you set multiple listeners.

> Was there a reason for moving away from using a plugin?

We'd still allow the use of a plugin for registering `cacheWillUpdate` if the developer explicitly chooses to do so.

The nuance here is that there are three different situations that need to be supported cleanly:

1. The developer explicitly opts-in to defining their own `cacheWillUpdate` behavior, most likely via passing in a `CacheableResponsePlugin` into the `RequestWrapper` constructor.
2. The developer doesn't define `cacheWillUpdate`, but we still need a sensible default behavior, and that default behavior is to require a `200` response.
3. The developer doesn't define `cacheWillUpdate`, and we want a different default behavior (for the `NetworkFirst` and `StaleWhileRevalidate` handlers), which allows for either `0` or `200` responses.

I jumped through some hoops in this PR to minimize the changes to the existing code path the gets executed for case 1, but it ended up making the code paths for 2 and 3 more complicated than they need to be.

My plan is to modify the approach so that specifying a `cacheWillUpdate` callback can be done as a parameter to `RequestWrapper.fetchAndCache()` instead of requiring that it's a property of the `RequestWrapper` itself. I think that should lead to a cleaner code path for 2 and 3.

addyosmani:
> My plan is to modify the approach so that specifying a cacheWillUpdate callback can be done as a parameter to RequestWrapper.fetchAndCache() instead of requiring that it's a property of the RequestWrapper itself. I think that should lead to a cleaner code path for 2 and 3.

This sounds reasonable to me. I'm going to adopt the Lighthouse chillin label to this PR so we know not to directly merge it but come back to it once you've had a chance to modify the approach based on learnings from the comments thread.

packages/sw-cacheable-response/src/lib/cacheable-response.js
jeffposnick: Okay, I'll remove it.

gauntface:
I'm sorry but I'm still really struggling to accept this API.

1. I'm still not following why you need to pick out a method from a fairly self contained object (i.e. given a request, tell me if I should cache it or not). Can you not just pass around the CacheableResponse object at necessary rather than do the bind dance and then pass around the function?
1. The order of the three scenarios concerns because it's really not clear what the priority is or should be from the API surface - i.e. should the plugin in the request wrapper really take precedence of over the plugin or response check passed into the fetchAndCache method?
1. The naming of some of the variables is also confusing at points (i.e. `const cacheableResponseCheck = this.cacheableResponseCheck || defaultCacheableResponseCheck || this.defaultCacheableResponseCheck;` - which ignores that this.defaultCacheableResponseCheck is a getter).

Maybe changing to:

`const cacheableResponseCheck = pluginCachableResponseCheck || cacheStrategyCachableResponseCheck || DEFAULT_CACHABLE_RESPONSE_CHECK` would help with the last point.

Then instead of binding everything just go:

`const cacheable = cacheableResponseCheck.isResponseCacheable({request, response});`

jeffposnick:
I don't like using `DEFAULT_CACHABLE_RESPONSE_CHECK` for the default `RequestWrapper` logic. I'd rather do use a method call that could lazy-instantiate a `CacheableResponse` instance, since it might never be needed. But instead of using the getter syntax, I could use a method named `getDefaultCacheableResponse()`.

Regarding the order of the three scenarios—that's an order that gives precedence to the developer explicitly setting their own criteria (which would be done via the `RequestWrapper` constructor), and when the developer doesn't set anything, allows specific handlers to override the default. Allowing a developer to change behavior by configuring `RequestWrapper` is already what's implicitly being done in `sw-lib`, and it's not a change to the API surface.

Right now, we're [calling `.bind()`](https://github.com/GoogleChrome/sw-helpers/blob/master/packages/sw-runtime-caching/src/lib/request-wrapper.js#L89) on all of the plugin callbacks that are passed into the `RequestWrapper` constructor before we save the callbacks for later use. Doing the same thing with the callback that's exposed by `CacheableResponse` is following that model. (Maybe I snuck that code initially by you, since I'm assuming you object to it on similar grounds?)

What I could do is change the model entirely and refactor the `RequestWrapper` constructor so that it doesn't save bound functions for any callbacks, but instead saves the context that exposes the methods, and then does a method invocation later on when it's time to use the callbacks. That will give you the syntax you want.

Refactoring all of the plugin callbacks is a significant change, so I'd feel better about landing the `sw-runtime-caching` tests first in a separate PR, and then revisiting the changes in this PR once we can use the passing tests as a baseline.

gauntface:
@jeffposnick yeah I never noticed that is what you were doing - apologies for missing it.

Writing tests first sound good.

Would it be possible to force plugins to provide a cacheableResponseCheck method that has a default, meaning we only have "2 options for defining a cachableResponseCheck".

1.) Developer sets the cachaeableResponsePlugin on the request wrapper.
2.) Caller of fetchAndCache must pass in a fallback cacheableResponsePlugin and then plugins can either extend a class that defines the default cacheableResponse behavior and extend / override OR each one just HAS to define a default cacheableResponse.

jeffposnick:
Here's another stab at this 😄 

Things that are different since the last review (some of this changes were included in already merged PRs):
- Tests for everything in `sw-runtime-caching`, including testing the new defaults and overrides.
- Lifecycle callbacks are now invoked directly on plugin references inside of `RequestWrapper`, which eliminates the usage of `.bind()` throughout this PR.
- I've adjusted some of the JSDocs per the feedback.
- The various possible sources of the effective `CacheableResponsePlugin` have all been renamed to hopefully clarify where each might be set. (There's also an extensive comment explaining the logic.)

What hasn't changed is the logic for determining what the effective `CacheableResponsePlugin` will be—it still checks three different sources in a specific order. @gauntface's asked about a different approach in https://github.com/GoogleChrome/sw-helpers/pull/399#issuecomment-291608103, but I don't like the overhead of adding in an additional required parameter to the `RequestWrapper` constructor when we could properly assume a default. That change would shift complexity onto the API surface with the benefit of (slightly) less complex internal logic.

packages/sw-runtime-caching/src/lib/network-first.js
```diff
@@ -21,6 +23,13 @@ import assert from '../../../../lib/assert';
  * An implementation of a [network first](https://developers.google.com/web/fundamentals/instant-and-offline/offline-cookbook/#network-falling-back-to-cache)
  * request strategy.
  *
+ * By default, `NetworkFirst` will cache [opaque responses](http://stackoverflow.com/q/39109789)
```
gauntface: Nit: Could you inverse this description. "Network first will cache 200 responses as well as opaque responses. Override with ..."

packages/sw-runtime-caching/src/lib/request-wrapper.js
```diff
@@ -242,21 +263,39 @@ class RequestWrapper {
...
+    //    constructor, which sets this._userSpecifiedCachableResponsePlugin.
+    // 2. Passing in a parameter to the fetchAndCache() method (done by certain
+    //    runtime handlers, like `StaleWhileRevalidate`), which sets
+    //    handlerDefaultCacheableResponsePlugin.
+    // 3. The default that applies to anything using the `RequestWrapper` class
+    //    that doesn't specify the custom behavior, which is accessed via
+    //    the this.getDefaultCacheableResponsePlugin().
+    const effectiveCacheableResponsePlugin =
+      this._userSpecifiedCachableResponsePlugin ||
+      handlerDefaultCacheableResponsePlugin ||
```
gauntface: Could this just be named cacheResponsePlugin. I can't think of a better way to structure this code, but multiple "defaults" is bugging me.

jeffposnick:
I believe that @addyosmani's earlier review feedback's been addressed, so I'm going to merge.

PR #408, opened 2017-04-03
Build fix

### LINKED ISSUE #403: generateFileManifest() starting from project root instead of app.
**Library Affected**:
sw-build*

My code looks right, according to Matt. It is constructing a manifest.js from my project root instead of from my `app` directory. To run the faulty code, clone [from this branch](https://github.com/jpmedley/redder2/tree/sw-build) and run `gulp build` (which will run the faulty code).

The output from this looks normal:

    [09:09:52] Using gulpfile ~/Documents/projects/redder2/gulpfile.js
    [09:09:52] Starting 'build'...
    [09:09:52] Starting 'clean'...
    Build file has been created.

gauntface:
R: @jeffposnick @addyosmani 

Fixes #403

Original bug reported by @jpmedley was down to current working directory being the actual command CWD rather than the rootdirectory.

This fixes that and neatens up the test which *should* have caught this - but it was poorly written by myself so didn't catch the issue.


jeffposnick:
👍  for the additional changes as well.

PR #409, opened 2017-04-04
webpack plugin for sw-build

prateekbh:

This PR has the webpack plugin for sw-build. I am working to add tests as well.
The index file is just a very minimal wrapper over sw-build, making fallbacks to webpack environment values where ever possible.


packages/sw-build-webpack-plugin/index.js
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Use the instance of this in the plugins array of the webpack config.
+ * @example
+ * const SwBuildWebpackPlugin = require('sw-build-webpack-plugin');
+ * .
+ * .
+ * plugins: [
+ * 	new SwBuildWebpackPlugin({
+ * 		// all options of `swBuild`
```
gauntface: Could you include a minimal example but still leave this comment in.

packages/sw-build-webpack-plugin/index.js
```diff
@@ -0,0 +1,65 @@
+const swBuild = require('sw-build');
+const path = require('path');
+const SERVICE_WORKER_NAME = 'sw.js';
+const GLOB_PATTERNS = '**\/*.{html,js,css}';
```
gauntface: I'm not sure it's wise having a set of defaults in the plugin, either they are required or the defaults should be moved into sw-build.

What do you think?

packages/sw-build-webpack-plugin/index.js
```diff
@@ -0,0 +1,65 @@
...
+			}
+			// If no dest is given, fallback to default service worker name in
+			// output path directory
+			if (!this._config.dest) {
+				this._config.dest =
+					path.join(this._config.rootDirectory, SERVICE_WORKER_NAME);
+			}
+			swBuild.generateSW(this._config)
+				.then(() => {
+					callback();
```
gauntface: This can be shorted to:
```
.then(() => callback())
.catch((err) => callback(err))
```

packages/sw-build-webpack-plugin/index.js
addyosmani: 👍 Let's include the smallest possible functional example in here.

packages/sw-build-webpack-plugin/index.js
```diff
@@ -0,0 +1,65 @@
+const swBuild = require('sw-build');
+const path = require('path');
+const SERVICE_WORKER_NAME = 'sw.js';
```
addyosmani: I agree with Matt that defaults might be better defined in sw-build otherwise we might set folks up for the wrong expectations. I'd imagine most folks are going to copy/paste whatever code snippets for using the Webpack plugin we're going to include in our docs and that might be a better place to include both of these bits (I think that would be useful).

packages/sw-build-webpack-plugin/index.js
prateekbh: Yes!
It'd be great if `sw-build` can have a set of defaults, that way every plugin does not have to define them

gauntface: Cool, could you open an issue with what you think is the best set of defaults - just so everyone can weigh in, then remove this and handle errors in the plugin if the developer doesn't define them?

It will suck for a bit, but should get fixed outside of the plugin soon :)

prateekbh: Why errors? If we are going to handle the defaults in `sw-build` I am not sure why is it a necessity for the dev to define these.

packages/sw-build-webpack-plugin/index.js
prateekbh: done, removed all the "defaults" login from webpack plugin

packages/sw-build-webpack-plugin/index.js
prateekbh: added, PTAL

packages/sw-build-webpack-plugin/index.js
gauntface: @addyosmani @jeffposnick should we default the dest to `<rootDirectory>/sw.js` in sw-build as well?

jeffposnick: It seems like that would be in keeping with the other changes to defer to `sw-build`'s defaults, so 👍

addyosmani:
Config changes LGTM. Are we going to add a test for the Webpack plugin so we know the interface between it and sw-build doesn't break over time? Fwiw, did this with the [preload](https://github.com/GoogleChrome/preload-webpack-plugin/blob/master/test/spec.js) plugin and didn't find it too bad to orchestrate Webpack's compiler for tests.

packages/sw-build-webpack-plugin/index.js
addyosmani: > should we default the dest to <rootDirectory>/sw.js in sw-build as well?

👍

gauntface:
Ignore AppVeyor errors for now - it's a total mess :(

prateekbh:
@gauntface @addyosmani this PR now contains tests and changes for injectManifestFile

packages/sw-build-webpack-plugin/index.js
```diff
@@ -0,0 +1,70 @@
+const swBuild = require('../sw-build/src/');
+const path = require('path');
+const SERVICE_WORKER_NAME = 'sw.js';
+
+/**
+ * Use the instance of this in the plugins array of the webpack config.
+ * @example
```
gauntface: Is it worth including an example and explanation of how to use the inject file or should we leave that up to some other docs?

cc @jpmedley

packages/sw-build-webpack-plugin/index.js
```diff
@@ -0,0 +1,70 @@
+const swBuild = require('../sw-build/src/');
+const path = require('path');
+const SERVICE_WORKER_NAME = 'sw.js';
```
gauntface: I don't think this is used anywhere

packages/sw-build-webpack-plugin/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "build"
+  ],
+  "engines": {
+    "node": ">=4.0.0"
+  },
+  "author": "Google's Web DevRel Team",
+  "license": "Apache-2.0",
+  "repository": "googlechrome/sw-helpers",
+  "bugs": "https://github.com/googlechrome/sw-helpers/issues",
+  "homepage": "https://github.com/GoogleChrome/sw-helpers/tree/master/packages/sw-build-webpack-plugin",
```
gauntface: I think comma on the end of this line is breaking travis.

packages/sw-build-webpack-plugin/index.js
jpmedley: How big is the example?

packages/sw-build-webpack-plugin/package.json
prateekbh: removed

packages/sw-build-webpack-plugin/index.js
prateekbh: removed

prateekbh:
@gauntface I see a test of sw-cli is broken

addyosmani:
cc @goldhand @thelarkinn in case there are any Webpack plugin best practices we're missing out on here.

packages/sw-build-webpack-plugin/index.js
prateekbh: @jpmedley its same size as the one there right now

addyosmani:
@gauntface What's the process for getting an npm package of this published using our current setup? Are we manually publishing each one?

addyosmani:
@prateekbh Nice work on getting this done. Can you update the thread on build system integration to share this is now available and seek any additional feedback that might be useful?

@jpmedley fyi on the Webpack plugin now landing.

PR #411, opened 2017-04-04
Manifest injection

### LINKED ISSUE #405: Formalize the manifest injection across sw-cli, Webpack, etc.
**Library Affected**:
*sw-cli, sw-build, Webpack plugin?*

Some developers will use build tools that work best when the revisioned manifest is directly injected into a source `service-worker.js` file. (Others might want to use build tools to generate a standalone manifest that gets pulled in via `importScripts()`, and that's outside the scope of this issue.)

Injecting into a source file presupposes that our tools know *where* to inject, and after some discussion, we seem to have settled on checking for the presence of `cacheRevisionedAssets([])`, and then handle injection by replacing the `[]` with the populated manifest. If `cacheRevisionedAssets([])` can't be found in the source, it would be a fatal error.

While we were previously thinking about baking this logic into the forthcoming Webpack plugin, this behavior will be useful for users of the other build tools, like `sw-cli`, so it might make sense to implement it inside of `sw-build`.

CC: @prateekbh @gauntface

gauntface:
R: @jeffposnick @addyosmani 

Fixes #405

An early implementation of the manifest injection. This currently expects a file to read in and outputs to dest.


packages/sw-build/src/lib/inject-manifest.js
```diff
@@ -0,0 +1,87 @@
...
+ *   rootDirectory: './build/',
+ *   staticFileGlobs: ['**\/*.{html,js,css}'],
+ *   globIgnores: ['admin.html'],
+ *   swFile: 'sw.js'
+ * })
+ * .then(() => {
+ *   console.log('Build Manifest generated.');
+ * });
+ *
+ * This method will generate a file manifest that can be used in a service
```
jeffposnick: "This method will modify an existing service worker file by finding `cacheRevisionedAssets([])` and replacing the empty `[]` array with an up to date precache manifest. This will allow the service worker to maintain a cache of local assets that will be available offline."

packages/sw-build/src/lib/inject-manifest.js
```diff
@@ -0,0 +1,87 @@
...
+      new Error(errors['invalid-inject-manifest-arg']));
+  }
+
+  const injectionPointRegex = /(\.cacheRevisionedAssets\()\s*\[\s*\]\s*(\))/g;
+
+  return getFileManifestEntries(input)
+  .then((manifestEntries) => {
+    let swFileContents = fs.readFileSync(
+      path.join(input.rootDirectory, input.swFile), 'utf8');
+    const injectionResult = injectionPointRegex.exec(swFileContents);
```
jeffposnick: This won't let you know whether there are multiple matches for `injectionPointRegex` within the string. If there are, I think that should lead to a failure with a specific error message.

You can switch to

```js
const matches = swFileContents.match(injectionPointRegex);
if (matches) {
  if (matches.length !== 1) {
    // throw error about multiple matches
  }
} else {
  // throw error about no matches
}
```

gauntface:
cc @jeffposnick added in the feature you requested and updated the comment.

packages/sw-build/src/lib/errors.js
```diff
@@ -62,4 +62,7 @@ module.exports = {
   'injection-point-not-found': `Unable to find a place to inject the ` +
     `manifest. Please ensure that you have 'swlib.cacheRevisionedAssets([])' ` +
     `somewhere in your service worker file.`,
+  'multiple-injection-points-found': `There can only be one injection point ` +
+    `in your service worker file. Please snure you only have one instance of ` +
```
jeffposnick: snure => ensure (I assume...)

packages/sw-build/src/lib/errors.js
```diff
@@ -62,4 +62,7 @@ module.exports = {
...
+    `'swlib.cacheRevisionedAssets([])' if your service worker file.`,
```
jeffposnick: if => in

packages/sw-build/src/lib/inject-manifest.js
```diff
@@ -56,10 +58,14 @@ const injectManifest = (input) => {
   .then((manifestEntries) => {
     let swFileContents = fs.readFileSync(
       path.join(input.rootDirectory, input.swFile), 'utf8');
-    const injectionResult = injectionPointRegex.exec(swFileContents);
-    if (!injectionResult) {
+    const injectionResults = swFileContents.match(injectionPointRegex);
+    if (!injectionResults) {
       throw new Error(errors['injection-point-not-found']);
     }
+    /* eslint-disable no-console */
```
jeffposnick: Remove this comment.

packages/sw-build/src/lib/inject-manifest.js
```diff
@@ -0,0 +1,85 @@
...
+ * .then(() => {
+ *   console.log('Build Manifest generated.');
+ * });
+ *
+ * This method will generate a file manifest that can be used in a service
+ * worker for caching assets offline.
+ * @param {Object} input
+ * @param {String} input.dest The name and path you wish to write your
+ * manifest file to.
+ * @param {String} input.rootDirectory The root of the files you wish to
```
addyosmani: The root -> The root directory

PR #413, opened 2017-04-04
Cli install doc fix

### LINKED ISSUE #344: fix sw-cli install / command instructions
**Library Affected**:
sw-cli

**Browser & Platform**:
all

[readme](https://github.com/GoogleChrome/sw-helpers#sw-cli) says to install sw-cli as a dev dep:

    npm install --save-dev sw-cli

and then the [sw-cli doc](https://googlechrome.github.io/sw-helpers/reference-docs/stable/v0.0.8/module-sw-cli.html) assumes that you can run it as a global command

    sw-cli generate-sw

Either switch the readme to instruct to install it globally or change the command to use the local dev dep (e.g. `node_modules/.bin/sw-cli generate-sw`) 

I'm also getting an error that I need to install `glob` globally. Add it explicitly as a dependency of sw-cli?

gauntface:
R: @jeffposnick @addyosmani 

Fixes #344 

Makes docs print out to install sw-cli with -g flag.


gulp-tasks/documentation.js
```diff
@@ -54,6 +54,11 @@ const documentPackage = (projectPath) => {
 
     handleBarsData.hasRefDocs = (projectMetadata.name !== 'sw-cli');
 
+    // sw-cli needs a custom install command
+    if (projectMetadata.name === 'sw-cli') {
+      handleBarsData.installCmd = `npm install -g sw-cli`;
+    }
+
```
addyosmani: Had to do a double take on this, but totally makes sense given this is a global CLI you're going to use differently to more local modules.

addyosmani: One ask: can we change this to `npm install --global sw-cli`? Experienced Node folks totally get the `-g` shorthand but those more new to it could benefit from knowing what we're doing here.

jpmedley: +1

gauntface:
@addyosmani @jeffposnick would appreciate a second pass on this - the documentation task now shares logic for documentation:repo and documentation:projects (i.e. no code duplication).

gulp-tasks/documentation.js
```diff
@@ -31,33 +31,40 @@ navigation_weight: 0
...
+    name: projectMetadata.name,
+    description: projectMetadata.description,
+    background: projectMetadata.background,
+  };
+
+  try {
+    const demoStats = fs.statSync(path.join(projectPath, 'demo'));
+    handleBarsData.hasDemo = demoStats.isDirectory();
+  } catch(err) {
+    handleBarsData.hasDemo = false;
```
addyosmani: Is there a reason this is in a try/catch block vs `.hasDemo = !!demoStats.isDirectory()`?

gauntface: If the file doesn't exist statSync will through.

addyosmani: Gotcha. Alright then.

PR #414, opened 2017-04-04
Modify url prefix

### LINKED ISSUE #358: Need way to change something in the root path
**Library Affected**:
sw-build

**Browser & Platform**:
all browsers

My server maps /assets to /static and so I need a way to replace assets with static. `rootDirectory` doesn't work as it only accepts actual paths that exist.

gauntface:
R: @jeffposnick @addyosmani

Fixes #358

Adds support for the modifyUrlPrefix optional parameter..


packages/sw-build/test/utils-modify-url-prefix.js
```diff
@@ -0,0 +1,94 @@
...
+  });
+
+  it('should only match at the start of a url', function() {
+    const modifiers = {
+      '/first-match': '/altered',
+    };
+    const newUrl = modifyUrlPrefix('/example/first-match/12345/hello', modifiers);
+    newUrl.should.equal('/example/first-match/12345/hello');
+  });
+});
```
addyosmani: Your test coverage demos that this captures all of the use-cases I had in mind for this feature. Nice work.

packages/sw-build/src/lib/utils/modify-url-prefix.js
```diff
@@ -0,0 +1,30 @@
...
+  });
+
+  const escapeRegExp = (string) => {
+    return string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+  };
+
+  const modifyRegex = new RegExp('^(' +
+    Object.keys(modifyUrlPrefix).map(escapeRegExp).join('|') +
+    ')');
+
```
addyosmani: It's a wee bit unclear how the modifyRegex and escapeRegExp pieces are used here but that's the only comment I had for this bit.

packages/sw-build/src/lib/utils/modify-url-prefix.js
```diff
@@ -0,0 +1,30 @@
...
+    throw new Error(errors['modify-url-prefix-bad-prefixes']);
+  }
+
+  Object.keys(modifyUrlPrefix).forEach((key) => {
+    if (typeof modifyUrlPrefix[key] !== 'string') {
+      throw new Error(errors['modify-url-prefix-bad-prefixes']);
+    }
+  });
+
+  const escapeRegExp = (string) => {
```
gauntface: Totally copied this from sw-precache ;)

PR #416, opened 2017-04-05
Tests for the sw-runtime-caching module

### LINKED ISSUE #246: Add networkTimeoutSeconds

### LINKED ISSUE #301: Tests for sw-runtime-caching
**Library Affected**:
*sw-runtime-caching*

The module is currently untested 😦

jeffposnick:
R: @addyosmani @gauntface

Fixes #301, fixes #246

This adds unit tests for all the individual classes and methods within the `sw-runtime-caching` module. I think we'll be in better shape to land #399 once this is in place.


jeffposnick:
(I'm cleaning up the test failures in the Travis CI environment.)

jeffposnick:
Yay! Green tests across all the platforms. Should be ready for review.

packages/sw-runtime-caching/src/lib/network-first.js
```diff
@@ -84,18 +84,20 @@ class NetworkFirst extends Handler {
...
 
-        return response ?
-          response :
-          Promise.reject('No response received; falling back to cache.');
-      })
-      .catch(() => this.requestWrapper.match({request: event.request}))
-    );
+      return response ?
+        response :
+        Promise.reject('No response received; falling back to cache.');
```
gauntface: Nit: Could you move this to an errors file.

packages/sw-runtime-caching/test/sw/cache-first.js
```diff
@@ -0,0 +1,33 @@
...
+  const REQUEST_WRAPPER = new goog.runtimeCaching.RequestWrapper({cacheName: CACHE_NAME});
+  const CACHE_FIRST = new goog.runtimeCaching.CacheFirst({requestWrapper: REQUEST_WRAPPER, waitOnCache: true});
+  const COUNTER_URL = new URL('/__echo/counter', location).href;
+
+  let cache;
+  let initialCachedResponse;
+
+  before(() => {
+    const event = new FetchEvent('fetch', {request: new Request(COUNTER_URL)});
+    return caches.delete(CACHE_NAME)
```
gauntface: This chain looks a little funky / difficult to reason with in the inidividual tests.

This seems dangerous to do just once before all tests given that you are setting a cache entry of some kind - are you expecting that value to exist  / be changed between tests? If so, I can see what breaking pretty quickly.

Could you refactor it to reset before every test and add a comment on what this chain is doing.

packages/sw-runtime-caching/test/sw/cache-first.js
```diff
@@ -0,0 +1,33 @@
...
+      .then(() => caches.open(CACHE_NAME))
+      .then((openedCache) => cache = openedCache)
+      .then(() => CACHE_FIRST.handle({event}))
+      .then(() => cache.match(COUNTER_URL))
+      .then((cachedResponse) => initialCachedResponse = cachedResponse);
+  });
+
+  it(`should add the initial response to the cache`, function() {
+    expect(initialCachedResponse).to.exist;
```
gauntface: This feels off - it's relying on something from the before() function to test in the actual unit test.

packages/sw-runtime-caching/test/sw/cache-first.js
gauntface: Looking through this, I'd just consolidate everything into a single unit test and use before() to delete any existing caches.

packages/sw-runtime-caching/test/sw/network-first.js
```diff
@@ -0,0 +1,80 @@
...
+      .then((cachedResponse) => initialCachedResponse = cachedResponse);
+  });
+
+  afterEach(function() {
+    globalStubs.forEach((stub) => stub.restore());
+    globalStubs = [];
+  });
+
+  it(`should add the initial response to the cache`, function() {
+    expect(initialCachedResponse).to.exist;
```
gauntface: Same as above - this feels off.

packages/sw-runtime-caching/test/sw/network-first.js
```diff
@@ -0,0 +1,80 @@
...
+    const event = new FetchEvent('fetch', {request: new Request(COUNTER_URL)});
+    let handleResponse;
+    return NETWORK_FIRST.handle({event})
+      .then((response) => handleResponse = response)
+      .then(() => expectSameResponseBodies(initialCachedResponse, handleResponse))
+      .then(() => cache.match(COUNTER_URL))
+      .then((currentCachedResponse) => expectSameResponseBodies(currentCachedResponse, initialCachedResponse));
+  });
+
+  it(`should return the cached response if the network request times out`, function() {
```
gauntface: This might be a good use for sinon fakeClocks.

All setTimeouts and date will use sinon and you can programmatically fast forward time.

Initialise:
https://github.com/gauntface/route/blob/master/packages/device-clock/test/node/device-clock.js#L17

Progress clock forward:
https://github.com/gauntface/route/blob/master/packages/device-clock/test/node/device-clock.js#L46

packages/sw-runtime-caching/test/sw/request-wrapper.js
```diff
@@ -0,0 +1,186 @@
...
+  const CACHE_WILL_UPDATE_PLUGIN = {cacheWillUpdate: () => {}};
+  const CACHE_WILL_MATCH_PLUGIN = {cacheWillMatch: () => {}};
+  const CACHED_URL = '/cached';
+
+  let globalStubs = [];
+
+  before(() => {
+    return caches.delete(CACHE_NAME)
+      .then(() => caches.open(CACHE_NAME))
+      .then((cache) => cache.put(CACHED_URL, new Response('')));
```
gauntface: Again this feels odd doing only at the beginning of the test suite. would make sense to move to beforeEach (Just to make sure one test doesn't mess up the rest).

packages/sw-runtime-caching/test/sw/stale-while-revalidate.js
```diff
@@ -0,0 +1,63 @@
...
+  let globalStubs = [];
+  let cache;
+  let initialCachedResponse;
+
+  before(() => {
+    const event = new FetchEvent('fetch', {request: new Request(COUNTER_URL)});
+    return caches.delete(CACHE_NAME)
+      .then(() => caches.open(CACHE_NAME))
+      .then((openedCache) => cache = openedCache)
+      .then(() => STALE_WHILE_REVALIDATE.handle({event}))
```
gauntface: Same as above - funky one time setup.

packages/sw-runtime-caching/test/sw/stale-while-revalidate.js
```diff
@@ -0,0 +1,63 @@
...
+      .then((cachedResponse) => initialCachedResponse = cachedResponse);
+  });
+
+  afterEach(function() {
+    globalStubs.forEach((stub) => stub.restore());
+    globalStubs = [];
+  });
+
+  it(`should add the initial response to the cache`, function() {
+    expect(initialCachedResponse).to.exist;
```
gauntface: Same as above.

utils/server-instance.js
```diff
@@ -31,6 +31,8 @@ class ServerInstance {
     this._app = express();
     this._app.use(cookieParser());
 
+    this._counter = 1;
```
gauntface: Could we just have this return Date.now() rather than have a counter and then have '/__echo/random/nocache/' ?

I'm just thinking it's less orchestrating, states we have no-cache header and "random" seems like it could be useful else where.

jeffposnick: I'm not clear on the benefit of making that change.

With a counter, you're guaranteed that each response will be different, eliminating the infinitesimal chance that two random numbers will be the same, and the slightly more realistic issue of two requests in quick succession returning the same `Date.now()`.

Re: putting in `/nocache/` in the URL, there are some endpoints that respond with a `max-age` of 1 day, and those don't have `/1day/` in the URL.

If you want to change the URL to reflect the idea that its going to return a different value each time a page makes a request, what if I renamed it to `/__echo/unique`?

packages/sw-runtime-caching/src/lib/network-first.js
jeffposnick: Done.

jeffposnick:
Thanks for the feedback, @gauntface!

High-level response to some of your inline comments: I didn't realize it was bad form to initialize test state inside a `before()` and then rely on that state inside each of the individual test cases. Ditto for relying on state from the previous test case inside the next one.

I'll collapse some of the smaller test cases that made assumptions about state into larger test cases that are self-contained.

addyosmani:
> High-level response to some of your inline comments: I didn't realize it was bad form to initialize test state inside a before() and then rely on that state inside each of the individual test cases. Ditto for relying on state from the previous test case inside the next one.

SG. I'll wait until these changes come through before reviewing further.

utils/server-instance.js
gauntface: Just leave it as is.

packages/sw-runtime-caching/test/sw/network-first.js
jeffposnick: I haven't been able to successfully get this working with the asynchronous tests 😦 

I've lowered the amount of time that the code is sleeping for and it finishes quickly and doesn't appear to be flaky using the actual system clock, so I'm just going to leave as-is.

jeffposnick:
PTAL. The test cases should now all rely on independent state, with the cache cleared before each is invoked. Most of them are also now making liberal use of `async`/`await` instead of Promises, which made the refactoring much more pleasant, and hopefully makes the test logic more readable.

jeffposnick:
Well, this looks fine on the Travis CI side of things after merging in the changes from #420, but the AppVeyor Windows test of `sw-cli` is failing.

This PR doesn't modify anything in `sw-cli`, and I don't remember AppVeyor even running for previous commits to this PR.

@gauntface, should I be concerned about AppVeyor? If there's some test flakiness there, can that be addressed independently of this PR?

gauntface:
@jeffposnick yeah ignore AppVeyor for now, it's up and down like a yo-yo for some reason.

addyosmani:
AppVeyor weirdness aside, LGTM.

PR #420, opened 2017-04-05
Test sweep

gauntface:
R: @jeffposnick @addyosmani

Not ready to be merged, but at least highlights how we can move all tests only a single path - removes a lot of duplicate code but removes some flexibility.


gulp-tasks/test.js
```diff
@@ -38,25 +37,25 @@ gulp.task('download-browsers', function() {
...
   }
-  const testGlob = `packages/${global.projectOrStar}/test/*.js`;
-  const testFiles = glob.sync(testGlob);
-  if (testFiles.length === 0) {
-    // Mocha fails when no tests are found so return early.
-    return;
-  }
-  return gulp.src(testGlob, {read: false})
+
+  return gulp.src('test/bootstrap.js', {read: false})
```
jeffposnick: I'm seeing instances where there are individual errors in tests, but the `gulp test` process exits with a 0 status:

E.g.
https://travis-ci.org/GoogleChrome/sw-helpers/jobs/219368904#L1829 is an error, but the Travis CI build is green because the test exited with a 0: https://travis-ci.org/GoogleChrome/sw-helpers/jobs/219368904#L2466

Can you double-check this modified code and see what might be causing the errors not to bubble up and trigger the `.once('error', ...)` listener?

gauntface: Eek - thanks for catching that!!!

addyosmani: Well spotted, Jeff!

gauntface: Just checked this out - it's not actually failing the test. It's actually retrying the tests and successfully getting a pass on one of the tries.

gauntface: Ill remove the retries from the test runner - forcing them to be in the browser tests - meaning that these kinds of logs will be cleaner / clearer.

jeffposnick: Is that to be expected? The fact that there are test that don't pass on the first attempt doesn't sound good either, albeit a separate issue.

gauntface: As an FYI - I injected an error locally and it is actually failing - more than happy to inject a failing test and ensure Travis fails.

gauntface: @jeffposnick Before this PR - yes it would have been expected because it's normally an issue with selenium web driver. With these tests - yes and no. Its going to put the tests under a different process - so may reveal some lucky situations that are no longer true - like the bg-sync-queue tests

utils/test-runner.js
```diff
@@ -211,11 +209,6 @@ class TestRunner {
   }
 
   _runIntegrationTests(webdriverCb, packagePath, getBaseTestUrl) {
-    /** it(`should pass '${path.basename(packagePath)}' integration tests`, function() {
-      this.timeout(60 * 1000);
-      this.retries(2);
-    });**/
-
     const integrationTests = glob.sync(`${packagePath}/test/integration/*.js`);
```
addyosmani: Moving these out to individual tests vs the runner totally makes sense.

addyosmani:
After the latest additions (post your move around) the tests are still all green. Was there further work you wanted addressed here, @jeffposnick?

gulp-tasks/test.js
jeffposnick: Thanks for verifying!

PR #424, opened 2017-04-07
Rethinking precaching, incorporating hackathon feedback

### LINKED ISSUE #347: warmRuntimecache + CDN Assets
@devnook hit this.

### LINKED ISSUE #353: Difference unclear between warmRuntimeCache/cacheRevisionedAssets
Couldn't figure out what the specific differences were from examples without looking at the source. I think it would be a lot more obvious if it were cacheUnrevisionedAssets/cacheRevisionedAssets or something. Also might want to showcase warmRuntimeCache more for those who only have revision information for secondary assets and not index html.

### LINKED ISSUE #365: What caching strategy does cacheRevisionedAssets use?
**Library Affected**:
sw-lib

**Browser & Platform**:
all

The docs for [cacheOnly()](https://googlechrome.github.io/sw-helpers/reference-docs/stable/v0.0.13/module-sw-lib.SWLib.html#cacheOnly), `cacheFirst()`, etc. are straightforward in terms of which caching strategy they use. Those map to Jake's offline cookbook.

But the doc for [cacheRevisionedAssets()](https://googlechrome.github.io/sw-helpers/reference-docs/stable/v0.0.13/module-sw-lib.SWLib.html#cacheRevisionedAssets) is unclear.

From what I understand, when I use `sw-build.generateSW()`, `sw-build` detects when any of the files specified in `globPatterns` change. It then automatically increments the `revision` number for those files in the generated service worker. Something Jeff said led me to believe that the generated SW will use `cacheRevisionedAssets()` in this scenario

But what caching strategy does the SW use to download these updated files?

### LINKED ISSUE #367: Add a public method to get the name of the cache(s)
**Library Affected**:
*sw-lib*

*Add a public method to get the name of the cache(s)*

### LINKED ISSUE #370: Remove warmRuntimeCache, rename cacheRevisionedAssets?
**Library Affected**:
*sw-lib*

Potentially controversial, but after observing developer confusion at the hackathon—https://github.com/GoogleChrome/sw-helpers/issues/353, but also confusion as to whether `warmRuntimeCache()` sets up routes (it doesn't) and whether it also deletes entries in the `activate` handler (it does)—I'm of the opinion that we should remove `warmRuntimeCache()` in favor of exposing the name of the default runtime cache (https://github.com/GoogleChrome/sw-helpers/issues/367) and encouraging developers who need that functionality to explicitly write:

```js
  const urls = [/* ... */];
  const cacheName = swLib.runtimeCacheName; // Or use your own name.
  event.waitUntil(caches.open(cacheName).then((cache) => cache.addAll(urls)));
});

// Setup runtime routes as needed.
```

At the same time, I'd like to see us rename `cacheRevisionedAssets()` to `precache()` or maybe `precacheAndRoute()` (since it also sets up routes), to steer developers down the path that adheres to best practices.

jeffposnick:
R: @addyosmani @gauntface

Apologies for the large PR, but it touches a number of related topics that should probably be addressed all-or-nothing.

- It removes the `warmRuntimeCache()` method from `sw-lib`. The underlying `UnrevisionedCacheManager` class is left as-is in `sw-precaching`, with the understanding that if a developer explicitly wants to use it as part of their own piecemeal service worker without using `sw-lib`, they're free to do so. (Fixes #370)

- It renames `cacheRevisionedAssets()` to `precache()`, which hopefully emphasizes that it's the "correct" method to use for most cases. (Fixes #347, fixes #353)

- It adds a JSDoc comment and logs a message letting developers know about the automatic cache-first route for URLs that are in the precache list (Fixes #365)

- It exposes a `defaultRuntimeCacheName` getter in`sw-lib`, and includes an example of how that could be used in conjunction with the basic Cache Storage API to add/query the default cache. Developers who really need to manage this default cache can still do so via this approach, without `warmRuntimeCache()`. (Fixes #367)

It's possible that this PR closes some other closely related open issues, but these are the ones I could track down.

There's also an extra log message when there are URLs passed in to `precache()` that don't explicitly include revisioning information, letting developers know that they need to make sure the URL itself contains versioning info, or else they'll have a Bad Time.


packages/sw-precaching/src/lib/controllers/revisioned-cache-manager.js
```diff
@@ -53,6 +54,21 @@ class RevisionedCacheManager extends BaseCacheManager {
   addToCacheList({revisionedFiles} = {}) {
     assert.isInstance({revisionedFiles}, Array);
     super._addEntries(revisionedFiles);
+
+    const urlsWithoutRevisionFields = revisionedFiles
+      .filter((entry) => typeof entry === 'string');
+    if (urlsWithoutRevisionFields.length > 0) {
+      logHelper.debug({
+        that: this,
+        message: `Some precache entries don't include a separate revision
```
gauntface: This message is awfully verbose for the point it's trying to make.

"Please ensure the following cache entries are revisioned correctly

  /index.1234.css
  /example/unsafe.html

  If these aren't revisioned (a.k.a. cache busted, your users won't receive updates."

jeffposnick: I'll cut down on the word count.

But there's nuance here, as this usage might either be legitimate or an error, and I want to provide enough context to help the developer understand what's going on and whether they need to take action.

If I read your proposed message, I'd probably just go to the list of URLs passed to `precache()` (which, maybe I generated by hand, despite the guidance to use `sw-build`) and append `?cacheBust=${Date.now()}` to the end of them.

jeffposnick: I've cut down on the "above the fold" message and tried to include as much helpful context as I could in the `data` fields. PTAL.

addyosmani: The revised wording lgtm

addyosmani:
cc @philipwalton on that last one as an fyi

Having looked over this twice, the changes sit relatively well with me so far. Going to give the comments another sweep through.

philipwalton:
Works for me (on the last one).

PR #428, opened 2017-04-11
Cacheid param

### LINKED ISSUE #417: sw-precaching is throwing incorrect errors
Looking through the tests - sw-precaching is throwing on older implementations of `PrecacheManager` but tests are passing because it was expected that the sw would fail to register.

gauntface:
R: @jeffposnick @addyosmani

Fixes #417

Adds cacheId to sw-build, sw-lib, sw-precaching, sw-runtime-caching. Also fixes some pre-caching bugs.


packages/sw-lib/src/lib/sw-lib.js
```diff
@@ -32,13 +32,22 @@ import logHelper from '../../../../lib/log-helper';
...
 
   /**
-   * You should never need to instantiate this class. The library instantiates
-   * an instance which can be accessed by `self.goog.swlib`.
+   * You should instantiate this class with `new self.goog.SWLib()`.
    */
-  constructor({cacheId}) {
-    this._revisionedCacheManager = new RevisionedCacheManager();
+  constructor({cacheId} = {}) {
+    /* eslint-disable no-console */
```
addyosmani: Let's remove this one

packages/sw-runtime-caching/src/lib/constants.js
```diff
@@ -25,11 +25,18 @@
  * @type {string}
  * @memberof module:sw-runtime-caching
  */
-let cacheName = `sw-runtime-caching`;
- if (self && self.registration) {
+export const getDefaultCacheName = ({cacheId} = {}) => {
```
addyosmani: Let's include some examples in the above comment of what the output from `getDefaultCacheName()` might look like. I got here from the constructor and still found myself wondering if a clear example was available.

From reading below it sounds like...

- If `cacheId` is specified (e.g `foo`, it's `foo-sw-runtime-caching`)
- If self/self.registration with scope `bar`, cacheName is `sw-runtime-caching-bar`. Does any naming related sanitization need to be done on scope here or do you think we should be fine as is?

Side: you explained why we're appending the scope in person and although this isn't the prettiest thing I understand why we're doing it the way we are.

packages/sw-lib/src/lib/sw-lib.js
```diff
@@ -32,13 +32,22 @@ import logHelper from '../../../../lib/log-helper';
...
 
   /**
-   * You should never need to instantiate this class. The library instantiates
-   * an instance which can be accessed by `self.goog.swlib`.
+   * You should instantiate this class with `new self.goog.SWLib()`.
    */
-  constructor() {
-    this._revisionedCacheManager = new RevisionedCacheManager();
+  constructor({cacheId} = {}) {
+    /* eslint-disable no-console */
```
jeffposnick: Is this leftover from debugging?

addyosmani:
Talked about this offline and we're going to merge ahead so Matt isn't blocked on further work that needs to be done.

PR #430, opened 2017-04-11
Small change to the redirected browser test URL

jeffposnick:
R: @addyosmani @gauntface

Small change to account for the current URL pattern used for kicking off browser tests. This redirect is useful when manually browsing the directory listing page for a package and starting tests manually.


addyosmani:
LGTM

PR #431, opened 2017-04-11
Save plugin references and call methods on them

jeffposnick:
R: @addyosmani @gauntface

I'm addressing some of the feedback from https://github.com/GoogleChrome/sw-helpers/pull/399 out of band, as it extends past the boundaries of that PR.

This moves away from `RequestWrapper` saving and then invoking references to the bound functions exposed by plugins, and instead saves references to the plugins themselves, and calls the methods on the plugin instances directly.

The tests confirm that this shouldn't change any behavior.

### DIFF
--- a/packages/sw-runtime-caching/src/lib/request-wrapper.js
+++ b/packages/sw-runtime-caching/src/lib/request-wrapper.js
@@ -74,7 +74,7 @@ class RequestWrapper {
       this.matchOptions = matchOptions;
     }
 
-    this.pluginCallbacks = {};
+    this.plugins = new Map();
 
     if (plugins) {
       assert.isArrayOfType({plugins}, 'object');
@@ -82,26 +82,23 @@ class RequestWrapper {
       plugins.forEach((plugin) => {
         for (let callbackName of pluginCallbacks) {
           if (typeof plugin[callbackName] === 'function') {
-            if (!this.pluginCallbacks[callbackName]) {
-              this.pluginCallbacks[callbackName] = [];
+            if (!this.plugins.has(callbackName)) {
+              this.plugins.set(callbackName, []);
             }
-            this.pluginCallbacks[callbackName].push(
-              plugin[callbackName].bind(plugin));
+            this.plugins.get(callbackName).push(plugin);
           }
         }
       });
     }
 
-    if (this.pluginCallbacks.cacheWillUpdate) {
-      if (this.pluginCallbacks.cacheWillUpdate.length !== 1) {
-        throw ErrorFactory.createError('multiple-cache-will-update-plugins');
-      }
+    if (this.plugins.has('cacheWillUpdate') &&
+        this.plugins.get('cacheWillUpdate').length !== 1) {
+      throw ErrorFactory.createError('multiple-cache-will-update-plugins');
     }
 
-    if (this.pluginCallbacks.cacheWillMatch) {
-      if (this.pluginCallbacks.cacheWillMatch.length !== 1) {
-        throw ErrorFactory.createError('multiple-cache-will-match-plugins');
-      }
+    if (this.plugins.has('cacheWillMatch') &&
+      this.plugins.get('cacheWillMatch').length !== 1) {
+      throw ErrorFactory.createError('multiple-cache-will-match-plugins');
     }
   }
 
@@ -149,9 +146,9 @@ class RequestWrapper {
     const cache = await this.getCache();
     let cachedResponse = await cache.match(request, this.matchOptions);
 
-    if (this.pluginCallbacks.cacheWillMatch) {
-      cachedResponse = this.pluginCallbacks.cacheWillMatch[0](
-        {cachedResponse});
+    if (this.plugins.has('cacheWillMatch')) {
+      const plugin = this.plugins.get('cacheWillMatch')[0];
+      cachedResponse = plugin.cacheWillMatch({cachedResponse});
     }
 
     return cachedResponse;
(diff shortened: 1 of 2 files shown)


jeffposnick:
(Stay tuned for a follow-up commit to this PR addressing the `sw-lib` tests.)

jeffposnick:
The `sw-lib` tests are sorted. Please take a look.

packages/sw-runtime-caching/src/lib/request-wrapper.js
```diff
@@ -74,34 +74,31 @@ class RequestWrapper {
...
         }
       });
     }
 
-    if (this.pluginCallbacks.cacheWillUpdate) {
-      if (this.pluginCallbacks.cacheWillUpdate.length !== 1) {
-        throw ErrorFactory.createError('multiple-cache-will-update-plugins');
-      }
+    if (this.plugins.has('cacheWillUpdate') &&
+        this.plugins.get('cacheWillUpdate').length !== 1) {
```
gauntface: Should we block on being able to add multiple cacheWillUpdate plugins rather than add them and then throw afterwards?

packages/sw-runtime-caching/src/lib/request-wrapper.js
```diff
@@ -74,34 +74,31 @@ class RequestWrapper {
...
+      throw ErrorFactory.createError('multiple-cache-will-update-plugins');
     }
 
-    if (this.pluginCallbacks.cacheWillMatch) {
-      if (this.pluginCallbacks.cacheWillMatch.length !== 1) {
-        throw ErrorFactory.createError('multiple-cache-will-match-plugins');
-      }
+    if (this.plugins.has('cacheWillMatch') &&
+      this.plugins.get('cacheWillMatch').length !== 1) {
```
gauntface: Same as above.

packages/sw-runtime-caching/src/lib/request-wrapper.js
```diff
@@ -184,11 +181,11 @@ class RequestWrapper {
...
     }
 
-    const clonedRequest = this.pluginCallbacks.fetchDidFail ?
+    const clonedRequest = this.plugins.has('fetchDidFail') ?
       request.clone() : null;
-    if (this.pluginCallbacks.requestWillFetch) {
-      for (let callback of this.pluginCallbacks.requestWillFetch) {
-        const returnedPromise = callback({request});
+    if (this.plugins.has('requestWillFetch')) {
+      for (let plugin of this.plugins.get('requestWillFetch')) {
```
gauntface: Nit: I have a feeling it would be more memory efficient to get the array once and then iterate over the local instance of the array rather than use this short hand.

packages/sw-runtime-caching/src/lib/request-wrapper.js
```diff
@@ -184,11 +181,11 @@ class RequestWrapper {
       assert.isInstance({request}, Request);
     }
 
-    const clonedRequest = this.pluginCallbacks.fetchDidFail ?
+    const clonedRequest = this.plugins.has('fetchDidFail') ?
```
gauntface: Nit: Can you move this to where it's used.

Reading this, you check for fetchDidFail, clone the request, then iterate over requestWillFetch (which threw me because I assumed it had failed), hence the cloning.

packages/sw-runtime-caching/src/lib/request-wrapper.js
```diff
@@ -275,13 +272,15 @@ class RequestWrapper {
...
 
-        for (let callback of (this.pluginCallbacks.cacheDidUpdate || [])) {
-          callback({
-            cacheName: this.cacheName,
-            oldResponse,
-            newResponse,
-            url: request.url,
-          });
+        if (this.plugins.has('cacheDidUpdate')) {
+          for (let plugin of this.plugins.get('cacheDidUpdate')) {
```
gauntface: Same as above, think it would be better to just get the array once.

packages/sw-runtime-caching/src/lib/request-wrapper.js
jeffposnick: I believe `for ... of` used with an iterable does the right thing here, and only evaluates the entries it will iterate over once at the start of the loop.

See the hacky example at https://jsbin.com/mesudupeze/edit?js,console in which demonstrates that.

packages/sw-runtime-caching/src/lib/request-wrapper.js
jeffposnick: See my previous comment.

packages/sw-runtime-caching/src/lib/request-wrapper.js
jeffposnick: This is @prateekbh's code from https://github.com/GoogleChrome/sw-helpers/pull/260, but from what I recall from that review, it's necessary to clone the request early, before the request might be modified by a  `requestWillFetch` plugin. We need to save a copy of the original request, not the modified request, to be potentially used by a `fetchDidFail` plugin.

Looking over the code again, I think the cloning is happening in the correct place. I can throw in some additional comments to clarify the logic.

packages/sw-runtime-caching/src/lib/request-wrapper.js
jeffposnick: Sure, I can switch that logic over.

packages/sw-runtime-caching/src/lib/request-wrapper.js
jeffposnick: Ditto.

jeffposnick:
(I made that last commit while logged in via a different computer than I normally use, which confused the CLA bot.)

PR #432, opened 2017-04-11
Trying to set concurrency to 1 on Appveyor

gauntface:
R: @jeffposnick @addyosmani 

This will sadly be a chillin PR until I can make AppVeyor happy.


addyosmani:
I've noticed a few other projects needing to get a better hold of concurrency to make their CI happy so let's go ahead with this change.

PR #434, opened 2017-04-17
Clients claim

gauntface:
R: @jeffposnick @addyosmani

Add clientsClaim option to sw-build and sw-lib.

### DIFF
--- a/packages/sw-build/src/lib/generate-sw.js
+++ b/packages/sw-build/src/lib/generate-sw.js
@@ -44,8 +44,7 @@ const errors = require('./errors');
  * used by sw-build. This is primarily useful for local development where
  * multiple sites may be served from `http://localhost`.
  * @param {Boolean} [input.clientsClaim] An optional boolean that indicates if
- * the page should skip waiting and claim current pages on a new service worker
- * install (Defaults to false).
+ * the new service worker should claim current pages (Defaults to false).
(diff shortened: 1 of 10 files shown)


packages/sw-build/src/lib/generate-sw.js
```diff
@@ -43,6 +43,9 @@ const errors = require('./errors');
  * @param {String} [input.cacheId] An optional ID to be prepended to caches
  * used by sw-build. This is primarily useful for local development where
  * multiple sites may be served from `http://localhost`.
+ * @param {Boolean} [input.clientsClaim] An optional boolean that indicates if
+ * the page should skip waiting and claim current pages on a new service worker
```
jeffposnick: The "skip waiting" bit in the JSDoc isn't accurate. That describes what happens if you call `self.skipWaiting()` inside of an `install` handler, and is independent of what `self.clients.claim()` does.

Some proposed wording:

"...indicates whether the newly activated service worker should take control of existing client pages, including pages already controlled by an older service worker."

gauntface: Sorry - this was a typo before I realised you had split out clients claim and skipwaiting options.

gauntface:
Please ignore the appveyor failing test. Something I'm exploring in global-ignores branch at the moment. Not sure what it's doing.

PR #435, opened 2017-04-18
Merge Docs

jpmedley:

I need a clean base to work from.


gulp-tasks/documentation.js
```diff
@@ -23,8 +23,8 @@ const {globPromise, taskHarness} = require('../utils/build');
 
 const FRONT_MATTER = `---
```
gauntface: You need to comment this value out to fix the lint errors. Test it's working with `gulp lint`.

jpmedley:
PTAL

README.md
```diff
@@ -27,34 +27,21 @@
...
 Each module is installed separately using the command line [as listed below](#the-libraries). To use a particular module:
 
 1. Install the module. For example:
 
-   `npm install --save-dev sw-lib`
+   `npm install --save-dev sw-build`
 2. Copy the module's JavaScript and map files to your serving directory. 
 
-   `cp node_modules/sw-lib/build/* app/`
+   `cp node_modules/sw-build/build/* app/`
```
gauntface: This makes sense for sw-lib, but not for sw-build.

sw-lib you need to import the file in your service worker, sw-build is a node module you'd use at build time.

README.md
```diff
@@ -27,34 +27,21 @@
...
 3. Import the modules to your service worker file. For example:
 
-   `importScripts('sw-lib.min.js');`
+   `importScripts('sw-build.min.js');`
```
gauntface: Same as above - this should be sw-lib.

docs/modules_overview.md
```diff
@@ -33,34 +33,21 @@ navigation_weight: 0
...
 Each module is installed separately using the command line [as listed below](#the-libraries). To use a particular module:
 
 1. Install the module. For example:
 
-   `npm install --save-dev sw-lib`
+   `npm install --save-dev sw-build`
 2. Copy the module's JavaScript and map files to your serving directory. 
 
-   `cp node_modules/sw-lib/build/* app/`
+   `cp node_modules/sw-build/build/* app/`
```
gauntface: Same as above. This should be sw-lib.

packages/sw-cache-expiration/README.md
```diff
@@ -1,5 +1,8 @@
 <!-- DO NOT EDIT. This page is autogenerated. -->
 <!-- To make changes, edit templates/Project-README.hbs, not this file. -->
+
+All of the modules are described below. Each has an _About_ page with basic usage instructions and a _Demo_ directory with an example. The main page of the documentation is [here](https://googlechrome.github.io/sw-helpers/#main).
```
gauntface: This is the packages readme, so there is only one module being referenced so the intro "All of the modules are described below" doesn't make sense in this context.

jpmedley:
PTAL

docs/Gemfile.lock
```diff
@@ -0,0 +1,53 @@
+GEM
```
jeffposnick: There's already https://github.com/GoogleChrome/sw-helpers/blob/master/Gemfile.lock

Is there a reason for this extra copy? Can you just move the old one if it needs to be at this location?

docs/index.md
```diff
@@ -1,226 +1,35 @@
...
-[![Dev Dependency Status][dev-dependency-image]][dev-dependency-url]
+# Welcome to Workbox
 
-# SW-Helpers
+Workbox is a suite of helper libraries for service workers and progressive web
+apps designed to simplify your development experience. 
 
-## Contents
+The easiest workbox deployments implement precaching in a manner of minutes.
+While the full set of features supports such things as runtime caching, routing,
```
jeffposnick: This sentence isn't grammatically correct, and the whole thing flows awkwardly.

How about:

_Using Workbox, you can implement precaching in a matter of minutes.
If you want to go further, Workbox provides routing logic, runtime caching, offline analytics, background syncing, and more."_

docs/index.md
```diff
@@ -1,226 +1,35 @@
...
+Generate a service worker in your build process
+using [sw-build](reference-docs/stable/latest/module-sw-build.html#main).
 
-## Setting Up
+## Much more
 
-Each module is installed separately using the command line [as listed below](#the-libraries). To use a particular module:
+Go beyond the basics. If you'd like to use the Workbox libraries, and you
+already have your own service worker, then
+[sw-lib](reference-docs/stable/latest/module-sw-lib.html#main) is a better option.
```
jeffposnick: "better option" than what? It sounds like we're telling the developer that `sw-lib` is a better option than their own service worker, and I don't think that's what you mean.

Can you reword along the lines of:

_...then using `sw-lib` is the ideal entry point._

docs/sw-build.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Generating a Service Worker
+
+### gulp
+
+Add the following to your `gulpfile.js`:
+
+    const swBuild = require('sw-build');
+
+    swBuild.generateSW({
```
jeffposnick: This code snippet is useful as a generic way of using `sw-build`, but it's not `gulp` specific.

If the code snippet were wrapped in a `gulp.task()` then it would be closer to a `gulp` example.

jpmedley:
PTAL

docs/sw-build.md
```diff
@@ -31,23 +31,25 @@ into your build script.
...
-      dest: './build/sw.js',
-      globPatterns: ['**\/*.{html,js,css}'],
-      globIgnores: ['admin.html'],
-      templatedUrls: {
-        '/shell': ['shell.hbs', 'main.css', 'shell.css'],
-      },
+    gulp.task('build', (callback) => {
+      const swBuild = require('sw-build');
+
+      swBuild.generateSW({
```
jeffposnick: You need to `return swBuild.generateSW({...` here, so that the promise chain can be used to signify that the `gulp` task has completed.

docs/sw-build.md
```diff
@@ -31,23 +31,25 @@ into your build script.
...
-
-    swBuild.generateSW({
-      rootDirectory: './build/',
-      dest: './build/sw.js',
-      globPatterns: ['**\/*.{html,js,css}'],
-      globIgnores: ['admin.html'],
-      templatedUrls: {
-        '/shell': ['shell.hbs', 'main.css', 'shell.css'],
-      },
+    gulp.task('build', (callback) => {
```
jeffposnick: Change `(callback)` to `()`, since you'll be returning a promise to signify task completion.

PR #439, opened 2017-04-19
adding replay All functionality

prateekbh:

This adds `replayAll` method to instances of backgroundSyncQueue

I have also added the same in example `sw.js`


packages/sw-background-sync-queue/test/browser/background-sync-queue.js
```diff
@@ -17,7 +17,10 @@
 'use strict';
 
 describe('background sync queue test', () => {
-  function onRes() {}
+	let responseAchieved = 0;
+  function onRes() {
```
addyosmani: The indentation appears a little off here.

PR #440, opened 2017-04-19
Properly wait on IDB deletion & E2E tests for cache expiration

### LINKED ISSUE #437: Investigate cache expiration bugs
**Library Affected**:
*sw-cache-expiration*

**Browser & Platform**:
*all browsers*

I need to dig a bit and see what might be causing the cache expiration logic using the `maxEntries` configuration to fail and delete entries in an overly aggressive manner. I'm seeing this incorrect behavior in a real-world web app that makes a number of requests in quick succession using a handler configured with cache expiration.

jeffposnick:
R: @addyosmani @gauntface

Fixes #437 

This was not fun to track down—the culprit was using `await` inside the function passed to `forEach()` not behaving like I thought. Switching to `for ... of` seems to have done the trick.

This also allows handlers that have opted-in to `waitForCache` behavior to wait until the cache expiration logic has completed before returning a response.

It adds some end-to-end tests to catch the original issue and as general hygiene.

### DIFF
--- a/packages/sw-cache-expiration/src/lib/cache-expiration.js
+++ b/packages/sw-cache-expiration/src/lib/cache-expiration.js
@@ -66,6 +66,13 @@ class CacheExpiration {
     // These are used to keep track of open IndexDB and Caches for a given name.
     this._dbs = new Map();
     this._caches = new Map();
+
+    // This is used to ensure there's one asynchronous expiration operation
+    // running at a time.
+    this._expirationMutex = false;
+    // If another expiration request comes in, the timestamp is saved here and
+    // re-run after.
+    this._timestampForNextRun = null;
   }
 
   /**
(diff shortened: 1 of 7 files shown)


jeffposnick:
The Travis CI failures are due to `/packages/sw-runtime-caching/build/sw-runtime-caching.js` not existing, since only the code for the package that's currently being tested is built. I'll need to find some way around that.

packages/sw-cache-expiration/src/lib/cache-expiration.js
```diff
@@ -193,17 +200,34 @@ class CacheExpiration {
    * Expires entries, both based on the the maximum age and the maximum number
    * of entries, depending on how this instance is configured.
    *
+   * A mutex is used to ensure that there is only one copy of this asynchronous
```
jeffposnick: The comments hopefully explain what's going on, but just to provide some more context for the review: running simultaneous copies of this method asynchronously *shouldn't* be a problem, but since they all are reading from the same IDB source of truth when deciding what to delete, it's an optimization to run one at a time.

But you need to account for new things being added to IDB that might not have been picked up by the in-progress invocation, so we want to schedule a single additional invocation following the completion of the current one, saving the most recent `now` value to use when recursively invoking it.

gauntface:
If you add a devDependency, lerna bootstrap should include it on travis.

LGTM on the rest of this PR

packages/sw-cache-expiration/src/lib/cache-expiration.js
addyosmani: Thanks for expanding beyond what was noted in the comment. Your notes on scheduling make sense.

addyosmani:
> If you add a devDependency, lerna bootstrap should include it on travis.

👍 This looks good minus the Travis failure.

jeffposnick:
Okay, Travis seems to be happy with the updated `devDependencies`.

PR #443, opened 2017-04-19
Adds handleFetch param to sw-router

gauntface:
R: @jeffposnick @addyosmani

Adds the handleFetch param to sw-routing.

### DIFF
--- a/packages/sw-routing/src/lib/router.js
+++ b/packages/sw-routing/src/lib/router.js
@@ -52,11 +52,24 @@ class Router {
   /**
    * Start with an empty array of routes, and set up the fetch handler.
    */
-  constructor() {
+  constructor({handleFetch} = {}) {
+    if (typeof handleFetch === 'undefined') {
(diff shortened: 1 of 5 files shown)


packages/sw-routing/src/lib/router.js
```diff
@@ -52,11 +52,24 @@ class Router {
   /**
    * Start with an empty array of routes, and set up the fetch handler.
    */
-  constructor() {
+  constructor({handleFetch} = {}) {
+    if (typeof handleFetch === 'undefined') {
+      handleFetch = true;
+    }
+
```
addyosmani: So the behavior here (reviewing the tests too) is `handleFetch` will always default to true unless the developer explicitly opts out. That's reasonable.

gauntface: Yeah - this is largely just a debug feature (i.e. Web Starter Kit for dev mode skipping fetch to always getting latest from network).

PR #445, opened 2017-04-20
Directory index

### LINKED ISSUE #334: sw-cli doesn't map /index.html => /
**Library Affected**:
sw-cli

**Browser & Platform**:
Node v7.7.2

The manifest generation caches "/index.html", but doesn't cache "/". So reloading the top-level URL/domain/etc 404's because that resource doesn't exist.

Perhaps we could have a 'for-free' rewriting rule (e.g., if we see a request for "/" and we _have_ "/index.html", then serve that).

gauntface:
R: @jeffposnick @addyosmani

Fixes #334

I had to add some additional parameters to the cacheWillMatch method (i.e. request parameter wasn't being set before).

### DIFF
--- a/packages/sw-lib/src/lib/sw-lib.js
+++ b/packages/sw-lib/src/lib/sw-lib.js
@@ -43,7 +43,7 @@ class SWLib {
    * check cache entries for a URLs ending with '/' to see if there is a hit
    * when appending the directoryIndex (i.e. '/index.html').
    */
-  constructor({cacheId, clientsClaim, directoryIndex} = {}) {
+  constructor({cacheId, clientsClaim, directoryIndex = 'index.html'} = {}) {
     if (cacheId && (typeof cacheId !== 'string' || cacheId.length === 0)) {
       throw ErrorFactory.createError('bad-cache-id');
     }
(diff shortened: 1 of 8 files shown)


jeffposnick:
Looks good to merge, modulo the `sw-lib.js` conflict.

packages/sw-lib/src/lib/sw-lib.js
```diff
@@ -39,14 +39,29 @@ class SWLib {
    * localhost.
    * @param {boolean} clientsClaim To claim currently open clients set
    * this value to true. (Default false).
+   * @param  {String} [input.directoryIndex]  The directoryIndex will
+   * check cache entries for a URLs ending with '/' to see if there is a hit
+   * when appending the directoryIndex (i.e. '/index.html').
    */
-  constructor({cacheId, clientsClaim} = {}) {
+  constructor({cacheId, clientsClaim, directoryIndex} = {}) {
```
jeffposnick: Actually, can you get away with using [default parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters) here?

`directoryIndex = 'index.html'` seems like it would be sufficient and is easier to read.

PR #446, opened 2017-04-20
Adding directory index to sw-build

gauntface:
R: @jeffposnick @addyosmani 

Adds directoryIndex to sw-build.


packages/sw-build/src/lib/generate-sw.js
```diff
@@ -45,6 +45,8 @@ const errors = require('./errors');
  * multiple sites may be served from `http://localhost`.
  * @param {Boolean} [input.clientsClaim] An optional boolean that indicates if
  * the new service worker should claim current pages (Defaults to false).
+ * @param {Boolean} [input.directoryIndex] An optional string that will
```
jeffposnick: `{Boolean}` -> `{String}`

PR #448, opened 2017-04-20
Adds handleFetch to sw-build

gauntface:
R: @jeffposnick @addyosmani 

Adds handleFetch to sw-build

### DIFF
--- a/packages/sw-build/src/lib/generate-sw.js
+++ b/packages/sw-build/src/lib/generate-sw.js
@@ -43,14 +43,11 @@ const errors = require('./errors');
  * @param {String} [input.cacheId] An optional ID to be prepended to caches
  * used by sw-build. This is primarily useful for local development where
  * multiple sites may be served from `http://localhost`.
-<<<<<<< HEAD
- * @param {Boolean} [input.skipWaiting] An optional boolean that indicates if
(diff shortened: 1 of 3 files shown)


packages/sw-build/src/lib/generate-sw.js
```diff
@@ -43,6 +43,9 @@ const errors = require('./errors');
  * @param {String} [input.cacheId] An optional ID to be prepended to caches
  * used by sw-build. This is primarily useful for local development where
  * multiple sites may be served from `http://localhost`.
+ * @param {Boolean} [input.handleFetch] Stops the generated service worker
```
jeffposnick: This JSDoc could use some work. How about:

"If `true`, the service worker's fetch handlers will respond to requests. If `false`, the fetch handlers will not respond, which can be useful in a development environment. (Defaults to `true`.)"

packages/sw-build/test/node/lib-write-sw.js
```diff
@@ -277,4 +277,68 @@ swlib.precache(fileManifest);
         cacheId: 'cache-id-example',
       });
   });
+
+  it('should be able to generate sw for template with cacheId', function() {
```
jeffposnick: "cacheId" -> "handleFetch"

PR #450, opened 2017-04-21
Webpack bug fix

prateekbh:
 current webpack plugin implemation pointed towards the source code of other plugin and 
also used the relative key from webpack config instead of absolute one.

### DIFF
--- a/gulp-tasks/test.js
+++ b/gulp-tasks/test.js
@@ -37,7 +37,7 @@ gulp.task('download-browsers', function() {
   });
 });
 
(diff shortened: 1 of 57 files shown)


prateekbh:
Oops, my bad. 
Will make fix this on weekend sometime

addyosmani:
The unrelated change (which I believe was for another PR we just landed a minute ago) should be dropped from here, but otherwise LGTM. @prateekbh do you want to go ahead and update and we can land this so you're unblocked for preact-cli?

PR #452, opened 2017-04-21
Support for a BroadcastCacheUpdate plugin for precached() assets

### LINKED ISSUE #309: BroadcastCacheUpdates configuration for swlib.precache()
**Library Affected**:
*sw-lib*

It's not currently possible to pass through configuration for the underlying `RequestWrapper` when using `swLib.precache()`. It would be nice to take that in as an optional second parameter.

The specific use case that I had in mind was being able to configure `sw-broadcast-cache-updates` so that you can be notified when a precached resource (like your App Shell HTML) gets updated.

(Not blocking for the hackathon, obviously, but just a nice-to-have that I wanted to track.)

### LINKED ISSUE #441: Debug-level logging for when BroadcastCacheUpdate can't determine freshness
**Library Affected**:
sw-broadcast-cache-update

Currently [sw-helpers/responses-are-same.js](https://github.com/GoogleChrome/sw-helpers/blob/master/packages/sw-broadcast-cache-update/src/lib/responses-are-same.js) differentiates two responses by looking at the etag, response-length, and last-modified headers.

Another approach would be to look at the response bodies. @samccone proposed using [SubtleCrypto.digest()](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest) on the two response bodies. It takes an `arrayBuffer` and we have those handy in `response.body.arrayBuffer().then(ab => ...`

This could be an alternative or done in addition to the headers check.
This approach would make sense if the server doesn't issue last-modified or etag headers, and the change resulted in the same response-length.

jeffposnick:
R: @addyosmani @gauntface

Fixes #309, fixes #441

This adds support for a `BroadcastCacheUpdates` plugin that's enabled by default in `SWLib`, and announces when previously cached resources are updated.

I had to account for the fact that `precache()` might append a cache-busting URL parameter when retrieving responses from the network, so there needed to be a bit of massaging via the `cacheKey` parameter inside of `RequestWrapper.fetchAndCache()` to ensure that the proper responses were compared, and the proper URL reported to `BroadcastCacheUpdates`.

### DIFF
--- a/packages/sw-broadcast-cache-update/src/lib/responses-are-same.js
+++ b/packages/sw-broadcast-cache-update/src/lib/responses-are-same.js
@@ -14,6 +14,7 @@
 */
 
 import ErrorFactory from './error-factory';
+import logHelper from '../../../../lib/log-helper.js';
 
 /**
  * Given two `Response`s, compares several header values to see if they are
@@ -42,9 +43,28 @@ import ErrorFactory from './error-factory';
  */
 function responsesAreSame({first, second, headersToCheck}={}) {
   if (!(first instanceof Response &&
-        second instanceof Response &&
-        headersToCheck instanceof Array)) {
+    second instanceof Response &&
+    headersToCheck instanceof Array)) {
     throw ErrorFactory.createError('responses-are-same-parameters-required');
+  }
+
+  const atLeastOneHeaderAvailable = headersToCheck.some((header) => {
+    return first.headers.has(header) && second.headers.has(header);
+  });
+  if (!atLeastOneHeaderAvailable) {
+    logHelper.log({
+      message: `Unable to determine whether the response has been updated
+        because none of the headers that would be checked are present.`,
+      data: {
+        'First Response': first,
+        'Second Response': second,
+        'Headers To Check': JSON.stringify(headersToCheck),
+      },
+    });
+
+    // Just return true, indicating the that responses are the same, since we
+    // can't determine otherwise.
+    return true;
   }
 
   return headersToCheck.every((header) => {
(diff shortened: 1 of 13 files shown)


packages/sw-broadcast-cache-update/src/lib/responses-are-same.js
```diff
@@ -42,11 +43,30 @@ import ErrorFactory from './error-factory';
...
+    return first.headers.has(header) && second.headers.has(header);
+  });
+  if (!atLeastOneHeaderAvailable) {
+    logHelper.log({
+      message: `Unable to determine whether the response has been updated
+        because none of the headers that would be checked are present.`,
+      data: {
+        'First Response': first,
+        'Second Response': second,
+        'Headers To Check': JSON.stringify(headersToCheck),
```
gauntface: Was there a reason this couldn't just be the array?

packages/sw-lib/src/lib/sw-lib.js
```diff
@@ -34,17 +36,23 @@ class SWLib {
...
-   * this value to true. (Default false).
+   * @param {string} [input.cacheId] Defining a cacheId is useful to ensure
+   * uniqueness across cache names. Useful if you have multiple sites served
+   * over localhost.
+   * @param {boolean} [input.clientsClaim] To claim currently open clients set
+   * this value to true. (Defaults to false).
    * @param  {String} [input.directoryIndex]  The directoryIndex will
    * check cache entries for a URLs ending with '/' to see if there is a hit
    * when appending the directoryIndex (i.e. '/index.html').
+   * @param {string} [input.precacheUpdatesChannel] This value will be used as
```
gauntface: Would `precacheChannelName` be better variable name here?

packages/sw-broadcast-cache-update/src/lib/responses-are-same.js
jeffposnick: I've found that there's some point (after the variable's gone out of scope in the main thread?) after which objects that are logged to the JS console are no longer "live", and arrays, for example, won't properly display their contents if you try to expand them. So logging the string representations of objects seems like the safest approach. We could theoretically move that logic into `logHelper.js`, but complex objects can't be `JSON.stringified()`.

packages/sw-lib/src/lib/sw-lib.js
jeffposnick: Sure.

Issue #453, opened 2017-04-22
Unable to use injectManifest() for a given build setup


jeffposnick:
CC: @gauntface 

**Library Affected**:
*sw-build*

Apologies as we may have touched on some of these options in code reviews, but it's hard to realize the pain points until you try something out in a real project.

I've got the following rough configuration:

```sh
src/
  sw.js
  sw.prod.js
  index.html
  static/
    js/
    css/
build/
  index.html
  static/
    js/
    css/
```

The development web server uses `src/` as root, and the production web server uses `build/` as root. `src/sw.js` is an empty, no-op service worker, and `src/sw.prod.js` uses the `swlib.precache([])` placeholder, along with some runtime caching rules.

I'm trying to come up with a configuration for `swBuild.injectManifest()` that will allow me to add the precaching info into `src/sw.prod.js`, with the result copied over to `build/sw.js`. The manifest generation takes place after the static files from `src/` have been versioned and copied over to `build/`.

Here's something that I've tried which doesn't work:

```js
swBuild.injectManifest({
  rootDirectory: 'build/',
  swFile: '../src/sw.prod.js',
  dest: '.',
  staticFileGlobs: [
    'index.html',
    'static/{js,css}/**/!(*map*)'
  ]
})
```

Setting `rootDirectory: 'build/'` seems like the right option, since that's the web root, and since the glob patterns that I've passed in assume that as the root directory. (Maybe I could set a different `rootDirectory` and then use `modifyUrlPrefix` to correct the manifest URLs, but `modifyUrlPrefix` isn't documented as being supported for the `injectManifest()` method.)

Setting `swFile: '../src/sw.prod.js'` is necessary because the `swFile` path is joined with `rootDirectory` in order to find the source service worker file.

I've tried setting `dest: '.'` or `dest: 'ignored'` and neither give me the output path that I want. `dest` is joined with `swFile` to compute the destination for the service worker file, and `swFile` includes `src/` as part of it's path (which I don't want as part of the output path), and it also has a filename of `sw.prod.js`, while I'd like the name to be `sw.js`.

I *think* the approach that would help is if the `swFile` and `dest` was considered as being relative to the current directory that the build script is running from, **not** relative to `rootDirectory`. It would also help if `dest` was required to be a path + filename, since otherwise you wouldn't be able to rename the output file.

prateekbh:
I am running into the exact same issue with sw-build-webpack plugin

I have a structure like
```
lib/
[...]

PR #459, opened 2017-04-25
Glob directory

### LINKED ISSUE #451: sw-build always copy sw-lib in root-directory, does not consider dest key
**Library Affected**:
*sw-build*

**Browser & Platform**:
All

Currently if the I do sw-build something like:

```js
 swBuild.generateSW({
   rootDirectory: '/folderA',
   dest: '/myserviceworker.js'
 });
```

This ends up copying `sw-lib-0.18.0.min.js` into "/folderA" and "myserviceworker" at the right place.
But the "myserviceworker.js" still has code

```js
  importScripts('sw-lib.v0.0.18.min.js');
```

which should have been 

```js
  importScripts('/folderA/sw-lib.v0.0.18.min.js');
```

gauntface:
R: @jeffposnick @addyosmani @prateekbh 

Fixes #451 #453 

This makes a few changes:
- rootDirectory is now globDirectory
- dest must now be a full path for the output file
- swFile is now swPath and is seperate from anything else.

This actually sanitzed a lot of logic which is normally a good sign that the API is a better way to go.

Let me know what you think (Prateek I changed the swFile to swPath in the plugin, please let me know if that's not ok).

### DIFF
--- a/packages/sw-build-webpack-plugin/index.js
+++ b/packages/sw-build-webpack-plugin/index.js
@@ -57,7 +57,7 @@ class SwBuildWebpackPlugin {
 	apply(compiler) {
 		compiler.plugin('after-emit', (compilation, callback) => {
 			const config = this.getConfig(compilation);
-			if (config.swFile) {
+			if (config.swPath) {
 				swBuild.injectManifest(config)
 					.then(() => callback())
 					.catch((e) => callback(e));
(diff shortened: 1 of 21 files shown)


jeffposnick:
Sorry for not bringing this up when we were discussing in the issue tracker, but looking at the code changes in context in this PR, I realize that we could get some additional naming equivalence if we used `swSrc` and `swDest`, or `swSrcPath` and `swDestPath`.

What do you think about that, @gauntface?

gauntface:
I'd probably be on the side of `swSrc` and `swDest` if we went with this. @addyosmani @prateekbh does that sound like a good idea for you for the injectManifest() function?

addyosmani:
I'm in favor of going for `swSrc` and `swDest`. Shall we go ahead with that?

addyosmani:
Per meeting today, let's do it.

PR #461, opened 2017-04-26
Throw an error if a glob pattern doesn't match anything

### LINKED ISSUE #449: Treat glob pattern with no file matches as a fatal error?
**Library Affected**:
*sw-build*

When using `sw-build` and passing in a glob pattern (either via `staticFileGlobs` or for a URL in `templatedUrls`), if that pattern doesn't match any files, then I *think* we should treat that as a fatal error that stops the build. It likely indicates that there's a developer error, perhaps because they don't realize how the `rootDirectory` comes into play, or because they don't know the glob pattern syntax.

I was thinking of something along the lines of adding this to https://github.com/GoogleChrome/sw-helpers/blob/master/packages/sw-build/src/lib/utils/get-file-details.js#L18:

```
if (globbedFiles.length === 0) {
  throw new Error(errors['useless-glob-pattern'] +
    JSON.stringify({rootDirectory, globPattern, globIgnores}));
}
```

How's that sound, @gauntface?

jeffposnick:
R: @addyosmani @gauntface

Fixes #449


packages/sw-build/src/lib/utils/get-file-details.js
```diff
@@ -16,6 +16,11 @@ module.exports = (globDirectory, globPattern, globIgnores) => {
     throw new Error(errors['unable-to-glob-files'] + ` '${err.message}'`);
   }
 
+  if (globbedFiles.length === 0) {
+    throw new Error(errors['useless-glob-pattern'] +
+      JSON.stringify({rootDirectory, globPattern, globIgnores}, null, 2));
```
gauntface: Just needs to be globDirectory to fix this :)

PR #465, opened 2017-04-26
Try a single build process instead of one divided by projects.

jeffposnick:
Another speculative change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,25 +10,12 @@ cache:
 node_js:
   - 'stable'
 
-env:
-  - PROJECT="sw-background-sync-queue"
-  - PROJECT="sw-broadcast-cache-update"
-  - PROJECT="sw-build"
-  - PROJECT="sw-build-webpack-plugin"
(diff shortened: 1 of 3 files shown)


jeffposnick:
Yeah, that doesn't work either. The Firefox webdriver appears to get unhappy after a while with this setup? Anyway, it was worth an experiment.

```
     WebDriverError: Unable to parse new session response: {"sessionId":"550efd2a-070e-4e74-8f04-d57def6c6f9d","value":{"acceptInsecureCerts":false,"browserName":"firefox","browserVersion":"53.0","moz:accessibilityChecks":false,"moz:processID":6787,"moz:profile":"/tmp/rust_mozprofile.DcikDQKt6k8i","pageLoadStrategy":"normal","platformName":"linux","platformVersion":"4.4.0-51-generic","rotatable":false,"specificationLevel":0,"timeouts":{"implicit":0,"pageLoad":300000,"script":30000}}}
```

gauntface:
@jeffposnick for what it's worth - this may have been fixed by #468

PR #474, opened 2017-04-28
staleWhileRevalidate() is the new fastest()

### LINKED ISSUE #316: sw-lib: unclear that staleWhileRevalidate() is the new fastest()
**Library Affected**:
sw-lib

Couldn't find a reference to this in the docs during migration. It would be nice to be able to hint that the old name for this has been changed :)

jpmedley:

Fixes #316 

Clarify that that staleWhileRevalidate() is the new fastest().

__Fails on files that I'm not touching in this PR.__

### DIFF
--- a/packages/sw-runtime-caching/src/lib/stale-while-revalidate.js
+++ b/packages/sw-runtime-caching/src/lib/stale-while-revalidate.js
@@ -24,7 +24,10 @@ import assert from '../../../../lib/assert';
  *
  * In addition to updating the appropriate caches, it will also trigger any
  * appropriate plugins defined in the underlying `RequestWrapper`.
- * XXXXXXXX
+ * 
+ * This class always returns the cache and returns it first regardless of how
+ * quickly a network response returns. This make it suitable for scenarios where
+ * a fastest strategy would otherwise be used.
  *
  * By default, `StaleWhileRevalidate` will cache responses with a 200 status
  * code as well as [opaque responses](http://stackoverflow.com/q/39109789)
--- a/packages/sw-runtime-caching/src/lib/stale-while-revalidate.js
+++ b/packages/sw-runtime-caching/src/lib/stale-while-revalidate.js
@@ -22,12 +22,14 @@ import assert from '../../../../lib/assert';
  * An implementation of a [stale-while-revalidate](https://developers.google.com/web/fundamentals/instant-and-offline/offline-cookbook/#stale-while-revalidate)
  * request strategy.
  *
- * In addition to updating the appropriate caches, it will also trigger any
- * appropriate plugins defined in the underlying `RequestWrapper`.
- * 
- * This class always returns the cache and returns it first regardless of how
- * quickly a network response returns. This make it suitable for scenarios where
- * a fastest strategy would otherwise be used.
+ * Resources are requested from both the cache and the network in parallel, then
+ * responds with the cached version. The cache is replaced with whatever returns
+ * from the network. In addition to updating the appropriate caches, it will
+ * also trigger any appropriate plugins defined in the underlying `RequestWrapper`.
+ *
+ * This strategy is the closest equivalent to the sw-toolbox
+ * [fastest](https://googlechrome.github.io/sw-toolbox/api.html#toolboxfastest)
+ * strategy.
  *
  * By default, `StaleWhileRevalidate` will cache responses with a 200 status
  * code as well as [opaque responses](http://stackoverflow.com/q/39109789)
--- a/packages/sw-runtime-caching/src/lib/stale-while-revalidate.js
+++ b/packages/sw-runtime-caching/src/lib/stale-while-revalidate.js
@@ -25,7 +25,8 @@ import assert from '../../../../lib/assert';
  * Resources are requested from both the cache and the network in parallel, then
  * responds with the cached version. The cache is replaced with whatever returns
  * from the network. In addition to updating the appropriate caches, it will
- * also trigger any appropriate plugins defined in the underlying `RequestWrapper`.
+ * also trigger any appropriate plugins defined in the underlying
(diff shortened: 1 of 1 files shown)


packages/sw-runtime-caching/src/lib/stale-while-revalidate.js
```diff
@@ -24,6 +24,7 @@ import assert from '../../../../lib/assert';
  *
  * In addition to updating the appropriate caches, it will also trigger any
  * appropriate plugins defined in the underlying `RequestWrapper`.
+ * XXXXXXXX
```
addyosmani: Looks like a draft commit slipped through here.

packages/sw-runtime-caching/src/lib/stale-while-revalidate.js
```diff
@@ -24,7 +24,10 @@ import assert from '../../../../lib/assert';
  *
  * In addition to updating the appropriate caches, it will also trigger any
  * appropriate plugins defined in the underlying `RequestWrapper`.
- * XXXXXXXX
+ * 
+ * This class always returns the cache and returns it first regardless of how
+ * quickly a network response returns. This make it suitable for scenarios where
+ * a fastest strategy would otherwise be used.
```
addyosmani: @jeffposnick does the current behavior of staleWhileRevalidate() match the old description from https://googlechrome.github.io/sw-toolbox/api.html#toolboxfastest? If so, I wonder if there's value in linking back to the legacy implementation of this inline in comments for those who are used to using toolbox.

```
Request the resource from both the cache and the network in parallel. Respond with whichever returns first. Usually this will be the cached version, if there is one. On the one hand this strategy will always make a network request, even if the resource is cached. On the other hand, if/when the network request completes the cache is updated, so that future cache reads will be more up-to-date.
```

jeffposnick:
> **Fails on files that I'm not touching in this PR.**

As per the [Travis CI output](https://travis-ci.org/GoogleChrome/sw-helpers/jobs/227018473#L492), the failure is directly related to your change in this PR, on line 23 of `stale-while-revalidate.js`

```sh
[23:57:36] Starting 'lint'...
[23:57:37] 
/home/travis/build/GoogleChrome/sw-helpers/packages/sw-runtime-caching/src/lib/stale-while-revalidate.js
  27:3  error  Trailing spaces not allowed  no-trailing-spaces
✖ 1 problem (1 error, 0 warnings)
[23:57:37] 'lint' errored after 502 ms
```

You must have a space or two following the `*` character on that line, which isn't allowed.

packages/sw-runtime-caching/src/lib/stale-while-revalidate.js
jeffposnick: The descriptions don't match. The `sw-toolbox`'s `fastest` was [Cache & network race](https://developers.google.com/web/fundamentals/instant-and-offline/offline-cookbook/#cache-and-network-race), and now we have an actual [Stale-while-revalidate](https://developers.google.com/web/fundamentals/instant-and-offline/offline-cookbook/#stale-while-revalidate).

I think the idea was that we were going to tell people who previously used `fastest` in `sw-toolbox` that `StaleWhileRevalidate` was the closest equivalent, despite not being exactly the same.

Using the phrase "fastest" here without any qualification doesn't make sense, since "fastest" doesn't have any generic meaning when it comes to cache strategies. "fastest" only make sense within the context of `sw-toolbox`, so the second sentence could be reworked to specifically link to the [`sw-toolbox.fastest` documentation](https://googlechrome.github.io/sw-toolbox/api.html#toolboxfastest), and make it clear that it's information that's useful for people migrating from `sw-toolbox`.

packages/sw-runtime-caching/src/lib/stale-while-revalidate.js
```diff
@@ -24,6 +24,10 @@ import assert from '../../../../lib/assert';
  *
  * In addition to updating the appropriate caches, it will also trigger any
  * appropriate plugins defined in the underlying `RequestWrapper`.
+ * 
+ * This class always returns the cache and returns it first regardless of how
```
jeffposnick: > "This class always returns the cache and returns it **first** regardless of how quickly a network response returns."

This phrasing isn't accurate—"returns it first" implies that the network response is also going to be returned in addition to the previously cached response, which isn't the case.

The network response is used to update the cache asynchronously, after the previously cached response is returned to fulfill the request.

If there wasn't a previously cached response, then the network response **is** returned to the client in addition to being used to update the cache, but in that scenario, there's also only a single response returned to the client, so the "first" phrasing is still misleading.

jpmedley: > The network response is used to update the cache asynchronously, after the previously cached response is returned to fulfill the request.

So I'd have to call this again to get what returned in the validate part of staleWhileRevalidate?

jpmedley:
Second attempt is up.

jeffposnick:
The revised wording looks good, but there's now a different [linting error](https://travis-ci.org/GoogleChrome/sw-helpers/jobs/227147679#L508):

```sh
[15:01:40] Starting 'lint'...
[15:01:41] 
/home/travis/build/GoogleChrome/sw-helpers/packages/sw-runtime-caching/src/lib/stale-while-revalidate.js
  28:1  error  Line 28 exceeds the maximum line length of 80  max-len
✖ 1 problem (1 error, 0 warnings)
[15:01:41] 'lint' errored after 510 ms
```

gauntface:
Shouldn't this have been in a migration doc rather than in the reference docs for the module? (i.e. when sw-toolbox is no longer a maintained project or someone who has no knowledge of sw-toolbox, this just raises more questions than it solves.)

jpmedley:
It will be. I have a migration doc in an unmerged branch. More later.

PR #475, opened 2017-04-28
Add gh-pages/ to .gitignore.

jpmedley:

Is there some reason `gh-pages/` _isn't_ in `.gitignore`?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ docs/Gemfile*
 docs/_*
 docs/reference-docs
 docs/themes
+gh-pages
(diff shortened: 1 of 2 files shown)


jeffposnick:
`gh-pages/` isn't a local directory. It's a branch name. My understanding of the docs-related build scripts is that they all use `docs/` as the local directory—what's the process you're following that results in a `gh-pages/` directory being created?

gauntface:
It's created when publishing the docs.

Only scenario where it might be left is if the build fails or the script execution is killed early.

PR #476, opened 2017-04-29
Clean redirected responses

### LINKED ISSUE #173: Account for new redirected response security restriction
CC: @gauntface @addyosmani as an FYI.

**Library Affected**:
*sw-precaching*

**Browser & Platform**:
Chrome 59+, Firefox

There's background at https://bugs.chromium.org/p/chromium/issues/detail?id=658249

There's a new security restriction in place that adds in a `Response.redirected` field to indicate whether a response came from an HTTP 3xx redirect. If so, that response can't be used to fulfill any request that is made with a `redirect === 'manual'`. All navigation requests have `redirect === 'manual'`.

In practice, this is most likely to come up with precaching, since those responses are what's normally going to be used for navigations, and since you're most likely to pass in a specific URL that results in a redirect in that use case. (Passing in `/index.html`, which then gets redirected by a server to `/`, is common.)

There's a discussion of how this is going to be addressed in `sw-precache` at https://github.com/GoogleChrome/sw-precache/issues/220#issuecomment-274837496, and whatever approach is taken there should be carried over to `sw-precaching`, here: https://github.com/GoogleChrome/sw-helpers/blob/master/packages/sw-precaching/src/lib/controllers/base-cache-manager.js#L121

jeffposnick:
R: @addyosmani @gauntface

Fixes #173 

This is analogous to the change in https://github.com/GoogleChrome/sw-precache/pull/241

### DIFF
--- a/packages/sw-runtime-caching/src/lib/clean-response-copy.js
+++ b/packages/sw-runtime-caching/src/lib/clean-response-copy.js
@@ -1,0 +1,47 @@
+/*
+ Copyright 2016 Google Inc. All Rights Reserved.
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+import assert from '../../../../lib/assert';
+
+/**
+ * Helper method to "clean" a redirected response, so that it could be used
+ * to fulfill a navigation request.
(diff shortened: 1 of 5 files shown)


packages/sw-runtime-caching/src/lib/clean-response-copy.js
```diff
@@ -0,0 +1,47 @@
...
+
+  // Not all browsers support the Response.body stream, so fall back to reading
+  // the entire body into memory as a blob.
+  const bodyPromise = 'body' in clonedResponse ?
+    Promise.resolve(clonedResponse.body) :
+    clonedResponse.blob();
+
+  return bodyPromise.then((body) => {
+    // new Response() is happy when passed either a stream or a Blob.
+    return new Response(body, {
```
addyosmani: Reading through https://github.com/GoogleChrome/sw-precache/pull/241/commits/1899583f40453ddc32af5c241ef9c7a0fd5935a5#diff-c1129c8b045390789fa8ff62f2c6b4a9R492 I was reminded that we had a conversation about globals @gauntface. We might want to revisit that again sometime.

PR #479, opened 2017-05-01
Explains why the config file is useful.

### LINKED ISSUE #315: sw-cli: consider explaining why saving to a config file is useful
**Library Affected**:
sw-cli

Users might find it useful to understand why saving to a config file is a good idea:
image

jpmedley:

Fixes #315

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -15,8 +15,9 @@ offline analytics, background syncing, and more.
 
 ## A service worker in two minutes
 
-Use [sw-cli](sw-cli)
-to add a service worker to an existing project in under two minutes.
+Use [sw-cli](sw-cli) to add a service worker to an existing project in under two
+minutes. Save your cli responses to a configuration file so that updates are
+even faster.
 
 ## A build script in ten minutes
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -16,8 +16,7 @@ offline analytics, background syncing, and more.
 ## A service worker in two minutes
 
 Use [sw-cli](sw-cli) to add a service worker to an existing project in under two
-minutes. Save your cli responses to a configuration file so that updates are
-even faster.
+minutes. 
 
 ## A build script in ten minutes
 
--- a/docs/sw-cli.md
+++ b/docs/sw-cli.md
@@ -36,6 +36,20 @@ After the command has run you'll have two new files, a
 `sw-lib.vX.X.X.min.js` and `sw.js` file (unless you changed
 the file name).
 
+## Creating a Config File
+
+One of the questions asked by `sw-cli generate:sw` is whether you'd like to save
+your responses to a configuration file. Answering `Y` will save your responses
+to `sw-cli-config.json`. Subsequent runs of the cli will regenerate the service
+worker based on the config file. Not only does this make building faster, it
+opens the possibility of running subsequent calls in a build script.
+
+For example, if you're using gulp you can pass `sw-cli generate:sw` to
+[gulp-shell](https://www.npmjs.com/package/gulp-shell) or
+[gulp-exec](https://www.npmjs.com/package/gulp-exec).
+
+## Using the Generated Service Worker
+
 To use the service worker, you'll need to register your
 newly generated service worker file in your web page,
 which you can do like so:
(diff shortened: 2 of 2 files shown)


jpmedley:
This doesn't look like my message. Am I missing something?

```[Failed] broadcast manager test > check broadcast
    Timeout of 100ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
    [Stack Trace]
    testReceiverChannel.onmessage@http://localhost:3004/packages/sw-background-sync-queue/test/browser/broadcast-manager.js:32:4```

**Comment by gauntface** [comment]
you've expressed concern over the config file, how it should work and how it should be altered - should we wait until that or do you want this landed regardless?

**Comment by jpmedley** [comment]
I withdrew my concern. I think it came about because of a work interruption that made me forget what I was doing.

**Comment by jpmedley** [comment]
There's a note to that affect in #315

File: docs/index.md [hunk 1]
```diff
@@ -15,8 +15,9 @@ offline analytics, background syncing, and more.
 
 ## A service worker in two minutes
 
-Use [sw-cli](sw-cli)
-to add a service worker to an existing project in under two minutes.
+Use [sw-cli](sw-cli) to add a service worker to an existing project in under two
+minutes. Save your cli responses to a configuration file so that updates are
```
gauntface: Just my two cents - this config file bit feels a bit too much info for the developer.

docs/sw-cli.md
```diff
@@ -36,6 +36,20 @@ After the command has run you'll have two new files, a
 `sw-lib.vX.X.X.min.js` and `sw.js` file (unless you changed
 the file name).
 
+## Creating a Config File
+
+One of the questions asked by `sw-cli generate:sw` is whether you'd like to save
+your responses to a configuration file. Answering `Y` will save your responses
+to `sw-cli-config.json`. Subsequent runs of the cli will regenerate the service
+worker based on the config file. Not only does this make building faster, it
+opens the possibility of running subsequent calls in a build script.
```
gauntface: Maybe make this a little more explicit "opens the possibility of running the CLI without needing user input, like running from a build script".

docs/sw-cli.md
```diff
@@ -36,6 +36,20 @@ After the command has run you'll have two new files, a
...
+
+For example, if you're using gulp you can pass `sw-cli generate:sw` to
```
gauntface: I think developers using gulp would find it easier to use sw-build @jeffposnick @addyosmani thoughts on this?

docs/index.md
jpmedley: I felt like it reinforced the idea of a simple workflow. I'd like some other opinions before removing it.

docs/sw-cli.md
jpmedley: Is that a false choice? Seems like both options might have their place.

addyosmani: I think both options have their place but that using sw-build would probably address the 90% case without them having to think about cli. Perhaps we steer them towards build and mention the cli workflow here as an alternative?

jpmedley: Removing text from docs/index.md.

jpmedley:
@gauntface @addyosmani Take another look, guys.

jpmedley:
Ping!

docs/sw-cli.md
```diff
@@ -36,6 +36,21 @@ After the command has run you'll have two new files, a
...
 the file name).
 
+## Creating a Config File
+
+One of the questions asked by `sw-cli generate:sw` is whether you'd like to save
+your responses to a configuration file. Answering `Y` will save your responses
+to `sw-cli-config.json`. Subsequent runs of the cli will regenerate the service
+worker based on the config file, which makes building faster. 
+
+If you don't need anything more complicated (in other words if you don't need
```
jeffposnick: I don't see using a `gulp` plugin that shells out to `sw-cli` as being a build process that we'd recommend. This paragraph could just be cut.

There's also no reason to assume that someone reading the docs for `sw-cli` is using `gulp`, so if you did feel like this paragraph needs to stay, I wouldn't make it `gulp`-specific.

jpmedley: @jeffposnick Removed. Please review.

PR #480, opened 2017-05-01
Nav fallback

gauntface:

Adds navigateFallback option.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,6 @@
     "local-docs": "npm-publish-scripts serve",
     "publish-docs": "npm-publish-scripts publish-docs",
     "publish-release": "npm run lerna-bootstrap && gulp test && lerna publish --concurrency 1",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -7,7 +7,6 @@
     "local-docs": "npm-publish-scripts serve",
     "publish-docs": "npm-publish-scripts publish-docs",
     "publish-release": "npm run lerna-bootstrap && gulp test && lerna publish --concurrency 1",
-    "postinstall": "npm run lerna-bootstrap",
```
addyosmani: Only thing that was unclear to me in this PR is why we're dropping the postinstall here.

gauntface: Windows tries to build ALL modules after npm install. This hits the heap issue on windows.

addyosmani: Gotcha. AppVeyor appears to be done and all fine here. Waiting for Travis to complete.

PR #485, opened 2017-05-02
Major revamp of the build process

### LINKED ISSUE #477: Fix sourcemap generation
As per https://github.com/GoogleChrome/sw-helpers/issues/361#issuecomment-286641694, the current sourcemaps we generate are missing some context/information. @paulirish used the following to generate better sourcemaps, and we can standardize on something closer to that.

```js
const fs = require('fs');	
const babel = require('babel-core');

const rollup = require('rollup');
const nodeResolve = require('rollup-plugin-node-resolve');
const commonjs = require('rollup-plugin-commonjs');

rollup
.rollup({
  entry: './packages/sw-lib/src/index.js',
  sourceMap: true,
  plugins: [
    nodeResolve({ jsnext: true, main: true, browser: true }),
    commonjs()
  ]
})
.then(bundle => bundle.write({ 
  dest: 'swbundle.js', 
  moduleName: 'goog.swlib',
  format: 'cjs',
  sourceMap: true,
}))
.catch(err => console.log(err.message, err.stack))

.then(_ => {

  const sourcejs = fs.readFileSync('swbundle.js', 'utf-8');
  const result = babel.transform(sourcejs, {
    presets: [
      ['babili', {
        comments: false,
        keepFnName: true,
        keepClassName: true,
      }],
    ],
    sourceMaps: true,
    inputSourceMap: JSON.parse(fs.readFileSync('swbundle.js.map', 'utf-8'))
  });

       
  fs.writeFileSync('./swbundle.min.js', result.code);
  fs.writeFileSync('./swbundle.min.js.map', JSON.stringify(result.map));

});
```

jeffposnick:
R: @addyosmani @gauntface

This is a substantial update to the build process, with the following changes:
- Moving from gulp wrappers on top of Rollup, babili, and others to directly using Rollup. This fixes #477, sourcemap generation.
- No longer including the async-to-generator transpilation, since async/await are natively supported in production Chrome and Firefox. We can turn this back on if we need.
- Changing the build process for `sw-offline-google-analytics` and `sw-background-sync-queue` to only generate unminimized `iife` bundles of the individual `.js` files that they need in order to run their tests, and to generate them all in parallel instead of sequentially.
- Adding in an `.npmignore` file to the directory that hosts the built tests for `sw-offline-google-analytics` and `sw-background-sync-queue` to ensure they're not published to `npm`.

As a result of these changes, we've saved a modest ~364 bytes on the gzip-ed size of `swlib.min.js`. But the built times should be much quicker across the board—I've seen improvements of around 10 seconds in `sw-lib`, and up to 40 seconds improvement for `sw-background-sync-queue` (which was really bad to begin with, due to a long sequential chain of bundled test code).

### DIFF
--- a/gulp-tasks/build.js
+++ b/gulp-tasks/build.js
@@ -23,7 +23,7 @@ const fsePromise = promisify('fs-extra');
 
 const printHeading = (heading) => {
   /* eslint-disable no-console */
-  process.stdout.write(chalk.inverse(`  ⚒️  ${heading}  `));
+  process.stdout.write(chalk.inverse(`  ⚒  ${heading}  `));
   /* eslint-enable no-console */
 };
 
@@ -32,7 +32,7 @@ const printBuildTime = (buildTime) => {
 };
 
 /**
- * Buids a given project.
+ * Builds a given project.
  * @param {String} projectPath The path to a project directory.
  * @return {Promise} Resolves if building succeeds, rejects if it fails.
  */
--- a/packages/sw-offline-google-analytics/test/browser/enqueue-request.js
+++ b/packages/sw-offline-google-analytics/test/browser/enqueue-request.js
@@ -17,9 +17,9 @@
 'use strict';
 
 describe('enqueue-request', () => {
-  const enqueueRequest = goog.offlineGoogleAnalytics.test.enqueueRequest;
-  const constants = goog.offlineGoogleAnalytics.test.constants;
-  const IDBHelper = goog.offlineGoogleAnalytics.test.IDBHelper;
+  const enqueueRequest = goog.offlineGoogleAnalytics.test.EnqueueRequest;
+  const constants = goog.offlineGoogleAnalytics.test.Constants;
+  const IDBHelper = goog.offlineGoogleAnalytics.test.IdbHelper;
 
   const idbHelper = new IDBHelper(constants.IDB.NAME, constants.IDB.VERSION,
     constants.IDB.STORE);
--- a/package.json
+++ b/package.json
@@ -30,10 +30,7 @@
     "update-notifier": "^2.1.0"
   },
   "devDependencies": {
-    "babel-plugin-external-helpers": "^6.8.0",
-    "babel-plugin-transform-async-to-generator": "^6.8.0",
     "babel-preset-babili": "0.0.12",
-    "babel-preset-es2015": "^6.16.0",
     "chai": "^3.5.0",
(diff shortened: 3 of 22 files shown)


gulp-tasks/build.js
```diff
@@ -23,7 +23,7 @@ const fsePromise = promisify('fs-extra');
 
 const printHeading = (heading) => {
   /* eslint-disable no-console */
-  process.stdout.write(chalk.inverse(`  ⚒️  ${heading}  `));
+  process.stdout.write(chalk.inverse(`  ⚒  ${heading}  `));
```
jeffposnick: There was some weird control character in this line that would result in a primitive smiley face being printed out to the console on my Mac.

packages/sw-offline-google-analytics/test/browser/enqueue-request.js
```diff
@@ -17,9 +17,9 @@
 'use strict';
 
 describe('enqueue-request', () => {
-  const enqueueRequest = goog.offlineGoogleAnalytics.test.enqueueRequest;
-  const constants = goog.offlineGoogleAnalytics.test.constants;
-  const IDBHelper = goog.offlineGoogleAnalytics.test.IDBHelper;
+  const enqueueRequest = goog.offlineGoogleAnalytics.test.EnqueueRequest;
```
gauntface: Why are you renaming here? Won't this break developers?

gauntface:
If the tests pass then go for it. Only one comment on the analytics stuff.

Regarding transpiling, go for it - only concern - older versions of these browsers and edge support.

packages/sw-offline-google-analytics/test/browser/enqueue-request.js
jeffposnick: These aren't part of the public interface. These are exports consumed via the tests, and the tests were also updated to account for the new naming.

(The motivation for renaming is automation—instead of having to manually create the exported names, they're [generated](https://github.com/GoogleChrome/sw-helpers/pull/485/files#diff-96ddb612cfff564769e651b776a01117R37) by running the individual source file through [`uppercamelcase`](https://www.npmjs.com/package/uppercamelcase).)

jeffposnick:
Adding back `async`/`await`, or bundling as `umd` instead of `iife` might end up being necessary, but I figured let's start with what will lead to the smallest size to begin with, and see if anyone files issues after launch with use cases that require changes.

PR #486, opened 2017-05-02
fixing async tests

### LINKED ISSUE #484: Silently failing sw-background-sync-queue tests
**Library Affected**:
*sw-background-sync-queue*

While rewriting the build process, I was examining the tests for `sw-background-sync-queue` to make sure they continued to pass. I noticed that in the current `master` branch there are a number of failures that are logged to the JS console during test execution; they appear to the "silent" in that they don't cause any of the tests to fail. It's hard to copy the JS console text, but here's a screenshot from the current Chrome stable:

screen shot 2017-05-01 at 10 26 20 pm

prateekbh:
R: @jeffposnick 

Fixes #484 

Two of the async tests did not returned the promise back, which was failing the assertion.

### DIFF
--- a/packages/sw-background-sync-queue/test/browser/background-sync-queue.js
+++ b/packages/sw-background-sync-queue/test/browser/background-sync-queue.js
@@ -63,18 +63,14 @@ describe('background sync queue test', () => {
 
   it('check push proxy', async () => {
     const currentLen = backgroundSyncQueue._queue.queue.length;
-    await backgroundSyncQueue.pushIntoQueue({request: new Request('http://lipsum.com')}).then( (e) => {
-      chai.assert.equal(backgroundSyncQueue._queue.queue.length,
-        currentLen + 1);
-    });
+    await backgroundSyncQueue.pushIntoQueue({request: new Request('http://lipsum.com')});
+    chai.assert.equal(backgroundSyncQueue._queue.queue.length, currentLen + 1);
   });
 
   it('check fetchDid fail proxy', async () => {
     const currentLen = backgroundSyncQueue._queue.queue.length;
-    await backgroundSyncQueue.fetchDidFail({request: new Request('http://lipsum.com')}).then( (e) => {
-      chai.assert.equal(backgroundSyncQueue._queue.queue.length,
-        currentLen + 1);
-    });
+    await backgroundSyncQueue.fetchDidFail({request: new Request('http://lipsum.com')});
+    chai.assert.equal(backgroundSyncQueue._queue.queue.length, currentLen + 1);
   });
 
 	it('check replay', async function() {
--- a/packages/sw-background-sync-queue/test/browser/background-sync-queue.js
+++ b/packages/sw-background-sync-queue/test/browser/background-sync-queue.js
@@ -61,18 +61,18 @@ describe('background sync queue test', () => {
       CALLBACKS);
   });
 
-  it('check push proxy', () => {
+  it('check push proxy', async () => {
     const currentLen = backgroundSyncQueue._queue.queue.length;
-    return backgroundSyncQueue.pushIntoQueue({request: new Request('http://lipsum.com')}).then( (e) => {
-      return chai.assert.equal(backgroundSyncQueue._queue.queue.length,
+    await backgroundSyncQueue.pushIntoQueue({request: new Request('http://lipsum.com')}).then( (e) => {
+      chai.assert.equal(backgroundSyncQueue._queue.queue.length,
         currentLen + 1);
     });
   });
 
-  it('check fetchDid fail proxy', () => {
+  it('check fetchDid fail proxy', async () => {
     const currentLen = backgroundSyncQueue._queue.queue.length;
-    return backgroundSyncQueue.fetchDidFail({request: new Request('http://lipsum.com')}).then( (e) => {
-      return chai.assert.equal(backgroundSyncQueue._queue.queue.length,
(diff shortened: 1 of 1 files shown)


packages/sw-background-sync-queue/test/browser/background-sync-queue.js
```diff
@@ -63,16 +63,16 @@ describe('background sync queue test', () => {
 
   it('check push proxy', () => {
     const currentLen = backgroundSyncQueue._queue.queue.length;
-    backgroundSyncQueue.pushIntoQueue({request: new Request('http://lipsum.com')}).then( (e) => {
-      chai.assert.equal(backgroundSyncQueue._queue.queue.length,
+    return backgroundSyncQueue.pushIntoQueue({request: new Request('http://lipsum.com')}).then( (e) => {
+      return chai.assert.equal(backgroundSyncQueue._queue.queue.length,
```
jeffposnick: I don't think `chai.assert.*` has a return value, so `return`ing it here isn't accomplishing anything.

packages/sw-background-sync-queue/test/browser/background-sync-queue.js
```diff
@@ -63,16 +63,16 @@ describe('background sync queue test', () => {
...
         currentLen + 1);
     });
   });
 
   it('check fetchDid fail proxy', () => {
     const currentLen = backgroundSyncQueue._queue.queue.length;
-    backgroundSyncQueue.fetchDidFail({request: new Request('http://lipsum.com')}).then( (e) => {
-      chai.assert.equal(backgroundSyncQueue._queue.queue.length,
+    return backgroundSyncQueue.fetchDidFail({request: new Request('http://lipsum.com')}).then( (e) => {
+      return chai.assert.equal(backgroundSyncQueue._queue.queue.length,
```
jeffposnick: I don't think `chai.assert.*` has a return value, so `return`ing it here isn't accomplishing anything.

packages/sw-background-sync-queue/test/browser/background-sync-queue.js
```diff
@@ -63,16 +63,16 @@ describe('background sync queue test', () => {
 
   it('check push proxy', () => {
```
jeffposnick: Consider marking this function as `async`, and then you can do

```js
await backgroundSyncQueue.pushIntoQueue({request: new Request('http://lipsum.com')});
chai.assert.equal(...);
```

See, e.g., https://github.com/GoogleChrome/sw-helpers/blob/master/packages/sw-runtime-caching/test/sw/request-wrapper.js#L214

That might make it clearer that you're not actually using the value that the first promise resolves with, and also clarify that the assertion doesn't return a value.

packages/sw-background-sync-queue/test/browser/background-sync-queue.js
```diff
@@ -61,17 +61,17 @@ describe('background sync queue test', () => {
       CALLBACKS);
   });
 
-  it('check push proxy', () => {
+  it('check push proxy', async () => {
     const currentLen = backgroundSyncQueue._queue.queue.length;
-    backgroundSyncQueue.pushIntoQueue({request: new Request('http://lipsum.com')}).then( (e) => {
+    await backgroundSyncQueue.pushIntoQueue({request: new Request('http://lipsum.com')}).then( (e) => {
```
jeffposnick: `await` is an alternative to attaching a `.then()`.

So this could be:

```js
await backgroundSyncQueue.pushIntoQueue({request: new Request('http://lipsum.com')});
chai.assert(...);
```

That will block until `pushIntoQueue()` completes, and then execute the assertion. Again, it's just an alternative to using `.then()`, but I think it ends up be clearer, especially for cases when you're ignoring the value that the async operation resolves with (like you are in this case).

packages/sw-background-sync-queue/test/browser/background-sync-queue.js
```diff
@@ -61,17 +61,17 @@ describe('background sync queue test', () => {
...
       chai.assert.equal(backgroundSyncQueue._queue.queue.length,
         currentLen + 1);
     });
   });
 
-  it('check fetchDid fail proxy', () => {
+  it('check fetchDid fail proxy', async () => {
     const currentLen = backgroundSyncQueue._queue.queue.length;
-    backgroundSyncQueue.fetchDidFail({request: new Request('http://lipsum.com')}).then( (e) => {
+    await backgroundSyncQueue.fetchDidFail({request: new Request('http://lipsum.com')}).then( (e) => {
```
jeffposnick: Ditto re: removing the `.then()`.

packages/sw-background-sync-queue/test/browser/background-sync-queue.js
prateekbh: done

packages/sw-background-sync-queue/test/browser/background-sync-queue.js
prateekbh: done

jeffposnick:
The changes look okay—just to confirm, though, there are still some errors logged in the console when running the modified code. Are they to be expected/safe to ignore?

screen shot 2017-05-02 at 1 30 35 pm

prateekbh:
@jeffposnick most of them yes!
I've added dummy calls which does not even support Cross domain XHR. Although adding them immediately fires the call and causes some errors.
Ill revisit to minimize these, however its cool if they stay for a while

PR #491, opened 2017-05-02
Dont cache bust urls

gauntface:
R: @jeffposnick @addyosmani 

Adds dontCacheBustURLsMatching parameter

### DIFF
--- a/packages/sw-build/src/lib/generate-sw.js
+++ b/packages/sw-build/src/lib/generate-sw.js
@@ -55,6 +55,9 @@ const errors = require('./errors');
  * @param {number} [input.maximumFileSizeToCacheInBytes] An optional number to
  * define the maximum file size to consider whether the file should be
  * precached. (Defaults to 2MB).
+ * @param {Regex} [input.dontCacheBustUrlsMatching] An optional regex that will
+ * return a URL string and exclude the revision details for urls matching this
+ * regex. Useful if you have assets with file revisions in the URL.
  * @param {string} [input.navigateFallback] An optional string that will
  * attempt to serve the response for the URL defined as this option from cache.
  * @return {Promise} Resolves once the service worker has been generated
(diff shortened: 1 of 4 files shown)


packages/sw-build/src/lib/generate-sw.js
```diff
@@ -55,6 +55,9 @@ const errors = require('./errors');
  * @param {number} [input.maximumFileSizeToCacheInBytes] An optional number to
  * define the maximum file size to consider whether the file should be
  * precached. (Defaults to 2MB).
+ * @param {Regex} [input.dontCacheBustUrlsMatching] An optional regex that will
```
jeffposnick: Regex => RegExp

packages/sw-build/src/lib/get-file-manifest-entries.js
```diff
@@ -30,6 +30,9 @@ const getStringDetails = require('./utils/get-string-details');
  * @param {number} [input.maximumFileSizeToCacheInBytes] An optional number to
  * define the maximum file size to consider whether the file should be
  * precached. (Defaults to 2MB).
+ * @param {Regex} [input.dontCacheBustUrlsMatching] An optional regex that will
```
jeffposnick: Regex => RegExp

packages/sw-build/src/lib/generate-sw.js
```diff
@@ -55,7 +55,7 @@ const errors = require('./errors');
  * @param {number} [input.maximumFileSizeToCacheInBytes] An optional number to
  * define the maximum file size to consider whether the file should be
  * precached. (Defaults to 2MB).
- * @param {Regex} [input.dontCacheBustUrlsMatching] An optional regex that will
+ * @param {Regexp} [input.dontCacheBustUrlsMatching] An optional regex that will
```
jeffposnick: I don't know how particular JSDoc is, or whether it even matters in terms of how it generates the docs, but the capitalization of the JavaScript class is [`RegExp`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp). Feel free to ignore if it doesn't matter.

addyosmani: I assumed it was RegExp but similarly don't know how well JSDoc would trip up on casing here. Deferring to Matt.

PR #493, opened 2017-05-02
Runtimecaching

gauntface:
R: @jeffposnick @addyosmani 

runtimeCaching parameter.


packages/sw-build/src/lib/generate-sw.js
```diff
@@ -84,6 +84,11 @@ const generateSW = function(input) {
       new Error(errors['invalid-dest']));
   }
 
+  if (input.runtimeCaching && !(Array.isArray(input.runtimeCaching))) {
```
jeffposnick: Can you add a JSDoc comment?

PR #496, opened 2017-05-03
Mention that minifiers must be ES6 aware.

### LINKED ISSUE #352: Document that sw-lib can't be used uglify + Provide presets
**Library Affected**:
*sw-lib*, but probably all

**Platform**:
Using webpack 2.2.0

**Why do I care?** 
I'm already using webpack to bundle the rest of my site and want to include a piece of logic in my service worker that might be shared elsewhere without resorting to `importScripts` and leveraging existing webpack bundling.

**Description**
I know there's #269, but I would expect to still be able to require sw-lib without having to whitelist it from uglify (adding it to babel-loader can get it to work, but that increases build by ~30 seconds and needs a lot of presets) :)

in sw.js
```js
import swlib from 'sw-lib' // yields first error
import swlib from 'sw-lib/build/sw-lib.min.js' // yields second error
import swlib from 'sw-lib/build/sw-lib.js' // yields third error
swlib.warmRuntimeCache(['myfile.html'])
```

**Errors**
```
ERROR in sw.js from UglifyJs
Unexpected token: name (ErrorFactory$1) [./~/sw-lib/build/sw-lib.min.mjs:15,0][sw.js:90,6]
```

```
ERROR in sw.js from UglifyJs
Unexpected character '`' [./~/sw-lib/build/sw-lib.min.js:15,354][sw.js:88,339]
```

```
ERROR in sw.js from UglifyJs
Unexpected token: name (ErrorFactory$1) [./~/sw-lib/build/sw-lib.js:40,0][sw.js:113,6]
```

jpmedley:

Fixes #352

### DIFF
--- a/docs/sw-lib.md
+++ b/docs/sw-lib.md
@@ -42,9 +42,9 @@ service worker script, and minimal, straight-forward configuration.
    ```
 
    **Note:** If you use a minifier on your service worker script, be aware that
-   **Workbox requires one that is ES6-aware. At the time of Workbox's first
-   **release (May 2017)
-   [Babili](https://github.com/babel/babili) is the only one that does so.
+   Workbox requires one that is ES2015-aware. At the time of Workbox's first
+   release (May 2017)
+   [Babili](https://github.com/babel/babili) is the only one that is.
 
 ## Precaching
 
--- a/docs/sw-lib.md
+++ b/docs/sw-lib.md
@@ -41,6 +41,11 @@ service worker script, and minimal, straight-forward configuration.
    importScripts('/third-party/sw-lib/sw-lib.min.js');
    ```
 
+   **Note:** If you use a minifier on your service worker script, be aware that
+   **Workbox requires one that is ES6-aware. At the time of Workbox's first
+   **release (May 2017)
+   [Babili](https://github.com/babel/babili) is the only one that does so.
+
 ## Precaching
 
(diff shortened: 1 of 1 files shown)


addyosmani:
Thanks for adding a note. The alternative here would be a user needing to transpile Workbox to ES5 then run a regular minifier (Uglify etc) against it (afaik). Is this something we also want to document?

jpmedley:
@addyosmani Can we discuss that in one of the meetings or an issue? If so, then please merge this.

jpmedley:
Also, I need an actual review so I can merge this.

docs/sw-lib.md
```diff
@@ -41,6 +41,11 @@ service worker script, and minimal, straight-forward configuration.
    importScripts('/third-party/sw-lib/sw-lib.min.js');
    ```
 
+   **Note:** If you use a minifier on your service worker script, be aware that
+   **Workbox requires one that is ES6-aware. At the time of Workbox's first
```
jeffposnick: ES6 => ES2015

docs/sw-lib.md
```diff
@@ -41,6 +41,11 @@ service worker script, and minimal, straight-forward configuration.
...
+   **release (May 2017)
+   [Babili](https://github.com/babel/babili) is the only one that does so.
```
jeffposnick: "is the only one that fully supports ES2015 syntax" would make more sense.

PR #498, opened 2017-05-03
Adding ignoreUrlParamsMatching to sw-lib

gauntface:
R: @jeffposnick @addyosmani 

This just adds the logic to sw-lib, will hook up with sw-build in a seperate PR.

@jeffposnick could you check my logic isn't crazy different from yours in sw-precache.

### DIFF
--- a/packages/sw-lib/src/lib/sw-lib.js
+++ b/packages/sw-lib/src/lib/sw-lib.js
@@ -49,10 +49,14 @@ class SWLib {
    * plugin sends a message whenever a precached URL is updated. To disable this
    * plugin, set `precacheChannelName` to an empty string.
    * (Defaults to `'precache-updates'`)
+   * @param {Array<RegExp>} [input.ignoreUrlParametersMatching] An array of
+   * regex's to remove search params when looking for a cache match.
    */
   constructor({cacheId, clientsClaim, handleFetch,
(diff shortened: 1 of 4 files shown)


packages/sw-lib/src/lib/sw-lib.js
```diff
@@ -323,28 +338,79 @@ class SWLib {
 
   /**
    * @private
+   * @param  {Array<RegExp>} ignoreUrlParametersMatching An of regex's to
```
addyosmani: An array of?

packages/sw-lib/src/lib/sw-lib.js
```diff
@@ -323,28 +338,79 @@ class SWLib {
...
+      });
     };
 
-    return {cacheWillMatch: directoryIndexFunc};
+    return {cacheWillMatch: cacheMatchFunction};
+  }
+
+  /**
+   * @param {string} originalUrl The original url to remove the search params.
+   * @param  {Array<RegExp>} ignoreUrlParametersMatching An of regex's to
```
addyosmani: An array of

jeffposnick:
I believe @addyosmani's comments were addressed, and I know he's traveling. so I'm going to merge.

PR #500, opened 2017-05-04
Omit error-stack-parser from the minified bundles

jeffposnick:
R: @addyosmani @gauntface

This is based off the discussion in https://github.com/GoogleChrome/sw-helpers/issues/361#issuecomment-298640521

It swaps out the real [`error-stack-parser`](https://www.npmjs.com/package/error-stack-parser) module import for an import that has the same interface but is effectively a no-op. You lose a bit of context if an assertion fails without the real module, but our assertions are really meant to catch errors in a development environment, not after the service worker has been deployed.

This gets the production `sw-lib.min.js` bundle size down from 50249 bytes (14938 bytes gzip-ed) to 45450 bytes (13468 bytes gzip-ed).

I had to change a few of the tests that specifically examined the assertion failures to use the unminified bundle, but other than that, it's much less of a hack than it could have been.

### DIFF
--- a/lib/assert-min.js
+++ b/lib/assert-min.js
@@ -1,0 +1,28 @@
+/*
+ Copyright 2017 Google Inc. All Rights Reserved.
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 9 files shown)


jeffposnick:
Yeah, I've got a few ideas about renaming the build targets that I'd like to address as part of the overall CDN/distribution question, tracked at https://github.com/GoogleChrome/sw-helpers/issues/470

PR #501, opened 2017-05-04
Bg q fixes

prateekbh:

This removes the explicit need of calling `initialize` before usage.
Now in order to use BGSyncQ its just `new goog/wb.BackgroundSyncQueue();`


packages/sw-background-sync-queue/demo/index.html
```diff
@@ -123,9 +123,8 @@
         const toastContainer = document.querySelector('.toast-container');
 
         function purchase(){
-            fetch('https://jsonplaceholder.typicode.com/posts/1')
-            .then(res=>res.json())
-            .then(data=>{
+            fetch('http://localhost:3000/packages/sw-build/')
```
gauntface: why are you fetch sw-build? What exactly does this need to achieve in the request?

If it just needs some data, could it use one of the test server end points like "/__echo/counter"

packages/sw-background-sync-queue/demo/sw.js
```diff
@@ -26,6 +24,7 @@ let bgQueue = new goog.backgroundSyncQueue.BackgroundSyncQueue({callbacks:
 		},
 		onRetryFailure: (hash) => {},
 	},
+	dbName: 'queues',
```
gauntface: Why have you added a dbName here? There wasn't one in the previous example.

packages/sw-background-sync-queue/src/lib/background-sync-queue.js
```diff
@@ -129,6 +143,19 @@ class BackgroundSyncQueue {
 	replayRequests() {
 		return this._requestManager.replayRequests();
 	}
+
+  /**
+	 * sets the dbName, which is used to store the queue and requests
+	 * defaults to bgQueueSyncDB
+	 * @param {String} dbName
```
gauntface: this should be ID.

Should this method be private? If so, could you make it start with an underscore so it's more obvious it's not public,

I'm assuming it should be public as this used to be exposed by the module.

packages/sw-background-sync-queue/demo/index.html
prateekbh: yeah, i'll change it to use the counter call

packages/sw-background-sync-queue/demo/sw.js
prateekbh: I was just testing if the new constructor working.
Its totally optional. 
Will it be better if I remove it?

packages/sw-background-sync-queue/src/lib/background-sync-queue.js
prateekbh: oh yeah, thanks for pointing these out..
I'll make the changes

prateekbh: done

packages/sw-background-sync-queue/demo/index.html
prateekbh: done

prateekbh:
@gauntface I only have one concern about this change.
Earlier `initialize` method used to call `cleanUp` automatically.
But now constructor seems to be the apt position, but `cleanUp` is an async function. I am not sure if calling an async function inside a constructor is the wisest thing to do.
Right now dev has to call `cleanUp` manually, should we leave it like this and call this out in Docs?

jeffposnick:
Relying on a developer to remember to call `cleanup()` sounds problematic.

Constructors aren't allowed to by marked `async`, but there is nothing stopping them from starting asynchronous processes that might complete at some point after the constructor has finished. It's just that you can't get the return value from the asynchronous process—I'm assuming `cleanup()` doesn't return anything, so that's not a problem.

But taking a step back, how often would the `cleanup()` function ideally run?  Is it expensive to call when there's nothing that needs to actually be cleaned up? Kicking it off "in the background" via either the constructor or as part of the `sync` event handler sounds like it wouldn't be too bad.

packages/sw-background-sync-queue/demo/sw.js
gauntface: I'd have a test with and without tbh.

gauntface:
Couldn't you kick it off in the constructor and use that promise in a similar fashion to `navigator.serviceworker.ready` i.e. just chain other functions off of this.

Personally I don't think it's too bad to cleanup when the queue is finishing actually doing something.

I'm not familiar with the API so can't comment on the affect of this change, but I do feel like this is something the library should handle or treat as an edge / uncommon case and leave the developer to manage it.

prateekbh:
@gauntface  @jeffposnick I agree a manual trigger is super upsetting. I'll start the call in constructor.
As per frequency, it checks all the requests in the queue if they are past their threshold. So running them on every sync would be an unnecessary overhead.

I am not sure if a resolving a separate promise like `navigator.sw.ready` would make sense as this isn't a on off thing and if needed can be called multiple times as well, hence skipping this option for this PR

prateekbh:
Also @jeffposnick I'll open a separate PR for separating the plugin class

gauntface:
@prateekbh Is there any race conditions introduced by doing this in the constructor without anything waiting for it to finish?

prateekbh:
@gauntface @jeffposnick the cleanup queue has been added to the constructor.

prateekbh:
@gauntface @jeffposnick included changes for PluginClass as well

jeffposnick:
The `Queue`/`QueuePlugin` implementation looks 👍

gauntface:
Assuming this doesn't raise any problems, LGTM.

> Is there any race conditions introduced by doing this in the constructor without anything waiting for it to finish?

prateekbh:
@gauntface i don't see any condition in which it would create a race condition.

jeffposnick:
The AppVeyor failures are unrelated, and tracked in #517

Issue #109, opened 2017-01-09
Author migration guide for sw-framework/goog


addyosmani:
**Library Affected**:
*sw-framework*

**Browser & Platform**:
*All browsers*

**Issue or Feature Request Description**:

Let's write a guide for folks already using sw-precache and sw-toolbox to help them transition their existing code to the newer sw-framework work (I imagine this will primarily be through the higher level library @gauntface has been working on).

addyosmani:
@jpmedley Does it make sense for us to track the docs work for the libraries via this issue tracker or would you prefer for them to live in Web Fundamentals? (I'm still trying to figure out the balance of where guides should live - in-repo vs. WF)

gauntface:
Doc is here: https://docs.google.com/document/d/1qN3dHKkTHG-jWjWTwNzIN7NhVzd555R6FGa3AqRya_c/edit?usp=sharing just needs writing

PR #504, opened 2017-05-05
Start migration guide.

### LINKED ISSUE #313: sw-cli: document workflow for sw-precache users
**Library Affected**:
sw-cli

My old workflow using sw-precache to generate a SW for each run of my build process for a particular app used to look a little like this:

```js
./node_modules/sw-precache/cli.js --root=public --config=sw-precache-config.json
```

Reading through the sw-cli docs, something that might be useful is explaining if your old workflow relied on doing this in a fully automated way with the above approach the new strategy is to walk through the wizard to generate your SW.

jpmedley:

Fixes #313 

Starts the content for #109

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -1,0 +1,15 @@
+---
+layout: index
+title: Contributing
+navigation_weight: 7
+---
+# Migration from sw-framework
+
+[Intro]
+
+## Command-Line Interface (sw-precache)
+
+Just like `sw-precache`, Workbox allows you generate a service worker using the command line using [`sw-cli`](sw-cli). There are a few differences between the two.
+
+* Information that was previously passed as command line arguments is now gathered by a wizard that provides the option of saving your choices to a config file (`sw-cli-config.json`).
+* `sw-precache` does **not** support runtime caching. For that you'll need [`sw-lib`](sw-lib) or one of the lower-level modules.
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -1,7 +1,7 @@
 ---
 layout: index
 title: Contributing
-navigation_weight: 7
+navigation_weight: 100
 ---
 # Contributing
 


jpmedley:
The error is in code that I didn't touch (unless I'm missing something).

docs/migration.md
```diff
@@ -0,0 +1,15 @@
...
+title: Contributing
+navigation_weight: 7
+---
+# Migration from sw-framework
+
+[Intro]
+
+## Command-Line Interface (sw-precache)
+
+Just like `sw-precache`, Workbox allows you generate a service worker using the command line using [`sw-cli`](sw-cli). There are a few differences between the two.
```
jeffposnick: "using the command line using" is a bit awkward due to the repetition of "using". How about "via the command line using"

docs/migration.md
```diff
@@ -0,0 +1,15 @@
...
+
+* Information that was previously passed as command line arguments is now gathered by a wizard that provides the option of saving your choices to a config file (`sw-cli-config.json`).
```
jeffposnick: `sw-cli` takes in command line flags as well as supporting a config file. `sw-precache` supported both, too.

`sw-precache` didn't have a wizard for initial configuration, though.

docs/migration.md
```diff
@@ -0,0 +1,15 @@
...
+* `sw-precache` does **not** support runtime caching. For that you'll need [`sw-lib`](sw-lib) or one of the lower-level modules.
```
jeffposnick: > `sw-precache` does **not** support runtime caching.

Do you mean `sw-cli` there? And in any case, @gauntface added in something to support `runtimeCaching` in https://github.com/GoogleChrome/sw-helpers/pull/493

docs/migration.md
addyosmani: "via the command line using" 👍

docs/migration.md
addyosmani: Jeff's correct here. As much as possible we're trying to maintain some level of parity in how sw-cli is configured compared to sw-precache and they both do support CLI flags and a config.

docs/migration.md
addyosmani: With 493 landed, it would be good to drop this call out or change it to say something akin to "Similar to sw-precache, sw-cli also supports runtime caching".

gauntface: I've got parity between sw-precache and sw-build in terms of behaving as modules, this week will be parity as CLI's (Fingers crossed won't be too much work).

Issue #502, opened 2017-05-05
Revamped build artifacts and versioning


jeffposnick:
I wanted to outline a proposal around the build output is for our two different types of projects, get consensus, and then implement the changes.

There are a few other issues out there (#470, #487, #444, #361 to some extent) related to the build output, and they would follow once these details are nailed down.

## CommonJS & node CLI Projects
This includes `sw-cli`, `sw-build`, and `sw-build-webpack-plugin`.

### Proposed `build/` output
No substantial changes—we currently copy over the (unbundled) source to the the `build/` directory as-is, without a distinction between `dev` and `prod`.

### Proposed `package.json` metadata
- `"version":` *version string from lerna*
- `"files": ["build"]`
We currently include `"src"` as well, but that's not needed and can be removed.
- `"main": "build/index.js"`

## Service Worker "Runtime" Projects
This includes code that's executed at runtime within the context of a service worker, most notably `sw-lib`, but also the other smaller projects like `sw-runtime-caching` or `sw-background-sync-queue`.

### Proposed `build/` output
```sh
build/
  importScripts/
    <pkg.name>.dev.v<pkg.version>.js
    <pkg.name>.prod.v<pkg.version>.js
  modules/
    <pkg.name>.dev.v<pkg.version>.mjs
    <pkg.name>.prod.v<pkg.version>.mjs
```

The `build/importScripts/` naming emphasizes that the files in that directory are meant to be consumed by the service worker via `importScripts()`.

The `build/modules/` naming, and the `.mjs` extension, emphasizes that the files there are meant to be consumed as ES2015 modules. I don't know if it's currently important to include the versioning info inside of those filenames, but including it is forward-looking for the time in the future in which a service worker could natively import the modules instead of using `importScripts()`.

Getting the correct version number included in the filenames might pose a challenge, as it could involve hooking into the `lerna` build and doing a renaming after `Rollup` bundling is complete but before `lerna` pushes things out to `npm`. I need to explore that more.

Assume that sourcemaps are published alongside each JavaScript bundle.

### Proposed `package.json` metadata
- `"version":` *version string from lerna*
- `"files": ["build"]`
We currently include `"src"` as well, but that's *probably* not needed and can be removed.
- `"main": "build/importScripts/<pkg.name>.prod.v<pkg.version>.js"`
Point to the `prod` `iife` bundle here, primarily because some JS CDNs use that info to determine what gets served by default.
[...]

PR #505, opened 2017-05-05
Build output normalization for CommonJS/node projects.

jeffposnick:
R: @prateekbh @gauntface

This is the first half of #502, with some small changes to the build and `npm` packaging for the projects that are consumed as CommonJS modules or are CLIs.

`fs-extra` v3+ [supports promises natively now](https://github.com/jprichardson/node-fs-extra#usage) (yay!) so I was able to get rid of some of the extra explicit promise wrapping.

### DIFF
--- a/gulp-tasks/build.js
+++ b/gulp-tasks/build.js
@@ -12,51 +12,39 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 */
-const path = require('path');
-const gulp = require('gulp');
-const chalk = require('chalk');
-const promisify = require('promisify-node');
 
+/* eslint-disable no-console */
+
+const chalk = require('chalk');
+const fse = require('fs-extra');
+const gulp = require('gulp');
+const path = require('path');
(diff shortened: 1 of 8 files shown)


gulp-tasks/build.js
```diff
@@ -12,51 +12,39 @@
...
+const fse = require('fs-extra');
+const gulp = require('gulp');
+const path = require('path');
+const {taskHarness, buildJSBundle} = require('../utils/build');
 
 const printHeading = (heading) => {
-  /* eslint-disable no-console */
-  process.stdout.write(chalk.inverse(`  ⚒  ${heading}  `));
-  /* eslint-enable no-console */
+  process.stdout.write(chalk.inverse(`  ⚒ ${heading}`));
```
addyosmani: `(`  ⚒  ${heading}  `)`?

jeffposnick: I'm assuming this comment is about spacing—I'll adjust to what you suggest.

PR #510, opened 2017-05-06
Build output normalization for service worker runtime projects

jeffposnick:
~~~DO NOT MERGE. Still speculative.~~~


gulp-tasks/build.js
```diff
@@ -47,6 +47,36 @@ const buildPackage = (projectPath) => {
...
 };
 
+/**
+ * Updates the fields in package.json that contain version string to match
+ * the latest version from lerna.json.
+ *
+ * @param {String} projectPath The path to a project directory.
+ * @return {Promise} Resolves if updating succeeds, and rejects if it fails.
+ */
+const updateVersionedBundles = (projectPath) => {
```
jeffposnick: 👓

packages/sw-build/src/lib/utils/copy-sw-lib.js
```diff
@@ -1,37 +1,13 @@
...
-      const stream = fs.createReadStream(swlibBuiltPath)
-        .pipe(fs.createWriteStream(swlibOutputPath));
-      stream.on('error', function(err) {
-        reject(new Error(errors['unable-to-copy-sw-lib'] +
-          ` '${err.message}'`));
-      });
-      stream.on('finish', function() {
-        resolve(swlibOutputPath);
-      });
+  const swlibSrcPath = require.resolve('sw-lib');
```
jeffposnick: 👓

utils/server-instance.js
```diff
@@ -78,6 +78,48 @@ class ServerInstance {
       res.send(JSON.stringify(req.cookies));
     });
 
+    // Used by the following two routes.
```
jeffposnick: 👓

jeffposnick:
Hey @addyosmani, @gauntface & co.:

This is a pretty big PR, but there's a lot of configuration/boilerplate changes, so I've specifically called out a few of the areas that involve actual new logic to make sure you didn't miss them.

I'm not 100% sure that this will cleanly update the version strings inside the `package.json`'s `main` and `module` fields when used as part of a real `lerna publish` process, but I'm hopeful that it will. I don't think you can do a dry-run of `lerna publish`, so I'm ready to confirm the behavior after this is committed and then adjust (or rollback, if needed) depending on what the end result is.

gulp-tasks/build.js
gauntface: This worries me a little.

Lerna should be responsible for bumping the individual package versions (including the bootstrapped modules).

If you manually bump the module versions could this upset lerna and make it skip bumping the dependency versions?

gauntface: I thought packages only update when they've changed - this blanket updates everything to the latest version of packages - which may not be the case for some packages.

If this is running after lerna, then each package should be referencing it's own version.

packages/sw-build/src/lib/utils/copy-sw-lib.js
gauntface: LGTM

packages/sw-cache-expiration/test/static/cache-expiration.js
```diff
@@ -1,5 +1,5 @@
-importScripts('/packages/sw-cache-expiration/node_modules/sw-runtime-caching/build/sw-runtime-caching.js');
-importScripts('/packages/sw-cache-expiration/build/sw-cache-expiration.js');
+importScripts('/__test/lernaBundle/sw-cache-expiration/sw-runtime-caching');
```
gauntface: What is lerna bundle vs bundle?

packages/sw-lib/test/sw/claim-clients-false.js
```diff
@@ -1,9 +1,8 @@
 importScripts('/node_modules/mocha/mocha.js');
 importScripts('/node_modules/chai/chai.js');
-importScripts('/node_modules/sinon/pkg/sinon.js');
+importScripts('/node_modules/sinon/pkg/sinon-no-sourcemaps.js');
```
gauntface: why no sourcemaps?

utils/server-instance.js
```diff
@@ -78,6 +78,48 @@ class ServerInstance {
...
+      const pkg = req.params.pkg;
+      const pattern = `packages/${pkg}/build/importScripts/*dev*.js`;
+      respondWithScriptMatchingPattern(pattern, res);
+    });
+
+    // Used to return the latest iife bundle for a given package, without
+    // having to specify the version string, for packages that are installed
+    // via lerna.
+    // This is used within unit tests.
+    this._app.get('/__test/lernaBundle/:parentPkg/:pkg', function(req, res) {
```
gauntface: why is this needed? Wouldn't we always want tests running against latest builds?

gulp-tasks/build.js
jeffposnick: Good point—I'll have it read `pkg.version` instead.

packages/sw-lib/test/sw/claim-clients-false.js
jeffposnick: Oh, that.

I was seeing weirdness with the previous `sinon.js` was pulled in via `importScripts` in my
 local Chrome installation—it would fail to evaluate properly, and cause service worker registration to fail. I wasn't able to figure out what specifically was causing it to choke, but using the `-no-sourcemaps.js` version worked fine, and is equivalent.

I can still reproduce it, so maybe I'll take some time and investigate further and then file a bug against https://github.com/sinonjs/sinon

utils/server-instance.js
jeffposnick: I *thought* that this was needed because, e.g., the `packages/sw-runtime-caching/build` wouldn't exist when a single-project build, like `gulp build --project='sw-cache-expiration'` was performed.

Now after experimenting, I realize that `lerna bootstrap` will build that, (and also symlink), so I can drop it. Thanks for calling that out.

packages/sw-cache-expiration/test/static/cache-expiration.js
jeffposnick: (Not actually needed.)

jeffposnick:
Thanks for the `lerna`-related tips. I'm going to merge given your previous approval, and will try to cut a `0.0.21` release.

PR #511, opened 2017-05-08
Update Readme.md - Fixed broken links

sethbergman:
I added the correct paths to the broken links in the Readme.

Fixes #*issue number*

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,25 +14,25 @@ more.
 
 ## A service worker in two minutes
 
-Use [sw-cli](sw-cli) to add a service worker to an existing project in under two
+Use [sw-cli](/packages/sw-cli/README.md) to add a service worker to an existing project in under two
 minutes. 
 
 ## A build script in ten minutes
 
 Generate a service worker in your build process
-using [sw-build](sw-build).
+using [sw-build](/packages/sw-build/README.md).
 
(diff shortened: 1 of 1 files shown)


addyosmani:
Can confirm these paths are currently broken. It would be useful to check with @gauntface if the current README is being automatically generated before landing this change in case we need to make it elsewhere.

sethbergman:
I'm wondering if these pages could be the culprit(s).

- [/templates/README.hbs](https://github.com/GoogleChrome/sw-helpers/blob/master/templates/README.hbs)

- [/jsdoc.conf](https://github.com/GoogleChrome/sw-helpers/blob/master/jsdoc.conf)

gauntface:
Yeah, sorry, you need to use the gulp documentation task to update the templates for the readme's.

In this case it'll be edit README.hbs, run gulp documentation and commit the changes.

jpmedley:
Those are actually not the intended paths. 

Also, since we moved the docs into another repo this isn't the right place to make the changes either.

gauntface:
I'm afraid things have moved to a point where this PR can't be used. I'll be changing the readme's in a seperate PR.

sethbergman:
Sounds good. Thanks for the updates.

PR #514, opened 2017-05-08
full expample of webpack

prateekbh:

Fixes #*issue number*

Adding full webpack usage example


packages/sw-build-webpack-plugin/index.js
```diff
@@ -6,17 +6,25 @@ const swBuild = require('sw-build');
...
+ *     app: './step1/app.js'
+ *   },
+ *   output: {
+ *     path: __dirname + '/step1/public/js',
+ *     publicPath: '/public/js/',
+ *     filename: '[name].js',
+ *   },
+ *   plugins: [
+ *   	new SwBuildWebpackPlugin({
+ *   		globDirectory: './build/',
```
jeffposnick: The indentation is weird here, as it is in a lot of your files—are you using tabs instead of spaces? Normally it wouldn't matter much, but this will be user-facing.

prateekbh: yeah... my bad. that also reminds me, i have to raise a pr for prettier.
Anyways i'll fix this first

addyosmani: This LGTM minus the indentation change.

prateekbh:
@addyosmani @jeffposnick fixed. please take a look :)

prateekbh:
@jeffposnick fixed :)

anilanar:
This example is quite confusing. What's the relation between `path`, `publicPath` and `globDirectory`?

prateekbh:
so now `globDirectory` be default will be your `path`.
`publicPath` however is a webpack config as the path from where your async chunk can be loaded.

I'll take this as a feedback and raise a PR for more self explanatory docs

PR #515, opened 2017-05-09
gulp wrappers for lerna boostrapping and publishing

jeffposnick:
R: @addyosmani @gauntface

I **think** this ends up accomplishing what we want for publishing, via `gulp lerna-publish`.

It does it by:
1. Running the publishing flow with git and npm integration turned off. This updates the `package.json` and `lerna.json` `version` fields for the modified packages in the local working directory, without committing anything.
2. Automatically updating the `main` and `module` fields of all packages to pick up any modified versions.
3. Re-running the publishing flow, this time with git and npm enabled, but with a forced version passed in, read from the current `lerna.json`. This will then commit everything to GitHub and push to npm.

Rather than sprinkling some of the build steps in `package.json`'s `scripts` and others in `gulp`, I moved everything `lerna`-related over to `gulp`. This required a light refactoring of the CI configuration.

This is all working for me without `--concurrency=1`, so I'm hoping we can move ahead with that and speed up `lerna` in general.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ env:
 install:
   - npm install -g gulp-cli
   - npm install
-  - npm run lerna-bootstrap-scoped -- "${PROJECT}"
+  - gulp lerna-bootstrap-scoped --project="${PROJECT}"
 
 # Read more here: https://docs.travis-ci.com/user/gui-and-headless-browsers/#Using-xvfb-to-Run-Tests-That-Require-a-GUI
 before_script:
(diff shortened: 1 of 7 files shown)


jeffposnick:
The AppVeyor failures apparently started a few days ago, when the first round of build changes for sw-cli was introduced: https://ci.appveyor.com/project/gauntface/sw-helpers/build/1.0.453

I'm not sure why they weren't flagged then—I guess an AppVeyor build wasn't triggered for that PR?

In any case, I'll try to fix them in this PR, but it's requiring some trial-and-error since I can't reproduce them locally.

PR #518, opened 2017-05-09
Workbox microsite + gulp task instead of npm-publish-scripts

gauntface:

This adds the workbox microsite to gh-pages - removes npm-publish-scripts.

Running gulp documentations:gh-pages will push to github.

Reference docs are statically in /docs/ and navigation is done via /_data/navigation.yml. This should remove complicated logic in jekyll template and make the whole process easier to manage / reason with.


gauntface:
Also worth noting: The styles and JS are using CSS variables and in ES2015. This will probably need trans-piling with postCSS etc but will address that in a future PR. This is blocking demos and Joe's doc work so would like to land something sooner rather than later.

package.json
```diff
@@ -4,7 +4,7 @@
     "node": ">=6.0.0"
   },
   "scripts": {
-    "local-docs": "npm-publish-scripts serve",
+    "local-docs": "jekyll serve --source ./docs/ --trace --config ./docs/_config.yml",
     "publish-docs": "npm-publish-scripts publish-docs",
```
jeffposnick: `publish-docs` should be removed.

package.json
```diff
@@ -4,7 +4,7 @@
     "node": ">=6.0.0"
   },
   "scripts": {
-    "local-docs": "npm-publish-scripts serve",
+    "local-docs": "jekyll serve --source ./docs/ --trace --config ./docs/_config.yml",
```
jeffposnick: Optional—this could be be a task defined in `gulp-tasks/documentation.js` that called `processPromiseWrapper('jekyll', ['serve', etc...])`

It would be nice to clear things out of `npm` `scripts`.

PR #522, opened 2017-05-09
Dev + prod logging

### LINKED ISSUE #444: Create an initial "intro" debug log message for localhost
**Library Affected**:
*log-helper.js*

It could improve DX to add in an initial, debug-level log message that's triggered when `logHelper` [detects that we're on `localhost`](https://github.com/GoogleChrome/sw-helpers/blob/master/lib/log-helper.js#L53), with the following (inside the "Extra Information" hidden content):

- An explanation of logs defaulting to "debug" level on `localhost`, with instructions on how to change that level.
- Link to instructions on building a "productionized" version of the libraries with assertions/extra logging removed via dead-code elimination triggered by an environment variable.
- Link to a general getting started guide in our docs

### LINKED ISSUE #487: Warn when using unoptimized sw-lib build in production
**Library Affected**:
*sw-lib*

Some straightforward heuristic like detecting whether or not we're being served from `localhost` might be enough. We could strip out the code that warns about this being an unoptimized build via the process outlined in https://github.com/GoogleChrome/sw-helpers/issues/361#issuecomment-298640521

jeffposnick:
R: @addyosmani @gauntface

Fixes #444, fixes #487 

This adds a new `environment.js` module which exposes a few functions to interrogate the current runtime.

It uses the same `rollup-replace-plugin` method we're using elsewhere to determine whether we're in a production or development bundle.

It adds some logging to the `SWLib` constructor based on various combinations of localhost/prod and dev/prod bundles. Here's an example of the output for localhost + dev bundle combo:

screen shot 2017-05-09 at 4 04 52 pm

Putting startup logging logic into `SWLib`'s constructor seemed like the cleanest approach, since we expect that to be the most typical entry point to the library. Putting the logic into a code path executed by the lower-level, standalone modules seemed like it would just confuse developers who were only using a smaller piece of the framework.

### DIFF
--- a/lib/environment.js
+++ b/lib/environment.js
@@ -1,0 +1,57 @@
+/*
+ Copyright 2016 Google Inc. All Rights Reserved.
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+/**
+ * @private
+ * @return {boolean} True, if we're running in the service worker global scope.
+ * False otherwise.
+ */
+function isServiceWorkerGlobalScope() {
+  return ('ServiceWorkerGlobalScope' in self &&
+          self instanceof ServiceWorkerGlobalScope);
+}
+
+/**
+ * @private
+ * @return {boolean} True, if we're running a development bundle.
+ * False otherwise.
+ */
+function isDevBuild() {
+  // The string `BUILD_TARGET` will be replaced during the build process.
+  return `BUILD_TARGET` === `dev`;
+}
+
+/**
+ * @private
+ * @return {boolean} True, if we're running on localhost or the equivalent IP
+ * address. False otherwise.
+ */
+function isLocalhost() {
+  return Boolean(
+    location.hostname === 'localhost' ||
+    // [::1] is the IPv6 localhost address.
+    location.hostname === '[::1]' ||
+    // 127.0.0.1/8 is considered localhost for IPv4.
+    location.hostname.match(
+      /^127(?:\.(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)){3}$/
+    )
+  );
+}
+
+export default {
+  isDevBuild,
+  isLocalhost,
+  isServiceWorkerGlobalScope,
+};
--- a/lib/environment.js
+++ b/lib/environment.js
@@ -29,8 +29,8 @@ function isServiceWorkerGlobalScope() {
  * False otherwise.
  */
 function isDevBuild() {
-  // The string `BUILD_TARGET` will be replaced during the build process.
-  return `BUILD_TARGET` === `dev`;
+  // `BUILD_PROCESS_REPLACE::BUILD_TARGET` is replaced during the build process.
+  return `BUILD_PROCESS_REPLACE::BUILD_TARGET` === `dev`;
 }
 
 /**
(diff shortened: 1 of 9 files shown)


lib/log-helper.js
```diff
@@ -50,8 +51,9 @@ class LogHelper {
    * LogHelper constructor.
    */
   constructor() {
-    this._defaultLogLevel = location.hostname === 'localhost' ?
-      self.goog.LOG_LEVEL.debug : self.goog.LOG_LEVEL.none;
+    this._defaultLogLevel = environment.isDevBuild() ?
+      self.goog.LOG_LEVEL.debug :
+      self.goog.LOG_LEVEL.warn;
```
jeffposnick: Defaulting to `warn` rather than `none` seemed safer for the production bundle.

addyosmani: I'm happy with us defaulting to `warn` here.

lib/environment.js
```diff
@@ -0,0 +1,57 @@
...
+}
+
+/**
+ * @private
+ * @return {boolean} True, if we're running a development bundle.
+ * False otherwise.
+ */
+function isDevBuild() {
+  // The string `BUILD_TARGET` will be replaced during the build process.
+  return `BUILD_TARGET` === `dev`;
```
gauntface: Nit: Could you add some kind of prefix along the lines of BUILD_PROCESS_REPLACE::BUILD_TARGET

Just so it's more obvious what is happening here.

packages/sw-lib/src/lib/sw-lib.js
```diff
@@ -55,8 +55,37 @@ class SWLib {
...
+      throw ErrorFactory.createError('not-in-sw');
+    }
+
+    if (environment.isDevBuild()) {
+      if (environment.isLocalhost()) {
+        // If this is a dev bundle on localhost, print a welcome message.
+        logHelper.debug({
+          message: 'Welcome to Workbox!',
+          data: {
+            '📖': 'https://googlechrome.github.io/sw-helpers/',
```
gauntface: Could you add a message before this?

"Find our guides and documentation at <url>"

packages/sw-lib/src/lib/sw-lib.js
```diff
@@ -55,8 +55,37 @@ class SWLib {
...
+            '❓': 'https://stackoverflow.com/questions/ask?tags=workbox',
```
gauntface: Same as above:

"If you have any questions, please raise them on stackoverflow with the workbox tag: <URL>"

packages/sw-lib/src/lib/sw-lib.js
```diff
@@ -55,8 +55,37 @@ class SWLib {
...
+            '🐛': 'https://github.com/GoogleChrome/sw-helpers/issues/new',
```
gauntface: "Found a bug? Let us know about it on Github: <url>"

packages/sw-lib/src/lib/sw-lib.js
```diff
@@ -55,8 +55,37 @@ class SWLib {
...
+          },
+        });
+      } else {
+        // If this is a dev bundle not on localhost, recommend the prod bundle.
+        logHelper.warn(`This appears to be a production server. Please switch
```
gauntface: Would be could to add a link here - worth raising an issue to track? I fear people will get caught out by this to be honest :(

packages/sw-lib/src/lib/sw-lib.js
```diff
@@ -55,8 +55,37 @@ class SWLib {
...
+          to the smaller, optimized production build of Workbox.`);
+      }
+    } else {
+      // If this is a prod bundle on localhost, recommend the dev bundle.
+      if (environment.isLocalhost()) {
+        logHelper.warn(`This appears to be a development server. Additional
```
gauntface: I'm not crazy about spamming people with this message. There isn't anything inherently wrong with using prod on localhost.

packages/sw-lib/src/lib/sw-lib.js
jeffposnick: When we have a good place to link to I'll modify the message. I'll track in an issue.

jeffposnick:
Here's the updated log display for dev/localhost:

screen shot 2017-05-10 at 1 40 34 pm

PR #528, opened 2017-05-10
Adding checks for dependencies and marking as required

### LINKED ISSUE #527: sw-build + sw-cli fs-extra needs to be an actual dependency
current versioning is broken because it's only dev and it's actually required by sw-cli and sw-build. Need to alter dependency test to prevent dev deps.

gauntface:
R: @jeffposnick @addyosmani

Fixes #527

fs-extra is required but marked as devDep. Added some more tests to dependency checks unit tests to guard against this in the future.

### DIFF
--- a/packages/sw-build/test/node/dependency-check.js
+++ b/packages/sw-build/test/node/dependency-check.js
@@ -40,7 +40,7 @@ describe('Test Dependencies', function() {
     // this module itself. So dependencies are checked above and devDependencies
     // can be put in top level.
     const pkg = require('../../package.json');
-    if (pkg.devDependencies && pkg.devDependencies > 0) {
+    if (pkg.devDependencies && Object.keys(pkg.devDependencies) > 0) {
       throw new Error('No devDependencies in this module.');
     }
   });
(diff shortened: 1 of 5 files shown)


packages/sw-build/test/node/dependency-check.js
```diff
@@ -26,11 +26,22 @@ describe('Test Dependencies', function() {
...
     });
   });
+
+  it('should have no devDependencies', function() {
+    // This test exists because there have been a number of situations where
+    // dependencies have been used from the top level project and NOT from
+    // this module itself. So dependencies are checked above and devDependencies
+    // can be put in top level.
+    const pkg = require('../../package.json');
+    if (pkg.devDependencies && pkg.devDependencies > 0) {
```
jeffposnick: `Object.keys(pkg.devDependencies).length > 0`

packages/sw-cli/test/node/dependency-check.js
```diff
@@ -32,4 +32,15 @@ describe('Test Dependencies', function() {
...
     });
   });
+
+  it('should have no devDependencies', function() {
+    // This test exists because there have been a number of situations where
+    // dependencies have been used from the top level project and NOT from
+    // this module itself. So dependencies are checked above and devDependencies
+    // can be put in top level.
+    const pkg = require('../../package.json');
+    if (pkg.devDependencies && pkg.devDependencies > 0) {
```
jeffposnick: `Object.keys(pkg.devDependencies).length > 0`

PR #529, opened 2017-05-10
Workbox renaming

gauntface:
R: @jeffposnick @addyosmani

Rename all the things!!!

PR #530, opened 2017-05-11
Use a TEST_BUNDLE env variable to toggle between testing dev & prod

### LINKED ISSUE #524: Run test suite against prod bundles (in addition to dev)
**Library Affected**:
*all*

Our server always responds with the `dev` bundles of our scripts: https://github.com/GoogleChrome/sw-helpers/blob/master/utils/server-instance.js#L106

It could be a good idea to support running the test suite with the `prod` bundles as well, possibly controlled by a command-line flag.

The wrinkle here is that a number of tests specifically check whether a method call throws an exception with a specific `name` property, and those names vary between the `dev` and `prod` bundles. (This is because the `prod` bundles don't include `error-stack-parser`, so the names can't be inferred.) So we would also want a way to disable those tests when we're in `prod` testing mode.

This is related to #523, as we would want to test the `prod` bundles to ensure that there are no regressions when switching between minifiers.

jeffposnick:
R: @addyosmani @gauntface

Fixes #524 

This uses an environment variable, `TEST_BUNDLE`, to toggle between serving the `dev` and `prod` bundles in our test server. Other approaches, like trying to append a search parameter to the URL passed to `importScripts()`, seemed like it would make it harder to toggle between the bundle types without modifying code that's checked in.

This also fixes a few of the tests that previously relied on knowing the exact assertion name that caused an error to be throw, since that additional context wasn't available in the `prod` bundle. We're just using `assertion-failed` as the error name across all environments now.

I understand that I'll probably have to fix merge conflicts once #529 goes in, but I wanted to get this in for review and see if there were any major complaints.

I am not changing the CI or release process right now to automatically run the tests against the prod bundles as well. I don't feel great about doubling the time it takes for CI to run, but I can see the value of changing the release process to test against both the `dev` and `prod` bundles in a follow-up PR.


gauntface:
MY gut says we should be testing prod on CI and run dev and prod for publish - thoughts?

lib/assert.js
```diff
@@ -115,7 +116,6 @@ function throwError(message) {
     // the assertion type that failed.
     error.message = `Invalid call to ${stackFrames[2].functionName}() — ` +
       message;
-    error.name = stackFrames[1].functionName.replace(/^Object\./, '');
```
gauntface: Could we move this to name to error.code and then do a check in tests like:

```
if (error.code) {
  error.code.should.equal('something awesome here');
}
```

addyosmani:
> MY gut says we should be testing prod on CI and run dev and prod for publish - thoughts?

Is your concern running dev on CI would introduce too long a wait time? I'm otherwise fine with going in this direction of prod for CI and both for publishing.

lib/assert.js
jeffposnick: `code` isn't a standard property of the `Error` object: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error

What's the motivation behind not using `name`? If we don't set `name`, all of the exceptions will end up with a `name` of `'Error'`.

jeffposnick:
Defaulting to `prod`, and using that for CI, sounds reasonable. I'll add in an extra step to this PR that will also run tests against `dev` prior to publishing.

Issue #536, opened 2017-05-12
Rename SWLib to WorkboxSW


jeffposnick:
The exported `SWLib` class needs to be renamed to `WorkboxSW`, and there are a number of examples and tests that need to pick up that new name.

PR #538, opened 2017-05-12
goog -> workbox, SWLib -> WorkboxSW

### LINKED ISSUE #288: Reassess scoping + naming of classes
**Library Affected**:
Primarily the modules that expose a Plugin interface, but I'm writing about `sw-broadcast-cache-update` for now.

This came up in the context of @gauntface's comment at https://github.com/GoogleChrome/sw-helpers/pull/287#discussion_r105302593, where a change was made to expose two classes, `BroadcastCacheUpdate` and `BroadcastCacheUpdatePlugin`. When consumed via `importScripts()`, these are exposed in the `goog.broadcastUpdate` scope, leading to clumsy `goog.broadcastUpdate.BroadcastCacheUpdatePlugin` usage.

(This is not a problem when using the classes via ES2015 imports.)

Some options:

1. Drop the package-level scoping for classes, leading to `goog.BroadcastCacheUpdate` and `goog.BroadcastCacheUpdatePlugin`. If we did that, we would probably still want to add in that extra level of scoping for the non-class exports, like [`goog.broadcastCacheUpdate.cachedUpdatedMessageType`](https://github.com/GoogleChrome/sw-helpers/blob/master/packages/sw-broadcast-cache-update/src/index.js#L68).

2. Keep the package-level scoping, but simplify the class names, leading to less redundancy. We might have `goog.broadcastUpdate.Base` for the standalone class, and `goog.broadcastUpdate.Plugin` for the subclass that acts as a `RequestWrapper` plugin. One drawback of this approach is degrading the ES2015 import experience, as folks would probably need to alias the exported names to make them less ambiguous (e.g. `import {Plugin as BroadcastCacheUpdatePlugin} from 'sw-broadcast-cache-update';`).

3. ??? I'm sure there are other suggestions.

jeffposnick:
R: @addyosmani @gauntface

Fixes #288 #536 

Hopefully nothing too controversial in here, as it's mainly a find-and-replace job with the symbols we agreed on.

There were a few places where I had to tweak `eslint` and `mocha` globals to account for the new global symbols.


packages/workbox-background-sync/build.js
```diff
@@ -35,7 +35,7 @@ libFiles.push(path.join('lib', 'idb-helper.js'));
 
 const testBuildConfigs = libFiles.reduce((configs, libFile) => {
   const className = upperCamelCase(path.basename(libFile, '.js'));
-  const moduleName = `goog.backgroundSyncQueue.test.${className}`;
+  const moduleName = `workbox.backgroundSyncQueue.test.${className}`;
```
gauntface: Should this be changed from 'backgroundSyncQueue' to 'backgroundSync'?

packages/workbox-background-sync/src/lib/background-sync-queue.js
```diff
@@ -11,7 +11,7 @@ import {getResponse} from './response-manager';
  *
  * @example
  * When you want to push the requests manually
- * let bgQueue = new goog.backgroundSyncQueue.Queue();
+ * let bgQueue = new workbox.backgroundSyncQueue.Queue();
```
gauntface: Same as above shouldn't this be 'backgroundSync'

packages/workbox-google-analytics/build.js
```diff
@@ -25,7 +25,7 @@ const productionBuildConfigs = generateBuildConfigs({
     iife: pkg.main,
   },
   baseDir: __dirname,
-  moduleName: 'goog.offlineGoogleAnalytics',
+  moduleName: 'workbox.offlineGoogleAnalytics',
```
gauntface: The package name for this is now workbox-google-analytics, should this be changed to workbox.googleAnalytics ?

jeffposnick:
I believe @gauntface's comments were addressed, and @addyosmani approved, so I'm going to merge this one.

PR #539, opened 2017-05-13
Don't calculate the relative path twice

### LINKED ISSUE #535: workbox-cli generate:sw issues
**Library Affected**:
*workbox-cli*

I noticed that running `workbox-cli generate:sw`, with `build` as my root, spits out a service worker that still uses the `SWLib` class name, instead of `WorkboxSW`.

I'm also seeing issues with the path it uses for `importScripts()`; it's using

```js
importScripts('../workbox-sw.prod.v0.0.1.js');
```

instead of 

```js
importScripts('./workbox-sw.prod.v0.0.1.js');
```

Here's my directory structure after running `workbox-cli`:

```sh
$ tree
.
├── build
│   ├── index.html
│   ├── sw.js
│   └── workbox-sw.prod.v0.0.1.js
└── workbox-cli-config.json
```

I'm going to try to put together a fix for #517, and I'll fix these issues in the same PR.

jeffposnick:
R: @gauntface

Fixes #535

I'm not sure if it was the switch from `rootDirectory` to `globDirectory`, or my fiddling with [`copy-workbox-sw.js`](https://github.com/GoogleChrome/workbox/blob/1697b830fb2b315d5a60f3e7eef0bad000ee1058/packages/workbox-build/src/lib/utils/copy-workbox-sw.js) that led to the breakage in #535 to begin with.

There's already a `path.relative()` performed at https://github.com/GoogleChrome/workbox/blob/1697b830fb2b315d5a60f3e7eef0bad000ee1058/packages/workbox-build/src/lib/generate-sw.js#L112, and the one I'm removing seems like it was actively causing trouble.

This *seems* to do the right thing for me now, in that when testing locally, basic use cases result in a generated service worker file that has a correct path in `importScripts()`

That being said, I'd really appreciate it if @gauntface could take a close look at what's changing, and confirm my assumptions. I've been poking around at this code without all the full context and I might be missing something important.

### DIFF
--- a/packages/workbox-build/src/lib/templates/sw.js.tmpl
+++ b/packages/workbox-build/src/lib/templates/sw.js.tmpl
@@ -1,4 +1,4 @@
-importScripts('<%= workboxSWPath %>');
+importScripts('<%= wbSWPathRelToSW %>');
 
 /**
  * DO NOT EDIT THE FILE MANIFEST ENTRY
--- a/packages/workbox-build/src/lib/write-sw.js
+++ b/packages/workbox-build/src/lib/write-sw.js
@@ -6,7 +6,7 @@ const template = require('lodash.template');
 const errors = require('./errors');
 
 module.exports =
-  (swSrc, manifestEntries, workboxSWPath, rootDirectory, options) => {
+  (swSrc, manifestEntries, wbSWPathRelToSW, globDirectory, options) => {
   options = options || {};
   try {
     mkdirp.sync(path.dirname(swSrc));
@@ -88,7 +88,7 @@ module.exports =
       }
       return template(templateString)({
         manifestEntries: manifestEntries,
-        workboxSWPath,
+        wbSWPathRelToSW,
         navigateFallback: options.navigateFallback,
         navigateFallbackWhitelist: options.navigateFallbackWhitelist,
         workboxSWOptionsString,
--- a/packages/workbox-build/src/lib/write-sw.js
+++ b/packages/workbox-build/src/lib/write-sw.js
@@ -29,8 +29,6 @@ module.exports =
     });
   })
   .then((templateString) => {
-    const relSwlibPath = path.relative(rootDirectory, workboxSWPath);
-
     const workboxSWOptions = {};
     if (options.cacheId) {
       workboxSWOptions.cacheId = options.cacheId;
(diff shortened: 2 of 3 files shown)


gauntface:
The problem that I have with this PR is that it's changing stuff, but I can't see the original issue being produced (if that makes sense). Ideally this should have shown itself with one of the end-to-end tests. My assumption is that I'm not doing something specific enough / weird enough that you did in your example - can you share a full example that resulted in this bug?

My assumption is that the following is where some of the weirdness is happening:

1. **Could you add this in your PR**: rootDirectory should be renamed to globDirectory here: https://github.com/GoogleChrome/workbox/blob/1697b830fb2b315d5a60f3e7eef0bad000ee1058/packages/workbox-build/src/lib/write-sw.js#L9
1. workbox-sw should be relative to the service worker not the glob directory (which you are removing) and hence the resulting issue.
1. Could you change the `workboxSWPath` variable name in `write-sw.js` to `wbSWPathRelToSW`. I know it's a disgusting name, but it will be extremely clear what it represents.

Ultimately this PR is good and makes sense now that I've noodled on it, but it would be good to safe guard this in the future. Paths for all this stuff has been a source of pain since the start so more tests to ensure correctness would be good.

jeffposnick:
Hey @gauntface—thanks for the feedback. There are steps about reproducing the bug, including what commands I ran and the directory structure, in the underlying issue, https://github.com/GoogleChrome/workbox/issues/535.

As for why the tests were still passing, based on what I could make out, the code that validates the generated service worker ends up ignoring the value that's passed in to `importScripts()`, so it doesn't pick up on the fact that the relative path to `workbox-sw` is incorrect.

Maybe instead of a [no-op for `importScripts`](https://github.com/GoogleChrome/workbox/blob/master/packages/workbox-cli/test/utils/e2e-sw-validator.js#L111) in that validation flow the test can take the string passed in and confirm that it corresponds to a file on the filesystem? If that sounds like the right approach then we could open another issue to track that enhancement.

gauntface:
@jeffposnick I've just commited a unit test to confirm the behavior. This test would fail on the master build. Doesn't on your PR.

Sorry to be a pain on this - only this stuff is fiddly and want to confirm issues are reproducible, captured and not regressed.

PR #540, opened 2017-05-13
Change payload.updatedUrl to payload.url

jeffposnick:
R: @addyosmani @gauntface

After using the cache update `BroadcastChannel` thing more, I realized that the naming `payload.url` is a lot more intuitive than `payload.updatedUrl`.

I might have initially went with `updatedUrl` because I though there might eventually be another `url` field used for something else, but that concern didn't play out.

### DIFF
--- a/packages/workbox-broadcast-cache-update/demo/index.html
+++ b/packages/workbox-broadcast-cache-update/demo/index.html
@@ -16,7 +16,7 @@
 
       const cacheUpdatesChannel = new BroadcastChannel('cache-updates');
       cacheUpdatesChannel.addEventListener('message', event => {
-        console.log(`There was an update to ${event.data.payload.updatedUrl}`);
+        console.log(`There was an update to ${event.data.payload.url}`);
       });
     </script>
   </body>
--- a/packages/workbox-broadcast-cache-update/src/index.js
+++ b/packages/workbox-broadcast-cache-update/src/index.js
@@ -52,7 +52,7 @@
  *
  * const updateChannel = new BroadcastChannel('cache-updates');
  * updateChannel.addEventListener('message', event => {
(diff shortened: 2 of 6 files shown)


gauntface:
Stupid question, why isn't this just event.data.updatedUrl?

jeffposnick:
I had to come up with some format for the message, and went with the [`flux-standard-action`](https://github.com/acdlite/flux-standard-action#example) pattern.

It's not explicitly called out that these are FSAs, though.

gauntface:
Got you. Personally I prefer updatedUrl. 'url' is too generic for me to understand what it could be referring to in this context.

My main fear is, broadcast channel name could be 'my-awesome-msg' and then payload.url doesn't really indicate it's a cache update. updatedUrl at least hints at it. If you were open to adding a msgType field and then having url, I'm +1 this change.

Happy to be over-ruled if anyone else is in favor of this change.

addyosmani:
I had a similar reaction to `url` (what url does this refer to?) when reading through and `updatedUrl` seemed a little more clear.

jeffposnick:
The full message looks like

```
{
  type: 'CACHE_UPDATED',
  meta: '<URL of SW>',
  payload: {
    cacheName: 'the-cache-name',
    updatedUrl: 'https://example.com/'
  }
}
```

(I was a little hard on myself about the lack of documentation; the [JSDocs does actually mention](https://github.com/GoogleChrome/workbox/blob/master/packages/workbox-broadcast-cache-update/src/lib/broadcast-update.js#L28) that this is a FSA-inspired format.)

gauntface:
To be honest, the more I sit on this, the less I like this change - sorry Jeff.

jeffposnick:
No worries. I'll close it and leave things as-is.

PR #544, opened 2017-05-15
Cli params

gauntface:
R: @jeffposnick @addyosmani

This adds a very naive approach for add cli params for swbuild for generate:sw.

Will add manifest in PR once this lands.

Main concern is that some flags, while they'll be past through, I'm not sure the sane way to input them, for example:

```
  --dontCacheBustUrlsMatching /./
  --ignoreUrlParametersMatching [/^example/, /^other/]
  --navigateFallbackWhitelist [/^\/guide\//, /^\/lolz\//]
  --staticFileGlobs ['**/*.{html,js,css}']
  --modifyUrlPrefix {'/styles': '/static/styles','/page': '/pages/page'}
  --runtimeCaching [{urlPattern: /^https:\/\/example\.com\/api/,handler: 'cacheFirst'}]
  --dynamicUrlToDependencies { '/template/url1': ['page-1.html', 'index.html'], '/template/url2': ['page-2.html', 'index.html'],}
```
But is this a common use case / something users of sw-precache actually did?


jeffposnick:
The `sw-precache` CLI attempted to do some massaging of the strings passed in as flags, like wrapping things that needed to be arrays in a `[]`, and in some cases, converting strings to corresponding RegExp objects:

https://github.com/GoogleChrome/sw-precache/blob/master/cli.js#L67

This code is really unpleasant. If the use case worked out really well for developers, I'd say it was worth trying to support it in `workbox-cli`, but command line parsing edge cases were a source of frustration for developers and a maintenance burden.

I think you could get away with supporting an extremely minimal number of parameters passed via the CLI, and make it clear that the robust way to configure `workbox-cli` is to use an external configuration file. I know that we support a JSON config file right now, but adding in support for a JS config file by reading in its `module.exports` would be a nice addition, since there's no native support in JSON for RegExp objects.

I'd see if you could get away with `globDirectory` and `staticFileGlobs` (and maybe a `config` option to point to a configuration file), and then if someone asks for additional ones, evaluate that on a case-by-case basis.

addyosmani:
> I think you could get away with supporting an extremely minimal number of parameters passed via the CLI, and make it clear that the robust way to configure workbox-cli is to use an external configuration file

👍 

> I'd see if you could get away with globDirectory and staticFileGlobs (and maybe a config option to point to a configuration file), and then if someone asks for additional ones, evaluate that on a case-by-case basis.

This sounds reasonable to me. As much as possible, I'd love to avoid us having to maintain input sanitization for what will probably end up being relatively complex structured input. Using config for that will probably be cleaner for both users and code health.

jeffposnick:
Is this ready for review, or are you still experimenting?

gauntface:
If you could give it another review I'd be very greatful

packages/workbox-cli/src/index.js
```diff
@@ -88,31 +88,70 @@ class SWCli {
...
+      config = Object.assign(config, flags);
     })
     .then(() => {
+      const requiredFields = [
+        'globDirectory',
+        'staticFileGlobs',
+        'swDest',
+      ];
+
+      let askQuestions = false;
```
jeffposnick: Optional refactor:

```js
let askQuestions = requiredFelds.some((field) => !config[field]);
```

PR #547, opened 2017-05-15
Add install instructions to JSDocs.

jpmedley:

This seems like the best way to get them into the micro site.

### DIFF
--- a/packages/workbox-background-sync/src/index.js
+++ b/packages/workbox-background-sync/src/index.js
@@ -17,6 +17,8 @@
  * Queues failed requests and uses the Background Sync API to replay those
  * requests at a later time when the network state has changed.
  *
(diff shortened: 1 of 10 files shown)


prateekbh:
@jpmedley  can you check if you are able to generate jsDocs from this?

I was about to open an issue about this as this page appears empty to me
https://workboxjs.org/reference-docs/latest/module-workbox-background-sync.html

jpmedley:
I already did. I visited every page that I touched. Background sync is thin, not because of a bug, but because no one has written it yet. 

I though the install might trigger some discussion so I wanted to submit it as it's own PR. I plan to go through the JSDocs and even them up, but that will be a separte PR.

prateekbh:
cool sure, though every js file in src has its jsDocs

jpmedley:
I mean the stuff in index.js's.

PR #559, opened 2017-05-17
Minor tweaks to cache-expiration-docs

gauntface:
R: @jeffposnick @addyosmani 

@jeffposnick is the updateTimestamp() useful in CacheExpiration class? If so, is there something more useful we can include in the docs?

### DIFF
--- a/packages/workbox-broadcast-cache-update/src/index.js
+++ b/packages/workbox-broadcast-cache-update/src/index.js
@@ -20,7 +20,7 @@
  * a caches entry is updated with a new response, allowing your web app
  * to listen for these updates and react to them.
  *
- * Most develpers will use this module will by instantiating a new
+ * Most develpers will use this module by instantiating a new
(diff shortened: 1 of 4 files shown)


packages/workbox-broadcast-cache-update/src/index.js
```diff
@@ -20,7 +20,7 @@
  * a caches entry is updated with a new response, allowing your web app
  * to listen for these updates and react to them.
  *
- * Most develpers will use this module will by instantiating a new
+ * Most develpers will use this module by instantiating a new
```
addyosmani: developers

jeffposnick:
`updateTimestap()` is needed for folks who choose not to use the `Plugin` and instead use the `CacheExpiration` class directly, i.e. https://github.com/GoogleChrome/workbox-microsite/blob/master/src/_examples/workbox-cache-expiration/sw.js#L15

I can add a commit to your PR with a better explanation.

PR #571, opened 2017-05-19
Webpack fixes

prateekbh:

This fixes the `rootDirectory` keyname, and the make some basic stuff optional for the user.

I would prefer if we can provide some default values of  `swDest` and `globPatterns` in `workbox-build` instead of webpack plugin.


prateekbh:
just an FYI... the current published plugin works well if given all the options, it just it isn't able to provide defaults.

packages/workbox-webpack-plugin/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "workbox-webpack-plugin",
-  "version": "1.0.0",
```
gauntface: Please undo this - the lerna publish process will handle bumping versions.

packages/workbox-webpack-plugin/test/node/test.js
```diff
@@ -61,13 +61,13 @@ describe('Tests for webpack plugin', function() {
 
   it('should mutate config accordin to webpack defaults', () => {
     let swWebpackPlugin = new SwWebpackPlugin({});
-    assert.equal(swWebpackPlugin.getConfig(webpackCompilation).rootDirectory,
+    assert.equal(swWebpackPlugin.getConfig(webpackCompilation).globDirectory,
```
gauntface: How was this test passing before? I would have thought it would fail given how long rootDirectory has been changed. Are you testing the final service worker output?

prateekbh: Actually this being the unit test, i have stubbed the sw-build details.
I'll open an issue and send the PR for end to end test on monday

packages/workbox-webpack-plugin/package.json
prateekbh: done :)

prateekbh:
opened #573 for end 2 end tests

addyosmani:
I talked to @gauntface about this in person the other day. Labelling https://github.com/GoogleChrome/workbox/issues/573 as a P0 follow-up. Going to land the current work here as we can't have folks require providing all of the options to get this working.

PR #587, opened 2017-05-25
Webpack end 2 end tests

### LINKED ISSUE #227: End-to-end test of sw-build's writeFileManifest()
**Library Affected**:
`sw-build`

There are some unit tests to confirm the low-level functionality of `sw-build`'s `writeFileManifest()`, but currently are not any higher level, end-to-end tests, confirming that it will output a valid manifest file in the specified format`.

See https://github.com/GoogleChrome/sw-helpers/pull/225#issuecomment-281494962

prateekbh:

Fixes #227 

Adding end to end tests for webpack plugin

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,5 +9,3 @@ packages/**/LICENSE
 temp
 tmp-*
 /_site/
-packages/workbox-webpack-plugin/**/webpack-assets/sw*
-packages/workbox-webpack-plugin/**/webpack-assets/workbox*
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ packages/**/LICENSE
 temp
 tmp-*
 /_site/
+packages/workbox-webpack-plugin/**/webpack-assets/sw*
+packages/workbox-webpack-plugin/**/webpack-assets/workbox*
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
     "sinon": "^2.2.0",
     "sw-testing-helpers": "^1.0.2",
     "tmp": "0.0.31",
-    "uppercamelcase": "^3.0.0"
+    "uppercamelcase": "^3.0.0",
+    "webpack": "^2.6.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "gulp-rename": "^1.2.2",
     "gulp-spawn-mocha": "^3.1.0",
     "handlebars": "^4.0.6",
+    "html-webpack-plugin": "^2.28.0",
     "jsdoc-strip-async-await": "^0.1.0",
     "lerna": "2.0.0-rc.4",
     "minimist": "^1.2.0",
--- a/packages/workbox-webpack-plugin/test/node/e2etest.js
+++ b/packages/workbox-webpack-plugin/test/node/e2etest.js
@@ -1,12 +1,14 @@
 /* global workbox, expect, describe */
 const assert = require('chai').assert;
-const WorkboxBuild = require('../../../workbox-build/src');
-const proxyquire = require('proxyquire');
 const webpack = require('webpack');
 const fs = require('fs');
 const path = require('path');
-
-let WorkboxWebpackPlugin;
+const validator = require('../../../workbox-cli/test/utils/e2e-sw-validator.js');
+const WorkboxWebpackPlugin = require('../../');
+const mkdirp = require('mkdirp');
+const testServerGen = require('../../../../utils/test-server-generator.js');
+const fsExtra = require('fs-extra');
+const HtmlWebpackPlugin = require('html-webpack-plugin');
 
 function runWebpack(webpackConfig) {
   return new Promise((resolve, reject) => {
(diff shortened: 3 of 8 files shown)


package.json
```diff
@@ -63,6 +63,7 @@
     "sinon": "^2.2.0",
     "sw-testing-helpers": "^1.0.2",
     "tmp": "0.0.31",
-    "uppercamelcase": "^3.0.0"
+    "uppercamelcase": "^3.0.0",
+    "webpack": "^2.6.0"
```
jeffposnick: I *think* that the Lerna best practice is to include this as a `devDependency` in `packages/workbox-webpack-plugin/package.json` instead of in the top-level `package.json`.

packages/workbox-webpack-plugin/test/node/e2etest.js
```diff
@@ -0,0 +1,84 @@
...
+}
+
+describe('Tests for webpack plugin', function() {
+  this.timeout(55000);
+  beforeEach(()=>{
+    // do a proxy require
+    WorkboxWebpackPlugin = proxyquire('../../', {
+      'workbox-build': WorkboxBuild,
+    });
+    new WorkboxWebpackPlugin();
```
jeffposnick: Isn't this a no-op? You're not assigning the newly-constructed instance to anything.

packages/workbox-webpack-plugin/test/node/e2etest.js
```diff
@@ -0,0 +1,84 @@
...
+  });
+
+  it('should generate sw, when `swSrc` is not present', () => {
+    this.timeout(35000);
```
jeffposnick: Why does this test have a timeout of 35 seconds, after you set a timeout of 55 seconds at the beginning of the `describe()` block? And what is the significance of those values—is there any reason why running a Webpack build would legitimately take 55 seconds to complete?

packages/workbox-webpack-plugin/test/node/e2etest.js
```diff
@@ -0,0 +1,84 @@
...
+    });
+  });
+}
+
+describe('Tests for webpack plugin', function() {
+  this.timeout(55000);
+  beforeEach(()=>{
+    // do a proxy require
+    WorkboxWebpackPlugin = proxyquire('../../', {
+      'workbox-build': WorkboxBuild,
```
jeffposnick: What is the effective difference between what would normally be `require`d and what's being `proxyquire`d here? As far as I can tell, you're swapping out `workbox-build` for... the same underlying `workbox-build` module.

packages/workbox-webpack-plugin/test/node/e2etest.js
prateekbh: sorry, removed it

packages/workbox-webpack-plugin/test/node/e2etest.js
prateekbh: this was left unintentionally removed it

packages/workbox-webpack-plugin/test/node/e2etest.js
prateekbh: since workbox-build isn't installed as a dependency, shouldn't this be required from the source file?

package.json
prateekbh: in that case, will that be installed before running tests?

gauntface: It will be on travis yes.

.gitignore
```diff
@@ -9,3 +9,5 @@ packages/**/LICENSE
 temp
 tmp-*
 /_site/
+packages/workbox-webpack-plugin/**/webpack-assets/sw*
```
gauntface: could you use a `tmp-*` file name if you are dynamically creating folders in tests?

package.json
gauntface: No top level is fine here (and preferable here).

the *-build and *-cli have tests to check dependencies and it's much easier if devDep's are in top level and only deps are in the package itself.

packages/workbox-webpack-plugin/test/node/e2etest.js
gauntface: workbox-build will be "npm linked" with the local workbox-build through lerna. So you only need to have workbox-box as a dependency in your package.json and then require as normal.

packages/workbox-webpack-plugin/test/node/e2etest.js
```diff
@@ -0,0 +1,81 @@
...
+    });
+  });
+
+  it('should generate sw, when `swSrc` is not present', () => {
+    const webpackConfig = {
+      entry: {
+        swApp: path.join(__dirname, 'webpack-assets/src.js'),
```
gauntface: Please change to `tmp-*` instead of a static folder.

Please add an after step to remove the directory and it's contents as well. Main advantage is that the directory should be different every test run - reduces risk of a false positive if an old test worked and a new one doesn't.

packages/workbox-webpack-plugin/test/node/e2etest.js
```diff
@@ -0,0 +1,81 @@
...
+      },
+      output: {
+        path: path.join(__dirname, '/webpack-assets/'),
```
gauntface: Please change to tmp-* instead of a static folder.

Please add an after step to remove the directory and it's contents as well. Main advantage is that the directory should be different every test run - reduces risk of a false positive if an old test worked and a new one doesn't.

prateekbh: done :)

PR #590, opened 2017-05-30
More robust runtimeCaching conversion.

### LINKED ISSUE #574: Service Worker generated by Workbox webpack plugin not creating write caches
**Library Affected**:
*workbox-webpack plugin 1.0.0*

**Browser & Platform**:
* "all browsers".*

**Issue Description**:
Options mentioned for runtime Caching in webpack config do not take effect in the generated Service worker.

**Webpack configuration**
 ```
new workboxPlugin({
       globDirectory: DIST_DIR,
       staticFileGlobs: ['**/*.{html,js,css}'],
       swDest: path.join(DIST_DIR, 'sw.js'),
       modifyUrlPrefix: {
                          '/': ''
                        },
        runtimeCaching: [{
                        urlPattern: /bbc-news/,
                        handler: 'networkFirst',
                        options: {
                          cache:{
                          name: 'bbc-cache',
                          maxEnteries: 10,
                          maxAgeSeconds: 300
                          }
                        }
                      },
                      {
                        urlPattern: /cnn/,
                        handler: 'cacheFirst',
                        options: {
                          cache:{
                          name: 'cnn-cache',
                          maxEnteries: 10,
                          maxAgeSeconds: 300
                        }
                        }
                      }]
     })
  ]
```
**Generated Service Worker**
```
const workboxSW = new self.WorkboxSW();
workboxSW.precache(fileManifest);
workboxSW.router.registerRoute(/bbc-news/, workboxSW.strategies.networkFirst({
  "cache": {
    "name": "bbc-cache",
    "maxEnteries": 10,
    "maxAgeSeconds": 300
  }
}));
workboxSW.router.registerRoute(/cnn/, workboxSW.strategies.cacheFirst({
  "cache": {
    "name": "cnn-cache",
    "maxEnteries": 10,
    "maxAgeSeconds": 300
  }
}));

```

The service worker is installed correctly however all request made to http url containing "bbc-news" are not cached. No new cache by name of 'bbc-cache' is created. Same for other specified options.

**Missing 'bbc-cache'**
image

**Http request going to network instead of cache**

image

**SW_Precache**
However when I try this with sw-precache things work fine.
Same runtimeCaching option given to sw-precache. 
_Generated Service Worker
```
toolbox.router.get(/bbc-news/, toolbox.networkFirst, {"cache":{"name":"bbc-cache","maxEnteries":10,"maxAgeSeconds":300}});
toolbox.router.get(/cnn/, toolbox.cacheFirst, {"cache":{"name":"cnn-cache","maxEnteries":10,"maxAgeSeconds":300}});

```
This service worker also have the register, fetch and the toolbox code embedded in it.
image

_

### PR DESCRIPTION
**Pull request by jeffposnick**
R: @addyosmani @gauntface

Fixes #574 

This should give us parity with the full set of options that `runtimeCaching` in `sw-precache` supported.

### DIFF
--- a/packages/workbox-build/src/lib/utils/runtime-caching-converter.js
+++ b/packages/workbox-build/src/lib/utils/runtime-caching-converter.js
@@ -27,8 +27,9 @@ function getOptionsString(options) {
   }
 
   // Everything should be copied to the corresponding new option names at this
-  // point, so delete the old-style `cache` property if it's present.
-  delete effectiveOptions.cache;
+  // point, so set the old-style `cache` property to undefined so that it
+  // doesn't show up in the JSON output.
+  effectiveOptions.cache = undefined;
 
   // JSON.stringify() will automatically omit any properties that are set to
   // undefined values.
--- a/packages/workbox-build/src/lib/utils/runtime-caching-converter.js
+++ b/packages/workbox-build/src/lib/utils/runtime-caching-converter.js
@@ -51,6 +51,9 @@ module.exports = (runtimeCaching) => {
       throw new Error(errors['handler-is-required']);
     }
 
+    // urlPattern might be either a string or a RegExp object.
+    // If it's a string, it needs to be quoted. If it's a RegExp, it should
+    // be used as-is.
     const matcher = typeof entry.urlPattern === 'string' ?
       `'${entry.urlPattern}'` :
       entry.urlPattern;
--- a/packages/workbox-build/src/lib/utils/runtime-caching-converter.js
+++ b/packages/workbox-build/src/lib/utils/runtime-caching-converter.js
@@ -1,0 +1,73 @@
+const errors = require('../errors');
+
+/**
+ * Given a set of options that configures `sw-toolbox`'s behavior, convert it
+ * into a string that would configure equivalent `workbox-sw` behavior.
+ *
+ * @param {Object} options See
+ *        https://googlechrome.github.io/sw-toolbox/api.html#options
+ * @return {String} A JSON string representing the equivalent options.
+ */
+function getOptionsString(options) {
+  const cacheOptions = options.cache || {};
+  // Start with a base of a few properties that need to be renamed, as well
+  // as copying over all the other source properties as-is.
+  const effectiveOptions = Object.assign({
+    cacheName: cacheOptions.name,
+  }, options);
+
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: packages/workbox-build/src/lib/utils/runtime-caching-converter.js [hunk 1]
```diff
@@ -0,0 +1,73 @@
...
+    effectiveOptions.cacheExpiration =
+      Object.assign(effectiveOptions.cacheExpiration || {}, {
+        maxEntries: cacheOptions.maxEntries,
+        maxAgeSeconds: cacheOptions.maxAgeSeconds,
+      });
+  }
+
+  // Everything should be copied to the corresponding new option names at this
+  // point, so delete the old-style `cache` property if it's present.
+  delete effectiveOptions.cache;
```
addyosmani: If this was `effectiveOptions.cache = null` or `undefined`, would it be just as effective for Line 35? `delete` used like this will likely cause V8 deopts.

addyosmani:
Unfortunately AppVeyor is still failing on a few PRs atm. Trying to dig into why.

packages/workbox-build/src/lib/utils/runtime-caching-converter.js
jeffposnick: Setting to `null` will cause the property to appear in the JSON output. Setting to `undefined` will remove the property from the JSON output, so I switched to that.

packages/workbox-build/src/lib/utils/runtime-caching-converter.js
```diff
@@ -0,0 +1,75 @@
...
+
+    if (!entry.urlPattern) {
+      throw new Error(errors['urlPattern-is-required']);
+    }
+
+    if (!entry.handler) {
+      throw new Error(errors['handler-is-required']);
+    }
+
+    const matcher = typeof entry.urlPattern === 'string' ?
```
gauntface: could you add a comment that this is to account for regexs. Reading this for first time I was a little lost.

jeffposnick:
The old code didn't work with Express-style strings for routes, or when cache names/expiration options were provided.

The new unit tests I added should cover both of those scenarios, but if you feel like e2e tests are in order, too, I wouldn't stop you.

gauntface:
Thats fine, if you are happy I'm happy ;)

PR #592, opened 2017-05-31
Moves to the current sinon syntax.

jeffposnick:
R: @addyosmani @gauntface

I ran https://github.com/hurrymaplelad/sinon-codemod against our codebase to automatically update to the currently-supported syntax.

This fixes the

```
sinon.stub(obj, 'meth', fn) is deprecated and will be removed from the public API in a future version of sinon.
 Use stub(obj, 'meth').callsFake(fn).
 Codemod available at https://github.com/hurrymaplelad/sinon-codemod
```

noise in our test output.


jeffposnick:
As per our discussion, merging despite the AppVeyor failures, which @gauntface said he'd look into.

PR #594, opened 2017-06-02
Looking at minor size optimisations.

gauntface:
R: @jeffposnick @addyosmani 

This PR does the following:
- Adds Babili back into the dev build.
- Moves assert and environment into individual exports (this helps reduce file size slightly)
- Tidies / simplifies the log helpers (and bring some basic manual tests back in for assert and log helper)
- Removes the error-stack-parser. While the logs are helpful it's size isn't worth it (looking at the logs from assert example is /test/assert.html the error stack trace does go back up to the source, so I'd rather explore how we keep that error useful (i.e. ensure maps are working or configuring babili for dev not to manipulate function names).

for workbox-precaching prod goes from 23KB -> 21.4KB. Dev goes from 86.9KB -> 21.4KB

### DIFF
--- a/gulp-tasks/build.js
+++ b/gulp-tasks/build.js
@@ -21,6 +21,8 @@ const gulp = require('gulp');
 const path = require('path');
 const runSequence = require('run-sequence');
 const {taskHarness, buildJSBundle, lernaWrapper} = require('../utils/build');
+const commonjs = require('rollup-plugin-commonjs');
+const resolve = require('rollup-plugin-node-resolve');
 
 const printHeading = (heading) => {
   process.stdout.write(chalk.inverse(`  ⚒  ${heading}  `));
@@ -71,15 +73,40 @@ const updateVersionedBundles = (projectPath) => {
 };
 
 gulp.task('build:shared', () => {
+  const basePlugins = [
+    resolve({
+      jsnext: true,
+      main: true,
+      browser: true,
+    }),
+    commonjs(),
+  ];
+
   return buildJSBundle({
     rollupConfig: {
       entry: path.join(__dirname, '..', 'lib', 'log-helper.js'),
-      format: 'umd',
-      moduleName: 'goog.logHelper',
       plugins: [],
     },
-    buildPath: 'build/log-helper.js',
-    projectDir: path.join(__dirname, '..'),
+    writeConfig: {
+      sourceMap: true,
+      format: 'umd',
+      moduleName: 'goog.logHelper',
+      dest: path.join(__dirname, '..', 'build', 'log-helper.js'),
+    },
+  })
+  .then(() => {
+    return buildJSBundle({
+      rollupConfig: {
+        entry: path.join(__dirname, '..', 'lib', 'assert.js'),
+        plugins: basePlugins,
+      },
+      writeConfig: {
+        sourceMap: true,
+        format: 'umd',
+        moduleName: 'goog.assert',
+        dest: path.join(__dirname, '..', 'build', 'assert.js'),
+      },
+    });
   });
 });
 
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -17,7 +17,7 @@
 
 import ErrorStackParser from 'error-stack-parser';
 
-function atLeastOne(object) {
+export function atLeastOne(object) {
   const parameters = Object.keys(object);
   if (!parameters.some((parameter) => object[parameter] !== undefined)) {
     throwError('Please set at least one of the following parameters: ' +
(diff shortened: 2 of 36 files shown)


addyosmani:
> Adds Babili back into the dev build.

Do we have a link to a PR or issue where we described why Babili was originally taken out? If memory serves it was breaking something but I can't recollect if that was indeed the reason.

> Moves assert and environment into individual exports (this helps reduce file size slightly)

sgtm

> Tidies / simplifies the log helpers (and bring some basic manual tests back in for assert and log helper)

These appear to be in a reasonable state.

> Removes the error-stack-parser. While the logs are helpful it's size isn't worth it (looking at the logs from assert example is /test/assert.html the error stack trace does go back up to the source, so I'd rather explore how we keep that error useful (i.e. ensure maps are working or configuring babili for dev not to manipulate function names).

I'm personally fine with this.

jeffposnick:
> > Adds Babili back into the dev build.

> Do we have a link to a PR or issue where we described why Babili was originally taken out? If memory serves it was breaking something but I can't recollect if that was indeed the reason.

If I recall correctly, we've never had `babili` turned on for the dev build. The dev build has, generally speaking, focused on minimal modifications to the original source code, other than the Rollup bundling needed to convert the individual ES2015 modules into `iife` bundles that could be consumed via `importScripts()`.

Who the target audience for this modified dev build? What's the primary differentiator between dev and prod if this PR is merged?

gauntface:
I always though dev build was a build that had "optional" stuff (i.e. the dev build would contain helpful logs, additional messages and assertions etc.

Prod was the stripped down version that removed these optional pieces.

Being close to the source code, in my opinion, offers nothing to the developer.

jeffposnick:
The "dev" vs. "prod" distinction has been fluid over the life of the project.

Up until last month, we were just using `min` and non-`min` naming, and the code was semantically the same. The extra `babili` step was the only thing that was different between the output.

Then with https://github.com/GoogleChrome/workbox/pull/500 and https://github.com/GoogleChrome/workbox/pull/522, we started adding in semantic differences, and went with the `dev` and `prod` naming, instead of the `min` and the implicitly not-`min`.

In my mind, "dev" is the bundle that maximizes logging, context and general debuggability, without regard for bundle size or execution speed. And "prod" is what we ship as being optimized for size (and speed).

> Just for the record, I'm not saying we drop the dev build at the moment, but I would like to know what wins developers get from replacing dev build with a prod build that uses NOOPs.

The "dev" build currently offers additional context for assertion failures, via `error-stack-parser`. It allows the assertion failures to refer back up the stack and identify the function that was called which lead to the assertion failure. In other words, `error-stack-parser` isn't something that helps unmangle symbols—it provides a cross-browser way of walking up the error stack.

gauntface:
@jeffposnick yeah I was thinking we'd going down the path of both 500 and 522 where it makes sense.

The logging from the error-stack-parser was better, but I'm honestly not convinced it was that much better given that browsers can call out the path. I'd rather invest time in the source map to make sure it comes out with a valid stack trace that way.

But that doesn't change that dev and prod should be treated with that same purpose, strip out unnecessary code and behaving more appropriately for the environment so I want to keep them for now so we can see what wins we can get between the two, but if dev ends up being the same say as prod and the only difference is minification, I'm not sure why we'd continue with two.

jeffposnick:
FWIW, here's what the difference is currently between the prod build without `error-stack-parser` (first entry) and the dev build with `error-stack-parser` (second entry):

screen shot 2017-06-06 at 4 31 23 pm

With `error-stack-parser`, we're able to prefix the underlying assertion failure with the additional context "Invalid call to new NetworkFirst() —", making it clearer to a developer who might not know how to interpret the signal of the underlying function call from the noise of the stack trace.

jeffposnick:
👍  with these changes. Feel free to merge when you're ready.

philipwalton:
> but if dev ends up being the same say as prod and the only difference is minification, I'm not sure why we'd continue with two.

Debugging minified code, even with source maps, is a pretty big pain in devtools because the symbols you type into the console won't necessarily match the symbols you see in your code view.

Here's an example. As you can see, referencing the local `regExp` and `handler` variables in the console throws:

screen shot 2017-07-20 at 3 15 00 pm

philipwalton:
And here are two devtools bugs related to this issue. Sounds like it's not going to be fixed any time soon:
https://crbug.com/672444
https://crbug.com/327092

gauntface:
@philipwalton added a note to V3 proposal doc on this to **not** minify dev build if everyone agrees the approach is simple enough to switch between the two builds.

PR #601, opened 2017-06-06
Bg sync replay fix

### LINKED ISSUE #597: Background sync is not rescheduled if replaying fails
**Library Affected**:
*workbox-background-sync*

**Browser & Platform**:
*all browsers*

*Background sync is not rescheduled if replaying fails.*

Sync is only triggered when a new request is pushed to the queue. However, one of the use cases (the most important one IMO) of background sync is deferring requests to when network state changes from offline to online. For that purpose, browsers reschedule syncs that have failed with exponential backoff. As far as I know, the condition for a failing sync event is that it must receive a rejected promise from `event.waitUntil`. However `workbox-background-sync` catches all failed promises.

prateekbh:
R: @jeffposnick 

Fixes #597

After this PR if during the entire replay process any request fails, we'll be passing a rejected promise to `event.waitUntil` of `sync` event, so that browser can replay the failed requests

### DIFF
--- a/packages/workbox-background-sync/src/lib/request-manager.js
+++ b/packages/workbox-background-sync/src/lib/request-manager.js
@@ -49,8 +49,9 @@ class RequestManager {
    * @memberOf RequestManager
    * @private
    */
-  replayRequests() {
-    return this._queue.queue.reduce((promise, hash) => {
+  async replayRequests() {
+    let allRequestsStatus = [];
+    await this._queue.queue.reduce((promise, hash) => {
       return promise
         .then(async (item) => {
           const reqData = await this._queue.getRequestFromQueue({hash});
@@ -66,6 +67,7 @@ class RequestManager {
           return fetch(request)
             .then((response)=>{
               if(!response.ok) {
+                allRequestsStatus.push(Promise.reject());
                 return Promise.resolve();
               } else {
                 // not blocking on putResponse.
@@ -77,14 +79,17 @@ class RequestManager {
                 });
                 this._globalCallbacks.onResponse
                   && this._globalCallbacks.onResponse(hash, response);
+                allRequestsStatus.push(Promise.resolve());
               }
             })
             .catch((err)=>{
+              allRequestsStatus.push(Promise.reject());
               this._globalCallbacks.onRetryFailure
                 && this._globalCallbacks.onRetryFailure(hash, err);
             });
         });
     }, Promise.resolve());
+    return Promise.all(allRequestsStatus);
   }
 }
 
--- a/packages/workbox-background-sync/src/lib/request-manager.js
+++ b/packages/workbox-background-sync/src/lib/request-manager.js
@@ -57,7 +57,7 @@ class RequestManager {
         idbRequestObject: reqData.request});
       const response = await fetch(request);
       if(!response.ok) {
-        return Promise.reject();
+        return Promise.reject(response.status);
       } else {
         // not blocking on putResponse.
         putResponse({
@@ -72,7 +72,7 @@ class RequestManager {
     } catch(err) {
       this._globalCallbacks.onRetryFailure
         && this._globalCallbacks.onRetryFailure(hash, err);
-        return Promise.reject(err);
+      return Promise.reject(err);
     }
   }
 
@@ -85,15 +85,17 @@ class RequestManager {
    * @private
    */
   async replayRequests() {
-    let failedItems = 0;
+    let failedItems = [];
     for (let hash of this._queue.queue) {
       try {
         await this.replayRequest(hash);
-      } catch (e) {
+      } catch (err) {
+        failedItems.push(err);
         failedItems++;
       }
     }
-    return failedItems > 0 ? Promise.reject() : Promise.resolve();
+    return failedItems.length > 0 ?
+      Promise.reject(failedItems) : Promise.resolve();
   }
 }
 
--- a/packages/workbox-background-sync/src/lib/request-manager.js
+++ b/packages/workbox-background-sync/src/lib/request-manager.js
@@ -63,7 +63,7 @@ class RequestManager {
       });
       const response = await fetch(request);
       if(!response.ok) {
-        return Promise.reject(response.status);
+        return Promise.reject(response);
       } else {
         // not blocking on putResponse.
         putResponse({
@@ -76,7 +76,7 @@ class RequestManager {
           this._globalCallbacks.onResponse(hash, response);
       }
     } catch(err) {
-      return Promise.reject(err);
+      return Promise.reject(err.stack);
     }
   }
 
@@ -90,7 +90,7 @@ class RequestManager {
    * @private
    */
   async replayRequests() {
-    let failedItems = [];
+    const failedItems = [];
     for (let hash of this._queue.queue) {
       try {
         await this.replayRequest(hash);
--- a/packages/workbox-background-sync/src/lib/request-manager.js
+++ b/packages/workbox-background-sync/src/lib/request-manager.js
@@ -42,54 +42,58 @@ class RequestManager {
   }
 
   /**
+   * function to play one single request
+   * given its hash
+   * @param {String=} hash
+   * @return {Promise}
+   *
+   * @memberOf RequestManager
+   * @private
+   */
+  async replayRequest(hash) {
+    try {
+      const reqData = await this._queue.getRequestFromQueue({hash});
+      const request = await getFetchableRequest({
+        idbRequestObject: reqData.request});
+      const response = await fetch(request);
+      if(!response.ok) {
+        return Promise.reject();
+      } else {
+        // not blocking on putResponse.
+        putResponse({
+          hash,
+          idbObject: reqData,
+          response: response.clone(),
+          idbQDb: this._queue.idbQDb,
+        });
+        this._globalCallbacks.onResponse
+          && this._globalCallbacks.onResponse(hash, response);
+      }
+    } catch(err) {
+      this._globalCallbacks.onRetryFailure
+        && this._globalCallbacks.onRetryFailure(hash, err);
+        return Promise.reject(err);
+    }
+  }
+
+  /**
    * function to start playing requests
    * in sequence
-   * @return {void}
+   * @return {Promise}
    *
    * @memberOf RequestManager
    * @private
    */
   async replayRequests() {
-    let allRequestsStatus = [];
-    await this._queue.queue.reduce((promise, hash) => {
-      return promise
-        .then(async (item) => {
-          const reqData = await this._queue.getRequestFromQueue({hash});
-          if(reqData.response) {
-            // check if request is not played already
-            return;
-          }
-
-          const request = await getFetchableRequest({
-            idbRequestObject: reqData.request,
-          });
-
-          return fetch(request)
-            .then((response)=>{
-              if(!response.ok) {
-                allRequestsStatus.push(Promise.reject());
-                return Promise.resolve();
-              } else {
-                // not blocking on putResponse.
-                putResponse({
-                  hash,
-                  idbObject: reqData,
-                  response: response.clone(),
-                  idbQDb: this._queue.idbQDb,
-                });
-                this._globalCallbacks.onResponse
-                  && this._globalCallbacks.onResponse(hash, response);
-                allRequestsStatus.push(Promise.resolve());
-              }
-            })
-            .catch((err)=>{
-              allRequestsStatus.push(Promise.reject());
-              this._globalCallbacks.onRetryFailure
-                && this._globalCallbacks.onRetryFailure(hash, err);
-            });
-        });
-    }, Promise.resolve());
-    return Promise.all(allRequestsStatus);
+    let failedItems = 0;
+    for (let hash of this._queue.queue) {
+      try {
+        await this.replayRequest(hash);
+      } catch (e) {
+        failedItems++;
+      }
+    }
+    return failedItems > 0 ? Promise.reject() : Promise.resolve();
   }
 }
 
--- a/packages/workbox-background-sync/src/lib/request-manager.js
+++ b/packages/workbox-background-sync/src/lib/request-manager.js
@@ -55,8 +55,12 @@ class RequestManager {
   async replayRequest(hash) {
     try {
       const reqData = await this._queue.getRequestFromQueue({hash});
+      if(reqData.response) {
+        return;
+      }
       const request = await getFetchableRequest({
-        idbRequestObject: reqData.request});
+        idbRequestObject: reqData.request,
+      });
       const response = await fetch(request);
       if(!response.ok) {
         return Promise.reject(response.status);
--- a/packages/workbox-background-sync/src/lib/request-manager.js
+++ b/packages/workbox-background-sync/src/lib/request-manager.js
@@ -76,7 +76,7 @@ class RequestManager {
           this._globalCallbacks.onResponse(hash, response);
       }
     } catch(err) {
-      return Promise.reject(err.stack);
+      return Promise.reject(err);
     }
   }
 
@@ -86,7 +86,6 @@ class RequestManager {
    * @return {Promise} Resolves if all requests are played successfully,
    * rejects if any of the request fails during the replay
    *
-   * @memberOf RequestManager
    * @private
    */
   async replayRequests() {
--- a/packages/workbox-background-sync/src/lib/request-manager.js
+++ b/packages/workbox-background-sync/src/lib/request-manager.js
@@ -42,10 +42,12 @@ class RequestManager {
   }
 
   /**
-   * function to play one single request
-   * given its hash
-   * @param {String=} hash
-   * @return {Promise}
+   * function to play one single request given its hash
+   *
+   * @param {String} hash
+   *
+   * @return {Promise} Resolves if the request corresponding to the hash is
+   * played successfully, rejects if it fails during the replay
    *
    * @memberOf RequestManager
    * @private
@@ -66,20 +68,19 @@ class RequestManager {
           response: response.clone(),
           idbQDb: this._queue.idbQDb,
         });
-        this._globalCallbacks.onResponse
-          && this._globalCallbacks.onResponse(hash, response);
+        if (this._globalCallbacks.onResponse)
+          this._globalCallbacks.onResponse(hash, response);
       }
     } catch(err) {
-      this._globalCallbacks.onRetryFailure
-        && this._globalCallbacks.onRetryFailure(hash, err);
       return Promise.reject(err);
     }
   }
 
   /**
-   * function to start playing requests
-   * in sequence
-   * @return {Promise}
+   * function to start playing requests in sequence
+   *
+   * @return {Promise} Resolves if all requests are played successfully,
+   * rejects if any of the request fails during the replay
    *
    * @memberOf RequestManager
    * @private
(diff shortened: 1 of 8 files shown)


packages/workbox-background-sync/src/lib/request-manager.js
```diff
@@ -42,49 +42,58 @@ class RequestManager {
   }
 
   /**
+   * function to play one single request
+   * given its hash
+   * @param {String=} hash
```
jeffposnick: Are you using `{String=}` to indicate that this is an optional parameter? If so, we've been using `[hash]` elsewhere in the docs (the first option listed in http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values)

packages/workbox-background-sync/src/lib/request-manager.js
```diff
@@ -42,49 +42,58 @@ class RequestManager {
...
+   * @return {Promise}
```
jeffposnick: Can you provide some context about what the promise represents? I.e. "Resolves when..., rejects when..."

packages/workbox-background-sync/src/lib/request-manager.js
```diff
@@ -42,49 +42,58 @@ class RequestManager {
...
+        return Promise.reject();
+      } else {
+        // not blocking on putResponse.
+        putResponse({
+          hash,
+          idbObject: reqData,
+          response: response.clone(),
+          idbQDb: this._queue.idbQDb,
+        });
+        this._globalCallbacks.onResponse
```
jeffposnick: Can you switch to `if(this._globalCallbacks.onResponse) {...}`. We've been using that syntax rather than `&&` elsewhere in the project.

packages/workbox-background-sync/src/lib/request-manager.js
```diff
@@ -42,49 +42,60 @@ class RequestManager {
...
+          && this._globalCallbacks.onResponse(hash, response);
+      }
+    } catch(err) {
+      this._globalCallbacks.onRetryFailure
```
jeffposnick: Similar to the previous comment about preferring `if()` to `&&`.

packages/workbox-background-sync/src/lib/request-manager.js
```diff
@@ -42,49 +42,60 @@ class RequestManager {
...
+   * @private
+   */
+  async replayRequest(hash) {
+    try {
+      const reqData = await this._queue.getRequestFromQueue({hash});
+      const request = await getFetchableRequest({
+        idbRequestObject: reqData.request});
+      const response = await fetch(request);
+      if(!response.ok) {
+        return Promise.reject(response.status);
```
jeffposnick: Just to confirm: you're not invoking the `this._globalCallbacks.onRetryFailure` callback here. Is that intentional?

packages/workbox-background-sync/src/lib/request-manager.js
```diff
@@ -42,49 +42,60 @@ class RequestManager {
...
-        });
-    }, Promise.resolve());
+  async replayRequests() {
+    let failedItems = [];
+    for (let hash of this._queue.queue) {
+      try {
+        await this.replayRequest(hash);
+      } catch (err) {
+        failedItems.push(err);
+        failedItems++;
```
jeffposnick: I'm not clear as to why you'd increment an array. Is this a typo?

packages/workbox-background-sync/src/lib/request-manager.js
```diff
@@ -42,49 +42,60 @@ class RequestManager {
...
-            })
-            .catch((err)=>{
-              this._globalCallbacks.onRetryFailure
-                && this._globalCallbacks.onRetryFailure(hash, err);
-            });
-        });
-    }, Promise.resolve());
+  async replayRequests() {
+    let failedItems = [];
+    for (let hash of this._queue.queue) {
```
jeffposnick: Does `this._globalCallbacks.onRetryFailure` end up adding in the failed entries to the end of `this._queue.queue`? If so, it seems like you may end up processing the same entry more than once while iterating through `this._queue.queue` if one of the earlier replay requests fails.

packages/workbox-background-sync/src/lib/request-manager.js
```diff
@@ -42,49 +42,60 @@ class RequestManager {
...
+        && this._globalCallbacks.onRetryFailure(hash, err);
+      return Promise.reject(err);
+    }
+  }
+
+  /**
    * function to start playing requests
    * in sequence
-   * @return {void}
+   * @return {Promise}
```
jeffposnick: Can you provide some context as to what this Promise represents?

packages/workbox-background-sync/src/lib/request-manager.js
gauntface: +1

packages/workbox-background-sync/src/lib/request-manager.js
```diff
@@ -42,49 +42,60 @@ class RequestManager {
...
+    let failedItems = [];
+    for (let hash of this._queue.queue) {
+      try {
+        await this.replayRequest(hash);
+      } catch (err) {
+        failedItems.push(err);
+        failedItems++;
+      }
+    }
+    return failedItems.length > 0 ?
```
gauntface: Since this is an async function, can you just throw an error?

gauntface:
My only comment on this PR, this feels like a fairly big change, without any test changes or new tests to come with it.

It sounds like something was broken, which meant it wasn't covered by a test, is there anyway we can add a test for it?

prateekbh:
@jeffposnick @gauntface yeah this code never had a test backing it up, I will be adding new tests for this change.

packages/workbox-background-sync/src/lib/request-manager.js
prateekbh: I am not sure about this, so rejecting the promise like this gives me an opportunity to pass the entire array of failes Requests and their causes. just in case developer decides to do something with it.

packages/workbox-background-sync/src/lib/request-manager.js
prateekbh: sorry, yeah typo

packages/workbox-background-sync/src/lib/request-manager.js
prateekbh: done

packages/workbox-background-sync/src/lib/request-manager.js
prateekbh: done

packages/workbox-background-sync/src/lib/request-manager.js
prateekbh: done

packages/workbox-background-sync/src/lib/request-manager.js
prateekbh: fixed this, moved this to parent function to catch all kind of rejections at one place

packages/workbox-background-sync/src/lib/request-manager.js
prateekbh: done

packages/workbox-background-sync/src/lib/request-manager.js
prateekbh: nah, its not an optional comment, removed the `=`

packages/workbox-background-sync/src/lib/request-manager.js
prateekbh: I am a little unclear as of what do you mean by end up adding in  failed entries?

packages/workbox-background-sync/src/lib/request-manager.js
```diff
@@ -49,8 +49,9 @@ class RequestManager {
    * @memberOf RequestManager
    * @private
    */
-  replayRequests() {
-    return this._queue.queue.reduce((promise, hash) => {
+  async replayRequests() {
+    let allRequestsStatus = [];
```
anilanar: This can be a `const` actually.

prateekbh: done

prateekbh:
@jeffposnick @gauntface the new test for failing promise is added

anilanar:
I wonder... Can we get confirmation from Chrome team about this? @addyosmani


On 7. June 2017 at 16:34:41, Prateek Bhatnagar (notifications@github.com<mailto:notifications@github.com>) wrote:

@jeffposnick<https://github.com/jeffposnick> @gauntface<https://github.com/gauntface> the new test for failing promise is added

prateekbh:
@anilanar confirmation about?

anilanar:
About what does passing a rejected promise do to the sync event. That blog post I shared in the original issue is too old, perhaps things have changed since then?

prateekbh:
@anilanar i just confirmed it is working like that, will re confirm before submitting

packages/workbox-background-sync/src/lib/request-manager.js
jeffposnick: The values in `this._queue.queue` are set somewhere—I'm not sure where. I'm asking whether they're set inside of `this._globalCallbacks.onRetryFailure`, and if they are, how they're being set.

What I'm worried about is the potential for an infinite loop while you iterate over `this._queue.queue`. I could imagine this happening if you're adding a new entry to the end of `this._queue.queue` every time there's a failure inside the loop.

prateekbh: oh ok I guess what you are asking, so `this._globalCallbacks.onRetryFailure` is actually just a callback from user land so that we can notify the user that hey request X has been failed,

user can then show a notification, do a console log or anything he/she wish for. This has nothing to do with requests getting stored, that portion is completely de-coupled from this.

bg-sync calls `this._globalCallbacks.onRetryFailure` every-time it fails replaying the request and calls `this._globalCallbacks.onResponse` whenever we successfully fetch a request

jeffposnick: So what's the code that removes items from `this._queue.queue`? Don't you have to remove them within one of the callbacks? And not remove them if there's a failure?

prateekbh: Shouldn't the request remain in the queue if it is failed, so that it can be replayed at later point?
And for the removal part, I remove them with a separate `. cleanupQueue` method, which is also auto-called while initialization.

packages/workbox-background-sync/src/lib/request-manager.js
```diff
@@ -42,49 +42,66 @@ class RequestManager {
...
-              }
-            })
-            .catch((err)=>{
-              this._globalCallbacks.onRetryFailure
-                && this._globalCallbacks.onRetryFailure(hash, err);
-            });
+  async replayRequest(hash) {
+    try {
+      const reqData = await this._queue.getRequestFromQueue({hash});
+      if(reqData.response) {
```
prateekbh: @jeffposnick this prevents replaying an already played request

jeffposnick: Gotcha.

jeffposnick:
The code changes look 👍 

There are various [CI test failures](https://travis-ci.org/GoogleChrome/workbox/jobs/240378531). I think some of them have to do with node no longer being happy with `throw 'Some string';`. You now need to do `throw Error('Some string');`

packages/workbox-background-sync/src/lib/request-manager.js
```diff
@@ -42,49 +42,66 @@ class RequestManager {
...
+      } else {
+        // not blocking on putResponse.
+        putResponse({
+          hash,
+          idbObject: reqData,
+          response: response.clone(),
+          idbQDb: this._queue.idbQDb,
         });
-    }, Promise.resolve());
+        if (this._globalCallbacks.onResponse)
```
gauntface: Not to be address in this PR, but I can't help but feel like this should be an event that is emitted rather than a callback function.

packages/workbox-background-sync/src/lib/request-manager.js
```diff
@@ -42,49 +42,66 @@ class RequestManager {
...
+          this._globalCallbacks.onResponse(hash, response);
+      }
+    } catch(err) {
+      return Promise.reject(err.stack);
```
gauntface: Why are you returning just the stack here? Why not the whole error which will include the message.

packages/workbox-background-sync/src/lib/request-manager.js
```diff
@@ -42,49 +42,66 @@ class RequestManager {
...
+    }
+  }
+
+  /**
+   * function to start playing requests in sequence
```
gauntface: Could you beef up this message up a little as it'll be in the reference docs.

i.e. When you need to replay all of the currently queued requests, call this method, it'll do XYZ, ....

packages/workbox-background-sync/src/lib/request-manager.js
```diff
@@ -42,49 +42,66 @@ class RequestManager {
...
+   *
+   * @return {Promise} Resolves if all requests are played successfully,
+   * rejects if any of the request fails during the replay
+   *
+   * @memberOf RequestManager
```
gauntface: This member of isn't needed for individual functions. The function is part of the class and whatever the class is a memberof is the important bit.

All of this is unnecessary because you've marked this class as private.

packages/workbox-background-sync/test/browser/background-sync-queue.js
```diff
@@ -71,17 +74,29 @@ describe('background sync queue test', () => {
 
   it('check push proxy', async () => {
     const currentLen = backgroundSyncQueue._queue.queue.length;
-    await backgroundSyncQueue.pushIntoQueue({request: new Request('http://lipsum.com')});
+    await backgroundSyncQueue.pushIntoQueue({request: new Request('http://localhost:3000/__echo/counter')});
+    await backgroundSyncQueue.replayRequests();
```
gauntface: Why are you replaying requests here but not before? My hunch from the code is that you are only checking the queue length, so whats the point of replaying?

packages/workbox-background-sync/test/browser/background-sync-queue.js
```diff
@@ -71,17 +74,29 @@ describe('background sync queue test', () => {
...
     chai.assert.equal(backgroundSyncQueue._queue.queue.length, currentLen + 1);
```
gauntface: The 'currentLen + 1' always concerned me a bit here, is this test reliant on other tests passing (i.e. if one test is incorrect and the length is wrong or currentLen is not updated, will this test fail?

packages/workbox-background-sync/test/browser/background-sync-queue.js
```diff
@@ -71,17 +74,29 @@ describe('background sync queue test', () => {
...
-    await backgroundSyncQueue.pushIntoQueue({request: new Request('https://jsonplaceholder.typicode.com/posts/2')});
+    await backgroundSyncQueue.pushIntoQueue({request: new Request('http://localhost:3000/__echo/counter')});
+    await backgroundSyncQueue.pushIntoQueue({request: new Request('http://localhost:3000/__echo/counter')});
     await backgroundSyncQueue.replayRequests();
     chai.assert.equal(responseAchieved, 2);
   });
 
+  it('check replay failure with rejected promise', async function() {
+    await backgroundSyncQueue.pushIntoQueue({request: new Request('http://localhost:3000/__echo/counter')});
+    await backgroundSyncQueue.pushIntoQueue({request: new Request('http://localhost:3002/__echo/counter')});
```
gauntface: It might be better to have an end point that just always fails. It's not clear from this test that you are referencing a failing URL.

Would '/__test/404/' endpoint work for your needs?

addyosmani:
Let's aim to address Matt's feedback and make sure Travis is happy before we merge this.

packages/workbox-background-sync/src/lib/request-manager.js
prateekbh: done!

packages/workbox-background-sync/src/lib/request-manager.js
prateekbh: done!

packages/workbox-background-sync/test/browser/background-sync-queue.js
prateekbh: cleaned this up, made `bgSyncQueue` to be reinitiated before every test, so that replay is not needed anymore.

packages/workbox-background-sync/test/browser/background-sync-queue.js
prateekbh: cleaned this up, made `bgSyncQueue` to be reinitiated before every test, so that no test is interdependent on each other

packages/workbox-background-sync/test/browser/background-sync-queue.js
prateekbh: done, thaks its much cleaner now :)

packages/workbox-background-sync/src/lib/request-manager.js
prateekbh: I tried adding some description though this one is internally called. The one available to the user to call is in `background-sync-queue`

packages/workbox-background-sync/test/browser/background-sync-queue.js
```diff
@@ -15,6 +15,10 @@
 /* global chai, workbox */
 
 'use strict';
+const testServerGen = require('../../../../utils/test-server-generator.js');
```
gauntface: This test makes no sense I'm afraid.

test files under **test/browser/ are unit tests that run in the browser. There is no require, there is no need to start a server and this code just fails :(

prateekbh: sorry, for the goof up, I managed to fix all the test cases. lets sync up on this sometime today

packages/workbox-background-sync/test/browser/background-sync-queue.js
```diff
@@ -38,20 +37,33 @@ describe('background sync queue test', () => {
     onRetryFail: onRetryFail,
   };
 
-  const backgroundSyncQueue
-    = new workbox.backgroundSync.test.BackgroundSyncQueue({
+  let backgroundSyncQueue;
+
+  beforeEach(async ()=>{
```
gauntface: Please use `function() {}` here. This is because mocha does some weird binding stuff that means fat arrows have unexpected results if you tried to do this.timeout or this.retries.

packages/workbox-background-sync/test/browser/background-sync-queue.js
```diff
@@ -38,20 +37,33 @@ describe('background sync queue test', () => {
...
+    responseAchieved = 0;
+    backgroundSyncQueue = new workbox.backgroundSync.test.BackgroundSyncQueue({
       maxRetentionTime: MAX_AGE,
-      queueName: QUEUE_NAME,
       callbacks: CALLBACKS,
     });
+  });
+
+  afterEach(async ()=> {
```
gauntface: Same as above - switch to function.

packages/workbox-background-sync/test/browser/background-sync-queue.js
prateekbh: cool

packages/workbox-background-sync/src/lib/request-queue.js
```diff
@@ -140,7 +140,8 @@ class RequestQueue {
     assert.isType({hash}, 'string');
 
     if(this._queue.includes(hash)) {
-      return await this._idbQDb.get(hash);
+      const req = await this._idbQDb.get(hash);
```
prateekbh: @gauntface just wanted to discuss this small change.
actually, this small change fixed everything failing on firefox, do you have any context?

gauntface: nope nothing from me. cc @jeffposnick as FYI

prateekbh:
@jeffposnick can you approve if all looks good?

PR #609, opened 2017-06-07
Default cli ignores

### LINKED ISSUE #326: sw-cli creates a manifest including node_modules/
**Library Affected**:
sw-cli

**Browser & Platform**:
Node v7.7.2

If I specify my root dir as the project path, sw-cli includes js/html etc files from node_modules/. I feel like this might not be desirable or could be an option.

gauntface:

Fixes #326

Ignore node_modules by default


gauntface:
@addyosmani which docs would you like this referenced in?

addyosmani:
As discussed this would be good to add a note to the microsite for https://workboxjs.org/get-started/npm-script.

PR #629, opened 2017-06-14
WorkboxError

gauntface:
R: @jeffposnick @addyosmani 

This is what I had in mind to manage errors in Workbox. It moves the error messages into a class which has a few nice features:

1. Errors will be displayed as 'WorkboxErrors' in the console.
1. Messages can be swapped out between dev and prod (hopefully reducing file size).
1. We can inline additional information in the message.
1. Add's consistency format of extra values

Major con is that is requires all messages to be accessible via this one WorkboxError class. We can organise the error-message-factory to pull in seperate files from packages or have it all in one file - not sure.

Let me know what you think.


gulpfile.js
```diff
@@ -16,8 +16,20 @@
 /* eslint-disable no-console, valid-jsdoc */
 
 const minimist = require('minimist');
+const fs = require('fs');
+const path = require('path');
 
 const options = minimist(process.argv.slice(2));
+
+if (options.project) {
+  // Ensure the project is invalid before running tasks
```
jeffposnick: invalid -> valid

lib/error-message-factory-no-op.js
```diff
@@ -0,0 +1,5 @@
+const errorMessageFactory = (errorCode, extras) => {
+  return `An error was thrown by workbox with error code: '${errorCode}' and extras: '${JSON.stringify(extras)}'.`;
```
jeffposnick: How about:

```js
const errorMessageFactory = (code, context) => {
  return `Workbox encountered an error:\n${JSON.stringify({code, context}, null, 2)}`;
};
```

If `context` isn't set, that will leave out the `context` key from the JSON output.

lib/error-message-factory.js
```diff
@@ -0,0 +1,74 @@
...
+};
+
+const errorMessageFactory = (errorCode, extras) => {
+  const errorDetails = ERRORS[errorCode];
+  if (!errorDetails) {
+    return `An error was thrown by workbox with error code: '${errorCode}'.`;
+  }
+
+  const errorMessage = errorDetails.message;
+  if (typeof errorMessage === 'function') {
```
jeffposnick: Can you switch this to:

```js
const errorMessage = typeof errorDetails.message === 'function' ?
  errorDetails.message(extras) :
  errorDetails.message;
return errorMessage.replace(/\s+/g, ' ');
```

Then, instead of having to concatenate multiple strings when defining our error messages, we can just use a single multi-line string, and the newlines and extra spaces will be stripped out. It's a little thing, but it makes it more pleasant to deal with error strings while respecting the line length limits.

gauntface:
@jeffposnick I bumped the chrome driver and it seems to have fixed the tests

PR #633, opened 2017-06-14
Strips the hash from a URL in a couple of places

### LINKED ISSUE #488: Account for hash fragments now being exposed in request.url
**Library Affected**:
*all*

**Browser & Platform**:
*Google Chrome v59+, Firefox 52+*

The service worker specification [was changed](https://github.com/w3c/ServiceWorker/issues/854) to start exposing hash fragments as part of the `Request.url` that's passed in to the `fetch` event handler. Firefox matches the spec as of release 52, and Chrome will match it as of release 59.

We may need to adjust the logic that determines whether a URL is being precached to account for this change, since it relies on string comparisons. The runtime caching logic shouldn't need to change, because it relies on `cache.match()`, which will automatically ignore the hash fragment.

(The corresponding issue for `sw-precache` is https://github.com/GoogleChrome/sw-precache/issues/290.)

jeffposnick:
R: @addyosmani @gauntface

The discussion in #488 talked about putting as much of the logic as possible inside reusable plugins; some of the logic that was changed in this PR is inside the `CacheExpiration` plugin.

However, there's no concept of plugins for routing (only for request handlers), and we needed to adjust the routing logic for precached URLs inside of `workbox-sw`. So that part is just done inline.

Fixes #488


canhphong:
@wibblymat hi hi

packages/workbox-sw/src/lib/workbox-sw.js
```diff
@@ -305,6 +305,9 @@ class WorkboxSW {
 
     const route = new Route({
       match: ({url}) => {
+        // See https://github.com/GoogleChrome/workbox/issues/488
```
gauntface: Can you put the actual problem description here?

jeffposnick:
PTAL.

PR #637, opened 2017-06-16
Include the URL's search when matching the NavigationRoute white/blacklist

### LINKED ISSUE #631: Allow NavigationRoute whitelisting/blacklisting to work with search parameters
**Library Affected**:
*workbox-routing*

**Browser & Platform**:
n/a

It would be nice if `NavigationRoute`'s `whitelist` and `blacklist` options supported matching on `url.pathname + url.search` instead of just on `url.pathname`. Performing server-based redirects based on a query parameter seems _somewhat_ common, and such a redirect breaks if we can't specify that an app shell should _not_ be served for URLs containing that query parameter.

Take the following route that we've implemented:

```js
workboxSW.router.registerNavigationRoute(APP_SHELL_ROUTE, {
  whitelist: [
    /^\/$/,
  ],
});
```

On the page, a user can change languages. The language picker redirects the browser to the current URL but with `locale=de` appended to the search query. Normally the server would intercept these requests and 302 redirect to `de.domain.com`. With a service worker installed and the above `NavigationRoute` registered, this redirect breaks; the user is served the app shell as usual and no redirect occurs.

It makes sense to me that we should be able to do the following:

```js
workboxSW.router.registerNavigationRoute(APP_SHELL_ROUTE, {
  whitelist: [
    /^\/$/,
  ],
  blacklist: [
    /[&?]locale=/,
  ],
});
```

Thoughts?

jeffposnick:
R: @addyosmani @gauntface
CC: @mikefowler

Fixes #631

### DIFF
--- a/packages/workbox-routing/src/lib/navigation-route.js
+++ b/packages/workbox-routing/src/lib/navigation-route.js
@@ -30,8 +30,10 @@ import logHelper from '../../../../lib/log-helper';
  * both lists are provided, and there's a navigation to a URL which matches
  * both, then the blacklist will take precedence and the request will not be
  * matched by this route. The regular expressions in `whitelist` and `blacklist`
- * are matched against the [`pathname`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLHyperlinkElementUtils/pathname)
- * portion of the requested URL.
+ * are matched against the concatenated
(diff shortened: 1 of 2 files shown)


addyosmani:
The Travis failures I'm seeing appear related to 614 landing rather than more recent work. Reviewing independent of that.

Issue #636, opened 2017-06-16
registerRoutes don't accept subclass of Route


jmtt89:
**Library Affected**:
*workbox-sw, workbox-routing*

**Browser & Platform**:
*all browsers.*

**Issue or Feature Request Description**:
registerRoutes don't accept subclass of Route

```js
import WorkboxSW from 'workbox-sw';
import {RegExpRoute, ExpressRoute} from 'workbox-routing';
import {QueuePlugin} from 'workbox-background-sync';

const workboxSW = new WorkboxSW();

workboxSW.precache([...]);

//backgroundSync
let bgQueue = new QueuePlugin();

// Network Stategies
const networkOnly = workboxSW.strategies.networkOnly({
  plugins: [bgQueue]
});
const cacheFirst = workboxSW.strategies.cacheFirst();

//RuntimeCache
workboxSW.router.registerRoutes({
  routes: [
    new RegExpRoute({
      regExp: new RegExp('^https://fonts.googleapis.com/*'),
      handler: cacheFirst
    }),
    new RegExpRoute({
      regExp: new RegExp('^https://code.jquery.com/*'),
      handler: cacheFirst
    }),
    //Device Endpoints
    new ExpressRoute({
      path: '/my/custom/route',
      handler: networkOnly
    })
  ]
});
```

*When reporting bugs, please include relevant JavaScript Console logs and links to public URLs at which the issue could be reproduced.*

** Console Output **
```shell
Uncaught assertion-failed: The 'routes' parameter should be an array containing one or more 'Route' instances.
```

** Function throw error **
```js
function isArrayOfClass(a, b) {
  const c = Object.keys(a).pop(), d = `The '${c}' parameter should be an array containing one or more '${b.name}' instances.`;
  Array.isArray(a[c]) || throwError(d); // Here all Good
  for (let e of a[c])
    e instanceof b || throwError(d); //This throw error
}
```

```shell
e ->
        RegExpRoute$1 {handler: CacheFirst, method: "GET", match: function}
          handler: CacheFirst
          match:({url:e})=> {…}
          method: "GET"
          __proto__:Route$1
```

```shell
b
       class Route{constructor({match:a,handler:b,method:c}={})
       {this.handler=normalizeHandler(b),assert.isType({match:a},'function'),this.match=a,c?
       (assert.isOneOf({method:c},validMethods),this.method=c):this.met…
```

```shell
e instanceof b
        false
```

jmtt89:
Extra update:
[...]

Issue #385, opened 2017-03-27
Failed assert.isInstance() checks due to namespacing of bundled classes


jeffposnick:
**Library Affected**:
*sw-routing*

There are some places in `sw-routing` where `assert.isInstance()` is used to check whether a parameter is of the correct class, e.g. [`assert.isInstance({route}, Route);`](https://github.com/GoogleChrome/sw-helpers/blob/master/packages/sw-routing/src/lib/router.js#L214).

I've seen situations where this fails when the class in question (here it's `Route`, but it might apply to other classes as well) is defined in multiple top-level modules, like in both `sw-routing` and `sw-lib`. Certain bundlers will rename what would otherwise be conflicting class names, so you end up with `Route$1` and `Route$2`, and as far as the assertion can tell, those are two different classes.

The real fix is probably to promote a way of telling the bundler that `Route` from each of those modules is actually the same symbol.

A more tactical fix might be to change the `assert.isInstance()` check so that it looked at the string version of the class names, and only required equality up until you reach a `'$'` character.

A brute-force fix would be to drop the assertion entirely.

addyosmani:
>  Certain bundlers will rename what would otherwise be conflicting class names, so you end up with Route$1 and Route$2, and as far as the assertion can tell, those are two different classes.

Is this happening with Rollup?

jeffposnick:
I'm fairly sure Rollup is adding in the namespaces, yes. But it's doing so in response to us passing in sources that it can't safely assume are the same underlying classes.

PR #639, opened 2017-06-19
Makes it easier to use custom Routes within WorkboxSW.

jeffposnick:
R: @addyosmani @gauntface
CC: @jmtt89 @mikefowler

This is something of a straw-man PR, and feedback is welcome.

This partially addresses #385 and #636, but I'm not sure that I'd call it a "fix", so I'm inclined to leave them open.

Still, it should make it easier to accomplish what @jmtt89 was trying to do, by letting you use a method other than `'GET'` when calling `workboxSW.router.registerRoute()`.

It also provides an escape hatch granting greater flexibility by letting developers import and use `Route` objects created from the `workbox.routing` namespace, by relaxing the assertions to just check whether the object passed to `registerRoute()` has the properties that a `Route` should have, not necessarily whether it's a subclass of a specific definition of a `Route` object.

This PR doesn't attempt to add in a factory method for creating new `Route` objects exposed via`workboxSW.router`, although that's an idea that's been discussed in the past and might still be on the table.

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -91,6 +91,36 @@ export function isArrayOfClass(object, expectedClass) {
   }
 }
 
+export function isInterface(object, expectedProperties) {
+  const parameter = Object.keys(object).pop();
+  const message = `The '${parameter}' parameter should be an object that has
+    '${expectedProperties}' properties.`;
+
+  for (let property of expectedProperties) {
+    if (!(property in object[parameter])) {
+      throwError(message);
+    }
+  }
+}
+
+export function isArrayOfInterface(object, expectedProperties) {
+  const parameter = Object.keys(object).pop();
+  const message = `The '${parameter}' parameter should be an array containing
+    objects that each have '${expectedProperties}' properties.`;
+
+  if (!Array.isArray(object[parameter])) {
+    throwError(message);
+  }
+
+  for (let item of object[parameter]) {
+    for (let property of expectedProperties) {
+      if (!(property in item)) {
+        throwError(message);
+      }
+    }
+  }
+}
+
 export function isValue(object, expectedValue) {
   const parameter = Object.keys(object).pop();
   const actualValue = object[parameter];
--- a/packages/workbox-routing/src/lib/router.js
+++ b/packages/workbox-routing/src/lib/router.js
@@ -13,10 +13,11 @@
  limitations under the License.
 */
 
-import Route from './route';
-import {isArrayOfClass, isInstance} from '../../../../lib/assert';
+import {isArrayOfInterface, isInterface} from '../../../../lib/assert';
 import logHelper from '../../../../lib/log-helper.js';
 import normalizeHandler from './normalize-handler';
+
+const ROUTE_INTERFACE = ['match', 'handler', 'method'];
 
 /**
  * The Router takes one or more [Routes]{@link Route} and registers a [`fetch`
@@ -220,7 +221,10 @@ class Router {
    * routes to register.
    */
   registerRoutes({routes} = {}) {
-    isArrayOfClass({routes}, Route);
+    // Rather than explicitly checking for Route subclasses, check to ensure
+    // that the object exposes the interface we need.
+    // See https://github.com/GoogleChrome/workbox/issues/385
+    isArrayOfInterface({routes}, ROUTE_INTERFACE);
 
     for (let route of routes) {
       if (!this._routes.has(route.method)) {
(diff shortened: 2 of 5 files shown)


mikefowler:
@jeffposnick this seems like a reasonable solution to me. 

What was the original/current thinking on not including `Route` in the public interface for an instance of `WorkboxSW`? I'll admit, it was confusing to see the class in the source code but have no way of accessing it.

jeffposnick:
I don't remember the original thinking—@gauntface and I had some back and forth about it, but I'm not even sure which side each of us took 😄 

It's slightly awkward to expose the `Route` class directly, because `workboxSW.router` is an automatically-constructed instance of a subclass of `Router`.

We could expose a `workboxSW.router.createRoute()` factory method (though that would preclude subclassing), or we could expose the `Route` class statically as `workboxSW.Route`.

I guess my vote is for the factory method, but that can be implemented separately from this PR.

lib/assert.js
```diff
@@ -91,6 +91,36 @@ export function isArrayOfClass(object, expectedClass) {
   }
 }
 
+export function isInterface(object, expectedProperties) {
+  const parameter = Object.keys(object).pop();
+  const message = `The '${parameter}' parameter should be an object that has
```
gauntface: Should you be creating this inside the for loop to update the message for the failing property?

lib/assert.js
```diff
@@ -91,6 +91,36 @@ export function isArrayOfClass(object, expectedClass) {
...
+    }
+  }
+}
+
+export function isArrayOfInterface(object, expectedProperties) {
+  const parameter = Object.keys(object).pop();
+  const message = `The '${parameter}' parameter should be an array containing
+    objects that each have '${expectedProperties}' properties.`;
+
+  if (!Array.isArray(object[parameter])) {
```
gauntface: not 100% of this check, the function name 'isArrayOfInterface' suggests that object should be an array and this asserts each entry in the array has an explicit interface.

lib/assert.js
```diff
@@ -91,6 +91,36 @@ export function isArrayOfClass(object, expectedClass) {
...
+    throwError(message);
+  }
+
+  for (let item of object[parameter]) {
+    for (let property of expectedProperties) {
+      if (!(property in item)) {
+        throwError(message);
```
gauntface: same as above, the message should be updated to handle the missing property

packages/workbox-routing/src/lib/router.js
```diff
@@ -13,11 +13,12 @@
  limitations under the License.
 */
 
-import Route from './route';
-import {isArrayOfClass, isInstance} from '../../../../lib/assert';
+import {isArrayOfInterface, isInterface} from '../../../../lib/assert';
 import logHelper from '../../../../lib/log-helper.js';
 import normalizeHandler from './normalize-handler';
 
+const ROUTE_INTERFACE = ['match', 'handler', 'method'];
```
gauntface: I'm not a huge fan of having this manually defined and it being defined away from the class it represents.

lib/assert.js
```diff
@@ -91,6 +91,36 @@ export function isArrayOfClass(object, expectedClass) {
...
 }
 
+export function isInterface(object, expectedProperties) {
+  const parameter = Object.keys(object).pop();
+  const message = `The '${parameter}' parameter should be an object that has
+    '${expectedProperties}' properties.`;
+
+  for (let property of expectedProperties) {
+    if (!(property in object[parameter])) {
+      throwError(message);
```
gauntface: Would it be worth adding the missing parameter to the message?

gauntface:
In the early versions we originally exposed it, but developer feedback was that it caused confusion have it exposed so closely to router.

packages/workbox-sw/src/lib/router.js
```diff
@@ -97,9 +99,9 @@ class Router extends SWRoutingRouter {
       if (capture.length === 0) {
         throw ErrorFactory.createError('empty-express-string');
       }
-      route = new ExpressRoute({path: capture, handler});
+      route = new ExpressRoute({path: capture, handler, method});
```
gauntface: why did you add the method?

gauntface:
Discussed with @jeffposnick , we can move the capture argument to accept a function, this will serve as the 'match' input for a Route that will be constructed behinds the scenes.

Sorry for this wrinkle.There is a largely re-think we should probably have about how we expose these values.

mikefowler:
@gauntface, to clarify, your proposal is to allow for the following? 

```js
workboxSW.router.registerRoute(
  
  // Capture
  ({ url, event }) => {
    return true;
  },
  
  // Handler
  ({ event }) => {
    return caches.match(event.request);
  },

);

jeffposnick:
That's correct, @mikefowler. I'm going to create a fresh PR that takes that approach.

PR #644, opened 2017-06-22
Cleaner way to create generic Route instances via registerRoute()

### LINKED ISSUE #385: Failed assert.isInstance() checks due to namespacing of bundled classes

### LINKED ISSUE #636: registerRoutes don't accept subclass of Route

jeffposnick:
R: @addyosmani @gauntface
CC: @jmtt89 @mikefowler

This is an alternative to https://github.com/GoogleChrome/workbox/pull/639, as discussed in that PR.

Fixes #385 and fixes #636

### DIFF
--- a/packages/workbox-precaching/src/lib/controllers/base-cache-manager.js
+++ b/packages/workbox-precaching/src/lib/controllers/base-cache-manager.js
@@ -280,7 +280,7 @@ class BaseCacheManager {
    * @return {Promise} Returns a Promise that resolves once the work is done.
    */
   _onEntryDeleted(url) {
-    throw ErrorFactory.createError('should-override');
+    throw new WorkboxError('requires-overriding');
   }
 }
 
--- a/packages/workbox-sw/src/lib/error-factory.js
+++ b/packages/workbox-sw/src/lib/error-factory.js
@@ -17,8 +17,8 @@ import ErrorFactory from '../../../../lib/error-factory';
 
 const errors = {
   'not-in-sw': 'workbox-sw must be loaded in your service worker file.',
-  'unsupported-route-type': 'Routes must be either a express style route ' +
-    'string, a Regex to capture request URLs or a Route instance.',
+  'unsupported-route-type': 'The first parameter to registerRoute() should be' +
+    ' either an Express-style path string, a RegExp, or a function.',
   'empty-express-string': 'The Express style route string must have some ' +
     'characters, an empty string is invalid.',
   'bad-revisioned-cache-list': `The 'precache()' method expects` +
--- a/packages/workbox-sw/src/lib/router.js
+++ b/packages/workbox-sw/src/lib/router.js
@@ -70,22 +70,26 @@ class Router extends SWRoutingRouter {
   }
 
   /**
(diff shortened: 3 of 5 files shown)


gauntface:
Just needs lint fix

PR #645, opened 2017-06-22
Removes packages/*/demos

### LINKED ISSUE #642: Deprecate `demo` directories?
**Library Affected**:
Multiple.

Several packages have `demo/` directories in them, which aren't generally maintained, discoverable, or useful for end users.

https://workboxjs.org/examples/ is, for the most part, the replacement for those `demo/` directories.

The code in `demo/` does sometimes come in handy when someone developing in this repo needs to test out a local change against `gulp serve`, but that use case is probably outweighed compared to the confusion it causes.

Related: #641, and @philipwalton was just complaining about something similar.

Unless there are objections (@gauntface?) I'll put together a PR to remove them.

jeffposnick:
R: @addyosmani @gauntface

Fixes #642


gauntface:
LGTM

PR #648, opened 2017-06-23
Use requestWrapper.fetch() instead of fetch() in BSQ

### LINKED ISSUE #646: Background sync should add the ability to modify request prior to retrying
I started looking into how to update the `workbox-google-analytics` library to use `background-sync`, and I discovered that there's no way to modify the request before it retries.

For the google analytics hit to work it needs to update the `qt` param to the difference between the current time and the original hit time. It also (for consistency with the existing API), needs to let developers set [param overrides](https://github.com/GoogleChrome/workbox/blob/v1.0.0/packages/workbox-google-analytics/src/lib/replay-queued-requests.js#L30-L35) as well as modify the existing params via a [hitFilter function](https://github.com/GoogleChrome/workbox/blob/v1.0.0/packages/workbox-google-analytics/src/lib/replay-queued-requests.js#L36-L41).

I see two options to make this work, and I wanted to get feedback before making any changes to the background sync code:

1. Add a new callback to the [`workbox.backgroundSync.Queue`](https://github.com/GoogleChrome/workbox/blob/v1.0.0/packages/workbox-background-sync/src/lib/background-sync-queue.js#L38-L39) so it can modify the request after it pulls it from the IDB cache and before it sends it.

2. Update the [`workbox.backgroundSync.RequestQueue`](https://github.com/GoogleChrome/workbox/blob/master/packages/workbox-background-sync/src/lib/request-queue.js#L84-L93) class to support queuing both plain `Request` objects as well as `Request` objects with a corresponding `RequestWrapper` object, then we could use a [`RequestWrapper`](https://github.com/GoogleChrome/workbox/blob/master/packages/workbox-runtime-caching/src/lib/request-wrapper.js) plugin's [`requestWillFetch`](https://github.com/GoogleChrome/workbox/blob/master/packages/workbox-runtime-caching/src/lib/request-wrapper.js#L185-L205) lifecycle method to modify the request prior to sending it.

I think both options are fine, but I slightly prefer option #2, if for no other reason than it seems wrong to duplicate behavior we already have in `RequestWrapper`. That being said, option #1 is probably much easier to implement.

What do others think? @jeffposnick @prateekbh

jeffposnick:
R: @philipwalton @prateekbh @addyosmani @gauntface

Fixes #646 

This also adds some missing JSDocs for the `requestWillFetch` lifecycle callback, and includes some cosmetic reformatting to adhere to our current style.


prateekbh:
Thanks for all the sanitization @jeffposnick, I hope the failing tests shouldn't be a big deal.

packages/workbox-background-sync/test/browser/request-manager.js
```diff
@@ -58,11 +64,31 @@ describe('request-manager test', () => {
...
+      callbacks,
+    });
     await backgroundSyncQueue.pushIntoQueue({request: new Request('/__echo/counter')});
     await backgroundSyncQueue.pushIntoQueue({request: new Request('/__echo/counter')});
     await backgroundSyncQueue._requestManager.replayRequests();
     chai.assert.equal(responseAchieved, 2);
   });
+
+  it(`will fetch() the request returned by the RequestWrapper's requestWillFetch plugin`, async function() {
+    const stub = sinon.stub(self, 'fetch').returns(new Response());
```
gauntface: does window.sinon fix this? Otherwise may need a .eslintrc file to work.

jeffposnick: Sorry, I missed the linting error. I've added `sinon` to the eslint globals config.

Issue #650, opened 2017-06-23
Output workbox-sw lib as part of build process


kaycebasques:
**Library Affected**:
workbox-build, workbox-sw

**Browser & Platform**:
All

**Issue or Feature Request Description**:
I had to "register a route" to handle pretty URLs, so I'm using `injectManifest` in a custom SW. Right now for my build process, I've installed workbox-sw as a node module, and just copy the prod version from the `importScripts` directory.

I'd prefer if workbox-build just output the workbox-sw library for me when I inject the manifest. Is there a way to do that?

```
function buildSw() {
  return wb.injectManifest({
    globDirectory: './_site/',
    globPatterns: ['**\/*'],
    swSrc: './src/sw.js',
    swDest: './_site/sw.js',
  }).then(() => {
    gutil.log('Build Manifest generated.');
  }).catch((error) => {
    gutil.log('Error: ' + error);
  });
}
```

gauntface:
At the moment there isn't. Open to suggestions to what you think would fit your workflow here.

This is for me, the biggest problem with workbox at the moment, getting the library file into your project and / or service worker.

I personally would like to see us encouraging CDN usage instead of pulling the file from NPM, this way you can just import the script with a known CDN and good to go.

gauntface:
This is also a problem with the webpack plugin (see #653)

gauntface:
We plan to add a helper method to copy over either all or some of the workbox files to make this easier.

gauntface:
The copyLibraries helper method and CDN by default should solve this.

PR #654, opened 2017-06-28
Copy workbox-sw lib as part of "manifest injection" build process

### LINKED ISSUE #653: copy node_modules scripts in webpack plugin
If I have some `importScripts('node_modules/.../workbox-sw.js');` in my `src/sw.js`, webpack plugin should copy those files into *destination folder*.

This "copying" feature is present when generating a serviceworker, but doesn't work when using a serviceworker template.

zhirzh:

Fixes #653 and #650 

There are 2 paths for building service-worker using `workbox-build`

1. generate-sw
2. inject-manifest

generate-sw automatically injects a copy of `workbox-sw` production build (from local `node_modules/`) into the build folder.
This feature is missing when using manifest injection. This PR adds the said missing feature.

### DIFF
--- a/packages/workbox-build/src/lib/inject-manifest.js
+++ b/packages/workbox-build/src/lib/inject-manifest.js
@@ -89,7 +89,8 @@ const injectManifest = (input) => {
   })
   .then((manifestEntries) => {
     let swFileContents = fs.readFileSync(input.swSrc, 'utf8');
-    swFileContents = `importScripts('${workboxSWImportPath}');\n` + swFileContents;
+    swFileContents = `importScripts('${workboxSWImportPath}');\n`
+                      + swFileContents;
     const injectionResults = swFileContents.match(injectionPointRegex);
     if (!injectionResults) {
       throw new Error(errors['injection-point-not-found']);
--- a/packages/workbox-build/src/lib/inject-manifest.js
+++ b/packages/workbox-build/src/lib/inject-manifest.js
@@ -2,6 +2,7 @@ const fs = require('fs');
 const mkdirp = require('mkdirp');
 const path = require('path');
 
+const copyWorkboxSW = require('./utils/copy-workbox-sw');
 const getFileManifestEntries = require('./get-file-manifest-entries');
 const errors = require('./errors');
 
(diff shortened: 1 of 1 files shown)


gauntface:
Thanks for the PR, there are a few things I'm concerned about:

1.This means the sw on dev environment fails (i.e. if the SW Template isn't built then there is no importScripts so any current code is broken).
1. Maybe this should be an optional thing / have an option to enable / disable.
1. No tests to ensure this is working / stay's working.

cc @jeffposnick what do you think?

zhirzh:
@gauntface I don't quite understand the issue.
 
when using generate-sw, a [template from `workbox-build`](https://github.com/GoogleChrome/workbox/blob/20c9da05294cbe8569f3b8e4a82683f197db41fd/packages/workbox-build/src/lib/templates/sw.js.tmpl) is used.

When opting for inject-manifest, the user supplies a path for their template (called `swSrc`). The build script then tries to read the file contents [here](https://github.com/zhirzh/workbox/blob/20c9da05294cbe8569f3b8e4a82683f197db41fd/packages/workbox-build/src/lib/inject-manifest.js#L70).

If `fs.readFileSync(...)` fails, it throws an error and the build process halts.

If nothing goes wrong here, that means we now have a user supplied template in `swFileContents`. So we can safely add the `importScripts(...)` line [at the top](https://github.com/zhirzh/workbox/blob/fba5ccd61301a8e48daf7f595400ca1bff32d365/packages/workbox-build/src/lib/inject-manifest.js#L92-L93).

Did I miss something or get something wrong?

gauntface:
When using `inject-manifest` it's not a case of just using a different manifest. It is simply a way of defining a place to put a manifest.

The original thinking behind this was to allow developers to have a working development service worker that would only precache on a final build (i.e. it's assumed the developer would decide how to get workbox-sw into their project files).

That being said, you aren't the first person to dislike or question this design decision and your proposal makes sense - just want to make sure we don't block the original thinking either.

gauntface:
@jeffposnick Ping

jeffposnick:
Thanks for the contribution, @zhirzh!

My overall thoughts:
- I think this would need to be optional, and probably default to being disabled. A pattern that seems to have become popular(?) is to pull in `workbox-sw` from a CDN (e.g. https://unpkg.com/workbox-sw@1.0.1) and I don't think we should forcibly include the local reference to a `workbox-sw` file.
- The additional code has a lot of overlap with what's already being used for `generate-sw`, and it would be best to refactor that out into a module that both flows relied on.
- At least some level of testing is a must.

That's asking a lot from you, so please let us know if you think it's more than you feel comfortable taking on.

gauntface:
@zhirzh would you be up for doing this refactoring or would you like one of us to look into it?

zhirzh:
@gauntface @jeffposnick 
I'll handle it. I have some idea already of what needs to be done; might as well take it all the way.

gauntface:
Perfect, thank you so much @zhirzh, if you get stuck or need any help please just feel to mention you me or @jeffposnick and we'll support as much as possible.

addyosmani:
This PR appears to no longer be getting updated (last update from July). If this is something that @zhirzh will have time to work on, we're happy to keep it open but we can also close it or hand off to someone else if time won't allow it to be worked on further.

gauntface:
At this stage I'd say close this for now. We have a proposal to offer something similar in v3 that is aimed at solving this problem.

PR #664, opened 2017-07-06
Adds support for manifestTransforms

### LINKED ISSUE #626: urlMapping option in workbox-build
**Library Affected**:
`workbox-build`

An idea that's come up a few times (see https://github.com/GoogleChrome/workbox/issues/604) is to provide an additional configuration option to `workbox-build` that would allow developers to manipulate each URL prior to writing it out to the final manifest.

It would give developers to more fine-grained control over the final URLs than what `modifyUrlPrefix` currently supports, and `modifyUrlPrefix` could probably be rewritten to use this approach under the hood.

The (hypothetically named) `urlMapping` would take a callback function that's passed a single parameter, `url`, and if it returns a string, that string would be used instead of `url` in the manifest entry. If it returned a false-y value, we could omit the manifest entry entirely, giving developers more control over excluding manifest entries without having to deal with glob patterns.

We could then implement CDN asset paths via something like:

```js
workboxBuild.generateSW({
  // Other options...
  urlMapping: (url) => {
    // url is an asset path, relative to the SW file.
    if (url === 'something-to-skip.js') {
      return null;
    }
    return url.endsWith('.html') ? url : CDN_PREFIX + url;
  },
});
```

jeffposnick:
R: @addyosmani @gauntface

Fixes #626

This is based on the discussion from #626 + some follow-up meetings.

It adds support for a new `manifestTransforms` option to `workbox-build`, which takes one or more `ManifestTransform` functions as a parameter. This opens the door for doing post-processing of the manifest that was generated from `globPatterns` + `templatedUrls`.

I've reimplemented the existing `dontCacheBustUrlsMatching` and `modifyUrlPrefix` options so that they use the same manifest transformation pipeline.

I've made it so that its valid for an object `ManifestEntry` not to have a `revision` field, and that's treated the same way we treat a string `ManifestEntry`. I haven't removed support for using strings to represent manifest entries that lack `revision` fields, though we should revisit deprecating strings in favor of objects without `revision` fields in a future release.


kevinphelps:
Any idea when this will land in an npm release?

kevinphelps:
I just updated my project to use this feature (using a custom build). It looks great. Thanks for the work on this!

packages/workbox-build/src/lib/utils/filter-files.js
```diff
@@ -1,14 +1,55 @@
...
 
-const logHelper = require('../log-helper');
-const constants = require('../constants');
-const modifyUrlPrefixes = require('./modify-url-prefix');
+/**
+ * A `ManifestTransform` function can be used to modify the modify the `url` or
+ * `revision` properties of some or all of the
+ * {@link module:workbox-build#ManifestEntry|ManifestEntries} in the manifest.
+ *
+ * Setting the `revision` property of an entry to a "falsey" value will cause
```
gauntface: Why not just return an object without a revision key, otherwise it must be a string with at least one character?


I always get nervous with "falsey" values because it just introduces possibility of a simple mistake in user code that just results in unexpected behavior.

packages/workbox-build/src/lib/utils/filter-files.js
```diff
@@ -1,14 +1,55 @@
...
+ *   if (entry.url.match(hashRegExp)) {
+ *     delete entry.revision;
+ *   }
+ *   return entry;
+ * });
+ *
+ * @callback ManifestTransform
+ * @param {Array<ManifestEntry>} manifest The manifest, prior to the current
+ * transformation.
+ * @return {Array<ManifestEntry>} The manifest with the transformation applied.
```
gauntface: Shouldn't the input and output be `ManifestEntry` instead of arrays of entries?

packages/workbox-build/src/lib/utils/filter-files.js
```diff
@@ -19,23 +60,28 @@ module.exports = (fileDetails, options) => {
...
-      url: url,
+      url: fileDetails.file.replace(path.sep, '/'),
       revision: fileDetails.hash,
     };
   });
+
+  const manifestTransforms = options.manifestTransforms || [];
+
+  if (options.modifyUrlPrefix) {
+    manifestTransforms.push(modifyUrlPrefixTranform(options.modifyUrlPrefix));
```
gauntface: I would have run the workbox transforms BEFORE the user transforms. i.e. my thinking of the API was that users would be a final chance to alter the urls / revisions.

packages/workbox-build/src/lib/utils/modify-url-prefix-transform.js
```diff
@@ -43,7 +42,15 @@ module.exports = (urlString, modifyUrlPrefix) => {
...
+  return (manifest) => manifest.map((entry) => {
+    if (typeof entry.url !== 'string') {
+      throw new Error(errors['manifest-entry-bad-url']);
+    }
+
+    const modifiedUrl = entry.url.replace(modifyRegex, (match) => {
+      return modifyUrlPrefix[match];
+    });
+
+    return Object.assign({}, entry, {url: modifiedUrl});
```
gauntface: why not just go `entry.url = modifiedUrl` ?

packages/workbox-precaching/test/sw/revisioned-caching.js
```diff
@@ -118,48 +118,6 @@ describe('sw/revisioned-caching()', function() {
...
-      }
-
-      if (!caughtError) {
-        throw new Error('Expected file manifest to cause an error.');
-      }
-      caughtError.message.indexOf(`The 'revision' parameter has the wrong type`).should.equal(0);
-    });
-  });
-
-  it(`should throw an errror for an empty string revision.`, function() {
```
gauntface: This still seems like a valid test, what was wrong with it?

packages/workbox-precaching/test/sw/revisioned-caching.js
```diff
@@ -118,48 +118,6 @@ describe('sw/revisioned-caching()', function() {
...
-    null,
-    undefined,
-    false,
-    true,
-    12345,
-    {},
-    [],
-  ];
-  invalidTypeRevisions.forEach((invalidRevision) => {
-    it(`should throw an errror for bad revision value '${JSON.stringify(invalidRevision)}'`, function() {
```
gauntface: This test seems like a valid set of tests.

packages/workbox-build/src/lib/utils/filter-files.js
addyosmani: Same question. Unless your intent is to transform multiple manifests, it's a little unclear why this is an array of `ManifestEntry` items.

packages/workbox-build/src/lib/utils/filter-files.js
```diff
@@ -1,14 +1,55 @@
...
+ * @return {Array<ManifestEntry>} The manifest with the transformation applied.
+ * @memberof module:workbox-build
+ */
 
 module.exports = (fileDetails, options) => {
   const maximumFileSize = options.maximumFileSizeToCacheInBytes ||
     constants.maximumFileSize;
   const filteredFileDetails = fileDetails.filter((fileDetails) => {
-    // Filter oversize files.
+    // Remove oversize files.
```
addyosmani: s/oversize/oversized

packages/workbox-build/src/lib/utils/filter-files.js
jeffposnick: The current implementation takes in a whole manifest (an array of one or more `ManifestEntry` objects) and returns a whole manifest.

Based on your comments, I think you're assuming that this behaves like a function that gets passed to `map()`, and which operates on one entry at a time. That's not what I ended up going with.

The function which gets passed in might use `map()` under the hood, or it might use `reduce()`, or it might just manually modify the array of `ManifestEntry` objects.

I can switch this over so that the function provided just gets used in `manifest.map(<yourFunction>)`, but there's more flexibility in this proposed implementation, and there are still checks in place to ensure that we end up with a valid manifest.

Thoughts?

packages/workbox-build/src/lib/utils/filter-files.js
jeffposnick: Okay, switching.

packages/workbox-build/src/lib/utils/modify-url-prefix-transform.js
jeffposnick: I was doing something stupid in a test that only would work if a copy of the original entry was returned. I've fixed the test and moved to your suggestion.

packages/workbox-build/src/lib/utils/filter-files.js
jeffposnick: I don't feel strongly about it as you do, so I'm happy to go with the strict requirement that you'd like to see.

packages/workbox-precaching/test/sw/revisioned-caching.js
jeffposnick: It was due to https://github.com/GoogleChrome/workbox/pull/664#discussion_r126388087, but I can reinstate it.

packages/workbox-precaching/test/sw/revisioned-caching.js
jeffposnick: It was due to https://github.com/GoogleChrome/workbox/pull/664#discussion_r126388087, but I can reinstate it.

packages/workbox-build/src/lib/utils/filter-files.js
gauntface: Could we change the  "manifest" variable name in all the examples to something like "allManifestEntries" just to make this extremely obvious?

packages/workbox-build/src/lib/utils/filter-files.js
```diff
@@ -19,23 +61,32 @@ module.exports = (fileDetails, options) => {
...
+  }
+
+  if (options.dontCacheBustUrlsMatching) {
+    manifestTransforms.push(
+      noRevisionForUrlsMatchingTransform(options.dontCacheBustUrlsMatching));
+  }
+
+  if (Array.isArray(options.manifestTransforms)) {
+    manifestTransforms.concat(options.manifestTransforms);
+  }
```
gauntface: Is it worth throwing an error if it's not an array but is defined? Originally I was expecting this to be "manifestTransform" and the use of this would be a single function (i.e. it would be up to the developer to manage multiple transforms etc. Fine with offering the API as an array, but I can expect people to set it to a function and then wonder why it's never called.

packages/workbox-build/src/lib/utils/filter-files.js
jeffposnick: I'm going to go with `manifestEntries` as the "all" could give the wrong impression if the developer thinks that they'll always be passed the initial, unmodified set of manifest entries into each callback. They're passed the current set of manifest entries in the pipeline of transformations.

jeffposnick:
@gauntface, please take another look.

kevinphelps:
I can confirm that the original changes in this pull request work in a production app. 💯

kevinphelps:
Any idea when a new version will be published with this feature? We are currently using the code in a custom build, and it works in production over a CloudFront CDN. 💯

coryrylan:
true

gauntface:
@kevinphelps @coryrylan we should we cutting a release tomorrow.

H/T to @jeffposnick for scheduling a more regular release schedule.

jeffposnick:
We're still finalizing the details, but the current idea is to track our release schedule (& any public meetings) using https://calendar.google.com/calendar/embed?src=gsmnmr45i797q67dfhp08t48no%40group.calendar.google.com&ctz=America/New_York

Stay tuned for more details.

jeffposnick:
It's in https://github.com/GoogleChrome/workbox/releases/tag/v1.2.0

PR #665, opened 2017-07-06
Adds a current package-lock.json

### LINKED ISSUE #606: Update to npm 5?
**Library Affected**:
all

`npm` 5 brings with it lockfiles, which AFAIK are similar to the lockfiles that `yarn` has been generating.

The [recommendation](https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/#libraries-vs-applications) from the `yarn` team is that lockfiles should always be checked in to GitHub, even for the "library" use case that our project falls under.

Presumably that also applies to `npm` 5 lockfiles, so once we start using that, we should check in the lockfile.

The other issue that's related to `npm` is whether it would play nicely with our `lerna` setup. The only issue I currently see in the `lerna` tracker that might be relevant is https://github.com/lerna/lerna/issues/856, but I *think* we're not using hoisting, so perhaps that's fine.

I'll update to `npm` 5 on a development machine and test things out.

jeffposnick:
R: @addyosmani @gauntface

Fixes #606

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2,23 +2,32 @@
   "name": "workbox",
   "version": "0.0.0",
   "lockfileVersion": 1,
+  "requires": true,
   "dependencies": {
     "JSONStream": {
       "version": "1.3.1",
       "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-1.3.1.tgz",
       "integrity": "sha1-cH92HgHa6eFvG8+TcDt4xwlmV5o=",
-      "dev": true
+      "dev": true,
+      "requires": {
+        "jsonparse": "1.3.1",
+        "through": "2.3.8"
+      }
     },
     "accepts": {
       "version": "1.3.3",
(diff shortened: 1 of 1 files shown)


gauntface:
I'm pro this - just restarted Travis to see if it was just having a funny moment. Merge once Travis is ok :)

gauntface:
Looks like one of the modules isn't included in the package lock and it's making travis cranky.

jeffposnick:
Okay, let's try again with the most recent iteration of my local lockfile.

jeffposnick:
Travis still isn't happy. Travis is using `npm` 5.0.3 and I have `npm` 5.2.0 locally. I'm going to look into whether that's a possible issue...

jeffposnick:
I deleted the `node_module` caches that Travis maintains and it seems to have helped. A few of the tests have come back green, and hopefully the whole PR will.

jeffposnick:
Okay, that seems to have done the trick. I'm going to merge.

Issue #478, opened 2017-04-29
Normalize eslint configs


jeffposnick:
We're currently using a mishmash of eslint config file + inline comments for various types of files. We should try to normalize things a bit and remove the one-offs whenever possible.

PR #666, opened 2017-07-07
Background sync queue callbacks

### LINKED ISSUE #646: Background sync should add the ability to modify request prior to retrying

philipwalton:
R: @jeffposnick @prateekbh 

Fixes #646

This PR updates the background sync queue callbacks to address the issues brought up in #646. Here's what's changed:

- renamed: `onResponse` => `replayDidSucceed`
- renamed: `onRetryFailure ` => `replayDidFail`
- added: `requestWillEnqueue`
- added: `requestWillDequeue`

I've also added a utility method to deprecate (and warn about) old object methods since I'm renaming existing callbacks, but I didn't feel like it warranted a major version bump.


packages/workbox-background-sync/src/lib/request-manager.js
```diff
@@ -18,8 +19,15 @@ class RequestManager {
...
    * @private
    */
-  constructor({callbacks, queue}) {
-    this._globalCallbacks = callbacks || {};
+  constructor({callbacks, queue} = {}) {
+    callbacks = callbacks || {};
+
+    // Rename deprecated callbacks.
+    const base = 'workbox-background-sync.RequestManager.callbacks';
+    deprecate(callbacks, base, 'onResponse', 'replayDidSucceed');
```
jeffposnick: Could you make these calls to `deprecate` conditional, and only invoke each one if the corresponding `callbacks.onResponse` / `callbacks.onRetryFailure` are set?

packages/workbox-background-sync/src/lib/request-queue.js
```diff
@@ -117,31 +126,42 @@ class RequestQueue {
...
+      return err;
     }
   }
 
   /**
    * get the Request from the queue at a particular index
    *
    * @param {string} hash hash of the request at the given index
-   * @return {Request} request object corresponding to given hash
+   * @return {Promise} request object corresponding to given hash
```
jeffposnick: `{Promise<Request>}`

packages/workbox-background-sync/src/lib/request-manager.js
philipwalton: Yeah, this was an oversight and noticed it when updating the GA stuff. It should be fixed now.

packages/workbox-background-sync/src/lib/request-queue.js
philipwalton: This doesn't fulfill to a `Request` instance, it fulfills to an object with `config`, `metadata`, and `request` properties, and the `request` property is a JSON-able version of the `Request` object.

I could update it to:

```js
/** @return {Promise<Object>} */
```

But I didn't see much of that in the existing code.

philipwalton:
The only callback documentation I saw in the original code was here, so that's where I updated it:
https://github.com/GoogleChrome/workbox/blob/1bf91ec22d666d5ce5b42d505e4cc850ea3222d5/packages/workbox-background-sync/src/lib/background-sync-queue-plugin.js#L10-L18

I agree that the JSDoc comments need to be more clear, but I wasn't sure what conventions we were using for documenting sub-properties (I'm not seeing a lot of consistency, so I was going to bring it up in the next meeting).

If we're going to be following Google JS style, we should be documenting them like this (but again, I didn't see that so I'm not sure if there was a reason we weren't doing it that way, e.g. doc generation, etc.):
https://google.github.io/styleguide/jsguide.html#features-objects-destructuring

philipwalton:
I was also going to suggest either I or @prateekbh do a clean up since most of the files in this project don't include the license, and there are many cases where the code doesn't match the style guide.

If I update the `gulp-eslint` and `eslint-config-google` versions I notice a lot more errors, most of which can be easily fixed via the `--fix` option.

jeffposnick:
(I'm in the middle of putting together a PR that updates some of our linting rules, tracked in #478)

philipwalton:
> (I'm in the middle of putting together a PR that updates some of our linting rules, tracked in #478)

Cool, happy to help out on that front if you want.

prateekbh:
>If I update the gulp-eslint and eslint-config-google versions I notice a lot more errors, most of which can be easily fixed via the --fix option.

Happy to do that. I tried adding prettier, but it conflicts with our rules a little too much(would like to discuss in the next meeting)

packages/workbox-background-sync/src/lib/request-queue.js
prateekbh: my bad it is indeed 
```
/** @return {Promise<Object>} */
```

philipwalton:
I've updated the jsdoc comments. @jeffposnick PTAL.

philipwalton:
One more issue:

The current `workbox-google-analytics` API accepts a `hitFilter` function, that -- if it throws -- aborts the hit and stops further retries. If we want to keep this behavior, we'd have to add an additional feature here to support aborting a retry.

The way queuing currently works, items are left in the queue even after a successful retry; they're only removed during [clean up](https://github.com/GoogleChrome/workbox/blob/master/packages/workbox-background-sync/src/lib/queue-utils.js#L58-L94).

If we want to support stopping a request from being replayed, we should probably add a utility method to remove an item from the queue rather than depending on a clean up cycle.

I think this makes sense since I know I was momentarily confused when debugging by the fact that successful replays remained in the queue.

prateekbh:
>If we want to support stopping a request from being replayed, we should probably add a utility method to remove an item from the queue rather than depending on a clean up cycle

I can take this up, do you want the item to be removed after successful replay or do you have anything else in mind?

P.S. `clean up` removes the request from IDB after its threshold time. But we can remove the request from the current instance after a successful replay.

prateekbh:
@philipwalton  As far as `hitFilter` is concerned, maybe we can add a `requestWillPlay` callback much like `requestWillDeque`, and returning false from this can make the request skip replaying.

jeffposnick:
@philipwalton, are there "legitimate" use cases for wanting to abort things via an exception in `hitFilter`? Or was that just an unintended side effect of the previous codebase? It sounds like it might be cleaner *not* to support that, and add in `catch()` as needed, along with logging to report the error.

lib/deprecate.js
```diff
@@ -0,0 +1,33 @@
...
+ * and aliases the old name to the new name.
+ * @param {Object} obj The object containing the methods.
+ * @param {string} oldName The method to deprecate.
+ * @param {string} newName The new method replacing the deprecated method.
+ * @param {string} ctx The context project/object to identify the method names.
+ */
+export default (obj, oldName, newName, ctx) => {
+  if (Object.prototype.hasOwnProperty.call(obj, oldName)) {
+    /* eslint-disable no-console */
+    console.warn(`In ${ctx}: ` +
```
jeffposnick: Sorry for the late comment, but I'd suggest using the `warn()` method of [`lib/log-helper`](https://github.com/GoogleChrome/workbox/blob/master/lib/log-helper.js) here. It will keep the messages in the console consistent, and you can include some of the detailed information in the optional second parameter, like:

```js
logHelper.warn(`${oldName} is deprecated; use ${newName} instead`, {Context: ctx});
```

philipwalton: Good to know. Done.

philipwalton:
> @philipwalton, are there "legitimate" use cases for wanting to abort things via an exception in `hitFilter`? Or was that just an unintended side effect of the previous codebase? It sounds like it might be cleaner *not* to support that, and add in `catch()` as needed, along with logging to report the error.

I'd be OK with not supporting it for the analytics use case. I originally implemented it primarily for compatibility with how [tasks](https://developers.google.com/analytics/devguides/collection/analyticsjs/tasks#aborting_task_processing) work in analytics.js, but I don't see it as a critical feature by any means.

> @philipwalton  As far as `hitFilter` is concerned, maybe we can add a `requestWillPlay` callback much like `requestWillDeque`, and returning false from this can make the request skip replaying.

From a generic background sync perspective, I can imagine cases where a developer wouldn't want to replay a hit. Perhaps in cases where they know the current state data is fresher than what this hit will send (though I suppose you could argue they could/should handle that on the back end).

If we add this method I'd recommend the name `requestShouldReplay`, and if should probably be async since you'd likely need to check IDB to know whether or not it should replay.

..in fact, perhaps all of these should be async :)

philipwalton:
@prateekbh and I discussed this offline, and if the rest of the team is OK with @jeffposnick's suggestion to drop support for aborting a replay by throwing in the `hitFilter` function, then we think we're ready to merge as is.

We decided it's probably better to wait for an actual use case for aborting a replay and design the API around that rather than trying to anticipate one.

So if everyone agrees, I think we should merge this and move the discussion of how to handling the API change in the workbox-google-analytics PR.

jeffposnick:
SGTM.

PR #667, opened 2017-07-07
Updates eslint and centralizes some config options.

jeffposnick:

This is part of #478, but it only attempts to normalize the usage of per-directory `.eslintrc*` files. It doesn't attempt to do anything about the use of inline `/* eslint-disable */` and similar rules. That will be done in a follow-up PR.


packages/workbox-runtime-caching/test/.eslintrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  globals: {
```
philipwalton: Did you mean to add this? I thought you wanted to remove all but the base eslintrc file.

I tend to prefer declaring globals via inline eslint comments on a per-file basis (and the only globals declared in eslintrc files are globals that apply generally to the entire project).

That way when you see a variable in a file, it's very clear that the intent is to make it global, and you don't have to look up the directory structure for eslintrc files. WDYT?

packages/workbox-background-sync/src/lib/request-queue.js
```diff
@@ -117,7 +117,7 @@ class RequestQueue {
         id: hash,
         url: request.url,
       });
-    } catch(e) {
+    } catch (e) {
```
philipwalton: Somewhat tangental, but in the future I'd love to see us standardize on `err` for `catch` blocks unless there's a good reason to use a more descriptive name. Maybe we could write an eslint plugin to enforce this...

.eslintrc.js
```diff
@@ -0,0 +1,45 @@
...
+    ecmaVersion: 2017,
+    sourceType: 'module',
+  },
+  globals: {
+    BroadcastChannel: false,
+    WorkboxSW: false,
+    workbox: false,
+  },
+  rules: {
+    'no-tabs': 2,
```
philipwalton: I think this is an oversight in `eslint-config-google`, as Google JS Style definitely requires spaces. I'll update that library (related: https://github.com/google/eslint-config-google/issues/40).

philipwalton: If you upgrade to 0.8.1 you'll get this rule now.

.eslintrc.js
```diff
@@ -0,0 +1,45 @@
...
+    'linebreak-style': 0,
+    'space-before-function-paren': ['error', {
```
philipwalton: I've also updated the  `eslint-config-google` rule for this, as I think it's what the Google JS style guide recommends. You can get the changes in `0.9.0`.

Also, in the future, if you're tempted to add rules/exceptions to your eslint config, feel free to ping me, as maybe it's a mistake :)

packages/workbox-runtime-caching/test/.eslintrc.js
gauntface: Personally I'm a fan of globals in eslint if all code being tested is run in that context.

.eslintrc.js
jeffposnick: Thanks!

packages/workbox-runtime-caching/test/.eslintrc.js
jeffposnick: This was a bit of a crutch as there are several files in this one directory that share a few globals, without them being used anywhere else.

There are similar issues throughout the codebase that are addressed via inline `/* eslint ... /*` config comments now, and I'm punting on addressing standardizing on that sort of thing for another PR.

Issue #669, opened 2017-07-10
Use babel to transpile the workbox-build codebase?


MaxLeiter:
**Library Affected**:
*workbox-build*

**Browser & Platform**:
*all browsers*

**Issue or Feature Request Description**:
Currently making a PR, should be as easy as throwing in a 'use strict';

Ref: https://travis-ci.org/thelounge/lounge/jobs/251861458
and thelounge/lounge#1317

jeffposnick:
Thanks for pointing that out! The intention is to get support node 4 for `workbox-build`, though we obviously haven't done a great job of testing that.

While we could approach this via adding in `'use strict';'` throughout the individual modules, I wanted to see how @gauntface & co. felt about running the `workbox-build` project through babel with [`babel-present-env`](https://github.com/babel/babel-preset-env) set up for node 4.

There are a number of times that I've wanted to use, e.g., `async`/`await` or spread operators when working on the `workbox-build` codebase, but didn't, because of legacy node support.

MaxLeiter:
Sounds good to me.

gauntface:
Personally I hate having to transpile anything just to run in a baseline environment so I'd rather just accept use strict.

There is a bigger question of what our support matrix should be. Should it be current stable and latest version? or should be follow the same as node's schedule where node 4 is in maintanence? https://github.com/nodejs/LTS

jeffposnick:
The initial impetus behind a minimum of node 4 was to match what `create-react-app` required, on the assumption that we will try to integrate with them at some point. It turns out `c-r-a` has since switched to requiring node 6: https://github.com/facebookincubator/create-react-app/pull/1897

That being said, since we can easily resolve node 4 support with minimal extra effort, I'm fine with keeping 4 as a minimum, at least until April 2018.

We're going to need `use strict` throughout `workbox-build` and `workbox-cli`, in that case.

MaxLeiter:
`use strict` is already present throughout workbox-cli

MaxLeiter:
Fixed as of #670

PR #670, opened 2017-07-10
Add use strict to generate-sw for Node 4 support

MaxLeiter:

Fixes #670

### DIFF
--- a/packages/workbox-build/src/index.js
+++ b/packages/workbox-build/src/index.js
@@ -1,3 +1,4 @@
+'use strict';
 // This will be the logic that powers both module and CLI
 const generateSW = require('./lib/generate-sw');
 const getFileManifestEntries = require('./lib/get-file-manifest-entries');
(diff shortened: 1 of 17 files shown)


MaxLeiter:
Signed

jeffposnick:
Thanks for the contribution, @MaxLeiter!

I'd like to move the discussion to #669 in lieu of merging this PR as-is. I know that there are other files inside of `workbox-build` that are using `const` and `let`, so I have a feeling that the change in this PR wouldn't be sufficient. We should talk about addressing it in a more holistic fashion.

gauntface:
Was looking into adding testing for node 4 but it's going to be a pretty big headache since selenium doesn't support anything lower than node 6.

@jeffposnick I'm fine with taking this PR, any other comments from you?

jeffposnick:
👍

MaxLeiter:
any chance of a `1.1.1` release?

jeffposnick:
It's in https://github.com/GoogleChrome/workbox/releases/tag/v1.2.0

PR #680, opened 2017-07-13
fixing queue initialization bug

### LINKED ISSUE #674: Background sync RequestQueue doesn't re-populate
The `initQueue` method calls [`.concat(idbQueue)` here](https://github.com/GoogleChrome/workbox/blob/d60bf9744dd89ecaafe80205c329225771caf0ab/packages/workbox-background-sync/src/lib/request-queue.js#L55) when it should be calling `.push(...idbQueue)`, and as a result the queue is not initialized with existing data in the store.

In addition to fixing this, we should probably also add a test covering this case. @prateekbh are you able to do this? If not I can take a look.

prateekbh:
R: @jeffposnick @philipwalton 

Fixes #674 

Fixes the repopulation of the list and also adds test coverage for it

### DIFF
--- a/packages/workbox-background-sync/test/browser/request-queue.js
+++ b/packages/workbox-background-sync/test/browser/request-queue.js
@@ -37,7 +37,20 @@ describe('request-queue tests', () => {
     chai.assert.isObject(queue._config);
   });
 
-  it('queueName is corrent', () =>{
+  it('initialize should re-fill the queue', async () => {
+    chai.assert.equal(queue._queue.length, 0);
+    const hash = await queue.push({
+      request: new Request('http://lipsum.com/generate'),
+    });
+    chai.assert.equal(queue._queue.length, 1);
+    queue._queue = [];
+    chai.assert.equal(queue._queue.length, 0);
+    await queue.initQueue();
+    chai.assert.equal(queue._queue.length, 1);
+    chai.assert.equal(queue._queue[0], hash);
+  });
+
+  it('queueName is correct', () =>{
     chai.assert.equal(queue._queueName, QUEUE_NAME);
   });
 
--- a/packages/workbox-background-sync/test/browser/request-queue.js
+++ b/packages/workbox-background-sync/test/browser/request-queue.js
@@ -37,6 +37,13 @@ describe('request-queue tests', () => {
     chai.assert.isObject(queue._config);
   });
 
+  it('initialize should not fail for null data', async () => {
+    chai.assert.equal(queue._queue.length, 0);
+    idbHelper.put(queue._queueName, null);
+    await queue.initQueue();
+    chai.assert.equal(queue._queue.length, 0);
+  });
+
   it('initialize should re-fill the queue', async () => {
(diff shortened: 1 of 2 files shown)


packages/workbox-background-sync/test/browser/request-queue.js
```diff
@@ -37,7 +37,20 @@ describe('request-queue tests', () => {
     chai.assert.isObject(queue._config);
   });
 
-  it('queueName is corrent', () =>{
+  it('initialize should re-fill the queue', async () => {
```
gauntface: I may be misunderstanding the original problem, but this test doesn't actually test what happens when you call initQueue when there are no entries to be added.

prateekbh: oh I can cover that. However earlier on reboot of sw the queue was not getting re-initialized, which this should cover now.

philipwalton: Yes, please add a test case for when `this._idbQDb.get()` resolves to a null value, as it will the first time the SW starts up.

philipwalton:
On a slight tangent, I think part of the reason this bug came about in the first place is you need to set `this._queue` to something in the constructor, but you can't run async code there, so you initialize it to an empty array and then kick off the async task.

However, I think this is an anti-pattern because we don't know when the initialization will complete and it could lead to race conditions if you invoke another prototype method before completion.

I think a better approach would be return the promise returned by `initQueue()`, and wait for its resolution in all other methods that need to reference `this._queue`.

gauntface:
+1 @philipwalton's comments here. You have a race condition that the external libraries can't detect or handle.

Would be good to ensure initQueue is called before other methods are used as well.

prateekbh:
I have fixed the test cases for this scenario. For the race condition I wanted to have a quick syncup before writing the code, will probably like to do it in a separate post.

prateekbh:
@gauntface I talked to @philipwalton about the idea of handling the race condition in a separate PR.
If its cool with you, would u want to merge this one first?

packages/workbox-background-sync/test/browser/request-queue.js
```diff
@@ -60,7 +80,7 @@ describe('request-queue tests', () => {
 
     chai.assert(callbacks.requestWillEnqueue.calledOnce);
     chai.assert(callbacks.requestWillEnqueue.calledWith(
-        sinon.match.has('request')));
+      sinon.match.has('request')));
```
philipwalton: This indentation of 4 spaces is intentional as it's a line continuation:
https://google.github.io/styleguide/jsguide.html#formatting-indent

prateekbh: our lint doesn't catch this?

prateekbh:
@philipwalton @gauntface, fixed the whitespaces. Can we make some eslint rule to catch what @philipwalton mentioned?

philipwalton:
There's no built-in ESLint rule config that would catch this indentation issue. I have it on my list of rules to write, but it's very low on the list.

PR #681, opened 2017-07-13
Enable testing on supported versions of node

gauntface:

Should be merged before #670

This enables tests on Travis for Node 4, 6 and latest version of node (8.X)


gauntface:
Scrapping this. It seems like our testing infra needs node 6.

PR #684, opened 2017-07-17
Adding defaults to workbox-build

### LINKED ISSUE #421: Defaults in sw-build
**Library Affected**:
sw-build

**Browser & Platform**:
N/A

Issue

So `sw-build` should have a defined set of defaults to it, which it can use if the developer passes nothing.

This will also avoid various build step plugins to define these defaults randomly on their own will.

gauntface:

Fixes #421

Sets default glob pattern to include js,css and html

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2,27 +2,12 @@
   "name": "workbox",
   "version": "0.0.0",
   "lockfileVersion": 1,
-  "requires": true,
   "dependencies": {
-    "JSONStream": {
-      "version": "1.3.1",
-      "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-1.3.1.tgz",
-      "integrity": "sha1-cH92HgHa6eFvG8+TcDt4xwlmV5o=",
-      "dev": true,
-      "requires": {
-        "jsonparse": "1.3.1",
-        "through": "2.3.8"
-      }
-    },
     "accepts": {
       "version": "1.3.3",
       "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.3.tgz",
       "integrity": "sha1-w8p0NJOGSMPg2cHjKN1otiLChMo=",
-      "dev": true,
-      "requires": {
-        "mime-types": "2.1.15",
-        "negotiator": "0.6.1"
-      }
+      "dev": true
     },
     "acorn": {
       "version": "5.1.1",
@@ -35,9 +20,6 @@
       "resolved": "https://registry.npmjs.org/acorn-dynamic-import/-/acorn-dynamic-import-2.0.2.tgz",
       "integrity": "sha1-x1K9IQvvZ5UBtsbLf8hPj0cVjMQ=",
       "dev": true,
-      "requires": {
-        "acorn": "4.0.13"
-      },
       "dependencies": {
         "acorn": {
           "version": "4.0.13",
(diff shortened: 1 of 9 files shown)


packages/workbox-build/src/lib/get-file-manifest-entries.js
```diff
@@ -63,7 +65,12 @@ const getFileManifestEntries = (input) => {
     return Promise.reject(
       new Error(errors['both-glob-patterns-static-file-globs']));
   }
-  const globPatterns = input.globPatterns || input.staticFileGlobs;
+
+  let globPatterns = input.globPatterns || input.staticFileGlobs;
+  if (typeof input.globPatterns === 'undefined' &&
```
jeffposnick: Would

```js
if (typeof globPatterns === 'undefined') {
  globPatterns = constants.defaultGlobPatterns;
}
```

be equivalent?

packages/workbox-build/src/lib/generate-sw.js
```diff
@@ -18,6 +18,8 @@ const constants = require('./constants');
  * `globPatterns` against.
  * @param {Array<String>} input.globPatterns Files matching against any of
  * these glob patterns will be included in the file manifest.
+ *
+ * Defaults to ['**\/*.{js,css}']
```
jeffposnick: Here and elsewhere, you can drop the `\` character in `['**\/*.{js,css}']`.

packages/workbox-build/test/node/get-file-manifest-entries.js
```diff
@@ -84,6 +83,22 @@ describe('Test getFileManifestEntries', function() {
     }, Promise.resolve());
   });
 
+  it('should use defaults for undefined globPatterns', function() {
+      let args = Object.assign({}, EXAMPLE_INPUT);
+      delete args.globPatterns;
+
+      try {
```
jeffposnick: I think you could accomplish the same thing with

```js
// const expect = require('chai').expect;

expect(
  () => wBuild.getFileManifestEntries(args)
).to.throw(`${errors['useless-glob-pattern']} ${JSON.stringify(args)}`);
```

package-lock.json
```diff
@@ -2,27 +2,12 @@
...
     "accepts": {
       "version": "1.3.3",
       "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.3.tgz",
       "integrity": "sha1-w8p0NJOGSMPg2cHjKN1otiLChMo=",
-      "dev": true,
-      "requires": {
-        "mime-types": "2.1.15",
-        "negotiator": "0.6.1"
-      }
+      "dev": true
```
jeffposnick: There's a really big diff for `package-lock.json`, with a lot of it made up of changes like this. Any idea why? Which version of npm are you using?

packages/workbox-build/src/lib/get-file-manifest-entries.js
gauntface: No sadly not.

```
input.globPatterns = '';
const globPatterns = input.globPatterns || input.staticFileGlobs;
// '' is a falsey value, that with the '||' means globPatterns is undefined.
```

packages/workbox-build/src/lib/generate-sw.js
gauntface: no you can't because JSDoc get's upset. Will look into filter this out in the code snippets somehow.

gauntface:
npm is version 5.3.0 now and bumped selenium assistant as it was given me issues - hence new package lock.

Why does this need to ship behind a major semver bump? It's not a change to the API, only a very edge case behavior (i.e. someone doesn't define something which will be none of our current users).

packages/workbox-build/src/lib/generate-sw.js
jeffposnick: Okay, gotcha. This was something that ESLint was complaining about when it made it into the actual code rather than the comments, so I was on the lookout for it.

PR #690, opened 2017-07-19
Add a build step to the mocha tests for easier cross-package integration

philipwalton:

This PR does a few primary things:

1. Replaces all instances of workbox.<package>.<Class> with just <Class>, which is imported at the top of the file and bundled into the test build.
2. Fixes up some inconsistencies that I noticed while doing my replacements in (1).
3. Includes part of @prateekbh's fix to the background sync queue initialization from IDB (without it some additional tests wouldn't pass).


gulp-tasks/build.js
```diff
@@ -44,11 +45,58 @@ const buildPackage = (projectPath) => {
...
+  return Promise.all([
+    buildJSBundle({
+      rollupConfig: {
+        entry: `./packages/${project}/test/browser/*.js`,
+        plugins: getPlugins(),
+      },
+      writeConfig: {
+        sourceMap: true,
+        format: 'iife',
+        dest: `./packages/${project}/build/test/browser.js`,
```
gauntface: Nit: Could we call this `browser-bundle.js` just so it hints that this isn't the source.

gulp-tasks/build.js
```diff
@@ -44,11 +45,58 @@ const buildPackage = (projectPath) => {
...
+        entry: [
+          `./utils/sw-test-utils.js`,
+          `./packages/${project}/test/sw/*.js`,
+        ],
+        plugins: getPlugins(),
+      },
+      writeConfig: {
+        sourceMap: true,
+        format: 'iife',
+        dest: `./packages/${project}/build/test/sw.js`,
```
gauntface: Same as above `sw-bundle.js`

gulp-tasks/build.js
```diff
@@ -44,11 +45,58 @@ const buildPackage = (projectPath) => {
...
     .then(() => printBuildTime(`${(Date.now() - startTime) / 1000}s`));
 };
 
+/**
+ * Builds the browser and service worker test bundles.
+ * @param {String} projectPath The path to a project directory.
+ * @return {Promise} Resolves if building succeeds, rejects if it fails.
+ */
+const buildTestBundles = (projectPath) => {
+  const getPlugins = () => [
```
gauntface: Nit: Could this just be an array of plugins rather than a function?

packages/workbox-background-sync/src/lib/request-queue.js
```diff
@@ -52,7 +52,9 @@ class RequestQueue {
    */
   async initQueue() {
     const idbQueue = await this._idbQDb.get(this._queueName);
-    this._queue.concat(idbQueue);
+    if (idbQueue) {
```
gauntface: Was this the fix from Prateeks other PR or something different?

packages/workbox-background-sync/test/browser/background-sync-queue-plugin.js
```diff
@@ -3,25 +3,38 @@
...
  distributed under the License is distributed on an "AS IS" BASIS,
  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  See the License for the specific language governing permissions and
  limitations under the License.
- */
+*/
 
 /* eslint-env mocha, browser */
-/* global chai, workbox */
+/* global chai */
```
gauntface: I think @jeffposnick made it so this wouldn't be needed anymore during his eslint refactor.

packages/workbox-background-sync/test/browser/background-sync-queue.js
```diff
@@ -39,15 +41,15 @@ describe('background sync queue test', () => {
 
   let backgroundSyncQueue;
 
-  beforeEach(async function() {
+  beforeEach(async () => {
```
gauntface: These should be functions. Mocha does some weird binding stuff that simply won't work if there are fat arrow functions.

packages/workbox-background-sync/test/browser/background-sync-queue-plugin.js
```diff
@@ -3,25 +3,38 @@
...
+import IDBHelper from '../../../../lib/idb-helper.js';
+import BackgroundSyncQueuePlugin
+    from '../../src/lib/background-sync-queue-plugin.js';
+import {defaultDBName} from '../../src/lib/constants.js';
 
 describe('background-sync-queue-plugin test', () => {
-  const backgroundSyncQueue =
-      new workbox.backgroundSync.test.BackgroundSyncQueuePlugin({});
+  const db = new IDBHelper(defaultDBName, 1, 'QueueStore');
+  const resetDb = async () => {
```
gauntface: Please make this a function to make mocha happy.

packages/workbox-background-sync/test/browser/background-sync-queue.js
```diff
@@ -39,15 +41,15 @@ describe('background sync queue test', () => {
...
     responseAchieved = 0;
-    backgroundSyncQueue = new workbox.backgroundSync.test.BackgroundSyncQueue({
+    backgroundSyncQueue = new BackgroundSyncQueue({
       maxRetentionTime: MAX_AGE,
       callbacks: CALLBACKS,
     });
   });
 
-  afterEach(async function() {
+  afterEach(async () => {
```
gauntface: function for mocha.

packages/workbox-background-sync/test/browser/background-sync-queue.js
```diff
@@ -58,22 +60,20 @@ describe('background sync queue test', () => {
   });
 
   it('check defaults', () => {
```
gauntface: This should also be a function as well.

packages/workbox-background-sync/test/browser/queue-utils.js
```diff
@@ -3,43 +3,43 @@
...
  distributed under the License is distributed on an "AS IS" BASIS,
  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  See the License for the specific language governing permissions and
  limitations under the License.
- */
+*/
 
 /* eslint-env mocha, browser */
-/* global chai, workbox */
+/* global chai */
```
gauntface: Don't think the chai global is needed.

packages/workbox-background-sync/test/browser/queue-utils.js
```diff
@@ -3,43 +3,43 @@
...
-      'bgQueueSyncDB', 1, 'QueueStore');
-    return idbHelper.getAllKeys()
-    .then((keys) => {
-      keys.forEach((key) => {
-        idbHelper.delete(key);
-      });
-    });
-  });
+  const db = new IDBHelper(defaultDBName, 1, 'QueueStore');
+  const resetDb = async () => {
```
gauntface: please check function for mocha methods throughout.

packages/workbox-broadcast-cache-update/test/sw/broadcast-update.js
```diff
@@ -1,22 +1,40 @@
...
   const channelName = 'test-channel';
   const channel = new BroadcastChannel(channelName);
   const cacheName = 'test-cache';
   const url = 'https://example.com';
   const source = 'test-source';
 
-  it(`should throw when broadcastUpdate() is called without any parameters`, function() {
-    expect(() => workbox.broadcastCacheUpdate.broadcastUpdate()).to.throw();
+  it(`should throw when broadcastUpdate() is called without any parameters`, () => {
+    expect(() => broadcastUpdate()).to.throw(TypeError);
```
gauntface: It sounds dodgy to me to be throwing a TypeError here, is it an exception we are throwing or is it an unexpected error from an edge case?

packages/workbox-precaching/test/sw/cookies.js
```diff
@@ -1,20 +1,38 @@
...
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+/* eslint-env mocha, browser */
+/* global expect */
+
+import RevisionedCacheManager
```
gauntface: What is the advantage of this vs testing the actual build files here?

I follow that for developing the tests with stubs etc, bundling is useful, but here we've switch from testing the dev and prod builds to a custom test build, which isn't necessarily bad, it's not going to cover some edge cases we've hit with the actual end builds.

packages/workbox-precaching/test/sw/failing-cache-tests.js
```diff
@@ -20,30 +38,28 @@ describe('Test Failing Cache Behavior', function() {
...
-    return unrevisionedCacheManager.install()
-    .then(() => {
-      throw new Error('Expected error to be thrown.');
-    }, (err) => {
-      expect(err).to.exist;
-      expect(err.name).to.equal('request-not-cached');
-      expect(err.extras.url).to.equal(new URL('/__test/404/', self.location).toString());
-    });
-  });
+  // it('should fail to install unrevisioned with 404 cache request', () => {
```
gauntface: Why comment when we can just skip?

utils/build.js
```diff
@@ -166,7 +166,7 @@ function generateBuildConfigs({formatToPath, baseDir, moduleName,
     buildConfigs.push({
       rollupConfig: {
         entry: entry || path.join(baseDir, 'src', 'index.js'),
-        plugins: [devReplacePlugin, ...basePlugins, babelPlugin],
```
gauntface: Why was this removed?

templates/test-sw.hbs
```diff
@@ -44,22 +44,22 @@
...
-      const expect = chai.expect;
+      window.expect = chai.expect;
     </script>
 
     <!-- Package-specific tests matching packages/<pkg>/test/browser/*.js -->
 
     <script>
     describe('{{pkg}} > test/sw/*.js', function() {
-      {{#each scripts}}
-      it(`should run '{{this}}' sw tests`, function() {
```
gauntface: This looks like you've removed the ability to test multiple test files in favor of a single file. Could we support both options? I.e. some files can be plain JS and some files can be bundled JS builds and the testing knows to look in two locations and load them accordingly?

templates/test-sw.hbs
```diff
@@ -44,22 +44,22 @@
 
     <script>
       mocha.setup({
-        ui: 'bdd'
+        ui: 'bdd',
+        timeout: 10000,
```
gauntface: Why set a globally long timeout when individual tests can (and IMO should) define the timeout. i.e. if  a test should pass / fail quickly, this will unnecessarily extend the test time.

templates/test-browser.hbs
```diff
@@ -55,13 +55,10 @@
...
       chai.should();
-      const expect = chai.expect;
+      window.expect = chai.expect;
     </script>
 
-    <!-- Package-specific tests matching packages/<pkg>/test/browser/*.js -->
-    {{#each scripts}}
-      <script src="{{this}}"></script>
-    {{/each}}
+    <script src="/packages/{{pkg}}/build/test/browser.js"></script>
```
gauntface: Again, would favor us being able to run both plain and bundle JS files

gulp-tasks/build.js
philipwalton: Done.

gulp-tasks/build.js
philipwalton: Done.

gulp-tasks/build.js
philipwalton: It can't, because each plugin has to be a fresh instance.

packages/workbox-background-sync/src/lib/request-queue.js
philipwalton: Yeah, as I mentioned in the PR description, I needed to include this to make the tests pass. What's strange is why it's still showing up in the diff since his change was merged.

I can take a look before the final merge.

packages/workbox-background-sync/test/browser/background-sync-queue-plugin.js
philipwalton: Ahh, it looks like @prateekbh is using `chai.assert` and everywhere else we use `chai.expect`, which is aliased to just `expect`.

I can remove this in all packages expect background sync, but we should probably be consistent everywhere.

packages/workbox-background-sync/test/browser/background-sync-queue.js
philipwalton: See my main comments for discussion on this.

packages/workbox-broadcast-cache-update/test/sw/broadcast-update.js
philipwalton: I think I just happened to notice it *was* a `TypeError` that was thrown, so I wanted to be a little more specific. I didn't want to get a false positive if we updated the code and a different error were thrown but the test still passed.

packages/workbox-precaching/test/sw/cookies.js
philipwalton: I'd be curious to hear what these edge cases were.

packages/workbox-precaching/test/sw/failing-cache-tests.js
philipwalton: IIRC, the `UnrevisionedCacheManager` file didn't build, so even just having the reference there was problematic.

I suppose I could leave it unchanged if you think that's preferable.

utils/build.js
philipwalton: Oh, I guess I forgot to ask this in the PR description.

I was going to mention that I removed it because I didn't see any reason to be minifying the dev build. I thought maybe it was an oversight?

templates/test-sw.hbs
philipwalton: I think that's fine, but then those tests would have to live in a separate directory, or have an identifiably different naming structure or something. Otherwise they'd be included in the test bundle.

Since we don't (AFAICT) have any instances of this currently, do you want to wait until the use case comes up?

templates/test-sw.hbs
philipwalton: I set this because when debugging it was annoying to get timeout errors. Then I noticed a 10 second timeout was in the browser template, so I kept it for consistency.

I can revert if you like.

templates/test-browser.hbs
philipwalton: Same comment as above.

philipwalton:
Individual responses inline, general responses below:

> 1.) The reason for it(), describe(), before() and after() all receiving functions rather than fat arrow functions is that Mocha does some weird binding to the functions to add methods like this.timeout() etc. I'd keep this functionality in rather than block it on fat arrow syntax.

Yeah, I'm aware of this (and annoyed by it), and I didn't convert `function` to `=>` anywhere `this` was referenced in a `describe` or `it` block. The reason I made the change in the first place is we were being inconsistent with usage, so I standardized on `=>`. I can standardize on `function` if that's preferred, but I think we should be consistent.

 > 2.)  I like the support of bundled test files, but I don't want the entire test suite requiring a single bundle file. The reason for this is that these tests are no longer testing the final builds which has caught subtle issues in the past (when we release we run the tests against dev and prod versions)

I'm not sure I understand the issue here. Couldn't we run the unit tests in prod mode to test that behavior, and then don't the webdriver tests check the final builds?

Also, if we're going to be doing away with individual package builds in the future, won't this be a non-issue?

packages/workbox-broadcast-cache-update/test/sw/broadcast-update.js
jeffposnick: The original code predated some of the changes to `assert.js`.

I think the best practice nowadays would be

```js
expect(() => broadcastUpdate()).to.throw('assertion-failed');
```

That will work with both the `prod` and `dev` bundles, as it checks the `name` property of the thrown exception, and in both cases that will be set to `'assertion-failed'`.

utils/build.js
jeffposnick: There's a back-and-forth about minification in the dev build in https://github.com/GoogleChrome/workbox/pull/594#issuecomment-305903950

packages/workbox-broadcast-cache-update/test/sw/broadcast-update.js
philipwalton: That's how I did the rest of them, so I can do that here as well. BTW, the syntax is slightly more verbose when referencing a property of the error rather than the message:

```js
expect(() => {
  broadcastUpdate();
}).to.throw().with.property('name', 'assertion-failed');
```

philipwalton: Hmmm, actually it looks like this error doesn't have a name:

```
Error: expected [TypeError: Cannot match against 'undefined' or 'null'.] to have property 'name' of 'assertion-failed', but got 'TypeError'
    at Context.it (http://localhost:3000/packages/workbox-broadcast-cache-update/build/test/sw-bundle.js:1441:24)
```

utils/build.js
philipwalton: Added my comments there. I'm in favor of *not* minifying in dev mode.

packages/workbox-background-sync/test/browser/background-sync-queue-plugin.js
gauntface: If you could make it consistent, that would be awesome.

packages/workbox-broadcast-cache-update/test/sw/broadcast-update.js
gauntface: This looks like it may be an uncaught error.

packages/workbox-precaching/test/sw/failing-cache-tests.js
gauntface: I would prefer, but it should build, if not that's a little more worrying.

utils/build.js
gauntface: Personally, I'm concerned developers will be shipping dev mode because it's a pain to switch between the files and if so we should probably minify.

templates/test-sw.hbs
gauntface: You have the use case for the bundling, my gut would be coming up with something than makes sense to you and following it.

Off the top of my head, have "bundled" under browser and sw directories, ignoring those from the test list, building them into browser and sw respectively as _bundled.js or something similar and making sure they are ignored with gitignore.

templates/test-sw.hbs
gauntface: can't argue with consistency. keep it in.

gauntface:
@philipwalton 

For 1 - yes go for consistency and yes to using function() everywhere, just to avoid weirdness if anyone tries to use describe() etc with timeouts or retries.

For 2 - We aren't doing away with package builds in the future, UMD will still be bundles of the packages, they just won't bundle the shared dependencies and how are they are used / served will be different. All of this is run through webdriver tests. The CLI and Build tests which are "closer to integration tests" will catch some errors, but the individual packages still catch different issues between the build types.

For now I'd rather add support to bundled tests along side the current testing approach and not replace it with bundling (i.e. forcing our tests to be bundled).

I would be happy to look at our testing infra in a different light for V3 but wouldn't want to make this big of a change at this point.

packages/workbox-background-sync/test/browser/background-sync-queue.js
philipwalton: Done.

packages/workbox-background-sync/test/browser/background-sync-queue.js
philipwalton: Done.

packages/workbox-background-sync/test/browser/queue-utils.js
philipwalton: It is in the background sync tests. I'll file an issue for @prateekbh to fix and make consistent.

packages/workbox-background-sync/test/browser/queue-utils.js
philipwalton: Done.

packages/workbox-broadcast-cache-update/test/sw/broadcast-update.js
philipwalton: Ahh, look like the `broadcastUpdate` function wasn't setting a default param. I changed:

```js
function broadcastUpdate({channel, cacheName, url, source}) {
```

to:

```js
function broadcastUpdate({channel, cacheName, url, source} = {}) {
```

packages/workbox-precaching/test/sw/failing-cache-tests.js
philipwalton: I reverted it back to how it was and kept the `.skip()`.

philipwalton:
> For now I'd rather add support to bundled tests along side the current testing approach and not replace it with bundling (i.e. forcing our tests to be bundled).

Ok, I think this is still possible. AFAICT I don't think any of my changes prohibit this. If you'd like me to revert any of the tests I've converted to bundling, let me know as I'm not sure which ones in particular you're concerned about.

Otherwise I think I've made all the requested changes, so pending the above we should be able to merge.

gauntface:
@philipwalton workbox-precaching is the only one I would like to be unbundled (largely because I have to maintain it and the un-bundling has helped me catch bugs in my tests). Will leave the other modules to @jeffposnick's discretion as the main author and fan of es2015 modules ;)

On a side not, I'd be curious if any of the browser tests could just take advantage of `script type="module"`. But that's for a future exploration and definitely not for this.

templates/test-browser.hbs
```diff
@@ -55,13 +55,10 @@
...
       chai.should();
-      const expect = chai.expect;
+      window.expect = chai.expect;
     </script>
 
-    <!-- Package-specific tests matching packages/<pkg>/test/browser/*.js -->
-    {{#each scripts}}
-      <script src="{{this}}"></script>
-    {{/each}}
+    <script src="/packages/{{pkg}}/build/test/browser-bundle.js"></script>
```
gauntface: Won't this change need to be undone to support multiple files?

templates/test-sw.hbs
```diff
@@ -44,22 +44,22 @@
...
     </script>
 
     <!-- Package-specific tests matching packages/<pkg>/test/browser/*.js -->
 
     <script>
     describe('{{pkg}} > test/sw/*.js', function() {
-      {{#each scripts}}
-      it(`should run '{{this}}' sw tests`, function() {
-        return window.goog.mochaUtils.registerServiceWorkerMochaTests('{{this}}');
+      it(`should run sw tests`, function() {
```
gauntface: Same as browser tests.

templates/test-browser.hbs
philipwalton: I was thinking all modules would have a bundle, but the `{{#each scripts}}` line above would allow it to also load other script files.

packages/workbox-background-sync/test/browser/background-sync-queue-plugin.js
philipwalton: It's consistent everywhere except background-sync. I'm going to open an issue for @prateekbh to do that.

templates/test-sw.hbs
philipwalton: These template changes work both for packages whose tests I changed as well as packages whose tests I didn't change, so I think it's good enough for now.

philipwalton:
@gauntface I reverted all test changes to workbox-precaching and the tests still pass, so I think we should be good. PTAL.

jeffposnick:
(I'm good if @gauntface is good.)

templates/test-browser.hbs
gauntface: Is this actually running all tests or is this just running the browser bundle for the modules that have it?

templates/test-sw.hbs
gauntface: Same as above - is this running all tests or just the bundle tests?

philipwalton:
Ahh, yes, sorry!

I updated the code to just use bundling for all modules (including precache), but since precache doesn't have any `import` statements, the bundling process is effectively just a concat, so it should be functionally identical to how it was before.

gauntface:
@philipwalton what are your plans for this?

philipwalton:
@gauntface Just waiting for an LGTM :)

gauntface:
The service worker tests are running on precache.

gauntface:
aren't*

philipwalton:
Re-reverted the test files in precaching.

PR #694, opened 2017-07-20
Fixes a bug with cacheFirst handler + maxAgeSeconds + CORS request

### LINKED ISSUE #691: Issue with cacheFirst + maxAgeSeconds for CORS resource
**Library Affected**:
*workbox-cache-expiration*

**Browser & Platform**:
Chrome, and others

@pbakaus reported an issue in which a route for a CORS resource that uses `cacheFirst` + `maxAgeSeconds` isn't behaving like it should. The configuration is https://github.com/ampproject/amp-publisher-sample/blob/master/amp-pwa-reader/src/sw.js#L18, and a live example is at https://amp.cards/

From what I can tell, the original response is reused indefinitely, instead of expiring after 5 minutes have passed.

jeffposnick:
R: @addyosmani @gauntface
CC: @pbakaus

Fixes #691

This fixes a bug that would prevent a response used in a cache-first handler from properly triggering `maxAgeSeconds`-based expiration iff that response lacked a `Date:` header. A CORS-response without the `Date:` header specifically whitelisted via [`Access-Control-Expose-Headers`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Access-Control-Expose-Headers) could trigger this scenario, as @pbakaus encountered.

This also changes the method signature of the `cacheWillMatch` lifecycle callback to be `async` (i.e. return a `Promise`), since we now want to wait for it to complete before returning a cached response to the client (to ensure that any expiration that's necessary gets performed). That should (arguably) be considered a breaking change, since `cacheWillMatch` is part of the public interface.

### DIFF
--- a/packages/workbox-cache-expiration/src/lib/cache-expiration.js
+++ b/packages/workbox-cache-expiration/src/lib/cache-expiration.js
@@ -161,11 +161,17 @@ class CacheExpiration {
         }
 
         const parsedDate = new Date(dateHeader);
+        const headerTime = parsedDate.getTime();
         // If the Date header was invalid for some reason, parsedDate.getTime()
-        // will return NaN, and the comparison will always be false. We want
-        // invalid dates to be treated as fresh. In order to ensure this
-        // behavior, we need to return the negation of this comparison.
-        return !((parsedDate.getTime() + (this.maxAgeSeconds * 1000)) < now);
+        // will return NaN. We want to treat that as a fresh response, since we
+        // assume fresh unless proven otherwise.
+        if (isNaN(headerTime)) {
+          return true;
+        }
+
+        // If we have a valid headerTime, then our response is fresh iff the
+        // headerTime plus maxAgeSeconds is greater than the current time.
+        return (headerTime + (this.maxAgeSeconds * 1000)) > now;
       } else {
         // TODO (jeffposnick): Change this method interface to be async, and
         // check for the IDB for the specific URL in order to determine
--- a/packages/workbox-cache-expiration/src/lib/cache-expiration.js
+++ b/packages/workbox-cache-expiration/src/lib/cache-expiration.js
@@ -131,21 +131,22 @@ class CacheExpiration {
    * be used.
    *
    * @param {Object} input
+   * @param {string} input.cacheName Name of the cache the responses belong to.
    * @param {Response} input.cachedResponse The `Response` object that's been
    *        read from a cache and whose freshness should be checked.
    * @param {Number} [input.now] A timestamp.
    *
    * Defaults to the current time.
-   * @return {boolean} Either `true` if the response is fresh, or `false` if the
-   * `Response` is older than `maxAgeSeconds` and should no longer be
-   * used.
+   * @return {Promise<boolean>} Either `true` if the response is fresh, or
+   * `false` if the `Response` is older than `maxAgeSeconds` and should no
+   * longer be used.
    *
    * @example
    * expirationPlugin.isResponseFresh({
    *   cachedResponse: responseFromCache
    * });
    */
-  isResponseFresh({cachedResponse, now} = {}) {
+  async isResponseFresh({cacheName, cachedResponse, now} = {}) {
     // Only bother checking for freshness if we have a valid response and if
     // maxAgeSeconds is set. Otherwise, skip the check and always return true.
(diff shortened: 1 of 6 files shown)


jeffposnick:
Actually, as an alternative to changing `cacheWillMatch` to be `async`, I could keep things as-is without affecting the real-world behavior for most scenarios. The downside is that it could introduce some flakiness into the new test and require longer timeouts to reduce the risk of that flakiness. That might be an acceptable trade-off, though.

packages/workbox-cache-expiration/src/lib/cache-expiration.js
```diff
@@ -165,6 +166,8 @@ class CacheExpiration {
           // Only return false if all the conditions are met.
           return false;
         }
+      } else {
+        this.expireEntries({cacheName, now});
```
gauntface: If there is no maxAgeSeconds, can we really regard the entry as fresh?

I'm just slightly confused at what this is solving.

jeffposnick: It's a little hard to line up the braces, but this `else` is associated with `if (dateHeader)`. I'll clarify in a comment what's going on.

gauntface: Shouldn't this return false?

packages/workbox-cache-expiration/src/lib/cache-expiration.js
```diff
@@ -159,15 +162,31 @@ class CacheExpiration {
 
         const parsedDate = new Date(dateHeader);
         // If the Date header was invalid for some reason, parsedDate.getTime()
-        // will return NaN, and the comparison will always be false. That means
-        // that an invalid date will be treated as if the response is fresh.
-        if ((parsedDate.getTime() + (this.maxAgeSeconds * 1000)) < now) {
-          // Only return false if all the conditions are met.
-          return false;
-        }
+        // will return NaN, and the comparison will always be false. We want
```
gauntface: Why not just check for NaN and treat as a special case? This negation is likely to cause confusion.

packages/workbox-cache-expiration/test/browser/end-to-end.js
```diff
@@ -37,4 +41,40 @@ describe(`End to End Test of Cache Expiration`, function() {
 
     expect(keys).to.have.length(EXPECTED_CACHE_SIZE);
   });
+
+  it(`should work properly when a cache-first strategy + maxAgeSeconds is used, and responses lack a Date header`, async function() {
+    // See the comment later about a source of potential flakiness.
+    this.retries(2);
```
gauntface: you should probably set the timeout of this test case to 5 or 6 with `this.timeout(6 * 1000);`

packages/workbox-cache-expiration/src/lib/cache-expiration.js
addyosmani: I found opting for the negation strategy here a little hard to parse. I'm similarly having trouble reasoning why we're not just special casing NaN per Matt's comment.

jeffposnick: Will switch.

jeffposnick:
PTAL.

packages/workbox-cache-expiration/src/lib/cache-expiration.js
```diff
@@ -161,11 +161,17 @@ class CacheExpiration {
...
-        // invalid dates to be treated as fresh. In order to ensure this
-        // behavior, we need to return the negation of this comparison.
-        return !((parsedDate.getTime() + (this.maxAgeSeconds * 1000)) < now);
+        // will return NaN. We want to treat that as a fresh response, since we
+        // assume fresh unless proven otherwise.
+        if (isNaN(headerTime)) {
+          return true;
+        }
+
+        // If we have a valid headerTime, then our response is fresh iff the
```
addyosmani: iff -> if

addyosmani:
Smallest nit in the world. LGTM otherwise.

packages/workbox-cache-expiration/src/lib/cache-expiration.js
jeffposnick: https://en.wikipedia.org/wiki/If_and_only_if

PR #697, opened 2017-07-24
fixing race condition

prateekbh:
R: @philipwalton @gauntface 

Fixes race condition in initializing `background-sync-queue`


packages/workbox-background-sync/test/browser/request-queue.js
```diff
@@ -30,6 +30,14 @@ describe('request-queue tests', () => {
     callbacks,
   });
 
+  const getDelayedPromise = () => {
```
philipwalton: You have a very similar test method with a different name in another test file:
https://github.com/GoogleChrome/workbox/blob/master/packages/workbox-background-sync/test/browser/background-sync-queue.js#L18-L24

I wonder if it makes sense to wait until my changes in #690 get merged so you don't have to repeat yourself here, and you can just import a common utility method that does this.

@gauntface WDYT?

packages/workbox-background-sync/test/browser/request-queue.js
```diff
@@ -30,6 +30,14 @@ describe('request-queue tests', () => {
...
+    return new Promise((resolve)=>{
+      setTimeout(()=>{
```
philipwalton: Please use consistent spacing for arrow functions here and in line 34. There should be a single whitespace character around the arrow, e.g. `() => {`.

packages/workbox-background-sync/test/browser/request-queue.js
```diff
@@ -69,7 +77,8 @@ describe('request-queue tests', () => {
 
   it('push is working', async () => {
     callbacks.requestWillEnqueue = sinon.spy();
-
+    const startTime = Date.now();
```
philipwalton: Use `performance.now()` here and throughout the file instead of `Date.now()`.

packages/workbox-background-sync/test/browser/request-queue.js
gauntface: It does make sense to share this because there are three cases of this within a short space of time.

packages/workbox-background-sync/src/lib/request-queue.js
```diff
@@ -96,6 +96,7 @@ class RequestQueue {
    * @private
    */
   async push({request}) {
+    await this._initializationPromise;
```
gauntface: does it make sense to have a getter for the current queue that does this await?

MY main concern here is that methods exist that use `this._queue` but aren't awaiting on the `initializationPromise` (i.e. `saveQueue()`)

philipwalton: I don't think we can use a getter because this is async, but +1 to all methods that reference `this._queue` should await its initialization.

gauntface: By getter I meant an internal method _getQueue() that was async

packages/workbox-background-sync/test/browser/request-queue.js
prateekbh: is this have to be taken care manually every time? just curious, is there a better way?

packages/workbox-background-sync/src/lib/request-queue.js
prateekbh: my bad, i'll add it to `saveQueue` as well.
Also would adding it to saveQueue be enough? where would we want to use an async getter?

gauntface: With an async method like `_getQueue()` you'd never reference `this._queue` and always use the `_getQueue` method, this means that no one can trip up like has been done with `saveQueue()`.

prateekbh: coolio, will do it today

prateekbh: @gauntface quick question: method
`get queue() {...}` is only used at one place, do you think making this async itself will be useful in longer term?

packages/workbox-background-sync/test/browser/request-queue.js
prateekbh: @philipwalton is there an alternate approach to do this now?

packages/workbox-background-sync/src/lib/request-queue.js
gauntface: @prateekbh Do what ever you think it best. It was more of a suggestion based off of reading the current code which references the queue multiple times and that we already missed a place where the initialisation promise should have been waited on but wasn't because it's not obvious where it is needed. A getQueue() method for all uses of the queue will force the waiting on the initialisation without any real thought from the developers of this module.

prateekbh:
@philipwalton @gauntface updated the PR

packages/workbox-background-sync/src/lib/request-queue.js
gauntface: This shouldn't be needed if you use `this.getQueue()` instead of referencing `this._queue` directly.

packages/workbox-background-sync/test/browser/request-queue.js
gauntface: Yes it does. We had the prettier discussion, but no one drove it through to be useful in our repo with our eslint config.

If you want to use that and can get it working, then feel free to add it. Until then, it's manual.

packages/workbox-background-sync/test/browser/request-queue.js
```diff
@@ -106,15 +115,16 @@ describe(`request-queue`, function() {
       expect(callbacks.requestWillEnqueue.calledOnce).to.be.true;
       expect(callbacks.requestWillEnqueue.calledWith(sinon.match.has('request')))
           .to.be.true;
-
+      expect(performance.now() - startTime).to.be.above(800);
```
gauntface: Why was this added and where did 800 come from? This feels super error prone (i.e. Travis is going to be very different to your machine in terms of performance).

gauntface: Can you please make use of sinon clock to very reliably add do what ever timeout and clock stuff you want.

http://sinonjs.org/releases/v3.2.1/fake-timers/

prateekbh: @gauntface i have this `getDelayedPromise` which resolves after 800ms. This was added to make sure that the function does not proceed before the initialization promise is resolved

gauntface: @prateekbh this will just introduce flakey tests. Please move 800 out into a static VARIABLE and please use startimers to force a 800ms step forward regardless of actual time.

packages/workbox-background-sync/test/browser/request-queue.js
philipwalton: @prateekbh not that I know of, I just meant it probably made sense to wait until my changes were merged before spending time on this, since I changed a lot of stuff. But that's happened now.

packages/workbox-background-sync/src/lib/request-queue.js
philipwalton: Yes, we should remove the `get queue()` method altogether and only use `this.getQueue()`.

packages/workbox-background-sync/src/lib/request-queue.js
```diff
@@ -96,6 +96,7 @@ class RequestQueue {
    * @private
    */
   async push({request}) {
```
philipwalton: I think this method should be renamed to `pushOntoQueue` to avoid confusion with `Array.prototype.push`. This is especially problematic since this method is async and `Array.prototype.push` is not.

For example, the following code is an easy bug to make:

```js
this._queue.push(...);
```

It should instead be:

```js
await this._queue.push(...);
```

addyosmani:
Marking as stalled until https://github.com/GoogleChrome/workbox/issues/769 lands.

prateekbh:
@addyosmani fixing #769 along with this itself

packages/workbox-background-sync/test/browser/request-queue.js
```diff
@@ -106,16 +116,17 @@ describe(`request-queue`, function() {
...
       delete callbacks.requestWillEnqueue;
     });
   });
 
   describe(`getRequestFromQueue method`, function() {
     it(`should get the proper Request back`, async function() {
       callbacks.requestWillDequeue = sinon.spy();
-
-      const hash = await queue.push({
+      const startTime = performance.now();
```
gauntface: Sorry for bugging you on this again. Could you add a comment explaining this and other performance.now() methods. I'm not sure why these are needed or whats it's checking. I'd still prefer it if these tests were use sinon clocks to avoid being flakey and slow.

packages/workbox-background-sync/test/browser/request-queue.js
```diff
@@ -92,30 +92,31 @@ describe(`request-queue`, function() {
   });
 
   describe(`push method`, function() {
-    it(`should push the Request given in the private array`, async function() {
+    it(`should push the given Request in the private array`, async function() {
       callbacks.requestWillEnqueue = sinon.spy();
-
+      const spy = sinon.spy(queue, 'getQueue');
```
gauntface: I think you want to use the spy to check it was called `expect(spy.callCount).to.equal(1)` at the end of this test.

packages/workbox-background-sync/test/browser/request-queue.js
```diff
@@ -92,30 +92,31 @@ describe(`request-queue`, function() {
...
       delete callbacks.requestWillEnqueue;
     });
   });
 
   describe(`getRequestFromQueue method`, function() {
     it(`should get the proper Request back`, async function() {
       callbacks.requestWillDequeue = sinon.spy();
-
-      const hash = await queue.push({
+      const spy = sinon.spy(queue, 'getQueue');
```
gauntface: Same check on spy here

packages/workbox-background-sync/test/browser/request-queue.js
```diff
@@ -124,7 +125,8 @@ describe(`request-queue`, function() {
       expect(reqData).to.have.all.keys(['request', 'config', 'metadata']);
       expect(callbacks.requestWillDequeue.calledOnce).to.be.true;
       expect(callbacks.requestWillDequeue.calledWith(reqData)).to.be.true;
-
+      expect(spy.called).to.be.true;
```
prateekbh: same here

packages/workbox-background-sync/test/browser/background-sync-queue.js
```diff
@@ -98,12 +98,12 @@ describe(`background sync queue`, function() {
...
       request: new Request('/__echo/counter'),
     });
-    expect(backgroundSyncQueue._queue.queue.length).to.equal(2);
+    expect((await backgroundSyncQueue._queue.getQueue()).length).to.equal(2);
     await backgroundSyncQueue.replayRequests();
     expect(responseAchieved).to.equal(2);
   });
 
-  it(`should rejected promise on replay failure`, async function() {
+  it(`should rejecte promise on replay failure`, async function() {
```
philipwalton: s/rejecte/reject/

addyosmani:
Landing and will directly update the spelling correction in master.

prateekbh:
Thanks @addyosmani 👍

PR #707, opened 2017-07-28
expect in bgSync tests

### LINKED ISSUE #705: workbox-background-sync tests are not consistent with all other packages
All packages in this monorepo follow a similar BDD style for mocha tests, with the exception of workbox-background-sync (as discussed in #690). Here are the differences I think should be fixed for the sake of consistency:

- We should use `expect()` instead of `chai.assert()` everywhere.
- We should follow the convention of starting `it()` blocks with a description that reads logically (e.g. `it('should push entries onto the queue')` rather than `it('test push')`.

prateekbh:
R: @philipwalton 

Fixes #705 

Adding expect and removing chai.assert in tests of background sync queue.


packages/workbox-background-sync/test/browser/background-sync-queue-plugin.js
```diff
@@ -12,19 +12,19 @@
  */
 
 /* eslint-env mocha, browser */
-/* global chai, workbox */
+/* global expect, workbox */
```
philipwalton: `expect` is in the project [ESLint config](https://github.com/GoogleChrome/workbox/blob/master/.eslintrc.js#L27), so it doesn't need to go here (or in any of the other test files).

packages/workbox-background-sync/test/browser/background-sync-queue-plugin.js
```diff
@@ -12,19 +12,19 @@
...
 
 'use strict';
 
 describe('background-sync-queue-plugin test', () => {
```
philipwalton: You probably shouldn't have the word "test" here. `describe` is supposed to describe the functionality of the module being tested, not the test itself.

packages/workbox-background-sync/test/browser/background-sync-queue.js
```diff
@@ -57,60 +57,65 @@ describe('background sync queue test', () => {
     await backgroundSyncQueue.cleanupQueue();
   });
 
-  it('check defaults', () => {
+  it('should get default values if nothing is given in constructor', () => {
```
philipwalton: It's not clear to me from this `describe`/`it` block what this testing.

I prefer to nest tests as @jeffposnick described in https://github.com/GoogleChrome/workbox/issues/705#issuecomment-318675972. Here's what I'd do in this case:

```js

describe(`BackgroundSyncQueue`, function() {
  describe(`constructor`, function() {
    it(`should use default values when nothing is passed`, function() {
      // ...
    });
  });

  describe(`someOtherMethod`, function() {
    it(`should do something the method does`, function() {
      // ...
    });
  });
});
```

packages/workbox-background-sync/test/browser/background-sync-queue.js
```diff
@@ -57,60 +57,65 @@ describe('background sync queue test', () => {
...
-      workbox.backgroundSync.test.Constants.defaultQueueName + '_1');
-    chai.assert.equal(defaultsBackgroundSyncQueue._queue._config.maxAge,
-      workbox.backgroundSync.test.Constants.maxAge);
-    chai.assert.equal(
-      JSON.stringify(
-        defaultsBackgroundSyncQueue._requestManager._globalCallbacks),
-      JSON.stringify({}));
+    expect(defaultsBackgroundSyncQueue._queue).to.be.an('object');
+    expect(defaultsBackgroundSyncQueue._requestManager).to.be.an('object');
+    expect(defaultsBackgroundSyncQueue._queue._queueName).to.be
```
philipwalton: Use `to.equal()` instead of `to.be.equal()` (here and throughout) as the latter doesn't read well.

packages/workbox-background-sync/test/browser/background-sync-queue.js
```diff
@@ -57,60 +57,65 @@ describe('background sync queue test', () => {
...
+    await backgroundSyncQueue.pushIntoQueue(
+        {request: new Request('/__echo/counter')});
+    await backgroundSyncQueue.pushIntoQueue(
+        {request: new Request('/__test/404')});
     try {
       await backgroundSyncQueue.replayRequests();
       throw new Error('Replay should have failed because of invalid URL');
     } catch (err) {
-      chai.assert.equal(404, err[0].status);
+      expect(404).to.be.equal(err[0].status);
```
philipwalton: `expect` comparisons should compare the expected result with the actual result, so this should be reversed.

```js
expect(err[0].status).to.equal(404);
```

packages/workbox-background-sync/test/browser/request-queue.js
```diff
@@ -30,62 +30,62 @@ describe('request-queue tests', () => {
...
 
-    chai.assert.isString(hash);
-    chai.assert.equal(queue._queue.length, queueLength + 1);
+    expect(hash).to.be.a('string');
+    expect(queue._queue.length).to.be.equal(queueLength + 1);
 
-    chai.assert(callbacks.requestWillEnqueue.calledOnce);
-    chai.assert(callbacks.requestWillEnqueue.calledWith(
-        sinon.match.has('request')));
+    expect(callbacks.requestWillEnqueue.calledOnce).to.be.true;
```
philipwalton: Hmm, minor nit, but it's unfortunate that the sinon assertion-style "called" properties don't read well with expect-style tests.

I'd use `to.be.ok` instead of `to.be.true` here and in the line below.

philipwalton:
Thanks for doing this! I made comments inline, but my other primary comment is we should probably hold off on merging this until #690 is merged, which makes some general test consistency changes that these changes should also follow.

It might be worth checking out the background sync changes in that PR to see what's coming.

packages/workbox-background-sync/test/browser/background-sync-queue.js
prateekbh: I see how that is making everything simpler, but that might block the fixes/ feature request for a longer time, can we skip this part and do it gradually instead?

philipwalton: Fine with me to do it gradually. I was mainly just showing how I like to structure my unit tests to more easily ensure the full public API space is covered.

prateekbh:
@philipwalton I have merged master with this and resolved conflicts..

packages/workbox-background-sync/test/browser/background-sync-queue-plugin.js
```diff
@@ -14,14 +14,15 @@
 */
 
 /* eslint-env mocha, browser */
-/* global chai */
+/* global workbox */
```
gauntface: This isn't needed.

packages/workbox-background-sync/test/browser/queue-utils.js
```diff
@@ -14,13 +14,13 @@
 */
 
 /* eslint-env mocha, browser */
-/* global chai */
+/* global workbox */
```
gauntface: Not needed. Workbox is not called anywhere in this code.

packages/workbox-background-sync/test/browser/request-manager.js
```diff
@@ -14,15 +14,15 @@
 */
 
 /* eslint-env mocha, browser */
-/* global chai */
+/* global sinon, workbox */
```
gauntface: Sinon should not be needed with the work to move everything into a single eslint file.

packages/workbox-background-sync/test/browser/response-manager.js
```diff
@@ -14,25 +14,27 @@
 */
 
 /* eslint-env mocha, browser */
-/* global chai */
+/* global workbox */
```
gauntface: Please delete.

jeffposnick:
👍

Issue #709, opened 2017-07-28
Normalize all RequestWrapper lifecycle methods to return a Promise<*>


jeffposnick:
**Library Affected**:
*workbox-runtime-caching*

The method signature for the public `cacheWillMatch()` interface is for it to return a `Response`, but there are use cases in which it would be ideal to perform some asynchronous operation prior to returning that response. In the next major release, we should change the interface to return a `Promise<Response>` instead.

The concrete use case for this is to perform an IndexedDB lookup prior to returning a response, which would have come in handy when fixing https://github.com/GoogleChrome/workbox/issues/691

addyosmani:
> In the next major release, we should change the interface to return a Promise<Response> instead.

I was about to ask about the use-case, but you have that covered :) Making this change in the next major release SGTM.

gauntface:
Should we make this consistent across all plugins, i.e. they all return a Promise?

jeffposnick:
That would probably future proof everything, yes.

addyosmani:
> Should we make this consistent across all plugins, i.e. they all return a Promise?

👍 Agree with going down this route for consistency sake. Do we need a separate issue for tracking doing this or did you want to reuse the same issue for it?

gauntface:
Reuse this one.

jeffposnick:
This was fixed in #736

PR #710, opened 2017-07-28
New library to support Range requests

### LINKED ISSUE #372: Add support for range requests?
As part of working on the sample media app I've had to handle range requests for large files, e.g. video files. Since there's no timeline for adding support for range requests to the Cache API so that a developer can get just part of a file, I wondered if we should consider making this a library in the helpers? Anyway, consider this a question that can be closed immediately if there's no appetite for it! 👍 

## The issue
A rough outline of the issue is:
1. Large files are stored in a cache.
2. A range request comes in to the SW, which matches the URL to cache.
3. The _entire file_ is pulled into memory and `.arrayBuffer()` is called to slice out the part needed to satisfy the range request
4. The ArrayBuffer is GC'd.

If the files are large then you can cause memory exhaustion and it's quite garbagey.

## The alternative approach
1. Large files are streamed into the cache in several chunks (chunk size being configurable)
2. A range request comes in to the SW, which matches the URL to cache.
3. The chunks needed to satisfy the range request are pulled and combined into a smaller ArrayBuffer.
4. Data is sliced out of this new ArrayBuffer to satisfy the range request.
5. The partial ArrayBuffer is GC'd.

FWIW the code I have for this, which would need to be generalized and so forth is here: https://github.com/GoogleChrome/sample-media-pwa/blob/master/src/client/scripts/ranged-response.js

jeffposnick:
R: @addyosmani @gauntface @philipwalton 
CC: @jakearchibald @beaufortfrancois @paullewis as an FYI.

Fixes #372

This adds a new library that exposes a static method which takes in a `Request` with a `Range:` header and a `Response`, and returns a new `Response` with the partial content specified via the `Range:` header (or an error `Response` if the `Range:` was invalid).

It also adds in a new plugin class, meant to be used statically, that implements `RequestWrapper`'s `cacheWillMatch` lifecycle method, meaning it could be used easily from within `workbox-sw`.

This relies on reading and manipulating the entire source response body as a `Blob`, which is not as efficient as it would be if there were full Streams API support across all service worker implementation. It also forgoes a method that @paullewis seemed to have pioneered, in which a larger source `Response` is broken up into smaller individual `Response` chunks before it's cached, meaning that each individual body is smaller. I did not want this library to require that the original `Response` be cached in that specific chunked format, and the Streams API approach seems like the long-term solution anyway.


packages/workbox-range-requests/src/index.js
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * - https://github.com/jakearchibald/range-request-test/blob/master/static/sw.js
+ * - https://github.com/GoogleChrome/sample-media-pwa/blob/master/src/client/scripts/ranged-response.js
+ */
+
+import {calculateEffectiveBoundaries, handleRangeRequest, parseRangeHeader}
+  from './lib/handle-range-request.js';
+import CachedRangeResponsePlugin from './lib/cached-range-response-plugin.js';
+
+export {
```
jeffposnick: There might be some controversy around naming, and this is also the first library that exposes body a plugin as a class and also some top-level standalone functions. I also don't know if there's an issue with this approach to exports and what @philipwalton is working on in #690

jakearchibald:
Reading it as a blob is actually really efficient in Chrome https://github.com/whatwg/fetch/issues/569

packages/workbox-range-requests/src/index.js
gauntface: Is there a use case for exposing the individual functions?

packages/workbox-range-requests/src/lib/cached-range-response-plugin.js
```diff
@@ -0,0 +1,73 @@
...
+ * {@link handleRangeRequest} function directly is recommended.
+ *
+ * Under the hood, this will call `handleRangeRequest()`, passing
+ * in the appropriate request and response objects.
+ *
+ * @example
+ * const workboxSW = new WorkboxSW();
+ *
+ * const cacheFirstRangeHandler = workboxSW.strategies.cacheFirst({
+ *   cacheName: 'range-requests',
```
gauntface: Is there a reason this can't go to the default cache?

packages/workbox-range-requests/src/lib/cached-range-response-plugin.js
```diff
@@ -0,0 +1,73 @@
...
+   * @param {Object} input
+   * @param {Request} input.request The original request, which may or may not
+   * contain a Range: header.
+   * @param {Response} input.cachedResponse The complete cached response.
+   * @return {Promise<Response>} If request contains a Range: header, then
+   * a Response with status 206 whose body is a subset of cachedResponse. If
+   * request does not have a Range: header, then cachedResponse is returned
+   * as-is.
+   */
+  static async cacheWillMatch({request, cachedResponse} = {}) {
```
gauntface: This is the first time I'm seeing this method used in a while and I'm struggling with what is expected.

i.e. should it return a "cached response" or a boolean to highlight if the cache will match a request or will it alter the request to match the request or ....

packages/workbox-range-requests/src/lib/handle-range-request.js
```diff
@@ -0,0 +1,188 @@
...
+ * @param {Response} input.response An original response containing the full
+ * content.
+ * @return {Promise<Response>} Either a `206 Partial Content` response, with
+ * the response body set to the slice of content specified by the request's
+ * `Range:` header, or a `416 Range Not Satisfiable` response if the
+ * conditions of the `Range:` header can't be met.
+ */
+export async function handleRangeRequest({request, response} = {}) {
+  try {
+    isInstance({request}, Request);
```
gauntface: Should the isInstance checks be moved outside the try/catch to ensure the intended error message is seen by the developer?

packages/workbox-range-requests/src/lib/handle-range-request.js
```diff
@@ -0,0 +1,188 @@
...
+      start: boundaries.start,
+      end: boundaries.end,
+    });
+
+    const slicedBlob = originalBlob.slice(effectiveBoundaries.start,
+      effectiveBoundaries.end);
+    const slicedBlobSize = slicedBlob.size;
+
+    const slicedResponse = new Response(slicedBlob, {
+      // See https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/206
```
gauntface: Could you extend the comment to:

```
// Status code 206 is for "Partial Content Reponse"
// See https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/206
```

packages/workbox-range-requests/src/lib/handle-range-request.js
```diff
@@ -0,0 +1,188 @@
...
+ * slice.
+ * @param {Number|null} input.end The offset to use as the end of the slice.
+ * @return {Object} An object with `start` and `end` properties, reflecting
+ * the effective boundaries to use given the size of the blob.
+ */
+export function calculateEffectiveBoundaries({blob, start, end} = {}) {
+  isInstance({blob}, Blob);
+  const blobSize = blob.size;
+
+  if (end > blobSize || start < 0) {
```
gauntface: Does end need to be less than the blob size? i.e. end >= blobSize?

packages/workbox-range-requests/src/lib/handle-range-request.js
```diff
@@ -0,0 +1,188 @@
...
+ * @private
+ * @param {Object} input
+ * @param {Blob} input.blob A source blob.
+ * @param {Number|null} input.start The offset to use as the start of the
+ * slice.
+ * @param {Number|null} input.end The offset to use as the end of the slice.
+ * @return {Object} An object with `start` and `end` properties, reflecting
+ * the effective boundaries to use given the size of the blob.
+ */
+export function calculateEffectiveBoundaries({blob, start, end} = {}) {
```
gauntface: Since this method is exported publicly, should these be private? Otherwise it won't have docs.

If it is public, can you add a description of the function.

packages/workbox-range-requests/src/lib/handle-range-request.js
```diff
@@ -0,0 +1,188 @@
...
+ */
+export function parseRangeHeader({rangeHeader} = {}) {
+  isType({rangeHeader}, 'string');
+
+  const normalizedRangeHeader = rangeHeader.trim().toLowerCase();
+  if (!normalizedRangeHeader.startsWith('bytes=')) {
+    throw ErrorFactory.createError('unit-must-be-bytes');
+  }
+
+  if (normalizedRangeHeader.includes(',')) {
```
gauntface: Could you add a comment on what this is checking for.

packages/workbox-range-requests/src/lib/handle-range-request.js
```diff
@@ -0,0 +1,188 @@
...
+ *           response: cachedResponse,
+ *         });
+ *       })
+ *     );
+ *   }
+ * });
+ */
+
+/**
+ * @private
```
gauntface: Should this be private since it's exported?

packages/workbox-range-requests/src/lib/handle-range-request.js
```diff
@@ -0,0 +1,188 @@
...
+
+    slicedResponse.headers.set('Content-Length', slicedBlobSize);
+    slicedResponse.headers.set('Content-Range',
+      `bytes ${effectiveBoundaries.start}-${effectiveBoundaries.end - 1}/` +
+      originalBlob.size);
+
+    return slicedResponse;
+  } catch (error) {
+    logHelper.warn({
+      message: `Unable to construct a sliced response; returning a 416 Range
```
gauntface: Would it be worth adding the range header to this string just to help with debugging in the future?

packages/workbox-range-requests/src/lib/handle-range-request.js
gauntface: I think this is not because end is inclusive, but not 100%

jeffposnick:
Sorry, I forgot that this is blocked on https://github.com/GoogleChrome/workbox/issues/709, i.e. we need the `cacheWillMatch` interface to change to return a promise. Adding the `chilling` label while I wait for that.

jeffposnick:
I've changed this PR to go against the `v2` branch, now that #709 is merged there.

packages/workbox-range-requests/src/index.js
jeffposnick: I'm going to see about dropping `calculateEffectiveBoundaries` and `parseRangeHeader` from the public exports as they were hack-y ways of doing testing. Presumably @philipwalton's new approach to bundling for tests will help here.

packages/workbox-range-requests/src/lib/cached-range-response-plugin.js
jeffposnick: Hopefully the new lifecycle method docs at https://github.com/GoogleChrome/workbox/blob/v2/packages/workbox-runtime-caching/src/lib/request-wrapper.js#L70 clarify that.

packages/workbox-range-requests/src/lib/handle-range-request.js
jeffposnick: Done.

packages/workbox-range-requests/src/lib/handle-range-request.js
jeffposnick: Done.

packages/workbox-range-requests/src/lib/handle-range-request.js
jeffposnick: The message should be in the Extra Details section, logged as `data.error`. I could add the error string to the top-most log message if you don't think that's sufficient.

packages/workbox-range-requests/src/lib/handle-range-request.js
jeffposnick: It's not exported from the top-level `index.js` anymore.

packages/workbox-range-requests/src/lib/handle-range-request.js
jeffposnick: It's not exported from the top-level `index.js` anymore.

packages/workbox-range-requests/src/lib/handle-range-request.js
jeffposnick: You should theoretically be able to get at it by examining the `request` that's logged in the `data.request` bit.

jeffposnick:
PTAL. Specifically, it should be cleaner now that @philipwalton's test bundling stuff means that I can avoid exporting things that are meant to be private.

packages/workbox-range-requests/src/index.js
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * If you are not already using Workbox, this library can be used in a
+ * standalone form via
+ * {@link handleRangeRequest|workbox.rangeRequests.handleRangeRequest}.
+ *
+ * If you are using Workbox for routing or request handling already, then
+ * {@link CacheRangeResponsePlugin} will offer the easiest integration.
+ *
+ * **Please note that the entire body of the response that's being manipulated
+ * is read into memory in order to create the final response. If your original
```
gauntface: Nit: Should we shorten and bold this? `*If you resource is large, this library should not be used.* This is due to the resource needing to be read into memory before being manipulated to respond to the request`

jeffposnick: So https://github.com/GoogleChrome/workbox/pull/710#issuecomment-318735722 suggests that this might not be a problem in Chrome, but it is in Firefox. I'll go with your language instead as at least a general warning, so that folks know to at least think about the behavior.

beaufortfrancois:
@jeffposnick Please ping me when workbox-range-requests library is publicly available so that we can promote it in https://github.com/google/WebFundamentals/pull/4744

packages/workbox-range-requests/src/index.js
jakearchibald: It doesn't read into memory in Firefox either, but it does read into a new place on disk, making it as slow as copying the whole blob.

jeffposnick:
@beaufortfrancois It'll be included in the `v2.0.0` release, planned for August 23.

PR #712, opened 2017-08-01
adding manifest transform for webpack public path

### LINKED ISSUE #689: workbox-webpack-plugin doesn't respect publicPath
**Library Affected**:
workbox-webpack-plugin

**Browser & Platform**:
Webpack 3.3.0, Node 6.9.1

The manifest generated by the plugin lists the file names only, not including any public path specified. Ideally, if the file is "vendor.js" and the public path is "dist/", the manifest entry should be for "dist/vendor.js". I'm unable to get the precache working using the plugin because of this; the requests are never picked up by the service worker and they're always brought down from the server.

Just as an example, this is what's currently being generated by the plugin for me:
```javascript
workboxSw.precache([
  {
    "url": "app.css",
    "revision": "7c6c1b5bc61102d9cdb3a3a64da0631f"
  },
  {
    "url": "app.js",
    "revision": "c49639a282ff7fc2b163f17fe7c8dd5a"
  },
  {
    "url": "immediate.css",
    "revision": "77fc2e53176fd54a733738ea9521c87c"
  },
  {
    "url": "styles.css",
    "revision": "11c6be3e42073db7cefae03c5b1cecf5"
  },
  {
    "url": "vendor.js",
    "revision": "52dd3b366cf368edc9be68234514dc5d"
  },
  {
    "url": "workbox-sw.js",
    "revision": "df86dfc69c6d017722ecb8a16d34c849"
  }
]);
```
And this is what I'd expect (and does work as expected):
```javascript
workboxSw.precache([
  {
    "url": "dist/app.css",
    "revision": "7c6c1b5bc61102d9cdb3a3a64da0631f"
  },
  {
    "url": "dist/app.js",
    "revision": "c49639a282ff7fc2b163f17fe7c8dd5a"
  },
  {
    "url": "dist/immediate.css",
    "revision": "77fc2e53176fd54a733738ea9521c87c"
  },
  {
    "url": "dist/styles.css",
    "revision": "11c6be3e42073db7cefae03c5b1cecf5"
  },
  {
    "url": "dist/vendor.js",
    "revision": "52dd3b366cf368edc9be68234514dc5d"
  },
  {
    "url": "dist/workbox-sw.js",
    "revision": "df86dfc69c6d017722ecb8a16d34c849"
  }
]);
```

prateekbh:

Fixes #689 and also fixes a bug for manifestTransforms


jeffposnick:
Not sure what the "[WIP]" indicates—is this ready for review?

prateekbh:
@jeffposnick nups, tests are remaining

addyosmani:
Let's try to consistently use `chillin` for work that isn't ready to be reviewed if we can

packages/workbox-build/src/lib/utils/filter-files.js
```diff
@@ -89,7 +89,8 @@ module.exports = (fileDetails, options) => {
 
   if (options.manifestTransforms) {
     if (Array.isArray(options.manifestTransforms)) {
-      manifestTransforms.concat(options.manifestTransforms);
+      manifestTransforms =
+          manifestTransforms.concat(options.manifestTransforms);
```
HenrikJoreteg: This bug just bit me too, would you folks take a PR that just fixed this? I already was about submit it, when i saw it was covered here. But... basically as it's written right now anything you pass in as a `manifestTransforms` option won't be run at all. Since `.concat` doesn't change an array in place. Effectively rendering this option useless.

HenrikJoreteg: i added it, just in case you want it: https://github.com/GoogleChrome/workbox/pull/715 

Feel free to toss it if you'd rather wait.

prateekbh: @HenrikJoreteg thank you so much, only this change was holding this PR back... was about to work on it, u saved me some effort :)

packages/workbox-webpack-plugin/test/node/e2etest.js
```diff
@@ -51,16 +51,10 @@ describe(`Tests for webpack plugin`, function() {
...
-    this.timeout(10 * 1000);
-
-    return fsExtra.remove(tmpDirectory);
-  });
-
   // Kill the web server once all tests are complete.
   after(function() {
     this.timeout(10 * 1000);
-
+    fsExtra.remove(tmpDirectory);
```
jeffposnick: Use `removeSync()` here (or return a promise chain that includes the `remove()` promise).

packages/workbox-webpack-plugin/test/node/e2etest.js
```diff
@@ -140,6 +134,69 @@ describe(`Tests for webpack plugin`, function() {
...
+    process.chdir(tmpDirectory);
+
+    const webpackConfig = {
+      entry: {
+        webpackEntry: path.join(tmpDirectory, 'webpackEntry.js'),
+      },
+      output: {
+        path: tmpDirectory,
+        filename: '[name].js',
+        publicPath: '/public',
```
jeffposnick: As per https://webpack.js.org/configuration/output/#output-publicpath, the `publicPath` value is just used as a string prefix, and should end in `/` to be a representative test case.

packages/workbox-webpack-plugin/test/node/e2etest.js
```diff
@@ -140,6 +134,69 @@ describe(`Tests for webpack plugin`, function() {
...
+
+    process.chdir(tmpDirectory);
+
+    const webpackConfig = {
+      entry: {
+        webpackEntry: path.join(tmpDirectory, 'webpackEntry.js'),
+      },
+      output: {
+        path: tmpDirectory,
+        filename: '[name].js',
```
jeffposnick: Would it be possible to include and output file that was a few levels deep in the directory structure, to confirm that this plugin works as intended when there are path separators in the generated asset? This will help identify problems that might exist when there's path matching on Windosw.

packages/workbox-webpack-plugin/index.js
```diff
@@ -62,6 +62,25 @@ class WorkboxBuildWebpackPlugin {
       config.globPatterns = ['**/*.{html,js,css}'];
     }
 
+    if (compilation.options.output.publicPath && !config.manifestTransforms) {
```
prateekbh: yeah I have some thought about this and I might be just too conservative in my approach.
So, if our transformation started having some conflict with a user land's transformation function, then there is no way to switch off our logic. but as I said, I might be just over thinking this... Thoughts?

jeffposnick: Running this transform before other user-specified transforms (by using `unshift()`) means that we give developers final control.

packages/workbox-webpack-plugin/test/node/e2etest.js
gauntface: Could it ever not include the slash? if so why not leave as is?

jeffposnick: It could exclude a trailing `/`, since the Webpack documentation is just a recommendation and isn't enforced by Webpack's build process.

If you use `publicPath: '/public'` in this test case, then the URL that should be present in the precache manifest, as per Webpack's documented behavior, is `'/publicwebpackEntry.js'` (which obviously looks weird).

packages/workbox-webpack-plugin/index.js
```diff
@@ -62,6 +62,25 @@ class WorkboxBuildWebpackPlugin {
...
+      const publicPath = compilation.options.output.publicPath;
+      const compiledAssets = [];
+      for (let key in compilation.assets) {
+        if (compilation.assets.hasOwnProperty(key)) {
+          compiledAssets.push(path.resolve(publicPath, './' + key));
```
goldhand: Don't need the `./` here. It's unnecessary because path.resolve, url.resolve and path.join will all treat `key` like its a directory anyway.

packages/workbox-webpack-plugin/index.js
```diff
@@ -62,6 +62,25 @@ class WorkboxBuildWebpackPlugin {
       config.globPatterns = ['**/*.{html,js,css}'];
     }
 
+    if (compilation.options.output.publicPath && !config.manifestTransforms) {
+      const publicPath = compilation.options.output.publicPath;
+      const compiledAssets = [];
```
goldhand: Is this an array of filesystem paths or web paths?

packages/workbox-webpack-plugin/index.js
```diff
@@ -62,6 +62,25 @@ class WorkboxBuildWebpackPlugin {
       config.globPatterns = ['**/*.{html,js,css}'];
     }
 
+    if (compilation.options.output.publicPath && !config.manifestTransforms) {
+      const publicPath = compilation.options.output.publicPath;
```
goldhand: favor destructuring if supported:
`const {publicPath} = compilation.options.output;`
(just being picky lol)

packages/workbox-webpack-plugin/index.js
```diff
@@ -62,6 +62,25 @@ class WorkboxBuildWebpackPlugin {
...
+      const compiledAssets = [];
+      for (let key in compilation.assets) {
```
goldhand: should be `const key` if you are not mutating it

packages/workbox-webpack-plugin/index.js
```diff
@@ -62,6 +62,25 @@ class WorkboxBuildWebpackPlugin {
...
+        if (compilation.assets.hasOwnProperty(key)) {
+          compiledAssets.push(path.resolve(publicPath, './' + key));
+        }
+      }
+      config.manifestTransforms= [(manifestEntries) =>
+          manifestEntries.map((entry) => {
+            if (compiledAssets.indexOf(path.resolve(publicPath, entry.url))
```
goldhand: Is this redundant? Doesn't checking for `compilation.assets[entry.url]` cover this?
I don't think you need the `compiledAssets` array, or do you?

packages/workbox-webpack-plugin/index.js
```diff
@@ -62,6 +62,25 @@ class WorkboxBuildWebpackPlugin {
       config.globPatterns = ['**/*.{html,js,css}'];
     }
 
+    if (compilation.options.output.publicPath && !config.manifestTransforms) {
+      const publicPath = compilation.options.output.publicPath;
+      const compiledAssets = [];
+      for (let key in compilation.assets) {
+        if (compilation.assets.hasOwnProperty(key)) {
```
goldhand: Should use `Object.prototype.hasOwnProperty.call(compilation.assets, key)`

packages/workbox-webpack-plugin/index.js
jeffposnick: Regarding how publicPath should be treated, my understanding based on https://webpack.js.org/configuration/output/#output-publicpath is that the value is treated as a string prefix, not as a value that should be joined with `url.resolve()` or `path.join()`. If the value ends in `/` as recommended, then this will be equivalent, but if the value doesn't end in `/` then you'll end up interpreting `publicPath` different in the precache manifest than how it's being interpreted by Webpack.

@goldhand was this a consideration when you wrote `sw-precache-webpack-plugin` (which uses `url.resolve()`)?

packages/workbox-webpack-plugin/index.js
addyosmani: I think you're right that just using `compilation.assets[entry.url]` should be enough here.

packages/workbox-webpack-plugin/index.js
addyosmani: Hah. Nothing wrong with that. I'm fine with us switching this over to destructuring.

packages/workbox-webpack-plugin/index.js
goldhand: I see what you mean @jeffposnick, and I agree, we should mirror webpack's usage. I haven't found examples of where publicPath is used in webpack yet but I think an example would be helpful in making sure its the same.

packages/workbox-webpack-plugin/test/node/e2etest.js
prateekbh: wouldn't it be nice if we can handle this for devs?
its pretty easy for anyone to forget the trailing `/`

packages/workbox-webpack-plugin/index.js
prateekbh: web paths

packages/workbox-webpack-plugin/test/node/e2etest.js
jeffposnick: If folks who know more about Webpack think that it's safe to interpret the `publicPath` differently than how it's documented, then I'll defer to that. But it smells funny to me.

goldhand: If you omit the trailing slash in your publicPath you should see other problems with your webpack build (not relating to workbox). If for some reason you needed to omit the trailing slash (I can't think of why you would but that doesn't mean there isn't any cause to) then this would not behave as expected and would be frustrating.

Should test different publicPaths and make sure we handle it the same way webpack does imo.

packages/workbox-webpack-plugin/test/node/e2etest.js
prateekbh: I tried addressing fixing all comments except this, I feel multi level can be achieved only by specifying configuration in user land. not sure about a good way of testing it and also about the specific thing to be tested.

We're mainly appending the public path which is already tested, any path beyond that would come from user land.

jeffposnick:
Thanks for addressing my feedback, @prateekbh!

Unless anyone else has unresolved issues, I think we're good to merge.

PR #719, opened 2017-08-03
V3 core

gauntface:
R: @jeffposnick @addyosmani @gauntface @philipwalton 

This is a PR start to build up for V3 so please hammer on this for anything you see and don't like / disagree with.

Notes in terms of changes...........

## Eslint

Been slightly more strict on where console is used - this is solely to force the use of a log helper which can add prefixs and colors to helper figure out where logs are coming from in build process and tests.

I moved us off of the gulp-eslint plugin as it couldnt be configured to use our gitignore (i.e. we'd have to duplicate ignores in eslintignore or gulp.src). I checked that we fail the build if there is a lint problem.

## Testing

To enable @philipwalton's use case of having access to raw classes I plan on tweaking his approach of creating bundle files where our build process will look for files under `test/bundle/node/`, `test/bundle/browser/` and `test/bundle/sw/` this will result in a built bundle consisting of all JS files under this directory and the test runner (which I will add in at a later stage) will basically add scripts from `browser/`, `node/` and `sw/` along side the bundles in the respective environment.

**Please Comment On This**: The bundled tests are currently built and run with no NODE_ENV as well as a NODE_ENV of 'production'. The idea here being that if code changes it's behavior between environments, the tests can check the NODE_ENV and alter their behavior accordingly (i.e. ensure something exists or doesn't exist based on the NODE_ENV). I'm happy to strip this out so we only build one set of tests OR we can use flags to determine what node_env is run OR add multiple gulp tasks (one for prod one for dev). Just wondering what people would want / not want for this.

## Gulp

I've switched to gulp 4 so we can get better Promise support.

I took the idea of task runner and altered it into packageRunner, the idea is that it wraps functions that perform the gulp work into an array of functions, one item for each package it should run on (i.e. --project vs no --project). This works well with gulp.series and gulp.parallel - I added a big readme to the file to explain.

### DIFF
--- a/gulp-tasks/lint.js
+++ b/gulp-tasks/lint.js
@@ -1,0 +1,10 @@
+const gulp = require('gulp');
+const spawn = require('./utils/spawn-promise-wrapper');
+
+// Use npm run lint to ensure we are using local eslint
+gulp.task('lint', () => {
+  return spawn('npm', ['run', 'lint'])
+  .catch((err) => {
+    throw new Error(`[Workbox Error Msg] 'gulp lint' discovered errors.`);
+  });
+});
--- a/gulp-tasks/test.js
+++ b/gulp-tasks/test.js
@@ -1,0 +1,44 @@
+const gulp = require('gulp');
+const mocha = require('gulp-spawn-mocha');
+const path = require('path');
+const oneLine = require('common-tags').oneLine;
+
+const constants = require('./utils/constants');
+const packageRunnner = require('./utils/package-runner');
+const logHelper = require('./utils/log-helper');
+const pkgPathToName = require('./utils/pkg-path-to-name');
+
+const runBundledTests = (packagePath, env) => {
+  logHelper.log(oneLine`
+    Running Test Bundles for
+    ${logHelper.highlight(pkgPathToName(packagePath))}.
+  `);
+
+  const bundleTestDirectory = path.posix.join(
+    packagePath, 'test', constants.BUNDLE_BUILD_DIRNAME);
+  return gulp.src(
+    path.posix.join(bundleTestDirectory, env, '*.js'), {read: false})
+  .pipe(mocha());
+};
+
+gulp.task('test:node', gulp.series(
+  packageRunnner('test:node [bundled tests]', runBundledTests, 'node')
+));
+
+gulp.task('test:browser', () => {
+  // TODO: This needs implementing
+  return Promise.resolve();
+});
+
+gulp.task('test:sw', () => {
+  // TODO: This needs implementing
+  return Promise.resolve();
+});
+
+gulp.task('test', gulp.series(
+  'build-test-bundles',
+  'test:node',
+  'test:browser',
+  'test:sw',
+  'lint'
+));
--- a/gulp-tasks/utils/package-runner.js
+++ b/gulp-tasks/utils/package-runner.js
@@ -1,0 +1,99 @@
+const path = require('path');
+const glob = require('glob');
+const oneLine = require('common-tags').oneLine;
+
+const pkgPathToName = require('./pkg-path-to-name');
+
+/**
+ * This methods only purpose is to call a function with
+ * the package path that needs to be acted upon.
+ *
+ * If we ran gulp as `gulp build` we would want the
+ * 'build' task to run against all packages.
+ *
+ * If we ran gulp as `gulp build --project workbox-core`
+ * we would want the 'build' task to run against `workbox-core`
+ * **only**.
+ *
+ * This method simplifies how we would write the 'build'
+ * task and should be the only way we write functions.
+ *
+ * ```javascript
+ * gulp.task('build',
+ *   gulp.series(
+ *     packageRunner(someFunction, someArg, someMoreArgs)
+ *   )
+ * );
+ * ```
+ *
+ * Package runner will return an array of functions that
+ * will call `someFunction` where the first argument
+ * is the absolute path to a package,
+ * like `/user/matt/workbox/packages/workbox-core` and
+ * the remaining arguments will be whatever is passed to
+ * packageRunner, in the above sample this would be
+ * `someArg` and `someMoreArgs`.
+ *
+ * In the example above, `someFunc` would be written as:
+ *
+ * ```javascript
+ * function someFunction(packagePath, arg1, arg2) {
+ *   // Return a Promise
+ *   return Promise.resolve().then(() => ...)
+ *
+ *   // OR
+ *
+ *   // Return a gulp stream
+ *   return gulp.src(path.posix.join(packagePath, 'example-dir', '*.js'))
+ *   .pipe(...)
+ *   .pipe(gulp.dest(...));
+ * }
+ * ```
+ *
+ * You can use this with both `gulp.series` and
+ * `gulp.parallel`.
+ *
+ * ```javascript
+ * gulp.series(packageRunner(someFunction))
+ * gulp.parallel(packageRunner(someFunction))
+ * ```
+ *
+ * If you need to call the runner with multiple functions
+ * you can do this and gulp will merge the arrays returned
+ * by the runner.
+ *
+ * For example:
+ *
+ * ```javascript
+ * gulp.parallel(
+ *   packageRunner(someFunction, true),
+ *   packageRunner(someFunction, false),
+ * )
+ * ```
+ *
+ * If we run the above, it would call `someFunction` for all
+ * packages in workbox twice, once with argument `true` and
+ * once with `false`. `gulp.parallel` will run all of these
+ * calls to `someFunction` in parallel.
+ */
+module.exports = (displayName, func, ...args) => {
+  const packagePaths = glob.sync(
+    `packages/${global.projectOrStar}/package.json`, {
+      absolute: true,
+    }
+  );
+
+  // `packageRoots` is an array of absolute paths to packages
+  // that should have func run against them.
+  const packageRoots = packagePaths.map((packagePath) =>
+    path.dirname(packagePath));
+
+  return packageRoots.map((packageRootPath) => {
+    const wrappedFunc = () => func(packageRootPath, ...args);
+    wrappedFunc.displayName = oneLine`
+      ${displayName}
+      (${pkgPathToName(packageRootPath)})
+      ${JSON.stringify(args)}`;
+    return wrappedFunc;
+  });
+};
--- a/infra/utils/generate-variant-tests.js
+++ b/infra/utils/generate-variant-tests.js
@@ -1,0 +1,22 @@
+/**
+ * This is a helper function that will auto-generate mocha unit tests
+ * for various inputs.
+ *
+ * @param {String} itTitle This is the title that will be passed to the it()
+ * function. The variant will be added to the end of this title to help
+ * idenfity the failing test.
+ * @param {Array<Objects>} variants This should be all the variations of the
+ * test you wish to generate.
+ * @param {function} func This is the function that will be called, with a
+ * variant as the only argument. This function should perform the desired test.
+ */
+export default (itTitle, variants, func) => {
+  variants.forEach((variant) => {
+    // We are using function() {} here and NOT ARROW FUNCTIONS
+    // to work with Mocha's binding for tests.
+    it(`${itTitle}. Variant: '${JSON.stringify(variant)}'`,
+      function() {
+        return func(variant);
+      });
+  });
+};
--- a/packages/workbox-core/package.json
+++ b/packages/workbox-core/package.json
@@ -1,0 +1,16 @@
+{
+  "name": "workbox-core",
+  "version": "1.2.0",
+  "license": "Apache-2.0",
+  "author": "Google's Web DevRel Team",
+  "description": "This module is used by a number of the other workboxjs.org modules to share common code.",
+  "repository": "googlechrome/workbox",
+  "bugs": "https://github.com/googlechrome/workbox/issues",
+  "homepage": "https://github.com/GoogleChrome/workbox/tree/master/packages/workbox-core",
+  "keywords": [
+    "workbox",
+    "workboxjs",
+    "service worker",
+    "sw"
+  ]
+}
--- a/packages/workbox-core/src/controllers/WorkboxCore.js
+++ b/packages/workbox-core/src/controllers/WorkboxCore.js
@@ -1,0 +1,24 @@
+import Assert from '../utils/Assert';
+import LogHelper from '../utils/LogHelper';
+
+/**
+ * WorkboxCore shares code across Workbox modules.
+ */
+export default class WorkboxCore {
+  /**
+   * @private
+   */
+  get INTERNAL() {
+    if (!this._internal) {
+      this._internal = {
+        logHelper: new LogHelper(),
+      };
+
+      if (process.env.NODE_ENV !== 'production') {
+        this._internal.assert = new Assert();
+      }
+    }
+
+    return this._internal;
+  }
+}
--- a/packages/workbox-core/test/bundle/node/controllers/test-WorkboxCore.js
+++ b/packages/workbox-core/test/bundle/node/controllers/test-WorkboxCore.js
@@ -1,0 +1,32 @@
+import {expect} from 'chai';
+
+import WorkboxCore from '../../../../src/controllers/WorkboxCore';
+
+describe(`WorkboxCore - ${process.env.NODE_ENV}`, function() {
+  describe(`constructor`, function() {
+    it(`should construct with no arguments`, function() {
+      new WorkboxCore();
+    });
+  });
+
+  describe(`INTERNAL.*`, function() {
+    it(`should expose INTERNAL`, function() {
+      const core = new WorkboxCore();
+      expect(core.INTERNAL).to.exist;
+    });
+
+    it(`should expose logHelper`, function() {
+      const core = new WorkboxCore();
+      expect(core.INTERNAL.logHelper).to.exist;
+    });
+
+    it(`should expose assertions only in non-production builds`, function() {
+      const core = new WorkboxCore();
+      if (process.env.NODE_ENV === 'production') {
+        expect(core.INTERNAL.assert).to.not.exist;
+      } else {
+        expect(core.INTERNAL.assert).to.exist;
+      }
+    });
+  });
+});
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -15,24 +15,36 @@ module.exports = {
   rules: {
   },
   overrides: [{
-    files: ['**/test/**/*.js', '**/test.js'],
+    files: ['packages/*/test/**/*.js'],
     env: {
       mocha: true,
     },
-    globals: {
-      expect: true,
-      sinon: true,
-    },
     rules: {
-      'no-console': 0,
       'valid-jsdoc': 0,
       'require-jsdoc': 0,
       'max-len': 0,
     },
   }, {
-    files: ['gulp-tasks/*.js'],
+    files: [
+      'gulp-tasks/utils/log-helper.js',
+      'packages/workbox-core/src/utils/LogHelper.js',
+      'packages/workbox-core/test/bundle/node/utils/test-LogHelper.js',
+    ],
     rules: {
       'no-console': 0,
     },
+  }, {
+    files: ['gulp-tasks/**/*.js'],
+    rules: {
+      'valid-jsdoc': 0,
+      'require-jsdoc': 0,
+    },
+  }, {
+    files: [
+      'infra/utils/generate-variant-tests.js'
+    ],
+    env: {
+      'mocha': true
+    }
   }],
 };
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ packages/**/LICENSE
 temp
 tmp-*
 !packages/workbox-cli/test/static/example-project-1/node_modules
+packages/**/test/bundle-build/
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,3 +21,7 @@ before_script:
 
 script:
   # Building is handled in gulp test.
+  - gulp test
+
+after_script:
+  - npm run bot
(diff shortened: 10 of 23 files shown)


gulp-tasks/test-bundles.js
```diff
@@ -0,0 +1,120 @@
...
+
+  let outputFilename = `${runningEnv}.js`;
+
+  if (nodeEnv) {
+    // Make a unique bundle file for this environment
+    outputFilename = path.fileNameWithPostfix(outputFilename, `.${nodeEnv}`);
+    // Replace allows us to input NODE_ENV and strip code accordingly
+    plugins.push(replace({
+      'process.env.NODE_ENV': JSON.stringify(nodeEnv),
+    }));
```
addyosmani: Could we add a comment to document what types of code will get stripped out?

gulp-tasks/utils/package-runner.js
```diff
@@ -0,0 +1,84 @@
...
+ * we would want the 'build' task to run against `workbox-core`
+ * **only**.
+ *
+ * This method simplifies how we would write the 'build'
+ * task and should be the only way we write functions.
+ *
+ * ```javascript
+ * gulp.task('build',
+ *   gulp.series(
+ *     packageRunner(someFunction, someArg, someMoreArgs)
```
addyosmani: I understand what you're trying to accomplish with `packageRunner()` in the abstract, but perhaps include an example of what `someFunction` could be? Is it a transform (e.g minification) that returns a stream?

packages/workbox-core/src/controllers/WorkboxCore.js
```diff
@@ -0,0 +1,20 @@
...
+      return this._internal;
+    }
+
+    this._internal = {
+      logHelper,
+    };
+
+    if (process.env.NODE_ENV !== 'production') {
+      this._internal.assert = assert;
+    }
```
addyosmani: So our code-stripping is wise enough to exclude assertions like this one from the final builds if the right environment is specified?

addyosmani:
> Been slightly more strict on where console is used - this is solely to force the use of a log helper which can add prefixs and colors to helper figure out where logs are coming from in build process and tests.

Encouraging use of the log helper class, especially if this makes it clearer for us to strip out particular levels of warnings in production builds at some point, sounds reasonable. 

> I took the idea of task runner and altered it into packageRunner, the idea is that it wraps functions that perform the gulp work into an array of functions, one item for each package it should run on (i.e. --project vs no --project). This works well with gulp.series and gulp.parallel - I added a big readme to the file to explain.

My main comment here is that re-reading the `packageRunner` source a few times, it would be nice to see a few more examples of what the typical function being passed to it might look like. I naively assumed it can literally be anything returning a stream.

packages/workbox-core/src/controllers/WorkboxCore.js
gauntface: Yes, if you look at the tests I believe there is a check where depending on the node_env var will determine if the test expects it to exist or not

gulp-tasks/test:build-bundles.js
```diff
@@ -0,0 +1,101 @@
...
+  const environmentPath = path.join(testPath, 'bundle', runningEnv);
+
+  // First check if the bundle directory exists, if it doesn't
+  // there is nothing to build (NOTE: Rollup + multientry will
+  // always create a file, even if the directory doesn't exist)
+  if (!fs.pathExistsSync(environmentPath)) {
+    return Promise.resolve();
+  }
+
+  logHelper.log(`Building Test Bundle for ` +
```
jeffposnick: I know I've asked about this in the past (and been shot down), but can we take this opportunity to either pull in https://github.com/declandewet/common-tags#oneline or implement similar logic in our own internal helper library?

We're using template strings, so we might as well take advantage full of them, and be able to write this as

```js
logHelper.log(oneLine`
  Building Test Bundle for
  ${logHelper.highlight(pkgPathToName(packagePath))}
  to run in '${logHelper.highlight(runningEnv)}'
  with NODE_ENV='${logHelper.highlight(nodeEnv)}'.
`);
```

(Your function call also uses a combination of both `+` and `,`, inadvertently mixing concatenation along with multiple string parameters, which emphasizes the fact that it's easy to be inconsistent with the current approach.)

gulp-tasks/test.js
```diff
@@ -0,0 +1,40 @@
...
+    `${logHelper.highlight(pkgPathToName(packagePath))}.`);
+
+  const bundleTestDirectory = path.join(
+    packagePath, 'test', constants.BUNDLE_BUILD_DIRNAME);
+  return gulp.src(path.join(bundleTestDirectory, env, '*.js'), {read: false})
+  .pipe(mocha());
+};
+
+gulp.task('test:node', gulp.series(
+  packageRunnner(runBundledTests, 'node'),
```
jeffposnick: (Asked, while ignorant to gulp 4 best practices:)

Is there an advantage to a single-function series vs. doing

```js
gulp.task('test:node', () => packageRunnner(runBundledTests, 'node'));
```

Or is this just a placeholder for more sub-tasks to come?

gulp-tasks/utils/package-runner.js
gauntface: If have the following underneath this line:

```javascript
 function someFunction(packagePath, arg1, arg2) {
   // Return a Promise or a Gulp Stream
 }
 ```

File: gulp-tasks/utils/package-runner.js [hunk 6]
```diff
@@ -0,0 +1,84 @@
...
+module.exports = (func, ...args) => {
+  const packagePaths = glob.sync(
+    `packages/${global.projectOrStar}/package.json`, {
+      absolute: true,
+    }
+  );
+
+  // `packageRoots` is an array of absolute paths to packages
+  // that should have func run against them.
+  const packageRoots = packagePaths.map((packagePath) => {
```
jeffposnick: Here and elsewhere, can we revisit the aversion to using single-line arrow functions? This:

```js
const packageRoots = packagePaths.map((packagePath) => path.dirname(packagePath));
```

would be my preferred syntax. From past reviews, I know that @gauntface doesn't like them, but I'm a fan and perhaps others can weigh in.

packages/workbox-core/src/controllers/WorkboxCore.js
gauntface: Test is here: https://github.com/GoogleChrome/workbox/blob/e7ac90b92664ecb5a2e357df474660497ab65346/packages/workbox-core/test/bundle/node/controllers/test-WorkboxCore.js#L27

gauntface:
@addyosmani packageRunner just needs to be given a function that returns a Gulp Stream or a Promise. What that stream does (or promise does) doesn't matter.

The main idea is to just stop any concern or thinking of what the task should be running against.

packages/workbox-core/src/controllers/WorkboxCore.js
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * WorkboxCore shares code across Workbox modules.
+ */
+export default class WorkboxCore {
+  /**
+   * @private
+   */
+  get INTERNAL() {
+    if (this._internal) {
```
jeffposnick: I don't feel strongly, but as an alternative to an early `return`, you can consider restructuring this code to match what some of the other getters do, which is to check for the negated value first, and if the value isn't present, initialize it, and then unconditionally return the value at the end.

C.f. https://github.com/GoogleChrome/workbox/blob/master/packages/workbox-cache-expiration/src/lib/cache-expiration.js#L116

Your call.

packages/workbox-core/src/utils/assert.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * A simple assertion class to use for development builds.
+ *
+ * This will be excluded from production builds.
+ */
+class Assertions {
+
+}
+
+export default new Assertions();
```
jeffposnick: I wanted to solicit feedback from the larger group on the design choice of writing these sorts of modules as a class and then having the only export be a static instance of that class.

We do it a lot in the current code base and it's always struck me as, if not odd, something that I don't understand the motivation for.

What's the advantage vs. exporting a static object that had the same properties?

packages/workbox-core/test/bundle/node/controllers/test-WorkboxCore.js
```diff
@@ -0,0 +1,34 @@
+import chai from 'chai';
```
jeffposnick: You should be able to do

```js
import {expect} from 'chai';
```

and then not have to assign `chai.expect` to a variable later on. The `commonjs` Rollup plugin synthesizes named exports for you: https://github.com/rollup/rollup-plugin-commonjs#custom-named-exports

packages/workbox-core/test/bundle/node/utils/test-logHelper.js
```diff
@@ -0,0 +1,339 @@
...
+    // TODO: Catch WorkboxError with error code
+    it(`should not allow log level greater than error`, function() {
+      expect(() => {
+        logHelper.logLevel = logHelper.LOG_LEVELS.error + 1;
+      }).to.throw();
+    });
+
+    // TODO: Catch WorkboxError with error code
+    it(`should not allow non-number log levels`, function() {
+      expect(() => {
```
jeffposnick: There's a bunch of places in the tests where we pass in a variety of bad inputs, and expect each to fail. It would be nice to standardize on something that made it easier to write those test cases, but also made it clear when there's a failure, what the underlying input was that caused the failure. (The way this test is written, there's no indication that 4 of the expects might have passed, but the one with `null` failed.)

Ideally, this would be done in some sort of registerTestCasesFromBadInput() function that took in a description, a set of bad input, and the logic that should be applied to each bad input, and then in our tests we could do:

```js
registerTestCasesFromBadInput(
  'should not allow non-number log levels',
  [undefined, null, '', [], {}],
  (badInput) => logHelper.logLevel = badInput
);
```

And then, under the hood, it would register a test case for each `badInput`:

```js
it(`should not allow non-number log levels (test input: ${badInput})`, function() {
  // logic goes here
});
```

I don't want to write the actual logic for `registerTestCasesFromBadInput()` within a comment in a PR, but we could leave that as a TODO or something.

packages/workbox-core/test/bundle/node/utils/test-logHelper.js
```diff
@@ -0,0 +1,339 @@
...
+const expect = chai.expect;
+
+describe(`logHelper [${process.env.NODE_ENV}]`, function() {
+  let stubs = [];
+
+  beforeEach(function() {
+    logHelper.logLevel = logHelper.LOG_LEVELS.verbose;
+  });
+
+  afterEach(function() {
```
jeffposnick: We could get into the habit of using Sinon's Sandboxes instead of our more manual approach with the global stubs array:

https://stackoverflow.com/a/14041593/385997

packages/workbox-core/test/bundle/node/utils/test-logHelper.js
```diff
@@ -0,0 +1,339 @@
...
+   * Why .calledWithMatch()?
+   *
+   * This method on sinon spies will ensure that the spied method was called
+   * "with matching arguments (and possibly others)." This means logHelper
+   * can add a prefix to the log and still pass the assertion.
+   */
+
+  describe('.log()', function() {
+    it('should work without input', function() {
+      const stub = sinon.spy(console, 'log');
```
jeffposnick: Could you name this `spy` instead of `stub`? (Here and elsewhere, whenever you're actually creating spies.)

gulp-tasks/utils/package-runner.js
addyosmani: That is a far easier example to reason through. Thanks for updating it!

packages/workbox-core/test/bundle/node/utils/test-logHelper.js
```diff
@@ -0,0 +1,339 @@
...
+      stubs.push(stub);
+
+      logHelper.log();
+
+      expect(console.log.callCount).to.equal(1);
```
jeffposnick: Since you're creating a spy object, I'd kind of expect the check to be done against `spy` rather than against the method that's being spied on.

Apparently they're equivalent, and I don't very strongly, but the following reads better to me:

```js
expect(spy.calledOnce).to.be.true;
expect(spy.calledWithMatch('', 'test', null, undefined, [], {})).to.be.true;
```

addyosmani:
@gauntface That's much better. Thanks!

gulp-tasks/test.js
gauntface: The above is equivalent to:
```javascript
gulp.task('test:node', gulp.series(
  [
    () => runBundledTests('/some-path/workbox/packages/workbox-core', 'node'),
    () => runBundledTests('/some-path/workbox/packages/workbox-precaching', 'node'),
    () => runBundledTests('/some-path/workbox/packages/workbox-*', 'node'),
  ]
)
```

I believe `gulp.series` and `gulp.parallel` will then manage running the functions and chaining Promises / gulp tasks accordingly.

When defining the gulp.task like so:

```javascript
gulp.task('<task name>', <input>);
```

The input variable needs to be a function returning a promise or gulp stream, and since `packageRunner` returns an array, it won't work. `gulp.series` and `gulp.parallel` however will return one of these for us.

I did consider if we should have `packageRunner.series()` and `packageRunner.parallel()` just to hide away the gulp calls, but wasn't sure if that was a good idea or not

gulp-tasks/test.js
```diff
@@ -0,0 +1,40 @@
...
+const logHelper = require('./utils/log-helper');
+const pkgPathToName = require('./utils/pkg-path-to-name');
+
+const runBundledTests = (packagePath, env) => {
+  logHelper.log(`Running Test Bundles for ` +
+    `${logHelper.highlight(pkgPathToName(packagePath))}.`);
+
+  const bundleTestDirectory = path.join(
+    packagePath, 'test', constants.BUNDLE_BUILD_DIRNAME);
+  return gulp.src(path.join(bundleTestDirectory, env, '*.js'), {read: false})
```
gauntface: Switch to path.posix.join which should force '/'.

The main reason for this is to avoid any concern with double forward slashes.

gulp-tasks/test.js
```diff
@@ -0,0 +1,40 @@
...
+  return Promise.resolve();
+});
+
+gulp.task('test:sw', () => {
+  // TODO: This needs implementing
+  return Promise.resolve();
+});
+
+gulp.task('test', gulp.series(
+  'test:build-bundles',
```
gauntface: I just tried this locally and it's working for me (both linux and OS X).

Do you have `gulp-cli` installed globally? That's the only thing I can think of that might be a difference.

jeffposnick:
```sh
$ gulp -v
[16:31:45] CLI version 1.3.0
[16:31:45] Local version 4.0.0-alpha.2
```

gulp-tasks/test.js
jeffposnick: You don't have to worry about `//` either. I just tested and confirmed that `glob`, and therefore `gulp.src()`, will normalize it all for you.

String concatenation with `/` is safe.

gulp-tasks/lint.js
```diff
@@ -0,0 +1,14 @@
+const gulp = require('gulp');
+const exec = require('child-process-promise').exec;
+
+const logHelper = require('./utils/log-helper');
+
+// Use npm run lint to ensure we are using local eslint
+gulp.task('lint', () => {
```
gauntface: You're right, it's much better. Just moved to `gulp-tasks/utils/spawn-promise-wrapper.js`

gulp-tasks/test:build-bundles.js
gauntface: The major fear I have is that if we add this into out gulp tasks, it's a short step for someone to want / ask / use it in the actual library, in which case we are taking on extra code for no real reason.

gauntface: Added in for gulp tasks - do not want in the actually library.

packages/workbox-core/src/controllers/WorkboxCore.js
gauntface: Switch to this (I think it's less code)

gulp-tasks/utils/package-runner.js
gauntface: Changed. Feel like its harder to read but it's fine.

packages/workbox-core/src/utils/assert.js
gauntface: So do you want exported functions, a class with static functions or a static instance of the class?

I've just switched to exporting the class itself and then workbox core instantiating, but I feel like you want something different.

For me, I just view this as a singleton (i.e. why would we ever want more than a single instance of this library? The plus of this being a class is that it makes it easier to ensure with it's being exported in the final output.

packages/workbox-core/test/bundle/node/utils/test-logHelper.js
gauntface: Good idea - will switch over now.

Do you think we should add some custom "project lint" checks (i.e. regexs for sinon.(spy|stub) to somewhat enforce these kinds of rules?)

I'm largely thinking on this as an interested considering that Phil's PR's have a lot of consistency stuff being done and it's a shame it's not enforced in any way.

packages/workbox-core/test/bundle/node/controllers/test-WorkboxCore.js
gauntface: Had to wrangle commonjs to do this with the name exports object.

packages/workbox-core/test/bundle/node/utils/test-logHelper.js
gauntface: Totally valid and awesome idea (Worth doing sooner rather than later).

I avoided that hear because I realised I complicated a number of the tests I had written but this seems like a nice and safe option.

See generate-variant-tests.js

packages/workbox-core/test/bundle/node/utils/test-logHelper.js
gauntface: I actually switched them to stubs so they didn't create pointless noise in the console.

packages/workbox-core/test/bundle/node/utils/test-logHelper.js
gauntface: I actually got this from the sinon docs, but you are right it's weird and your version is better.

gulp-tasks/test.js
gauntface: urgh so weird that it's working for me locally and on Travis. I was hoping we wouldn't need to babysit the gulp task files in gulpfile.ks

packages/workbox-core/src/utils/assert.js
jeffposnick: My default for this sort of thing, where there's no internal state information to be stored, would be exporting a static object with properties. (That's basically what the old-school `assert.js` [used to be](https://github.com/GoogleChrome/workbox/blob/7c3a4f505fb2462a9afce851a2ba0abd4fab55a5/lib/assert.js#L124).)

> The plus of this being a class is that it makes it easier to ensure with it's being exported in the final output.

Yeah, I just want to understand more about that benefit—what assurance does it offer?

I don't want to block the PR on this, and I'm mainly asking for my own education.

packages/workbox-core/test/bundle/node/controllers/test-WorkboxCore.js
jeffposnick: Huh, that's weird. I assumed that it would work by default. Not sure if it was worth the extra effort, then, but thanks for addressing.

packages/workbox-core/test/bundle/node/utils/test-logHelper.js
jeffposnick: I'd love to see more consistency.

Speaking for myself, I've learned a lot about testing best practices while working on v1, and I know that my newer tests have evolved quite a bit from what I was writing initially.

I worry about veering too far down the path of writing our own wrappers on top of `mocha`/`chai`/`sinon`, and then we've ended up effectively writing our own testing framework, but... that wouldn't be the worst idea?

I don't know about linting-style rules for test consistency at this point, because we don't need linting to tell us that the tests are inconsistent. We could impose some rigor via linting  once we have a baseline of consistency.

packages/workbox-core/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "homepage": "https://github.com/GoogleChrome/workbox/tree/master/packages/workbox-core",
+  "keywords": [
+    "workbox",
+    "workboxjs",
+    "service worker",
+    "sw"
+  ],
+  "scripts": {
+    "pretest": "node ../../infra/build-test-bundle.js",
+    "test": "mocha './test/bundle-build/**/*.js'"
```
jeffposnick: I know that `gulp test` does a lot of setup—starting up our test server, running the multi-browser testing harness, etc.

Is it sufficient to just run `mocha` directly here rather than going through the gulp task? Or is this just meant to be a subset of the tests?

infra/utils/generate-variant-tests.js
```diff
@@ -0,0 +1,18 @@
...
+ * idenfity the failing test.
+ * @param {Array<Objects>} variants This should be all the variations of the
+ * test you wish to generate.
+ * @param {function} func This is the function that will be called, with a
+ * variant as the only argument. This function should perform the desired test.
+ */
+export default (itTitle, variants, func) => {
+  variants.forEach((variant) => {
+    it(`${itTitle}. Variant: '${JSON.stringify(variant)}'`,
+      () => func(variant));
```
jeffposnick: This should be `function() { return func(variant); }` due to the prohibition on using arrow functions for the `it()` test logic.

jeffposnick:
This is great, @gauntface! I left a couple of other small questions after another round of review, but I think we're generally good to move ahead.

infra/utils/generate-variant-tests.js
gauntface: Great spot!

packages/workbox-core/package.json
gauntface: Originally was trying to set this up without gulp, but it was going to be to restrictive (at least I think it would be in the long run). Switching to Gulp 4 was the middle ground IMO. Will Remove.

gulp-tasks/test.js
jeffposnick: Just to confirm what we discussed via chat, a fresh clone + `npm install` has things working for me.

PR #720, opened 2017-08-04
Fix for multiple cacheId prefixes in the precache handler

### LINKED ISSUE #714: trouble with precaching when supplying `cacheId`
**Library Affected**:
workbox-precache

**Browser & Platform**:
Build related issue

It's not entirely clear how this is happening, but when i supply a `cacheId` i'm ending up with a second misnamed cache being created, that never gets populated.

On initial load everything seems fine:
screen shot 2017-08-02 at 2 42 39 pm

When I refresh, a second, misnamed/empty cache gets created:
screen shot 2017-08-02 at 2 42 58 pm

The trouble is, this second cache gets referenced when you go offline, and so it thinks it has these assets but seemingly looks for them in the wrong cache, causing complete failure to load.

This seems like a pretty major issue so I wanted to raise it, I was hoping to do a PR rather than file a vague bug, but didn't have the time right now, so I figured it was better to file something. 

To be clear the exact same SW works perfect if I just remove the cacheId.

jeffposnick:
R: @addyosmani @gauntface 
CC: @HenrikJoreteg @liuruoran88

This fixes #714 by explicitly changing the method call inside of `WorkboxSW` to indicate that for the internal `CacheOnly` handler use for precached requests, `cacheId` shouldn't be prepended when determining the corresponding cache name.

I decided to go with that approach rather than making a chance in the `RequestWrapper` constructor that checked to see if `cacheId` was already present in the `cacheName`, since you could theoretically imagine calling

```js
new RequestWrapper({
  cacheName: 'my-prefix',
  cacheId: 'my-prefix',
}});
```

In that situation, the user would expect the effective cache name to be `'my-prefix-my-prefix'`, and we'd want to still allow that.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2,7 +2,18 @@
   "name": "workbox",
   "version": "0.0.0",
   "lockfileVersion": 1,
+  "requires": true,
   "dependencies": {
+    "JSONStream": {
+      "version": "1.3.1",
+      "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-1.3.1.tgz",
+      "integrity": "sha1-cH92HgHa6eFvG8+TcDt4xwlmV5o=",
+      "dev": true,
+      "requires": {
+        "jsonparse": "1.3.1",
+        "through": "2.3.8"
+      }
+    },
     "accepts": {
       "version": "1.3.3",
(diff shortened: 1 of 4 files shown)


jeffposnick:
The interplay of `cacheId` and `cacheName`, while useful, definitely complicates the logic, too.

Issue #696, opened 2017-07-22
Webpack Discussion


goldhand:
**Library Affected**:
*workbox-webpack-plugin*

I wanted to discuss the idea of supporting the same configuration from [sw-precache-webpack-plugin](/goldhand/sw-precache-webpack-plugin) inside workbox-webpack-plugin.

I think it would make it easier for users of sw-precache-webpack-plugin to transition to using workbox-webpack-plugin if they are able to use the same configuration with workbox-webpack-plugin. 

I also wanted to consolidate a list of any other feature suggestions that might help the workbox / webpack integration feel seamless.

A few I'd like to see are:
* Integrating with html-webpack-plugin to generate a register script ( goldhand/sw-precache-webpack-plugin#62 )
* generating a register script and emitting it as a webpack chunk

There's also a lot of requests for webpack-dev-middleware/dev-server support as seen in #577 and goldhand/sw-precache-webpack-plugin#17

gauntface:
@goldhand sounds good to me, do you have a list of things that aren't currently supported? Would help me get a feel for what is missing in `workbox-build`.

goldhand:
@gauntface 

## Any sw-precache-webpack-plugin options that were added to support sw-precache

These are found under the **plugin options:** section of the [sw-precache-webpack-plugin configuration docs](https://github.com/goldhand/sw-precache-webpack-plugin#configuration).

I think workbox-build covers:

```
    staticFileGlobsIgnorePatterns: 'globIgnores'
    filepath: 'swDest', 'globDirectory'
```

`filename` isn't really needed.

which only leaves `minify` (minifies/uglifies) and `mergeStaticsConfig` (lets you keep webpack options and specify your own additional options for `staticFileGlobs` and `stripPrefixMulti`)


## Any default behaviors that sw-precache-webpack-plugin is passing to sw-precache.
I think we should try to provide the same default behavior mapped to the workbox equivalent, if possible.

These are found under the **sw-precache options:** section of the [sw-precache-webpack-plugin configuration docs](https://github.com/goldhand/sw-precache-webpack-plugin#configuration).

I need to look more into it but my connection is pretty spotty at the moment (sw cache use case 😆) but it looks like currently there is:
* globDirectory
* swDest
* globPatterns

That provide default options for workbox-build. I don't think all the functionality is the same but I looks like some are covered.

A couple missing options that are awesome IMHO are:
[...]

PR #724, opened 2017-08-08
Allow webpack plugin to use generateFileManifest, split into 3 plugins

anilanar:

This PR splits webpack plugin into three plugins:
- `GenerateSWPlugin`
- `GenerateFileManifestPlugin`
- `InjectManifestPlugin`

Thus this PR adds capability of manifest file generation.


prateekbh:
@anilanar any specific reason to have a separate plugin for just manifest generation?

anilanar:
As a user, I find it easier to understand 3 separate plugins that have slightly different purposes than a single plugin. For the latter, it becomes harder to document and understand which configuration properties belong to which "mode" (mode as in sw generation, sw injection, manifest generation). 

As a developer, I find it easier to unit/e2e test 3 different plugins than a single one. When you merge 3 plugins into one, number of edge cases and complexity increases.

The only downside I see is that this would be a breaking change. However it is quite trivial to swap the old plugin with one of the new plugins.

anilanar:
I should also explain why I think a `generateFileManifest` mode is needed:

- `injectManifest` mode is hard to use. There must be a specific precache call with an empty array. 
- Let's If you want to modify the manifest before calling precache, you must call a fake `precache` function which then calls the real `precache`.
- Injection does not clear contents of the array that was filled before. That makes development experience sub-optimal if you don't generate service worker from scratch as part of the build process (consider webpack watch mode).
- You can create a file that handles the whole hurdle mentioned in above points and import that file from the actual service worker. That somewhat mimicks `generateFileManifest` but all the disadvantages mentioned above remains.

gauntface:
@anilanar thanks for the extra comments, this is super interesting, sounds like you are hitting the following issues:

1. Regarding altering the manifest - we have added manifestTransforms to the workbox-build module and hopefully the webpack plugin will have support soon.
1. I personally never considered that a developer would over-write the service worker when using inject manifest. I think that is an issue we should detect and prevent. The idea if we create a "built" service worker.
1. We should make the generateFileManifest available via the plugin, but offering three plugins definitely feels like a bad approach. Is there any other API we should take?

@prateekbh @goldhand do you have any thoughts here?  Personally, this sounds like an issue but I'm not convinced this is the right approach.

addyosmani:
@anilanar Thanks for taking the time to think through and prototype your ideas here.

@goldhand is going to be leading up the rethink of our Webpack support so I'll defer to him for thoughts on going more granular. I think the points raised about it being easier to end to end test and reason about more focused plugins are sound, but I'm also conscious of any changes that make it more complex to explain the plugin and usage to end users. 

> I personally never considered that a developer would over-write the service worker when using inject manifest. I think that is an issue we should detect and prevent.

I concur!

goldhand:
Wow, this represents a lot of work and, at a glance, this is very well written code 👏 

The current generate / inject mode functionality is confusing and I think you do a good job of addressing it here. Adding an explicit option to the existing webpack plugin interface could also solve this though and there are probably other approaches too.

I like the idea of using a BasePlugin class. As a monorepo, workbox has a lot of modules and this could be useful for supporting other kinds of webpack integrations in the future. The downside of this pattern is the increased complexity multiple webpack plugins introduces.

I'm not totally convinced that creating three webpack plugins is the best approach. I see the merits of it, as others have mentioned, but the additional complexity concerns me. I would like to consider and discuss further what other apis we could use here.

anilanar:
I've shared my views in #696. Nevertheless from this point, we can merge 3 plugins into one by 1 plugin delegating functionality to separated classes based on a "mode" property.

gauntface:
Gentle ping @goldhand for what he'd like to do with this? @anilanar we are currently looking towards v3 for next major version and should give us a lot of freedom in terms of API, would you be interested in working with us to improve things there?

addyosmani:
From today's Workbox team meeting it sounds like we can close this PR. Deferring to @goldhand for an update on his progress and take on it.

goldhand:
Hey @anailar, closing this in favor of keeping the single webpack plugin api. Working on the requirements defined in #725 under **workbox-web-plugin**. Any interest in helping on those?

PR #732, opened 2017-08-10
V3 package build

gauntface:

This adds a browser bundle build step.

Currently builds to `<package>/build/browser-bundles/<package name>.<dev|production>.js` The reasoning for this is that `build/` could be where we put our ES2015 module code (i.e. dump files that we want to be referenced as `import '<package name>/<classname or path to file>'`

Minifier has been switched to uglify-es, this is so that we can minify private variables, which for the small amount of code got use from ~900B down to ~750B. I'm running the bundled tests through the minifier as well just to try and account for the fact that the source in our library code will ultimately be different and hopefully catch us out but this does make me nervous if we rely on the bundled tests too much.

Build process for packages is using lerna from start to finish. `gulp build` will simply run `lerna bootstrap` behind the scenes. Lerna should figure out the dependencies between modules and will then run `npm run prepublish` on each package. Each package should then run `gulp build-packages --package <package name>`. This should mean are working *with* lerna rather than against / next to it.

PR-Bot should now show the built file sizes in this PR.


gulp-tasks/build-packages.js
```diff
@@ -0,0 +1,115 @@
...
+const logHelper = require('./utils/log-helper');
+const pkgPathToName = require('./utils/pkg-path-to-name');
+
+/**
+ * To test sourcemaps are valid and working, use:
+ * http://paulirish.github.io/source-map-visualization/#custom-choose
+ */
+
+const buildPackage = (packagePath, buildType) => {
+  const srcPath = path.posix.join(packagePath, 'src');
```
jeffposnick: Since we're not using `gulp.src()` here, and therefore not using globs, shouldn't this be `path.join()` instead of always using POSIX paths?

gulp-tasks/build-packages.js
```diff
@@ -0,0 +1,115 @@
...
+};
+
+gulp.task('build-packages:clean', gulp.series(
+  packageRunnner('build-packages:clean', cleanPackages)
+));
+
+// This will create one version of the tests for each buildType.
+// i.e. we'll have a browser build for no NODE_ENV and one for 'production'
+// NODE_ENV and the same for sw and node tests.
+const packageBuilds = [];
```
jeffposnick: Good opportunity to use `.map()`:

```js
const packageBuilds = constants.BUILD_TYPES.map((buildType) => {
  return packageRunner('build-package', buildPackage, buildType);
});
```

gulp-tasks/build.js
```diff
@@ -0,0 +1,20 @@
+const gulp = require('gulp');
+
+const lernaWrapper = require('./utils/lerna-wrapper');
+
+gulp.task('lerna-bootstrap', () => {
+  // If it's a start, build all projects (I.e. bootstrap everything.)
```
jeffposnick: 'start' => 'star'

gulp-tasks/build.js
```diff
@@ -0,0 +1,20 @@
...
+  if (global.packageOrStar === '*') {
+    return lernaWrapper.bootstrap();
+  }
+
+  // If it's not a start, we can scope the build to a specific project.
```
jeffposnick: 'start' => 'star'

gulp-tasks/utils/constants.js
```diff
@@ -1,7 +1,48 @@
...
+  BROWSER_BUILD_DIRNAME: 'browser-bundles',
+  TEST_BUNDLES_BUILD_DIRNAME: 'bundle-builds',
+
+  // This is the environments that we should use for NODE_ENV.
+  BUILD_TYPES: [undefined, 'production'],
+
+  // Every use of rollup should have minification and the replace
+  // plugin set up and used to ensure as consist set of tests
+  // as possible.
+  getDefaultRollupPlugins: (buildType) => {
```
jeffposnick: It's slightly odd to have this in a `constants.js` file. I think it would make more sense conceptually if you put it in with the other Rollup code.

gauntface: The reason for this is that the rollup plugins that are shared between test bundles and browser bundles *should* be the same. We can inline into the rollup plugins but then we risk the two coming out of sync meaning our bundled tests are now running a more different build than the browser builds.

What do you reckon is best path forward for this?

package.json
```diff
@@ -29,21 +29,28 @@
...
     "rollup": "^0.45.2",
+    "rollup-plugin-babili": "^3.1.0",
     "rollup-plugin-commonjs": "^8.0.2",
     "rollup-plugin-multi-entry": "^2.0.1",
     "rollup-plugin-node-resolve": "^3.0.0",
     "rollup-plugin-replace": "^1.1.1",
+    "rollup-plugin-uglify": "^2.0.1",
     "rollup-stream": "^1.23.1",
     "sinon": "^2.4.1",
+    "uglify-es": "^3.0.27",
```
addyosmani: > Minifier has been switched to uglify-es, this is so that we can minify private variables, which for the small amount of code got use from ~900B down to ~750B

Speaking to developers a little closer to Babili, although both it and uglify-es are maintained, the latter is considered a little more stable. I'm in favor of the switch in minifier here.

PR #736, opened 2017-08-10
Adds JSDocs for the lifecycle methods, and awaits the return of each.

### LINKED ISSUE #533: Better documentation around lifecycle methods + plugins
**Library Affected**:
*workbox-runtime-caching*

There's some basic info about what lifecycle methods are exposed by `RequestWrapper`, but it would be good longer-term to create something more formal using the `@callback` JSDoc construct. A guide for writing your own plugins would follow from that initial work.

### LINKED ISSUE #709: Normalize all RequestWrapper lifecycle methods to return a Promise<*>

jeffposnick:
R: @addyosmani @gauntface

Fixes #709, fixes #533 

`await` is nice in that it doesn't require the expression to actually evaluate to a promise So both `await Promise.resolve(true)` and `await true` are both valid.

Because of this, I just had to change the code that calls the lifecycle methods to consistently use `await`, and the existing plugins that don't need to do anything asynchronous don't have to be updated.

This opens the doors for new plugins that weren't previously able to return promises, like the Range request helper in #710.

I took the opportunity to add in formal JSDoc for the lifecycle methods, via `@callback`, since that was a longstanding issue.

### DIFF
--- a/packages/workbox-runtime-caching/src/lib/request-wrapper.js
+++ b/packages/workbox-runtime-caching/src/lib/request-wrapper.js
@@ -22,25 +22,136 @@ import {pluginCallbacks, getDefaultCacheName} from './constants';
 import cleanResponseCopy from './clean-response-copy';
 
 /**
+ * Called prior to a response being written to the cache. This allows you to
+ * prevent the cache from being updated if the response doesn't meet your
+ * custom criteria.
+ *
+ * @example <caption>Determines whether a response is cacheable based on
+ * whether its Cache-Control header contains the string 'no-cache'.</caption>
+ *
+ * async function cacheWillUpdate({response}) {
+ *   return !response.headers.get('cache-control').includes('no-cache');
+ * }
+ *
+ * @callback cacheWillUpdate
+ * @param {Object} input
+ * @param {Request} input.request The original request.
+ * @param {Response} input.response The response to the request, based on the
+ * configured strategy.
+ * @return {Promise<Boolean>} `true` if the response meets your criteria for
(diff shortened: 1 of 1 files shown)


packages/workbox-runtime-caching/src/lib/request-wrapper.js
```diff
@@ -219,9 +330,7 @@ class RequestWrapper {
 
     if (this.plugins.has('requestWillFetch')) {
       for (let plugin of this.plugins.get('requestWillFetch')) {
-        const returnedPromise = plugin.requestWillFetch({request});
-        isInstance({returnedPromise}, Promise);
-        const returnedRequest = await returnedPromise;
+        const returnedRequest = await plugin.requestWillFetch({request});
```
jeffposnick: This logic is equivalent to what was there before; because I'm now using `await`, the return value from `plugin.requestWillFetch()` will always be a promise.

addyosmani: Thanks for the note. The return value here always being a promise looks fine with your update.

PR #737, opened 2017-08-10
V3 Analyse Browser Bundles

gauntface:

Here it comes . . . . . . I promise I'm not crazy :-/

giphy 4

This is a tool that pulls out property names and function names. The idea being that if we have a tonne of variables being used that are public or not marked as private, the minifier is unable to rename it. This means the end file size will be larger than it needs to be.

I have no idea if this will be used going forward or not. But after manually going through some of the workbox-core code's and refactor slightly to shave off bundle size ~850 bytes to ~ 750 bytes. Moving forward with much larger code, I suspect that this **might** be useful.

I'd like to keep this in the infra directory just in case it does proof handy and I've added a comment to remove if it doesn't get used / isn't useful.

It filters out words which are only reference once in the source (i.e. public functions / properties) and will exclude single character variables names from the output. Currently we have all variables stripped apart from our namespace which occurs several times to initialize the namespace.

```shell
$ gulp build && node ./infra/tools/analyse-properties.js 
[14:57:41] Using gulpfile ~/Projects/Code/workbox/gulpfile.js
[14:57:41] Starting 'build'...
[14:57:41] Starting 'lerna-bootstrap'...
lerna info version 2.0.0
lerna info Bootstrapping 1 packages
lerna info lifecycle preinstall
lerna info Symlinking packages and binaries
lerna info lifecycle postinstall
lerna info lifecycle prepublish
lerna success Bootstrapped 1 packages
[14:57:43] Finished 'lerna-bootstrap' after 2.02 s
[14:57:43] Finished 'build' after 2.02 s

Results for 'packages/workbox-core/build/browser-bundles/workbox-core.dev.js'

    google   5
    workbox  3


Results for 'packages/workbox-core/build/browser-bundles/workbox-core.production.js'

    google   5
    workbox  3


```

### DIFF
--- a/infra/tools/analyse-properties.js
+++ b/infra/tools/analyse-properties.js
@@ -1,0 +1,104 @@
+/**
+ * This file should be run as a node script to analyse the content of the
+ * browser bundles.
+ *
+ * It's an extremely naive approach to picking property names out and seeing
+ * how often they are used in the final browser bundles. This can be
+ * used to detect variable names that are long and used repeatedly meaning
+ * that a small refactor allowing `uglify-es` to mangle these properties,
+ * could lead to reasonable improvement in final bundle size.
+ *
+ * This is very rough and vague. It needs to be run manually and if it's not
+ * used can and should be removed from Workbox repo.
+ */
+const glob = require('glob');
+const path = require('path');
+const fs = require('fs-extra');
+
+const PROPERTY_REGEX = /\.(\w+)/g;
+
+class AnalyseBuildForProperties {
+  run() {
+    const filePaths = this.getBuildFiles();
+    return Promise.all(filePaths.map((filePath) => {
+      const rawAnalysis = this.analyzeFile(filePath);
+      const analysis = this.tidyData(rawAnalysis);
+
+      return {
+        filePath,
+        analysis,
+      };
+    }));
+  }
+
+  getBuildFiles() {
+    const buildGlob = path.join(__dirname, '..', '..', 'packages',
+      '*', 'build', 'browser-bundles', '*.js');
+    return glob.sync(buildGlob);
+  }
+
+  analyzeFile(filePath) {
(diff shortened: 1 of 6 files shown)


jeffposnick:
I'd like to try it out, but I'm stuck on getting it running.

From what I can tell, your script is meant to be run inside a package's `build` directory and then finds all of the `browser-bundles` for all projects:

    const buildGlob = path.join(__dirname, '..', '..', 'packages',
      '*', 'build', 'browser-bundles', '*.js');

The only build task that's defined is `gulp build-test-bundles`, and I don't see that creating `browser-bundles` directories for any of the projects.

(My completely uninformed, gut reaction is that this is the sort of thing that needs to be done via a parser that understands the code, like https://github.com/babel/babylon, rather than via RegExps, but who knows? I do like RegExps!)

gauntface:
Babylon is AMAZING

Just switched to it - got more results which is great and I added the gulp task that needs to be run to have the files to run against.

infra/tools/analyse-properties.js
```diff
@@ -0,0 +1,123 @@
...
+      const numberOfSpaces = longestPropertyName - entry.propertyName.length;
+      const extraSpace = ' '.repeat(numberOfSpaces);
+      console.log(`    ${entry.propertyName} ` +
+        `${extraSpace} ${entry.propertyCount}`);
+    });
+      console.log();
+  }
+}
+
+new Promise((resolve, reject) => {
```
jeffposnick: Putting this into its own `gulp` task, and using `async`/`await`, would lead to cleaner code:

```js
gulp.task('analyze-properties', ['build'], async () => {
  const analysisTool = new AnalyseBuildForProperties();
  const results = await analysisTool.run();
  for (entry of results) {
    analysisTool.printDetails(entry);
  }
});
```

PR #741, opened 2017-08-11
V3 coverage

gauntface:
R: @jeffposnick @addyosmani

Again, no enforcement, but I got code coverage working and I think it'll be good to see if others find this useful as they are writing / porting code.

```shell
     ✓ should work without input
      ✓ should work several inputs
      ✓ should log with verbose log level
      ✓ should log with error log level
      ✓ should log with warning log level
      ✓ should log with error log level


  78 passing (116ms)

-----------------------------------------------------|----------|----------|----------|----------|----------------|
File                                                 |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
-----------------------------------------------------|----------|----------|----------|----------|----------------|
 infra/utils/                                        |      100 |      100 |      100 |      100 |                |
  generate-variant-tests.js                          |      100 |      100 |      100 |      100 |                |
 packages/workbox-core/src/controllers/              |      100 |       75 |      100 |      100 |                |
  WorkboxCore.js                                     |      100 |       75 |      100 |      100 |                |
 packages/workbox-core/src/utils/                    |      100 |      100 |      100 |      100 |                |
  Assert.js                                          |      100 |      100 |      100 |      100 |                |
  LogHelper.js                                       |      100 |      100 |      100 |      100 |                |
 packages/workbox-core/test/bundle/node/controllers/ |      100 |      100 |      100 |      100 |                |
  test-WorkboxCore.js                                |      100 |      100 |      100 |      100 |                |
 packages/workbox-core/test/bundle/node/utils/       |      100 |      100 |      100 |      100 |                |
  test-LogHelper.js                                  |      100 |      100 |      100 |      100 |                |
-----------------------------------------------------|----------|----------|----------|----------|----------------|
All files                                            |      100 |    92.86 |      100 |      100 |                |
-----------------------------------------------------|----------|----------|----------|----------|----------------|

```


packages/workbox-core/test/bundle/node/controllers/test-WorkboxCore.js
```diff
@@ -15,6 +15,12 @@ describe(`WorkboxCore - ${process.env.NODE_ENV}`, function() {
       expect(core.INTERNAL).to.exist;
     });
 
+    it(`should return the same INTERNAL instance after first call`, function() {
+      const core = new WorkboxCore();
+      const internal = core.INTERNAL;
+      expect(core.INTERNAL).to.equal(internal);
```
jeffposnick: Not sure if it matters in this context, but my default is to use [`eql`](http://chaijs.com/api/bdd/#method_eql) (which tests for deep-equality) instead of `equal` whenever possible.

gauntface: In this case I actually wanted it to check it was the same variable, but that being said the test isn't actually useful anymore so I'll remove it.

gauntface:
Just tweaked Travis yaml to only run pr-bot at the end of multi-package builds.

https://travis-ci.org/GoogleChrome/workbox/builds/264444809

I'm hoping that once the staged builds, which is the "jobs" keyword in the yaml, is better supported, we can more explicitly define environments etc per stage and move the scripts etc into Jobs.

PR #746, opened 2017-08-15
V3 upload to cdn

### LINKED ISSUE #743: [V3] Handling Releasing
Next work load that will be added to my queue:

- [x] Create a project for CDN
- [x] Create a script that:
    - [x] Gets all tagged builds from github
    - [x] Figures out which tagged builds *aren't* on the CDN
    - [x] Collects all `dev.js`, `prod.js` and `.js.map` files for each tag and create directories for each tag
    - [x] Upload these collections into the CDN.
- [x] Create a script that will collect the `dev.js`, `prod.js` and `.js.map` files for a specific tag, create a tar.gz and upload them to a github release.
- [x] Create a script that will collect the `dev.js`, `prod.js` and `.js.map` files for a specific tag, create a .zip and upload them to a github release.
- [x] Finally find either an elegant way to add this script to the end of the lerna publish script OR add the lerna publish step as a pre-requisite *in* this script.

gauntface:
R: @jeffposnick @addyosmani

This PR adds the following:

- Creates draft release for a github tag.
- Pulls down the source for a git tab >= v3.0.0, builds it and creates tar.gz containing the browser files.
- Adds the tar.gz as a download asset to the github release
- Publishes the same files as the tar.gz to our CDN.

The Github draft looks like:

github-draft

The CDN Upload looks like:

gcloud-release

The public link for this kind of thing will be:

https://storage.googleapis.com/workbox-cdn/v3.0.0-alpha/workbox-core.dev.js

We can apparently use our workboxjs.org URL to point to the CDN storage if we really wanted, but that can come at a later stage.

Fixes #743


gulp-tasks/publish-bundle.js
```diff
@@ -0,0 +1,211 @@
...
+  const pattern = path.posix.join(projectPath, 'packages', '**',
+    constants.PACKAGE_BUILD_DIRNAME, constants.BROWSER_BUILD_DIRNAME,
+    '*.{js,map}');
+
+  const tarName = `workbox-${tagName}.tar.gz`;
+
+  const filesToIncludeInBundle = glob.sync(pattern);
+  const bundleDirectory = path.join(projectPath, '..', 'github-bundle');
+
+  filesToIncludeInBundle.forEach((fileToInclude) => {
```
jeffposnick:
```js
await Promise.all(filesToIncludeInBundle.map(
  (file) => fs.copy(file, path.join(bundleDirectory, path.basename(fileToInclude)));
```

gulp-tasks/publish-bundle.js
```diff
@@ -0,0 +1,211 @@
...
+      path.join(bundleDirectory, path.basename(fileToInclude))
+    );
+  });
+
+  const tarball = new TarGz({}, {
+    // Do not inlcude the top level in the tar.gz
+    fromBase: true,
+  });
+  const tarPath = path.join(bundleDirectory, '..', tarName);
+  return tarball.compress(bundleDirectory, tarPath)
```
jeffposnick:
```js
await tarball.compress(...);
return {bundleDirectory, tarPath};
```

gulp-tasks/publish-bundle.js
```diff
@@ -0,0 +1,211 @@
...
+  .then(() => {
+    return {
+      bundleDirectory,
+      tarPath,
+    };
+  });
+};
+
+const handleGithubAndCDNRelease = (tagName, gitBranch, release) => {
```
jeffposnick: Same `async`/`await` suggestions—I think you get the drift 😄

gulp-tasks/publish-bundle.js
```diff
@@ -0,0 +1,211 @@
...
+      name: path.basename(tarPath),
+      label: path.basename(tarPath),
+    })
+    .then(() => {
+      return uploadBundleToCDN(tagName, bundleDirectory);
+    });
+  });
+};
+
+const findTagsWithoutBundles = () => {
```
jeffposnick: Ditto the previous suggestions about `async`/`await` usage.

gulp-tasks/publish-bundle.js
```diff
@@ -0,0 +1,211 @@
...
+    repo: GITHUB_REPO,
+  })
+  .then((tagsResponse) => {
+    // We only want tags that are v3.0.0 or above.
+    const tagsData = tagsResponse.data;
+    return tagsData.filter((tagData) => {
+      // The "-alpha" means semver will match against tags with prelease
+      // info (although the prelease tag must be a string >= 'alpha'
+      // alphabetically speaking).
+      // TODO: Change to v3.0.0 when v3 is launched.
```
jeffposnick: Can you put this in a CONSTANT at the top of the file for increased visibility as something that needs to change?

gulp-tasks/publish-bundle.js
```diff
@@ -0,0 +1,211 @@
...
+        // asset
+        return false;
+      }
+
+      return true;
+    });
+  });
+};
+
+gulp.task('publish-bundle:generate-from-tags', () => {
```
jeffposnick: `gulp` tasks can use `async`/`await` too.

gulp-tasks/utils/upload-bundle-to-cdn.js
```diff
@@ -0,0 +1,34 @@
+const path = require('path');
+const glob = require('glob');
+
+const spawnPromiseWrapper = require('./spawn-promise-wrapper');
+
+const PROJECT_ID = 'workbox-bab1f';
+const BUCKET_NAME = 'workbox-cdn';
+
+module.exports = (tagName, directoryToUpload) => {
+  return spawnPromiseWrapper('gcloud', [
```
jeffposnick: Since there's an official `node` client library for the Google Cloud stuff, what do you think about switching to that? It also means we'll have everything needed to publish listed as dependencies in `package.json`, rather than needing to install the Google Cloud CLI out of band.

https://cloud.google.com/appengine/docs/flexible/nodejs/using-cloud-storage

gulp-tasks/publish-bundle.js
```diff
@@ -0,0 +1,211 @@
...
+  return findTagsWithoutBundles()
+  .then((tagsAndReleases) => {
+    return tagsAndReleases.reduce((promiseChain, tagAndRelease) => {
```
jeffposnick: With `async`/`await`, you can use a `for ... of` loop to execute asynchronous code in the proper sequence. `.reduce()` isn't needed anymore!

https://stackoverflow.com/a/37576787/385997

gulp-tasks/utils/upload-bundle-to-cdn.js
```diff
@@ -0,0 +1,34 @@
...
+    'config', 'set', 'project', PROJECT_ID,
+  ])
+  .then(() => {
+    const globPattern = path.posix.join(
+      directoryToUpload, '*');
+    const filePaths = glob.sync(globPattern, {
+      absolute: true,
+    });
+
+    return filePaths.reduce((promiseChain, filePath) => {
```
jeffposnick: Prefer `for ... of` + `await`: https://stackoverflow.com/a/37576787/385997 (or switch to the `node` client library).

gulp-tasks/publish-bundle.js
```diff
@@ -0,0 +1,211 @@
...
+    '--depth', '1',
+    `http://github.com/${GITHUB_OWNER}/${GITHUB_REPO}.git`,
+    tempPath,
+  ];
+
+  return spawnPromiseWrapper('git', params)
+  .then(() => tempPath);
+};
+
+const buildProject = (projectPath) => {
```
gauntface: This is running the build task in the context of what ever the tagged code is (i.e. any missing tags are pulled down, installed their dependencies and running gulp in that project).

I.e. this doesn't necessarily mean we are going to have the same build process.

Happy to change approach if you think it would be better (i.e. a node script that handles the github and CDN publishing seperately from our build process?)

gulp-tasks/utils/cdn-helper.js
```diff
@@ -0,0 +1,102 @@
...
+        destination: `${tagName}/${path.basename(filePath)}`,
+        public: true,
+      });
+    }
+  }
+}
+
+module.exports = new CDNHelper();
+
+/** module.exports = (tagName, directoryToUpload) => {
```
jeffposnick: Do you want to leave this commented out, or just delete it?

PR #748, opened 2017-08-15
Rename cacheWillMatch to cachedResponseWillBeUsed

### LINKED ISSUE #713: Rename cacheWillMatch lifecycle method
**Library Affected**:
*workbox-runtime-caching*

There's currently a lifecycle method triggered by `RequestWrapper.match()` called `cacheWillMatch()`, which gives a plugin the opportunity to inspect the cached response that would normally be returned for a given request, and either return `null` if the cached response isn't valid (because it's expired, for instance) or return an alternative response instead (like a 206 Partial Content version of it).

There's been some internal discussion about renaming this method in the next major semver release, with `cachedResponseWillBeUsed()` as the current leading candidate. If folks have alternative suggestions, add them here.

jeffposnick:
R: @addyosmani @gauntface

Fixes #713

### DIFF
--- a/packages/workbox-range-requests/test/sw/handle-range-request.js
+++ b/packages/workbox-range-requests/test/sw/handle-range-request.js
@@ -31,7 +31,7 @@ describe('Tests for the handle-range-request.js functions', function() {
     const reader = new FileReader();
     const finished = new Promise((resolve) => {
       reader.addEventListener('loadend', (event) => {
-        resolve(event.srcElement.result);
+        resolve(event.target.result);
       });
     });
     reader.readAsText(blob);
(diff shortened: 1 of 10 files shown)


packages/workbox-range-requests/test/sw/handle-range-request.js
```diff
@@ -31,7 +31,7 @@ describe('Tests for the handle-range-request.js functions', function() {
     const reader = new FileReader();
     const finished = new Promise((resolve) => {
       reader.addEventListener('loadend', (event) => {
-        resolve(event.srcElement.result);
+        resolve(event.target.result);
```
jeffposnick: This isn't related to the core change in this PR, but the Firefox tests were flakey when I ran them locally before switching to `target`. Not sure why it wasn't an issue before, though...

PR #751, opened 2017-08-17
V3 errors

gauntface:
R: @jeffposnick @addyosmani 

This was a bit of an exploration PR, so sorry for some of the sweeping changes.

I started to dig into few different things:

1. How we'd publish this to NPM such that modules would have a sane import without making our packages/<module name> feel cluttered
2. Not ship the entry point for rollup
3. Add a custom class for errors that will handle the code -> message conversion.

On dev, the messages.mjs class will be included (i.e. the messages associated with the code, however on prod, only the error code and details is displayed and the messages.mjs is not included in the final bundle.

Things that would be good for feedback:
1. Should we use .mjs? Seems like it'll be the future spec, but does require us to explicitly use .mjs on the imports.
2. Should the tests live in top level  `test` or in the packages? I moved them out since having the module source code in the root (which will be need to support `import ClassName from 'module/ClassName'`) was getting messy with test directory.
3. Should we do something to make it clear to developers where files aren't intended for public use? I.e. I've moved some classes on workbox-core to "INTERNAL", the idea being that this should make it obvious that it's not meant for public use but doesn't prevent developers from using it.


jeffposnick:
I get this when I run any `gulp` command:

```sh
jeffy [~/git/workbox]$ gulp test
/Users/jeffy/git/workbox/node_modules/github/lib/index.js:363
            throw new Error("Token authentication requires a token to be set");
            ^

Error: Token authentication requires a token to be set
    at module.exports.__dirname.authenticate (/Users/jeffy/git/workbox/node_modules/github/lib/index.js:363:19)
    at Object.<anonymous> (/Users/jeffy/git/workbox/gulp-tasks/utils/github-helper.js:8:8)
    at Module._compile (module.js:573:30)
    at Object.Module._extensions..js (module.js:584:10)
    at Module.load (module.js:507:32)
    at tryModuleLoad (module.js:470:12)
    at Function.Module._load (module.js:462:3)
    at Module.require (module.js:517:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/Users/jeffy/git/workbox/gulp-tasks/publish-github.js:6:22)
```

It looks like the `github.authenticate()` call happens unconditionally when `gulp` starts up and `gulp-tasks` are loaded, regardless of which task you actually plan on running. I don't have `process.env.GITHUB_TOKEN` set and most other folks who want to run non-publishing tasks won't either. This behavior was introduced in a previous PR.

I'm going to commit a fix to your open PR to make it so that the authentication only happens when you actually run a task that needs it, if you don't mind.

(And then I'll get down to actually reviewing.)

jeffposnick:
A response to your high-level questions before I get into the line-level feedback:

> Should we use .mjs? Seems like it'll be the future spec, but does require us to explicitly use .mjs on the imports.

I'm 👍  with this based on my current understanding of where tooling best practices are now, and having to include the `.mjs` extension in our `import` statements isn't a big deal.

> Should the tests live in top level test or in the packages? I moved them out since having the module source code in the root (which will be need to support import ClassName from 'module/ClassName') was getting messy with test directory.

Top-level `test/` looks fine to me, if it makes things easier for other parts of the infrastructure.

> Should we do something to make it clear to developers where files aren't intended for public use? I.e. I've moved some classes on workbox-core to "INTERNAL", the idea being that this should make it obvious that it's not meant for public use but doesn't prevent developers from using it.

I might go with `internal` rather than `INTERNAL`, just for aesthetics, but otherwise this seems like a reasonable approach.

gulp-tasks/utils/constants.js
```diff
@@ -18,4 +15,12 @@ module.exports = {
   // TODO: Change to v3.0.0 when v3 is launched to avoid doing anything
   // with prelease versions.
   MIN_RELEASE_TAG_TO_PUBLISH: 'v3.0.0-alpha',
+  GITHUB_OWNER: 'GoogleChrome',
+  GITHUB_REPO: 'workbox',
+
+  // This is the namespace of the workbox module.
+  NAMESPACE_PREFIX: 'google.workbox',
```
jeffposnick: I'm not sure what the best place to have this discussion is at this point (happy to move to the tracking issue or the Google Doc), but what are the benefits of `google.workbox.` vs. `workbox.` as the namespace prefix?

gulp-tasks/build-packages.js
```diff
@@ -14,59 +14,72 @@ const logHelper = require('./utils/log-helper');
 const pkgPathToName = require('./utils/pkg-path-to-name');
 const rollupHelper = require('./utils/rollup-helper');
 
+const ERROR_NO_BROWSER_BUNDLE = `Could not find the browser bundle: `;
+const ERROR_NO_NAMSPACE = oneLine`
+  You must define a 'browserNamespace' parameter in the 'package.json'.
+  Exmaple: 'workbox-precaching' would have a browserNamespace param of
+  'precaching' in 'package.json'. This will be appended to 'google.workbox'
```
jeffposnick: `google.workbox` -> `${constants.NAMESPACE_PREFIX}`

(Just in case it changes in the future.)

gulp-tasks/build-packages.js
```diff
@@ -14,59 +14,72 @@ const logHelper = require('./utils/log-helper');
...
+  meaning developers would use 'google.workbox.precaching' in their
```
jeffposnick: `google.workbox` -> `${constants.NAMESPACE_PREFIX}`

(Just in case it changes in the future.)

packages/workbox-core/.npmignore
```diff
@@ -0,0 +1 @@
+browser.mjs
```
jeffposnick: Just checking: we can't use [`package.files`](https://docs.npmjs.com/files/package.json#files) to whitelist what we publish to `npm`, because that's going to include a bunch of other `.mjs` files, including some at the same level as `browser.mjs`. So it's easier for you to just blacklist that one file?

packages/workbox-core/INTERNAL/models/messages/messageGenerator.mjs
```diff
@@ -0,0 +1,29 @@
+import messages from './messages.mjs';
+import core from '../../../index.mjs';
+
+const fallback = (code, ...args) => {
+  let msg = `${code}`;
```
jeffposnick: This looks odd—can't you just say `let msg = code;`?

packages/workbox-core/INTERNAL/models/messages/messageGenerator.mjs
```diff
@@ -0,0 +1,29 @@
...
+  const message = messages[code];
+  if (!message) {
+    core.INTERNAL.logHelper.warn(`Unable to find message for code '${code}'`);
+    return fallback(code, ...args);
+  }
+
+  try {
+    return message(...args);
+  } catch (err) {
+    return fallback(code, ...args);
```
jeffposnick: Can you add a

```js
console.error(`Unable to generate full error message: ${err}`);
```

here so that we can figure out why `message(...args)` failed?

packages/workbox-core/INTERNAL/models/messages/messages.mjs
```diff
@@ -0,0 +1,19 @@
+export default {
+  'invalid-type': ({paramName, expectedType, value}) => {
```
jeffposnick: Add  `= {}` to gracefully deal with the case when someone calls this without passing in any parameters.

packages/workbox-core/INTERNAL/utils/LogHelper.mjs
```diff
@@ -43,13 +45,21 @@ export default class LogHelper {
...
+      });
     }
 
     if (newLevel > ERROR_LOG_LEVEL ||
       newLevel < VERBOSE_LOG_LEVEL) {
-      // TODO: Throw workbox error + error code.
-      throw new Error(`Invalid new level`);
+      throw new WorkboxError('invalid-value', {
+        paramName: 'logLevel',
+        validValueDescription: `Please use a value from 'LOG_LEVEL', i.e ` +
```
jeffposnick: `'LOG_LEVEL'` -> `LOG_LEVELS`

(Drop the single-quotes around the getter name and add the `S`.)

gulp-tasks/utils/github-helper.js
```diff
@@ -15,28 +12,39 @@ github.authenticate({
...
   uploadAsset: (args) => {
-    args.owner = GITHUB_OWNER;
-    args.repo = GITHUB_REPO;
+    args.owner = constants.GITHUB_OWNER;
+    args.repo = constants.GITHUB_REPO;
     return github.repos.uploadAsset(args);
   },
 
-  getReleases: () => {
+  getTaggedReleases: () => {
```
jeffposnick: Another nudge towards `async`/`await`.

gulp-tasks/utils/constants.js
gauntface: Only reason I did this (And I'm not too fussed either way) is that I had imaginged the common use case with the `workbox-loader` / CDN will be that we'd ship a 'Workbox' class that you'd use as:

```javascript
const workbox = new Workbox();
workbox.precaching.precache([]);
workbox.router.registerRoute(.....);
```

With this in mind, it gets messy if we use `workbox.` as the name space.

Happy to change it though.

packages/workbox-core/.npmignore
gauntface: Exactly. We can use an unfriendly file name for this like `._browser.mjs` which npm will exclude by default, but I think this in cleaner.

That being said, I'm quickly leaning on auto-generating this file so we can ensure public classes / files are exported in the expected way.

packages/workbox-core/INTERNAL/models/messages/messages.mjs
gauntface: The bad scenario with this is that you end up with a message that doesn't make any sense.

> The parameter 'undefined' have an unexpected type. Expect type 'undefined'. Receive 'undefined'.

As an alternative, we can have a default message that is super generic (i.e. "A parameter had an unexpected type, please see stack trace for more info." then swap it out for "The paramter '${paramName}' ..." but the code quickly got verbose and just felt unusual).

I'm still not 100% happy with this way of generate useful error codes. Do you have any ideas of how this could be improved / what a better API would be?

gauntface:
@jeffposnick I responded to the log comment but Github collapsed it, its still not changed. Also left a comment on the namespace question.

jeffposnick:
This direction looks 👍 

Not sure whether you expect the tests to all pass right now, but Travis CI is unhappy.

PR #752, opened 2017-08-17
No longer register a fetch handler for each Router by default

### LINKED ISSUE #721: Router.handleRequest() Proposal
At the moment the Router is fairly all encompassing meaning that if developers wanted to use the routing logic with logic before and / or after the routers log, they couldn't.

**NOTE:** This is largely a suggestion from @jakearchibalds proposal from #700 .

## Add explicit "handleRequest()" method to Router class

At the moment the Router class itself adds a fetch handler.

We should alter it to take a similar approach where the "instance" of the module (i.e. the friendlier / 90% use case of the classes) creates the fetch handler and under the hood it calls a "handleRequest()" method.

So instead of in the Router code:

```javascript
const router = new Router();
router.registerRoute(...);
```

We'd have to do the following when using the Router directly:

```javascript
const router = new Router();
router.registerRoute(...);

self.addEventListener('fetch', (event) => {
  event.waitUntil(async () => {
      const response = await router.handleRequest(event);
      if (response) {
          await event.respondWith(response);
      }
  });
});
```

The end result for most users is that they'd use an instance of the router module that would handle this on behalf of the developer (i.e. they'd never write this code).

**Pros:**
- Provides flexibility to developers who want to manipulate requests before or after router behaviour
- Matches the precache model in terms of the lowest level code creates no events.

**Cons:**
- Unclear of the immediate use cases, but can imagine scenarios where other libraries could be used with Workbox by being able to manipulate requests / responses around the router.

jeffposnick:
R: @addyosmani @gauntface
CC: @jakearchibald @surma

Fixes #721

This changes the default behavior of `Router`, as discussed in the proposal.


packages/workbox-routing/src/lib/router.js
```diff
@@ -19,14 +19,19 @@ import logHelper from '../../../../lib/log-helper.js';
...
- * that will respond to network requests if there's a matching route.
+ * The Router takes one or more [Routes]{@link Route} and allows you to apply
+ * that routing logic to determine the appropriate way of handling requests
+ * inside of a service worker.
  *
  * It also allows you to define a "default" handler that applies to any requests
  * that don't explicitly match a `Route`, and a "catch" handler that responds
  * to any requests that throw an exception while being routed.
  *
+ * By default, the `Router` class won't register a `fetch` event handler,
```
gauntface: Nit: Could we rephrase to:

You can use the handleRequest() method to pass a fetch event through the router and ultimately get a "Routed response" back. If you'd like this to be handled automatically for you, calling `addFetchListener()` will cause the Router to respond to requests

packages/workbox-routing/src/lib/router.js
```diff
@@ -46,113 +51,152 @@ import normalizeHandler from './normalize-handler';
...
+   * const imageRoute = new RegExpRoute({
+   *   regExp: /images/,
+   *   handler: new CacheFirst(),
+   * });
+   *
+   * const router = new Router();
+   * router.registerRoute({route: imageRoute});
+   * router.addFetchListener();
+   */
+  addFetchListener() {
```
gauntface: We should check to see if this method has already been called and if so, display a warning and return early.

packages/workbox-routing/src/lib/router.js
```diff
@@ -46,113 +51,152 @@ import normalizeHandler from './normalize-handler';
...
+    self.addEventListener('fetch', (event) => {
+      const responsePromise = this.handleRequest({event});
```
gauntface: Q: Do you need to call waitUntil at all? or does a Promise that resolves to a Response cause the service worker to wait and respond correctly?

packages/workbox-routing/src/lib/router.js
```diff
@@ -46,113 +51,152 @@ import normalizeHandler from './normalize-handler';
...
+          return this.catchHandler.handle({url, event, error});
+        });
+      }
+      return responsePromise;
     }
   }
 
   /**
-   * This method will actually add the fetch event listener.
+   * Checks the incoming even.request against the registered routes, and if
```
gauntface: Nit: minor typo even**t**.request

packages/workbox-routing/src/lib/router.js
```diff
@@ -46,113 +51,152 @@ import normalizeHandler from './normalize-handler';
...
+   * there's a match, returns the corresponding handler along with any params
+   * generated by the match.
+   *
+   * @param {FetchEvent} input.event
+   * @param {URL} input.url
+   * @return {Object} Returns an object with `handler` and `params` properties
```
gauntface: Nit: Can drop the "set to appropriate values".

packages/workbox-routing/src/lib/router.js
```diff
@@ -46,113 +51,152 @@ import normalizeHandler from './normalize-handler';
...
+   * set to appropriate values if there was a match, and set to `undefined` if
+   * there was not matching route.
```
gauntface: Nit: Change "not" to "no"

packages/workbox-routing/test/sw/router.js
```diff
@@ -73,4 +49,32 @@ describe(`Test of the Router class`, function() {
     expect(router._routes.get('GET')).to.have.members([getRoute1]);
     expect(router._routes.get('PUT')).to.have.members([putRoute1]);
   });
+
+  it(`should call self.addEventListener('fetch') when addFetchListener() is called`, function() {
+    const stub = sinon.stub(self.__proto__.__proto__.__proto__, 'addEventListener');
```
gauntface: why so many __proto__?

packages/workbox-routing/src/lib/router.js
```diff
@@ -46,113 +51,152 @@ import normalizeHandler from './normalize-handler';
...
 class Router {
   /**
-   * Start with an empty array of routes, and by default, sets up the `fetch`
-   * handler.
-   *
-   * @param {Boolean} [handleFetch=true] Whether or not the Router should
-   * register a `fetch` handler and respond to requests. This can be set to
-   * `false` in a development environment, to prevent the service worker
-   * from responding with cached responses.
+   * Constructs a new instance, without any registered routes.
```
addyosmani: Nit: constructs a new `Router` instance

packages/workbox-routing/src/lib/router.js
addyosmani: Out of interest, what happens with this implementation if `addFetchListener` is called multiple times?

packages/workbox-routing/src/lib/router.js
addyosmani: 👍

packages/workbox-routing/src/lib/router.js
addyosmani: 👍 Matt's suggestion.  If I might suggest one further edit "Returns an object with `handler` and `params `properties. They are populated if a matching route was found or `undefined` otherwise".

packages/workbox-routing/src/lib/router.js
addyosmani: 👍

packages/workbox-routing/test/sw/router.js
addyosmani: This seems super strange :) I'm curious about the proto depth here, but imagine there's a good reason.

packages/workbox-routing/src/lib/router.js
jeffposnick: Done.

packages/workbox-routing/src/lib/router.js
jeffposnick: Sorry, I left that bit out—we agreed on returning a `boolean` based on whether that call actually registered the listener, and to prevent the listener from being registered multiple times. Will adjust.

packages/workbox-routing/src/lib/router.js
jeffposnick: The latter.

`waitUntil` is only needed if there are multiple asynchronous things that are part of separate promise chains going on inside the event handler. When there's just a single promise chain that resolves with the `Response`, it can be passed directly to `respondWith()`.

packages/workbox-routing/src/lib/router.js
jeffposnick: Done.

packages/workbox-routing/src/lib/router.js
jeffposnick: Done, with @addyosmani's language.

packages/workbox-routing/test/sw/router.js
jeffposnick: I've added a comment explaining what's going on. If there's a better approach for stubbing out `self.addEventListener()` without triggering a `mocha` global leak warning, I'm definitely open, as I had to bang my head for a while before I was able to get even that to work.

jeffposnick:
PTAL.

PR #753, opened 2017-08-18
use strict to support Node 4

Lyricalz:
> /home/travis/build/twbs/bootstrap/node_modules/workbox-build/build/lib/utils/get-file-details.js:9
  let globbedFiles;
  ^^^
SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode

build failed on node 4, due to "let".
this should fix it!

### DIFF
--- a/packages/workbox-build/src/lib/utils/get-file-details.js
+++ b/packages/workbox-build/src/lib/utils/get-file-details.js
@@ -1,3 +1,5 @@
+'use strict';
+
 const glob = require('glob');
 const path = require('path');
 
--- a/packages/workbox-build/src/lib/utils/no-revision-for-urls-matching-transform.js
+++ b/packages/workbox-build/src/lib/utils/no-revision-for-urls-matching-transform.js
@@ -1,3 +1,5 @@
+'use strict';
+
 const errors = require('../errors');
 
 module.exports = (regexp) => {


Lyricalz:
Signed it

Lyricalz:
can't figure out why appveyor failed, not sure either as it's a direct fork of the master branch..

packages/workbox-build/src/lib/utils/get-file-details.js
```diff
@@ -1,3 +1,4 @@
+'use strict';
```
gauntface: Could you add single new line under use strict for each of these.

gauntface:
@Lyricalz thanks for this. Don't worry about appveyor, it's been cranky for a while now. I need to re-think some of the tests on windows.

If you can add a new line under each 'use strict'; and I'll merge away.

Lyricalz:
Should be good now

gauntface:
Thanks for this @Lyricalz. 

We'll be cutting v2 of Workbox on Wednesday and this should be a part of this change.

XhmikosR:
@gauntface @addyosmani: any news on the new release?

addyosmani:
2.0 should now be published as latest.

PR #762, opened 2017-08-22
Ensure that ignoreUrlParametersMatching works with directoryIndex

### LINKED ISSUE #758: Using ignoreUrlParametersMatching has a bug when testing with `directoryIndex`
**Library Affected**:
workbox-sw

**Browser & Platform**:
all browsers

Issue: using option `ignoreUrlParametersMatching` does not work as expected.

Expected behavior is for the `service-worker` to serve the following urls `//localhost/` & `//localhost/?launcher=true`. But `//localhost/?launcher=true` doesn't get served from cache despite the `launcher` parameter being added to be ignored.

Used options:
```js
new self.WorkboxSW({
  "skipWaiting": true,
  "ignoreUrlParametersMatching": [/launcher/]
});
```

I believe this is a bug with the `capture` function which tries to add the `directoryIndex` and test against the original `url` instead of the `strippedUrl`.

Suspected culprit line is: https://github.com/GoogleChrome/workbox/blob/master/packages/workbox-sw/src/lib/workbox-sw.js#L329

N/A

jeffposnick:
R: @addyosmani @gauntface

Fixes #758

This fixes a small bug due to using the wrong variable inside of our WorkboxSW precache route.

### DIFF
--- a/packages/workbox-sw/test/sw/directory-index-custom.js
+++ b/packages/workbox-sw/test/sw/directory-index-custom.js
@@ -59,4 +59,32 @@ describe(`Test Directory Index`, function() {
       self.dispatchEvent(fetchEvent);
     });
   });
+
+  it(`should take the ignoreUrlParametersMatching setting into account when using the directoryIndex`, function() {
+    const EXAMPLE_URL = '/example/url/';
+    const DIRECTORY_INDEX = 'custom.html';
+    const URL_PARAM = 'test';
+    const IGNORE_URL_PARAMETERS_MATCHING = [new RegExp(URL_PARAM)];
+
+    const workboxSW = new WorkboxSW({
+      directoryIndex: DIRECTORY_INDEX,
+      ignoreUrlParametersMatching: IGNORE_URL_PARAMETERS_MATCHING,
+    });
+
+    workboxSW.precache([`${EXAMPLE_URL}${DIRECTORY_INDEX}`]);
+
+    return new Promise((resolve) => {
+      const fetchEvent = new FetchEvent('fetch', {
+        // Fire a request for the URL without the directory index, but with the URL param.
+        request: new Request(`${EXAMPLE_URL}?${URL_PARAM}`),
+      });
+
+      // cache.match() will only be called if the precache route's capture
+      // function returned true, which implies that the URL param was ignored.
+      const stub = sinon.stub(Cache.prototype, 'match').callsFake(resolve);
+      stubs.push(stub);
+
+      self.dispatchEvent(fetchEvent);
+    });
+  });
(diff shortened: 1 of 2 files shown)


jeffposnick:
Something about the new test is causing Firefox to crash... tracking that down now.

jeffposnick:
I've added some additional comments and refactored some of the variables, and hopefully it's mergeable now. I had initially tried to come up with a more straightforward way of testing the actual functionality, and what I ended up with was about as direct as I could figure out.

(Stubbing out some of the `Route` methods didn't seem possible, as they're not exposed anywhere in the public interface for `WorkboxSW`.)

packages/workbox-sw/test/sw/directory-index-custom.js
```diff
@@ -59,4 +59,45 @@ describe(`Test Directory Index`, function() {
...
+    return new Promise((resolve, reject) => {
+      const fetchEvent = new FetchEvent('fetch', {
+        // Request the URL without the directory index, but with the URL param.
+        request: new Request(urlToRequest),
+      });
+      fetchEvent.respondWith = (promiseChain) => {
+        promiseChain.then(() => {
+          // Make sure that the cache.match() stub was called, indicating that
+          // there was a match against the precached URL.
+          if (stub.called) {
```
gauntface: What is caches.match called with? Should it be with or without the directoryIndex value or not?

packages/workbox-sw/test/sw/directory-index-custom.js
```diff
@@ -59,4 +59,45 @@ describe(`Test Directory Index`, function() {
...
+      ignoreUrlParametersMatching,
+    });
+    workboxSW.precache([urlToPrecache]);
+
+    return new Promise((resolve, reject) => {
+      const fetchEvent = new FetchEvent('fetch', {
+        // Request the URL without the directory index, but with the URL param.
+        request: new Request(urlToRequest),
+      });
+      fetchEvent.respondWith = (promiseChain) => {
```
gauntface: Can you stub this out? Would make it a little easier to read  / follow if you could

packages/workbox-sw/test/sw/directory-index-custom.js
jeffposnick: Added in a check.

packages/workbox-sw/test/sw/directory-index-custom.js
jeffposnick: I don't think I can stub out `respondWith()` and still get things to work, no. (I copied this pattern from other tests in this suite, e.g. the previous test in this file.)

PR #766, opened 2017-08-23
V3 browser bundle

gauntface:
R: @addyosmani @philipwalton 

This is nearly the last piece for having an initial ES2015 + Browser Bundle - Code duplication set up.

For `workbox-core`'s current code it'll generate a browser bundle file of:

``` javascript
import modulesDefaultExport from '../index.mjs';
import messageGenerator from '../internal/models/messages/messageGenerator.mjs';
import messages from '../internal/models/messages/messages.mjs';
import WorkboxError from '../internal/models/WorkboxError.mjs';
import LogHelper from '../internal/utils/LogHelper.mjs';

export default {
  default: modulesDefaultExport,
  internal: {
    models: {
      messages: {
        messageGenerator,
        messages
      },
      WorkboxError
    },
    utils: {
      LogHelper
    }
  }
}
```

Which matches the file structure and matches what it shown in the browser.


addyosmani:
Pardon a top-level question as I'm likely misreading this: in the code snippet above you have fully qualified local system paths to each dependency being imported. What happens when this bundle gets deployed? Is it assumed there will be a 1:1 mapping between the server directory structure and what was built locally? or am I reading this incorrectly?

gauntface:
@addyosmani I'll make them relative paths, should be a simple change, but for the record:

This code should be never commited or published anywhere. In otherwords, developers should never see this, they are expected to use the browser bundle OR the actual source code.

The filename `._browser-entry.mjs` prevents it from being included by npm pack.

I'll probably need to double check the github release and CDN upload.

gauntface:
@addyosmani Fixed the original comment with the new output with the latest changes.

gulp-tasks/build-packages.js
```diff
@@ -31,11 +32,12 @@ const ERROR_NO_NAMSPACE = oneLine`
...
 const buildPackage = (packagePath, buildType) => {
   const packageName = pkgPathToName(packagePath);
-  const browserBundlePath = path.join(packagePath, 'browser.mjs');
+  const browserEntry = path.join(packagePath, constants.PACKAGE_BUILD_DIRNAME,
+    constants.BROWSER_ENTRY_FILENAME);
 
   // First check if the bundle file exists, if it doesn't
   // there is nothing to build
-  if (!fs.pathExistsSync(browserBundlePath)) {
+  if (!fs.pathExistsSync(browserEntry)) {
```
addyosmani: It's unclear from the name `browserEntry` what this is expected to be. I assume as you're checking `pathExistsSync(...)` that this could perhaps be `browserEntryPath`. Anything to improve readability would be appreciated.

gulp-tasks/build-packages.js
```diff
@@ -116,13 +116,81 @@ gulp.task('build-packages:clean', gulp.series(
...
+      outputStrings.push(`${key}: ${valueString}`);
+    }
+  });
+  const padding = '  '.repeat(levels + 1);
+  const closePadding = '  '.repeat(levels);
+  const joinString = `,\n${padding}`;
+  return `{\n${padding}${outputStrings.join(joinString)}\n${closePadding}}`;
+};
+
+const generateBrowserEntry = (pkgPath) => {
```
addyosmani: Minor: For `generateBrowserEntry()` and `convertExportObject()`, could we add a short comment summary to remind folks what they're accomplishing? It would be useful to get a thousand foot view when scrolling through without having to parse the relationships between these different pieces.

PR #770, opened 2017-08-25
Update workbox-google-analytics to use background sync

philipwalton:

One thing to note is this introduces a breaking change to the API (that we discussed and agreed upon offline). Previously a user could throw in the `hitFilter()` configuration option function to prevent a hit from being replayed. We now no longer support that.


addyosmani:
Super minor nit: The queue of changes here includes several commits that already landed in master. To update a branch to reflect the latest changes there you can usually follow something like this: https://stackoverflow.com/a/8965803

packages/workbox-google-analytics/package.json
```diff
@@ -10,20 +10,22 @@
...
-  },
   "scripts": {
     "prepublish": "gulp build --project workbox-google-analytics"
   },
+  "author": "Google's Web DevRel Team",
   "license": "Apache-2.0",
   "repository": "googlechrome/workbox",
   "bugs": "https://github.com/googlechrome/workbox/issues",
   "homepage": "https://github.com/GoogleChrome/workbox/tree/master/packages/workbox-google-analytics",
+  "devDependencies": {
```
gauntface: Shouldn't these be actual dependencies?

packages/workbox-google-analytics/src/index.js
```diff
@@ -56,83 +49,122 @@ import replayQueuedRequests from './lib/replay-queued-requests.js';
  * // might be useful is if you wanted to track the amount of time that elapsed
  * // between when the hit was attempted and when it was successfully replayed.
  * workbox.googleAnalytics.initialize({
- *   hitFilter: searchParams =>
+ *   hitFilter: (params) =>
  *     // Sets the `qt` param as a custom metric.
```
gauntface: Stupid comment: but could you add a small depscription of what 'qt' and 'cm1' is?

packages/workbox-google-analytics/src/index.js
```diff
@@ -56,83 +49,122 @@ import replayQueuedRequests from './lib/replay-queued-requests.js';
...
-          caches.open(constants.CACHE_NAME).then((cache) => {
-            return fetch(request).then((response) => {
-              return cache.put(request, response.clone()).then(() => response);
-            }).catch((error) => {
-              logHelper.error(error);
-              return cache.match(request);
-            });
-          })
-        );
+const createRequestWillDequeueCallback = (config) => {
```
gauntface: When is this method run? the naming doesn't tell me if it's run before the request is removed from the queue to attempt the request or if the request re-attempt has been done and successful and now it should be de-queued.

packages/workbox-google-analytics/src/index.js
```diff
@@ -56,83 +49,122 @@ import replayQueuedRequests from './lib/replay-queued-requests.js';
...
+  return (reqData) => {
+    const {request} = reqData;
```
gauntface: What is this line doing?

packages/workbox-google-analytics/src/index.js
```diff
@@ -56,83 +49,122 @@ import replayQueuedRequests from './lib/replay-queued-requests.js';
...
+    const hitTime = reqData.metadata.creationTimestamp;
+    const queueTime = Date.now() - hitTime;
+    const params = request.body ? new URLSearchParams(request.body) :
```
gauntface: A description here would be useful

packages/workbox-google-analytics/src/index.js
```diff
@@ -56,83 +49,122 @@ import replayQueuedRequests from './lib/replay-queued-requests.js';
...
+
+    if (config.parameterOverrides) {
+      for (const param of Object.keys(config.parameterOverrides)) {
+        const value = config.parameterOverrides[param];
+        params.set(param, value);
       }
     }
+
+    if (typeof config.hitFilter == 'function') {
+      config.hitFilter.call(null, params);
```
gauntface: Is there a reason for this approach of executing the function?

packages/workbox-google-analytics/src/index.js
```diff
@@ -56,83 +49,122 @@ import replayQueuedRequests from './lib/replay-queued-requests.js';
...
+    }
+
+    request.body = params.toString();
+    request.method = 'POST';
+    request.mode = 'cors';
+    request.credentials = 'omit';
```
gauntface: Isn't this the default mode for requests?

packages/workbox-google-analytics/src/index.js
```diff
@@ -56,83 +49,122 @@ import replayQueuedRequests from './lib/replay-queued-requests.js';
...
+  };
+};
+
+/**
+ * Creates GET and POST routes to catch failed Measurement Protocol hits.
+ * @param {Function} requestWillDequeueCallback The callback used by
+ *     background sync when requests are dequeued and retried.
+ * @return {Array<Route>} The created routes.
+ */
+const createCollectRoutes = (requestWillDequeueCallback) => {
```
gauntface: Why `createCollectRoutes`? Where did "collect" come from? I was expecting `createAnalyticsRoutes`

packages/workbox-google-analytics/src/index.js
```diff
@@ -56,83 +49,122 @@ import replayQueuedRequests from './lib/replay-queued-requests.js';
...
+ *     expressed as key/value pairs, to be added to replayed Google Analytics
+ *     requests. This can be used to, e.g., set a custom dimension indicating
+ *     that the request was replayed.
+ * @param {Function} [config.hitFilter] A function that allows you to modify
+ *     the hit parameters prior to replaying
+ *     the hit. The function is invoked with the original hit's URLSearchParams
+ *     object as its only argument.
+ * @memberof module:workbox-google-analytics
+ */
+export const initialize = (config = {}) => {
```
gauntface: what happens if I call this multiple times?

packages/workbox-google-analytics/src/index.js
```diff
@@ -56,83 +49,122 @@ import replayQueuedRequests from './lib/replay-queued-requests.js';
...
+};
+
+/**
+ * Creates a route with a network first strategy for the analytics.js script.
+ * @return {Route} The created route.
+ */
+const createAnalyticsJsRoute = () => {
+  return new Route({
+    match: ({url}) => {
+      return url.hostname == constants.URL.HOST &&
```
gauntface: constants.URL.HOST is too generic. What URL, what HOST? Why not just constants.ANALYTICS_HOST_URL.

packages/workbox-google-analytics/src/index.js
```diff
@@ -56,83 +49,122 @@ import replayQueuedRequests from './lib/replay-queued-requests.js';
...
+const createCollectRoutes = (requestWillDequeueCallback) => {
+  const bgSyncQueuePlugin = new QueuePlugin({
+    dbName: constants.IDB.NAME,
+    maxRetentionTime: constants.STOP_RETRYING_AFTER,
+    callbacks: {requestWillDequeue: requestWillDequeueCallback},
   });
 
-  replayQueuedRequests(config);
+  const match = ({url}) => url.hostname == constants.URL.HOST &&
+      url.pathname == constants.URL.COLLECT_PATH;
```
gauntface: should URL.COLLECT_PATH be MEASURE_PROTOCOL_URL?

packages/workbox-google-analytics/src/index.js
```diff
@@ -56,83 +49,122 @@ import replayQueuedRequests from './lib/replay-queued-requests.js';
...
+
+  const handler = new NetworkOnly({
+    requestWrapper: new RequestWrapper({plugins: [bgSyncQueuePlugin]}),
+  });
+
+  return [
+    new Route({method: 'GET', match, handler}),
```
gauntface: Can't this be a single Router where the match function checks and matches the requests?

packages/workbox-google-analytics/test/browser/index.js
```diff
@@ -0,0 +1,64 @@
...
+          resolve();
+        } else {
+          reject(new Error('replay:failure'));
+        }
+        channel.close();
+      };
+    });
+
+    try {
+      iframe.contentWindow.navigator.sendBeacon(
```
gauntface: Why sendBeacon and a fetch request?

packages/workbox-google-analytics/test/browser/index.js
```diff
@@ -0,0 +1,64 @@
...
+          `https://${constants.URL.HOST}${constants.URL.COLLECT_PATH}`,
+          HIT_PAYLOAD + 'foo=1');
```
gauntface: why foo=1?

packages/workbox-google-analytics/test/browser/index.js
```diff
@@ -0,0 +1,64 @@
...
+
+      await iframe.contentWindow.fetch(`https://${constants.URL.HOST}` +
+          `${constants.URL.COLLECT_PATH}?${HIT_PAYLOAD}&foo=2`);
```
gauntface: why foo=2?

packages/workbox-google-analytics/test/browser/index.js
```diff
@@ -0,0 +1,64 @@
...
+
+  it(`should add failed requests to a background sync queue and replay them`,
+      async function() {
+    const iframe = await window.goog.swUtils.activateSW(
+        `${STATIC_ASSETS_PATH}/default.js`);
+
+    // A promise that resolves once the sync event fires and the requests
+    // are replayed in the service worker.
+    const syncMessageReceived = new Promise((resolve, reject) => {
+      const channel = new BroadcastChannel('workbox-google-analytics-sync');
```
gauntface: Why is there a broadcast channel being used here? The test should fail requests, check queue for request and then replay them. I'm not seeing that from this test.

packages/workbox-google-analytics/test/static/default.js
```diff
@@ -0,0 +1,57 @@
...
+
+importScripts('/node_modules/sinon/pkg/sinon.js');
+importScripts('/__test/bundle/workbox-google-analytics');
+
+self.addEventListener('install', () => self.skipWaiting());
+self.addEventListener('activate', (evt) => evt.waitUntil(clients.claim()));
+
+// Add a sync event prior to calling googleAnalytics.initialize(), so this
+// listener runs before any other sync listeners.
+self.addEventListener('sync', (evt) => {
```
gauntface: Why are you setting a sync here at all?

packages/workbox-google-analytics/test/sw/index.js
```diff
@@ -0,0 +1,349 @@
...
+  } else {
+    await sleep(100);
+    return waitUntil(test);
+  }
+};
+
+describe(`initialize`, function() {
+  let idbMap = {};
+
+  const stubIdbHelper = () => {
```
gauntface: Why are you stubbing all of this out?

packages/workbox-google-analytics/test/sw/index.js
```diff
@@ -0,0 +1,349 @@
...
+    trackAddedEventListeners();
+    stubIdbHelper();
+  });
+
+  afterEach(function() {
+    removeAllEventListeners();
+    restoreIdbHelper();
+  });
+
+  it(`should register a handler to cache the analytics.js script`, function() {
```
gauntface: This test could have been a browser level test (i.e.  register a sw that just init's the analytics library, requests analytics JS and then checks the cache).

packages/workbox-google-analytics/test/sw/index.js
```diff
@@ -0,0 +1,349 @@
...
+    self.dispatchEvent(new FetchEvent('fetch', {
+      request: new Request(
+          `https://${constants.URL.HOST}${constants.URL.ANALYTICS_JS_PATH}`, {
+        mode: 'no-cors',
+      }),
+    }));
+
+    expect(NetworkFirst.prototype.handle.calledOnce).to.be.ok;
+
+    NetworkFirst.prototype.handle.restore();
```
gauntface: Could you use sinon sandboxing?

you create a sandbox and then you'd go sandbox.spy(NetworkFirst.prototype, 'handle') then in afterEach just go sandbox.restore()

packages/workbox-google-analytics/test/sw/index.js
```diff
@@ -0,0 +1,349 @@
...
+
+    self.dispatchEvent(new FetchEvent('fetch', {
+      request: new Request(`https://${constants.URL.HOST}` +
+          `${constants.URL.COLLECT_PATH}?${PAYLOAD}`, {
+        method: 'GET',
+      }),
+    }));
+
+    expect(self.fetch.calledOnce).to.be.ok;
+    expect(self.fetch.firstCall.args[0].url).to.equal(`https://` +
```
gauntface: I'm lost as to what you are checking here. You make a request with a url and check the request had that URL.

packages/workbox-google-analytics/test/sw/index.js
```diff
@@ -158,11 +158,10 @@ describe(`initialize`, function() {
...
     googleAnalytics.initialize();
-    const payload = 'v=1&t=pageview&tid=UA-12345-1&cid=1&dp=%2F';
 
     self.dispatchEvent(new FetchEvent('fetch', {
       request: new Request(`https://${constants.URL.HOST}` +
-          `${constants.URL.COLLECT_PATH}?${payload}`, {
+          `${constants.URL.COLLECT_PATH}?${PAYLOAD}`, {
         method: 'GET',
       }),
     }));
```
addyosmani: In general adding better comment descriptions to summarize what you are testing here would help. Especially if this is unclear from a read through.

packages/workbox-google-analytics/src/index.js
gauntface: Could this single file be split up into individual files / classes that perform the individual pieces of work? I.e. an AnalyticsRoute, AnalyticsPlugin (for the createRequestWillDequeueCallback)

packages/workbox-google-analytics/test/static/default.js
philipwalton: I need to unstub (and then restub) `fetch` in the service worker before the background sync queue is retried.

Since I know a `sync` event will happen before the queue is retried, listening for that was the easiest way.

packages/workbox-google-analytics/test/sw/index.js
philipwalton: We discussed this briefly the other day, but I was having issues with multiple open connections to this IDB database at the same time, and that was preventing me from deleting the DB between tests.

I should confirm that this is still an issue if I delete all the records rather than deleting the DB itself.

packages/workbox-google-analytics/test/sw/index.js
philipwalton: I didn't know about sandboxing, I'll look into it.

packages/workbox-google-analytics/test/sw/index.js
philipwalton: Yeah, I'm making sure the `initialize()` logic doesn't add anything that modifies the request URL in the non-failure case. Basically I'm making sure that the replay logic doesn't inadvertently apply to non-replayed requests.

philipwalton:
@addyosmani rebased. I originally avoided that because I'd merged with master, but rebasing seems to have made the history cleaner, so I pushed with that.

packages/workbox-google-analytics/src/index.js
gauntface: I think it's the `requestWillDequeue` that is most upsetting me.

packages/workbox-google-analytics/src/index.js
gauntface: that or just `const request = reqData.request` (assuming that is all it's doing).

packages/workbox-google-analytics/src/index.js
gauntface: Why shouldn't we protect them from it?

packages/workbox-google-analytics/src/index.js
gauntface: Yes please.

packages/workbox-google-analytics/src/index.js
gauntface: Can we rethink the naming for all this. I'm so far remove from google analytics that these names aren't overly helpful. 
```
constants.ANALYTICS_URLS.<COLLECT_ENDPOINT | BATCH_ENDPOINT | JS_SCRIPT_PATH >, 
```
helps me at least guess what is being checked or done here.

packages/workbox-google-analytics/src/index.js
gauntface: That really sucks. Opened: https://github.com/GoogleChrome/workbox/issues/777

packages/workbox-google-analytics/test/browser/index.js
gauntface: Can you add a comment explaining that.

Could these be pulled out into seperate tests with additonal comments to explain?

packages/workbox-google-analytics/test/browser/index.js
gauntface: Am I right in assuming there is no way to force a bg "sync" event and then getting an update once the request was successfully made?

packages/workbox-google-analytics/test/static/default.js
gauntface: Could you extend the comment to explain this.

packages/workbox-google-analytics/test/sw/index.js
gauntface: comments in the file other wise this will be totally forgotten and cause someone headaches later on.

packages/workbox-google-analytics/test/sw/index.js
gauntface: Yeah Jeff only recently put me onto it, but seems good.

packages/workbox-google-analytics/test/sw/index.js
gauntface: Another comment request please.

packages/workbox-google-analytics/src/index.js
gauntface: It's easier to pull apart the causes of conerns. At the moment it's a lot of scrolling up and down findng methods, for return values and generally it just breaks up the parsing of the code.

packages/workbox-google-analytics/src/index.js
jeffposnick: FWIW, I like the desctructuring syntax.

packages/workbox-google-analytics/package.json
```diff
@@ -10,20 +10,22 @@
...
-  },
   "scripts": {
     "prepublish": "gulp build --project workbox-google-analytics"
   },
+  "author": "Google's Web DevRel Team",
   "license": "Apache-2.0",
   "repository": "googlechrome/workbox",
   "bugs": "https://github.com/googlechrome/workbox/issues",
   "homepage": "https://github.com/GoogleChrome/workbox/tree/master/packages/workbox-google-analytics",
+  "dependencies": {
```
jeffposnick: It looks like `devDependencies` rather than `dependencies` is used elsewhere in the monorepo for cross-package dependencies. Would `devDepencencies` work for your use case?

packages/workbox-google-analytics/src/index.js
```diff
@@ -55,84 +50,128 @@ import replayQueuedRequests from './lib/replay-queued-requests.js';
...
-        );
+const createRequestWillDequeueCallback = (config) => {
+  return (reqData) => {
+    const {request} = reqData;
+    const hitTime = reqData.metadata.creationTimestamp;
+    const queueTime = Date.now() - hitTime;
+
+    // Measurement protocol requests can set their payload parameters in either
+    // the URL query string (for GET requests) or the POST body.
+    const params = request.body ? new URLSearchParams(request.body) :
```
jeffposnick: This relies on [Chrome 61](https://www.chromestatus.com/feature/5737549880360960), which just made it into stable yesterday. I think that's fine, since it's also in Firefox 54+, but it's cutting it close 😄

packages/workbox-google-analytics/src/index.js
```diff
@@ -55,84 +50,128 @@ import replayQueuedRequests from './lib/replay-queued-requests.js';
...
+        new URL(request.url).searchParams;
+
+    // Set the qt param prior to apply the hitFilter or parameterOverrides.
+    params.set('qt', queueTime);
+
+    if (config.parameterOverrides) {
+      for (const param of Object.keys(config.parameterOverrides)) {
```
jeffposnick: You could switch to

```js
for (const [param, value] of Object.entries(config.parameterOverrides)) {
  params.set(param, value);
}
```

packages/workbox-google-analytics/src/index.js
jeffposnick: If it's clearer, you could switch to using `RegExpRoute` and specify the criteria as a RegExp instead. (That might not be clearer!)

packages/workbox-google-analytics/src/index.js
jeffposnick: This might be clearer if you used a `RegExpRoute` with a RegExp that matched against the full URL.

packages/workbox-google-analytics/src/index.js
jeffposnick: Only the first `fetch` event handler would end up being invoked, since it would call `respondWith()` when an incoming request matches a `Route`. So there's no reason why you'd want to call `initialize()` multiple times, but there wouldn't be any effective negative impact.

packages/workbox-google-analytics/test/browser/index.js
```diff
@@ -0,0 +1,64 @@
...
+const HIT_PAYLOAD = 'v=1&t=pageview&tid=UA-12345-1&cid=1&dp=%2F';
+
+describe(`workbox-google-analytics`, function() {
+  const db = new IDBHelper(constants.IDB.NAME, 1, 'QueueStore');
+  const resetDb = async () => {
+    const keys = await db.getAllKeys();
+    return Promise.all(keys.map((key) => db.delete(key)));
+  };
+
+  before(function() {
```
jeffposnick: Do you need this cleanup, and the earlier IDB cleanup? If not, I'd recommend removing it.

packages/workbox-google-analytics/test/browser/index.js
```diff
@@ -0,0 +1,64 @@
...
+    const iframe = await window.goog.swUtils.activateSW(
+        `${STATIC_ASSETS_PATH}/default.js`);
+
+    // A promise that resolves once the sync event fires and the requests
+    // are replayed in the service worker.
+    const syncMessageReceived = new Promise((resolve, reject) => {
+      const channel = new BroadcastChannel('workbox-google-analytics-sync');
+      channel.onmessage = (evt) => {
+        if (evt.data == 'replay:success') {
+          channel.close();
```
jeffposnick: You've got another `channel.close()` outside of the `if()` block, so this could presumably be removed.

packages/workbox-google-analytics/src/index.js
philipwalton: `request.body` here will be a string, so it should be OK. Is that what you're referencing?

packages/workbox-google-analytics/test/browser/index.js
philipwalton: I'm not sure about the `cleanState()` stuff, I just copied that from another test :) so @gauntface can probably better answer that.

As for the IDB cleanup, yes that's needed.

packages/workbox-google-analytics/src/index.js
jeffposnick: Okay, if it's a string then it's probably fine for older browsers too. Chrome 61 added in support for passing in an Object.

packages/workbox-google-analytics/test/browser/index.js
philipwalton: Ahh, yes. Good catch.

packages/workbox-google-analytics/package.json
gauntface: This should have been dependencies since day 1.

philipwalton: Yeah, it should be `dependencies` or it won't get included in browser bundles.

packages/workbox-google-analytics/src/index.js
philipwalton: Is `Object.entries` supported in all our environments? If not are we transpiling it to ES2015?

philipwalton:
This PR has been replaced in v3 with https://github.com/GoogleChrome/workbox/pull/927.

PR #774, opened 2017-08-28
Fix backslashes replacement.

### LINKED ISSUE #771: `entry.url` is using Windows separator
Hi, there.

We are trying to precache our static assets for Bootstrap in this PR: https://github.com/twbs/bootstrap/pull/23533

With `workbox-build`  v2.0.0 on Windows, I'm getting this:

```js
  {
    "url": "about\\index.html",
    "revision": "f9307e727e85be5a2a420ca9a44a4f01"
  },
```

I can just do a replace for the backslashes, but I wonder if this should happen in the first place? I know that URLs do not use backslashes but here maybe this is normal?

Thanks!

XhmikosR:
This resulted in backslashes in the URL entries on Windows.

Fixes #771

### DIFF
--- a/packages/workbox-build/src/lib/utils/filter-files.js
+++ b/packages/workbox-build/src/lib/utils/filter-files.js
@@ -6,7 +6,6 @@ const logHelper = require('../log-helper');
 const modifyUrlPrefixTranform = require('./modify-url-prefix-transform');
 const noRevisionForUrlsMatchingTransform =
   require('./no-revision-for-urls-matching-transform');
-const path = require('path');
 
 /**
  * A `ManifestTransform` function can be used to modify the modify the `url` or
@@ -71,7 +70,7 @@ module.exports = (fileDetails, options) => {
   // {url, revision} objects, with path.sep replaced with /.
   const normalizedManifest = filteredFileDetails.map((fileDetails) => {
     return {
-      url: fileDetails.file.replace(path.sep, '/'),
+      url: fileDetails.file.replace(/\\/g, '/'),
       revision: fileDetails.hash,
     };
   });


XhmikosR:
The AppVeyor failure seems unrelated to this change.

On a side note, I could add a test if you could give me some hints, i.e. which file the test should be, if I should just check for backslashes in `entry.url`s, etc.

gauntface:
Thanks for this @XkhmikosR.

I'm waiting on a Windows machine to test this a little more. I'm largely curious to see if I can add some tests that "fake" windows files paths so we can test on linux for Window "style" support.

XhmikosR:
@gauntface: you have AppVeyor so you can actually test Windows at any time https://www.appveyor.com/docs/how-to/rdp-to-build-worker/

It shouldn't be too hard to have a test for that. You could use https://nodejs.org/api/path.html#path_path_win32 or just hardcode the backslashes in a test case and just check if there are any in the output regardless of the OS.

packages/workbox-build/src/lib/utils/filter-files.js
```diff
@@ -71,7 +70,7 @@ module.exports = (fileDetails, options) => {
   // {url, revision} objects, with path.sep replaced with /.
   const normalizedManifest = filteredFileDetails.map((fileDetails) => {
     return {
-      url: fileDetails.file.replace(path.sep, '/'),
+      url: fileDetails.file.replace(/\\/g, '/'),
```
addyosmani: Thanks for the PR, @XhmikosR! In theory this should be  addressing the issue but AppVeyor appears to currently be timing out. We'll need to wait on that before moving ahead.

@gauntface I tried to login to restart the build but don't appear to be on the AV for Workbox. Can has privs?

addyosmani:
@gauntface Change here LGTM, but I don't have access to our AppVeyor account to follow the windows test instructions @XhmikosR linked up to. If you get a chance can you add me or test it out? Would love to see if we can unblock https://github.com/twbs/bootstrap/pull/23533

XhmikosR:
Would be nice if we had a test to ensure this doesn't happen again in the future, though :)

gauntface:
Gonna accept this merge and dig into a way of testing this better. Windows has been awful for a while and frankly I'm stumped on a sane way to test for it.

@addyosmani tried adding you to AppVeyor but no idea how the hell it's "teams" works

gauntface:
Thanks again for this @XhmikosR - this will be cut in the release tomorrow.

Lyricalz:
@gauntface thanks, can we get 2.0.1 published to npm too please

jeffposnick:
Sorry, there were a few hiccups with the `2.0.1` release, but https://www.npmjs.com/package/workbox-build (and `workbox-cli`/`workbox-webpack-plugin`) are published.

PR #776, opened 2017-08-29
V3 globals

gauntface:

My previous PR would generate the browser bundle. This PR changes rollup to dynamically mark modules and external modules and will dynamically replace the module import with a global namespace value.

This should be the last piece of infra to support the proposed build flow in: https://github.com/GoogleChrome/workbox/wiki/V3-:-ES2015-Modules-&-Browser-Bundles.

I'll look at CommonJS ASAP.

### DIFF
--- a/gulp-tasks/build-packages.js
+++ b/gulp-tasks/build-packages.js
@@ -8,6 +8,7 @@ const source = require('vinyl-source-stream');
 const sourcemaps = require('gulp-sourcemaps');
 const rename = require('gulp-rename');
 const buffer = require('vinyl-buffer');
+const glob = require('glob');
 
 const constants = require('./utils/constants');
 const packageRunnner = require('./utils/package-runner');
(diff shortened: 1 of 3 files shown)


gulp-tasks/build-packages.js
```diff
@@ -26,6 +26,22 @@ const ERROR_NO_NAMSPACE = oneLine`
...
+const testObjectValueExists = (object, nestedPath) => {
+  const pieces = nestedPath.split('.');
+  let currentRoot = object;
+  pieces.forEach((piece) => {
+    if (!currentRoot.piece) {
+      return false;
+    }
+    currentRoot = object[piece];
+  });
+  return true;
```
philipwalton: This function is always going to return `true` since the inner `return false` statement will only return `false` from the `forEach()` loop.

gulp-tasks/build-packages.js
```diff
@@ -68,14 +84,98 @@ const buildPackage = (packagePath, buildType) => {
...
+    const packageName = result[1];
+    let importFilePath = null;
+    if (result.length === 3) {
+      importFilePath = result[2];
+      if (importFilePath && importFilePath.indexOf('/') === 0) {
+        importFilePath = importFilePath.replace('/', '');
+      }
+    }
+
+    // Get a packages browser namespace so we know where it will be
```
philipwalton: s/packages/package's/

gulp-tasks/build-packages.js
```diff
@@ -68,14 +84,98 @@ const buildPackage = (packagePath, buildType) => {
...
+    } catch (err) {
+      logHelper.error(`Unable to get browserNamespace for package: ` +
+        `'${packageName}'`);
+      logHelper.error(err);
+      throw err;
+    }
+
+    let globalNamespace = `${constants.NAMESPACE_PREFIX}.${browserNamespace}`;
+    let fileNamespace = '';
+    // If the module pulls in a specific files we'll need to add this
```
philipwalton: Check the grammar. I'm not sure if you want to say "a specific file" + "it" or "specific files" + "them".

philipwalton:
Mostly LGTM after you make sure `testObjectValueExists()` will fail when it's supposed to.

gauntface:
@philipwalton should be fixed now.

PR #779, opened 2017-08-31
V3: Second Module

gauntface:
cc @addyosmani @philipwalton @jeffposnick 

Adding a second module.

I started touching more things than I suspected so wanted to PR this in sooner rather than later.

The one thing I'm still pondering is the right dance between:

- using names exports vs file imports
- the whole 'internal' folder exposure.

Will grab you on monday to discuss an idea. (would simplify the build).


packages/workbox-core/internal/models/messages/messageGenerator.mjs
```diff
@@ -12,14 +12,14 @@ const fallback = (code, ...args) => {
 const generatorFunction = (code, ...args) => {
   const message = messages[code];
   if (!message) {
-    core.INTERNAL.logHelper.warn(`Unable to find message for code '${code}'`);
+    logger.warn(`Unable to find message for code '${code}'`);
```
addyosmani: As a user, if logger was unable to find a message for a specific code, what's the next actionable step?

gauntface: It would basically be a case of raising an issue with this and hopefully we can figure out which use of the code is invalid and help from there.

We can (and should) probably print the args as well, will PR that in.

Do think it would be worth adding a message asking developers to raise an issue in this case?

gauntface:
@addyosmani there is still a question on your comment - please ignore Githubs collapsing of your message.

test/workbox-core/node/utils/test-logger.mjs
```diff
@@ -0,0 +1,303 @@
...
+      [],
+      {},
+    ], (variant) => {
+      expect(() => {
+        logger.logLevel = variant;
+      }).to.throw(WorkboxError).that.has.property('name').that.equals('invalid-type');
+    });
+  });
+
+  /**
```
philipwalton: Minor nit, but this comment block should probably be either `/* ... */` or multiple `// ...`. Using `/**` (with two stars) signifies a JSDoc block, so it's best to not use it unless writing a JSDoc block (this is also required by google JS style, though not enforced by the eslint package).

test/workbox-precaching/node/test-environment.mjs
```diff
@@ -0,0 +1,52 @@
...
+  beforeEach(function() {
+    clearRequire.all();
+  });
+
+  afterEach(function() {
+    sandbox.restore();
+  });
+
+  describe(`Used in a window`, function() {
+    it(`should throw in dev builds when no in SW`, async function() {
```
philipwalton: s/no/not/

PR #783, opened 2017-09-05
Bundling idea...

gauntface:
In the original proposal for publishing files, there was an awkward set up where file names with underscore are auto-excluded from browser builds and the browser builds would be generated from an auto-generated file exporting all files. I wasn't happy with this and generally no one liked the "internal" name.

After digging a little more into Rollup (and I'm still learning) it seems like we could still offer the same build size wins to developers with just normal named exports on the main export (i.e. what we are doing today).

We would still need to export classes / objects from one module to another even if we wanted them to be private (i.e. logger / assertions). Instead of "internal" I've just moved to an `_private.mjs` file that has multiple exports under `_private` variable.


gulp-tasks/build-packages.js
```diff
@@ -25,39 +30,63 @@ const ERROR_NO_NAMSPACE = oneLine`
...
-  let currentRoot = object;
-  for (const piece of pieces) {
-    if (!currentRoot[piece]) {
-      return false;
-    }
-    currentRoot = currentRoot[piece];
+// This makes Rollup assume workbox-* will be added to the global
+// scope and replace references with the core namespace
+const globals = (moduleId) => {
+  // This regex matches for (workbox-*)(/any/path/here.mjs)
```
jeffposnick: Is there a need to use `path.sep` here to support Windows?

Also, could you get away with using `moduleId.split(path.sep)` instead of a RegExp? I love RegExps, but for this use case, they seem more fragile.

gulp-tasks/build-packages.js
```diff
@@ -25,39 +30,63 @@ const ERROR_NO_NAMSPACE = oneLine`
...
+    `);
+  }
+
+  // Get a package's browserNamespace so we know where it will be
+  // on the global scope (i.e. workbox.????)
+  let browserNamespace = null;
+  const packagePath = path.join(__dirname, '..', 'packages', packageName);
+  try {
+    const pkg = require(path.join(packagePath, 'package.json'));
+    browserNamespace = pkg.workbox.browserNamespace;
```
jeffposnick: Nit: you could make this `const browserNamespace = ...` and then move

```js
return `${constants.NAMESPACE_PREFIX}.${browserNamespace}`;
```

to the next line, and not have to declare `let browserNamespace = null;` outside of the `try{}` block.

gulp-tasks/build-packages.js
```diff
@@ -25,39 +30,63 @@ const ERROR_NO_NAMSPACE = oneLine`
...
+  return `${constants.NAMESPACE_PREFIX}.${browserNamespace}`;
 };
 
-/*
- * To test sourcemaps are valid and working, use:
- * http://paulirish.github.io/source-map-visualization/#custom-choose
- */
+// This ensures all workbox-* modules are treated as external and are
+// referenced as globals.
+const externalAndPure = (moduleId) => {
```
jeffposnick: If you want to get all ES2015-y:

```js
const externalAndPure = (moduleId) => moduleId.startsWith('workbox-');
```

packages/workbox-core/_private.mjs
```diff
@@ -0,0 +1,5 @@
+import logger from './utils/logger.mjs';
+import WorkboxError from './models/WorkboxError.mjs';
+
+export {logger};
+export {WorkboxError};
```
jeffposnick: Out of curiosity, is this equivalent to

```js
export {logger, WorkboxError};
```

or is there a semantic difference when they're split into two `export` statements?

packages/workbox-core/index.mjs
```diff
@@ -18,7 +17,6 @@ class WorkboxCore {
   }
 }
 
-export {logger};
-export {WorkboxError};
+export {_private};
```
jeffposnick: Can this be written as `export _private;`, since `_private` is already an object with the named exports as keys? Or does it have to be wrapped in another `{}`?

gauntface: Has to be wrapped.

addyosmani:
🎉

PR #784, opened 2017-09-06
#782 add jsdoc for conflicting routes

### LINKED ISSUE #782: Setting priority for conflicting glob patterns
**Library Affected**:
`workbox-sw`

**Browser & Platform**:
all.

It is unclear what the order of conflicting routes is when caching for multiple routes.
e.g. If I have a `workbox.strategies.networkFirst('/*' // ...` and then later a `workbox.strategies.cacheFirst('/index.html // ...`, the priority appears to be set by the order in which the different strategies are called.

**Is this a feature? Can I rely on it in my build process?**

See [this gist](https://gist.github.com/sircelsius/69f49d60051c120a0181dfc8faa697d5) which applies to [googlecodelabs/workbox-lab](https://github.com/googlecodelabs/workbox-lab).

sircelsius:
Fixes #782 

Added some JSDoc

### DIFF
--- a/packages/workbox-routing/src/lib/router.js
+++ b/packages/workbox-routing/src/lib/router.js
@@ -32,6 +32,9 @@ import normalizeHandler from './normalize-handler';
  * If you'd like this to be handled automatically for you, calling
  * `addFetchListener()` will cause the `Router` to respond to `fetch` events.
  *
+ * If a request matches multiple routes, precedence will be given to the 
+ * last-registered route.
+ *
  * @memberof module:workbox-routing
  *
  * @example
@@ -41,6 +44,7 @@ import normalizeHandler from './normalize-handler';
  * // Both the routes are registered with the router, and any requests that
  * // don't match either route will be handled using the default NetworkFirst
  * // strategy.
+ * // If a request matches both routes, the "images" handler will be used. 
  * const assetRoute = new RegExpRoute({
  *   regExp: /assets/,
  *   handler: new workbox.runtimeCaching.StaleWhileRevalidate(),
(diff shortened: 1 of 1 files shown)


sircelsius:
@jeffposnick any clue why Travis is failing?

prateekbh:
thats my bad @sircelsius. there's a due PR that should fix this

sircelsius:
ah cool. For a second there I started wondering why adding comments would break tests :)

packages/workbox-routing/src/lib/router.js
```diff
@@ -32,6 +32,9 @@ import normalizeHandler from './normalize-handler';
  * If you'd like this to be handled automatically for you, calling
  * `addFetchListener()` will cause the `Router` to respond to `fetch` events.
  *
+ * If a request matches multiple routes, precedence will be given to the
```
gauntface: @jeffposnick is this intended behavior? Seems like the opposite of what I'd expect.

packages/workbox-routing/src/lib/router.js
```diff
@@ -41,6 +44,7 @@ import normalizeHandler from './normalize-handler';
  * // Both the routes are registered with the router, and any requests that
  * // don't match either route will be handled using the default NetworkFirst
  * // strategy.
+ * // If a request matches both routes, the "images" handler will be used.
```
gauntface: Could you tweak the wording slightly here to 

"If a request matches both routes, the last route will be used to handle the request, in this case, the "images" handler would take precedence."

sircelsius: Sure

packages/workbox-routing/src/lib/router.js
jeffposnick: It is the intended behavior, yes: https://github.com/GoogleChrome/workbox/blob/master/packages/workbox-routing/src/lib/router.js#L294

I can't find the relevant discussion that preceded that code, but I assume we discussed it at the time 😄 

We could change it in the next major release if folks would prefer the opposite behavior.

PR #785, opened 2017-09-06
avoiding flaky test to unblock PRs

### LINKED ISSUE #754: Flaky workbox-background-sync test in Firefox Beta 56
**Library Affected**:
*workbox-background-sync*

I see the following failure at https://travis-ci.org/GoogleChrome/workbox/jobs/265987692#L770

```sh
❌ [Failed] background sync queue > should rejected promise on replay failure
    expected undefined to equal 404
    [Stack Trace]
    AssertionError@http://localhost:3004/node_modules/chai/chai.js:9315:13
    [3]</module.exports/Assertion.prototype.assert@http://localhost:3004/node_modules/chai/chai.js:239:13
    assertEqual@http://localhost:3004/node_modules/chai/chai.js:1375:7
    methodWrapper@http://localhost:3004/node_modules/chai/chai.js:7704:18
    @http://localhost:3004/packages/workbox-background-sync/build/test/browser-bundle.js:2100:7
```

Restarting the job leads to a successful test.

prateekbh:
R: @jeffposnick 

Fixes #754 and unblocks any pending PRs.

P.S. I'll look into it if I can get more verbose test around this

### DIFF
--- a/packages/workbox-background-sync/test/browser/background-sync-queue.js
+++ b/packages/workbox-background-sync/test/browser/background-sync-queue.js
@@ -110,12 +110,11 @@ describe(`background sync queue`, function() {
     await backgroundSyncQueue.pushIntoQueue({
       request: new Request('/__test/404'),
     });
-    try {
-      await backgroundSyncQueue.replayRequests();
+    return backgroundSyncQueue.replayRequests().then(()=>{
       throw new Error('Replay should have failed because of invalid URL');
-    } catch (err) {
-      expect(err[0].status).to.equal(404);
-    }
+    }).catch(()=>{
+       // The promise gets rejected as expected
+    });
   });
 
   it(`should remove requests from queue which are post threir maxRetentionTime`, async function() {
--- a/packages/workbox-background-sync/test/browser/background-sync-queue.js
+++ b/packages/workbox-background-sync/test/browser/background-sync-queue.js
@@ -110,10 +110,11 @@ describe(`background sync queue`, function() {
     await backgroundSyncQueue.pushIntoQueue({
       request: new Request('/__test/404'),
     });
-    return backgroundSyncQueue.replayRequests().then(()=>{
+    return backgroundSyncQueue.replayRequests()
+    .then(() => {
       throw new Error('Replay should have failed because of invalid URL');
-    }).catch(()=>{
-       // The promise gets rejected as expected
+    }, () => {
+       // The promise rejected as expected
     });
   });
 


packages/workbox-background-sync/test/browser/background-sync-queue.js
```diff
@@ -114,7 +114,7 @@ describe(`background sync queue`, function() {
       await backgroundSyncQueue.replayRequests();
       throw new Error('Replay should have failed because of invalid URL');
     } catch (err) {
-      expect(err[0].status).to.equal(404);
+      expect(err).to.be.an('Array');
```
gauntface: This change still worries me - a lot.

Why would err be an array? Shouldn't it just be a single error that is thrown?

You should be checking for an expected, specific error.

To unblock could you change this to:

```
return backgroundsyncQueue.replayRequests()
    .then(() => {
        throw new Error('Expected an error to be thrown);
    }, (err) => {
        // TODO: Check for specific error.
    });
```

prateekbh: done

packages/workbox-background-sync/test/browser/background-sync-queue.js
```diff
@@ -114,7 +114,7 @@ describe(`background sync queue`, function() {
       await backgroundSyncQueue.replayRequests();
       throw new Error('Replay should have failed because of invalid URL');
     } catch (err) {
-      expect(err[0].status).to.equal(404);
+      // TODO: Check for specific error.
```
gauntface: This won't actually test anything - you throw inside the try block meaning: replayRequest() doesn't throw, you then throw 'Replay should have failed...' meaning the catch will be caught and the test will pass.

Changing to:

```
return backgroundSyncQueue.replayRequests()
.then(() => {
    // Expected test to throw an error but it didn't
    throw new Error('...');
}, (err) => {
    // The function throw an error as expected - nothing to do.
})
```

will fix this

prateekbh: fixed :)

packages/workbox-background-sync/test/browser/background-sync-queue.js
```diff
@@ -110,12 +110,11 @@ describe(`background sync queue`, function() {
...
       request: new Request('/__test/404'),
     });
-    try {
-      await backgroundSyncQueue.replayRequests();
+    return backgroundSyncQueue.replayRequests().then(()=>{
       throw new Error('Replay should have failed because of invalid URL');
-    } catch (err) {
-      expect(err[0].status).to.equal(404);
-    }
+    }).catch(()=>{
```
gauntface: This still will always pass. You need to have `.then(() => {}, () => {})` and throw in the first callback.

PR #787, opened 2017-09-07
V3 logging

gauntface:
- Adds an internal logging API to core.
    - Default log level is verbose on dev.
    - Default log level is warn on prod.
- Updated to latest rollup version to fix a bug in the rollup-stream version.
- gulp.watch is added for any manual testing you want to do in the browser (see test/workbox-core/static/logger.html)
- This PR also changes the build so that dev builds are no longer minified.

API Example for external developers using browser bundles:
```javascript
workbox.core.default.logLevel = workbox.core.LOG_LEVELS.error;
```
API Example for external developers using ES2015 modules:
```javascript
import core from 'workbox-core';
import {LOG_LEVELS} from 'workbox-core';

core.logLevel = LOG_LEVELS.error;
```

For internal workbox-libraries (which will do the logging):

```javascript
import {_private} from  'workbox-core';
_private.logger.log('Example');
_private.logger.debug('Example');
_private.logger.warn('Example');
_private.logger.error('Example');

_private.logger.groupCollapsed('Group title');
_private.logger.groupEnd();
```

**Example of the .log() -> .error() messages**
workbox-logs

**Example of the groupCollapsed()**
workbox-log-groups

### DIFF
--- a/test/workbox-core/node/controllers/test-WorkboxCore.mjs
+++ b/test/workbox-core/node/controllers/test-WorkboxCore.mjs
@@ -1,6 +1,7 @@
 import {expect} from 'chai';
 import clearRequire from 'clear-require';
 
+import expectError from '../../../../infra/utils/expectError';
 import constants from '../../../../gulp-tasks/utils/constants.js';
 import generateVariantTests from '../../../../infra/utils/generate-variant-tests';
 import WorkboxError from '../../../../packages/workbox-core/models/WorkboxError.mjs';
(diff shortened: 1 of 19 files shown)


test/workbox-core/node/controllers/test-WorkboxCore.mjs
```diff
@@ -27,4 +30,69 @@ describe(`WorkboxCore`, function() {
       expect(core.assert).to.not.exist;
     });
   });
+
+  describe(`core.logLevel (getter)`, function() {
+    it(`should initialise to 'verbose' log level for dev build`, async function() {
+      process.env.NODE_ENV = 'dev';
+
+      const coreModule = await import('../../../../packages/workbox-core/index.mjs');
```
jeffposnick: Just confirming: this taking advantage of `@std/esm`'s support for `import()`s, right? And it gets switched on at some point during the initialization of the test harness?

gauntface: You are correct. The magic trick is: `mocha --require @std/esm <Glob> <Args>` which is done behind the scenes of `gulp test`.

PR #789, opened 2017-09-07
Warn about options that are invalid for a given workbox-build mode

### LINKED ISSUE #708: Warn when injectManifest({runtimeCaching, navigateFallback}) is called
**Library Affected**:
*workbox-build*

There are a few options that could be passed in to `workbox-build.injectManifest()` (either directly, or via `workbox-cli` or `workbox-webpack-plugin`) that don't make any sense, since in that mode, we're only replacing the contents of `precache([])`, and not making any other changes to the `swSrc` file. The ones that I can think of include:

- `runtimeCaching`
- `navigateFallback`
- `navigateFallbackWhitelist`

but there might be others—as part of #695 I might come across more.

Instead of silently ignoring them, leading to developer confusion when they don't have any effect, we should throw a fatal exception and explain why they shouldn't be used in the `injectManifest()` mode.

jeffposnick:
R: @addyosmani @gauntface

Fixes #708

The decision reached in that issue was to log a warning, but not fail the build.

### DIFF
--- a/packages/workbox-build/src/lib/utils/warn-about-config.js
+++ b/packages/workbox-build/src/lib/utils/warn-about-config.js
@@ -16,7 +16,8 @@ module.exports = (blacklist, fullConfig, methodName) => {
     .filter((configKey) => blacklist.indexOf(configKey) !== -1);
 
   if (invalidConfigKeys.length > 0) {
-    logHelper.warn(`The configuration options '${invalidConfig.join(',')}' ` +
-      `will be ignored in the context of Workbox's ${methodName} mode.`);
+    logHelper.warn(`The configuration options ` +
+      `'${invalidConfigKeys.join(',')}' will be ignored in the context of ` +
+      `Workbox's ${methodName} mode.`);
   }
 };
--- a/packages/workbox-build/src/lib/utils/warn-about-config.js
+++ b/packages/workbox-build/src/lib/utils/warn-about-config.js
@@ -1,0 +1,20 @@
+'use strict';
+
+const logHelper = require('../log-helper');
+
+/**
+ * A helper to find and log any configuration options that were set which are
+ * not valid for a given usage mode.
+ *
+ * @param {Array<String>} blacklist The names of invalid configuration options.
+ * @param {Object<String,String>} fullConfig The configuration options to check.
+ */
+module.exports = (blacklist, fullConfig) => {
+  const invalidConfig = Object.keys(fullConfig)
+    .filter((configKey) => blacklist.indexOf(configKey) !== -1);
+
+  if (invalidConfig.length > 0) {
+    logHelper.warn(`These Workbox configuration options are not valid for ` +
+      `the current mode, and have been ignored: ${invalidConfig}`);
+  }
+};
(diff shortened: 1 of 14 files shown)


packages/workbox-build/src/lib/utils/warn-about-config.js
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * A helper to find and log any configuration options that were set which are
+ * not valid for a given usage mode.
+ *
+ * @param {Array<String>} blacklist The names of invalid configuration options.
+ * @param {Object<String,String>} fullConfig The configuration options to check.
+ */
+module.exports = (blacklist, fullConfig) => {
+  const invalidConfig = Object.keys(fullConfig)
+    .filter((configKey) => blacklist.indexOf(configKey) !== -1);
```
gauntface: Could you use includes?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

packages/workbox-build/src/lib/utils/warn-about-config.js
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * A helper to find and log any configuration options that were set which are
+ * not valid for a given usage mode.
+ *
+ * @param {Array<String>} blacklist The names of invalid configuration options.
+ * @param {Object<String,String>} fullConfig The configuration options to check.
+ */
+module.exports = (blacklist, fullConfig) => {
+  const invalidConfig = Object.keys(fullConfig)
```
gauntface: Nit: Rename to invalidConfigKeys (Just reads better for if (invalidConfigKeys.length > 0)

packages/workbox-build/src/lib/utils/warn-about-config.js
```diff
@@ -0,0 +1,20 @@
...
+    .filter((configKey) => blacklist.indexOf(configKey) !== -1);
+
+  if (invalidConfig.length > 0) {
+    logHelper.warn(`These Workbox configuration options are not valid for ` +
+      `the current mode, and have been ignored: ${invalidConfig}`);
```
gauntface: Nit: Swap "have been" with "will be" as I'm assuming this log will be at the start of the build task (whatever that will be).

Would also make sense to reword this slightly away from "These Workbox configuration options". It's more to highlight that in this context it won't be used, so maybe:

> The configuration options '${invalidConfig.join(',')}' will be ignored by `workbox-build.${methodName}`.

packages/workbox-build/src/lib/utils/warn-about-config.js
jeffposnick: We're still aiming for Node 4 compatibility with these modules, so no `Array.prototype.includes()`: http://node.green/#ES2016-features-Array-prototype-includes

packages/workbox-build/src/lib/utils/warn-about-config.js
jeffposnick: Done.

packages/workbox-build/src/lib/utils/warn-about-config.js
jeffposnick: Folks might end up calling the methods indirectly via the Webpack or CLI, so I don't want to put `workbox-build.` in the message. But I'll reword along those lines.

PR #791, opened 2017-09-07
V3 precaching add cache list

gauntface:
This adds `workbox-precaching.PrecacheManager.addToCacheList()`;

At the moment it's only add support for strings and objects to be added (Will add Requests in next PR - this one was just growing too big already).

- Minor tweak to logHelper to useful outside of gulp tasks (i.e. in test runner)
- Added `expectError()` this is largely to be a "nicer" solution to expect().to.throw().<Long chain of checks>(). Generally we'll always want to detect a 'WorkboxError' with a specific name and values. `expectError` checks the constructor name, the error name and provides a callback function in case you want to do any additional checks
- Tidied up the assertion slightly to throw is the assertion is not true (before I was being silly and returning a boolean and throwing an error - cluttering the code).


packages/workbox-core/models/messages/messageGenerator.mjs
```diff
@@ -12,17 +12,14 @@ const fallback = (code, ...args) => {
 const generatorFunction = (code, ...args) => {
   const message = messages[code];
   if (!message) {
-    logger.warn(`Unable to find message for code '${code}'.`, args);
-    return fallback(code, ...args);
+    throw new Error(`Unable to find message for code '${code}'.`);
```
philipwalton: Did you intend to remove the logging here?

packages/workbox-core/models/messages/messages.mjs
```diff
@@ -23,4 +23,33 @@ export default {
...
+    if (!moduleName || !className || !funcName || !paramName) {
+      throw new Error(`Unexpected input to 'not-an-array' error.`);
+    }
+    return `The parameter '${paramName}' passed into ` +
+      `'${moduleName}.${className}.${funcName}()' must be an array.`;
+  },
+
+  'add-to-cache-list-unexpected-type': ({entry}) => {
+    return `An unexpected entry was passed to ` +
+    `'workbox-precaching.PrecacheManager.addToCacheList()'. The entry ` +
```
philipwalton: Is there period supposed to be there at the end of ".addToCacheList()'."? Same question for line 50.

packages/workbox-core/utils/assert.mjs
```diff
@@ -1,11 +1,27 @@
+import WorkboxError from '../models/WorkboxError.mjs';
+
 /*
  * This method returns true if the current context is a service worker.
  */
-const isSWEnv = () => {
-  return (typeof ServiceWorkerGlobalScope !== 'undefined' &&
-    self instanceof ServiceWorkerGlobalScope);
+const isSWEnv = (moduleName) => {
```
philipwalton: Minor nit I noticed today, Google JS style says this should be named `isSwEnv`. I think we do this in other places as well, and in general I think we should be consistent with our style guide.

packages/workbox-core/utils/assert.mjs
```diff
@@ -1,11 +1,27 @@
...
+  if (!(typeof ServiceWorkerGlobalScope !== 'undefined' &&
+    self instanceof ServiceWorkerGlobalScope)) {
+      throw new WorkboxError('not-in-sw', {moduleName});
+  }
+};
+
+const isArray = (value, {moduleName, className, funcName, paramName}) => {
```
philipwalton: You're destructuring `{moduleName, className, funcName, paramName}` only to restructure it again later. Why not just just an object?

packages/workbox-precaching/controllers/PrecacheManager.mjs
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * Performs efficient precaching of assets.
+ */
+export default class PrecacheManager {
+  /**
+   * Create a new PrecacheManager Instance
+   */
+  constructor() {
+    this._entriesToCacheMap = new Map();
+    if (process.env.NODE_ENV !== 'prod') {
```
philipwalton: Another thing I noticed today when looking at the v3 branch is we're using `'prod'` instead of `'production'`, which I think is more common in the npm community. We should confirm this and stick with whatever is most common.

packages/workbox-precaching/controllers/PrecacheManager.mjs
```diff
@@ -0,0 +1,99 @@
...
+      this.checkEntryRevisioning = true;
+    }
+  }
+
+  /**
+   * This method will add items to the precache list, removing duplicates
+   * and ensuring the information is valid.
+   * @param {Array<object|string>} userEntries Array of entries to
```
philipwalton: s/object/Object/

packages/workbox-precaching/utils/showWarningsIfNeeded.mjs
```diff
@@ -0,0 +1,41 @@
...
+    // No warnings needed.
+    return;
+  }
+
+  let pluralString = `${urlOnlyEntries.length} precache entries`;
+  if (urlOnlyEntries.length === 1) {
+    pluralString = `1 precache entry`;
+  }
+
+  _private.logger.groupCollapsed('Are you precached assets revisioned?');
```
philipwalton: s/you/your/

packages/workbox-precaching/utils/showWarningsIfNeeded.mjs
```diff
@@ -0,0 +1,41 @@
...
+    `workbox-precaching.PrecacheManager.addToCacheList() ` +
+    `found ${pluralString} that consist of just a URL meaning we can't ` +
+    `confirm it's revisioned.`);
+  _private.logger.debug(`For example, these entries are just URLs:`);
+  _private.logger.debug(`    /styles/example.css`);
+  _private.logger.debug(`    { url: '/index.html' }`);
+  _private.logger.debug(`These entries are URLs with revision information:`);
+  _private.logger.debug(`    /styles/example.1234.css`);
+  _private.logger.debug(`    { url: '/index.html', revision: '123' }`);
+  _private.logger.debug(`The following URL's should be checked: `);
```
philipwalton: s/URL's/URLs/

packages/workbox-core/models/messages/messageGenerator.mjs
gauntface: yes. This now throws to ensure we aren't silently failing to show a message when we intended to.

packages/workbox-core/utils/assert.mjs
gauntface: Only reason for this is that if you were looking to use this at some point, this shows what you are expected to supply to call this method, compared to:

```javascript
isArray(value, details) => {
 ...
}
```

Shall I add a comment just to explain this?

gauntface:
Unrevisioned message starts off collapsed and results in the following when expanded:

precaching-expanded

packages/workbox-core/utils/assert.mjs
philipwalton: I don't think it's a huge deal either way. For a public method, I'd expect to see a jsdoc comment for the param explaining the available options. If you're going to comment I'd use a jsdoc comment, otherwise I think it's fine as is.

test/workbox-precaching/node/controllers/test-PrecacheManager.mjs
```diff
@@ -0,0 +1,304 @@
+import {expect} from 'chai';
+import sinon from 'sinon';
+import clearRequire from 'clear-require';
+
+import expectError from '../../../../infra/utils/expectError';
+import generateTestVariants from '../../../../infra/utils/generate-variant-tests'
+
+// This sets Request mock on the Global scope
+import '../../../mocks/Request.mjs';
```
jeffposnick: How do you feel about adopting https://github.com/pinterest/service-workers/tree/master/packages/service-worker-mock for our Node tests? I'm assuming it works as expected, but if we find any bugs/issues, we could upstream them and the community would benefit.

test/workbox-precaching/static/addToCacheList.html
```diff
@@ -0,0 +1,24 @@
+<html>
```
jeffposnick: I think I've see a few other of these `static/*.html` files checked in now. How are they used?

test/workbox-precaching/node/controllers/test-PrecacheManager.mjs
gauntface: I didn't realise it had all the mocks - will investigate switching over.

test/workbox-precaching/static/addToCacheList.html
gauntface: At the moment they are just manual tests.

If you run gulp build and start a static server, you can see what the end developer would see (It's where the screenshots in this and the log PR came from). Happy to remove. I am mindful that they could break without anyone knowing.

packages/workbox-core/utils/assert.mjs
gauntface: This will be all internal (assertions are stripped completely from production builds)

Issue #757, opened 2017-08-19
Question: Converting sw-toolbox to workbox piece of code


ayumihamsaki:
Hello,

Before I write my question, I am sure there are a couple of coding errors of this page: https://workboxjs.org/examples/workbox-sw/

```javascript
workboxSW.precache([{
  url: 'precached.txt',
  revision: '43011922c2aef5ed5ee3731b11d3c2cb', <- Why is there a comma here?
}]);

Corrected Version:

workboxSW.precache([{
  url: 'precached.txt',
  revision: '43011922c2aef5ed5ee3731b11d3c2cb'
}]);
```

Also this bit of code:

```javascript
workboxSW.router.registerRoute(
  'https://httpbin.org/image/(.*)',
  workboxSW.strategies.cacheFirst({
    cacheName: 'images',
    cacheExpiration: {
      maxEntries: 2,
      maxAgeSeconds: 7 * 24 * 60 * 60, <- Why is there a comma here?
    },
    cacheableResponse: {statuses: [0, 200]}, <- Why is there a comma here?
  })
);

Corrected Version:

workboxSW.router.registerRoute(
  'https://httpbin.org/image/(.*)',
  workboxSW.strategies.cacheFirst({
    cacheName: 'images',
    cacheExpiration: {
      maxEntries: 2,
      maxAgeSeconds: 7 * 24 * 60 * 60
    },
    cacheableResponse: {statuses: [0, 200]}
  })
);
```

Ok on to my question, I am trying to convert my service worker from sw-toolbox to workbox, my code I am stuck on converting is here:

```javascript
global.toolbox.router.get('/(.*)', function(request, values, options) { 
  return global.toolbox.fastest(request, values, options).catch(function(error) {
    if (request.method === 'GET' && request.headers.get('accept').includes('text/html')) {
      return toolbox.cacheOnly(new Request('offline'), values, options);
    }
    throw error;
  });
}, {networkTimeoutSeconds: 4});
```

Looking at Workbox Caching list found here: https://workboxjs.org/reference-docs/latest/module-workbox-runtime-caching.html

You don't seem to have a **toolbox.fastest** is there one in toolbox that does the same thing?

toolbox.fastest = Request the resource from both the cache and the network in parallel.

I'm guessing toolbox.fastest has been replaced with workboxSW.strategies.staleWhileRevalidate.

I think my biggest problem is the method GET, been reading other issues and seems not straight forward with workbox.

[...]

PR #798, opened 2017-09-13
v3: workbox-routing.Route

jeffposnick:

I'm dipping my toes into v3 with an initial PR for just the `Route` class of `workbox-routing`. I'll follow-up with more after going through this process.


package.json
```diff
@@ -40,40 +40,41 @@
   "devDependencies": {
```
jeffposnick: I can ditch the `devDependencies` updates from this PR if you'd prefer, but I had done updated things locally while developing, and none of this updates seems to have broken anything.

package.json
```diff
@@ -40,40 +40,41 @@
...
-    "rollup-plugin-replace": "^1.1.1",
+    "rollup-plugin-replace": "^2.0.0",
     "rollup-plugin-uglify-es": "0.0.1",
-    "rollup-stream": "^1.23.1",
+    "rollup-stream": "^1.24.1",
     "semver": "^5.4.1",
     "service-worker-mock": "^1.2.2",
-    "sinon": "^2.4.1",
+    "sinon": "^3.2.1",
+    "text-encoding": "^0.6.4",
```
gauntface: I'm fine all plugins, where did this come from? Can't see it used anywhere.

packages/workbox-core/models/messages/messages.mjs
```diff
@@ -32,6 +32,22 @@ export default {
       `'${moduleName}.${className}.${funcName}()' must be an array.`;
   },
 
+  'not-of-type': ({paramName, expectedType, moduleName, funcName}) => {
```
gauntface: In the above code sample I included the ClassName. I know it's verbose, but would likely help diagnose issues (both for us and developers).

What do you think?

packages/workbox-routing/lib/Route.mjs
```diff
@@ -0,0 +1,170 @@
...
+ * [Route Constructor]{@link module:workbox-routing.Route}).
+ *
+ * @callback Route~matchCallback
+ * @param {Object} input
+ * @param {URL} input.url The request's URL.
+ * @param {FetchEvent} input.event The event that triggered the `fetch` handler.
+ * @return {Object|null} To signify a match, return a truthy value, otherwise
+ * return null if the route shouldn't match. If you return an Object with
+ * contents it will be passed to the `handler` in the `Route` constructor.
+ * @memberof module:workbox-routing
```
gauntface: Can you move this out into a seperate file?

Main reason is that it's really difficult to parse the code.

packages/workbox-core/models/messages/messages.mjs
```diff
@@ -32,6 +32,22 @@ export default {
...
+    }
+    return `The parameter '${paramName}' passed into ` +
+      `'${moduleName}.${funcName}()' must be of type ${expectedType}.`;
+  },
+
+  'not-a-method': ({paramName, expectedMethod, moduleName, funcName}) => {
+    if (!paramName || !expectedMethod || !moduleName || !funcName) {
+      throw new Error(`Unexpected input to 'not-a-method' error.`);
+    }
+    return `${moduleName}.${funcName}() was called with a ${paramName} that ` +
```
gauntface: For:

```
core.assert.hasMethod(handler, 'handle', {
    moduleName: 'workbox-routing',
    funcName: 'Route',
    paramName: 'handler',
});
```

Would result in:

```
workbox-routing.Route() was called with a handler that doesn't expose a handle method.
```

Could we tweak to:

```
workbox-routing.Router() expected the 'handler' parameter to expose a 'handle' method.
```

packages/workbox-core/models/messages/messages.mjs
gauntface: Could have function name as 'constructor' or special case special case 'constructor' to nothing.

packages/workbox-core/utils/assert.mjs
```diff
@@ -27,7 +27,37 @@ const isArray = (value, {moduleName, className, funcName, paramName}) => {
   }
 };
 
+const hasMethod = (object, expectedMethod,
+                   {moduleName, funcName, paramName}) => {
+  const type = typeof object[expectedMethod];
+  if (type !== 'function') {
+    throw new WorkboxError('not-a-method', {paramName, expectedMethod,
```
gauntface: Nit: could rename the error key to 'missing-a-method'.

packages/workbox-core/utils/assert.mjs
```diff
@@ -27,7 +27,37 @@ const isArray = (value, {moduleName, className, funcName, paramName}) => {
...
+      moduleName, funcName});
+  }
+};
+
+const isType = (object, expectedType,
+                {moduleName, funcName, paramName}) => {
+  if (typeof object !== expectedType) {
+    throw new WorkboxError('not-of-type', {paramName, expectedType,
```
gauntface: Nit: rename error key to `incorrect-type`

packages/workbox-routing/lib/Route.mjs
```diff
@@ -0,0 +1,170 @@
...
+
+import {defaultMethod, validMethods} from './constants.mjs';
+
+/**
+ * This is the definition of the `match` callback passed into the
+ * `Route` constructor.
+ *
+ * This callback is used to determine if a new `fetch` event can be served
+ * by this `Route`. Returning a truthy value indicates that this `Route` can
+ * handle this `fetch` event. Return `null` if this shouldn't match against
```
gauntface: This may be an implementation detail or a comment detail but presumably a "non-truthy" value will indicate the `Route` can't handle the fetch event.

If so would 'Return `false`' be better than 'Return `null`' just for simplicity?

packages/workbox-routing/lib/Route.mjs
```diff
@@ -0,0 +1,170 @@
...
+ * If you do return a truthy value, the object will be passed to the
+ * Route's `handler` (see the
+ * [Route Constructor]{@link module:workbox-routing.Route}).
+ *
+ * @callback Route~matchCallback
+ * @param {Object} input
+ * @param {URL} input.url The request's URL.
+ * @param {FetchEvent} input.event The event that triggered the `fetch` handler.
+ * @return {Object|null} To signify a match, return a truthy value, otherwise
+ * return null if the route shouldn't match. If you return an Object with
```
gauntface: Nit: Could you change "If you return an Object with contents it will be passed to the `handler` in the `Route` constructor." to "If you return an Object with contents, it will be passed to the Route's `handler` (see the [Route Constructor]{@link module:workbox-routing.Route}).

packages/workbox-routing/lib/Route.mjs
```diff
@@ -0,0 +1,170 @@
...
+
+/**
+ * A `Route` allows you to tell a service worker that it should handle
+ * certain network requests using a specific response strategy.
+ *
+ * A consists or a matcher and a handler. A matcher needs to determine if a
+ * route should be used for a request. A handler should handle the request
+ * if it does match a Router.
+ *
+ * Instead of implementing your own handlers, you can use one of the
```
gauntface: We could drop this and just demonstrate it in samples / generated code.

packages/workbox-routing/lib/Route.mjs
```diff
@@ -0,0 +1,170 @@
...
+ * pre-defined runtime caching strategies from the
+ * {@link module:workbox-runtime-caching|workbox-runtime-caching} module.
+ *
+ * There are also pre-defined Route's provided by this library:
```
gauntface: Same as previous comment.

packages/workbox-routing/lib/Route.mjs
```diff
@@ -0,0 +1,170 @@
...
+ * {@link module:workbox-routing.RegExpRoute|RegExpRoute}
+ * and {@link module:workbox-routing.ExpressRoute|ExpressRoute} subclasses
+ * which provide a convenient wrapper with a nicer interface for using regular
+ * expressions or Express-style routes as the `match` criteria.
+ *
+ * @example
```
gauntface: Given we are moving to DevSite. I'm wondering if it still makes sense to include code samples in JSDocs vs just moving to actual written docs for code samples and keep this just for reference.

packages/workbox-routing/lib/Route.mjs
```diff
@@ -0,0 +1,170 @@
...
+   * handle the `fetch` event.
+   *
+   * See [handlerCallback]{@link module:workbox-routing.Route~handlerCallback}
+   * for full details on using a callback function as the `handler`.
+   * @param {string} [input.method] Only match requests that use this
+   * HTTP method.
+   *
+   * Defaults to `'GET'`.
+   */
+  constructor({match, handler, method} = {}) {
```
gauntface: Can we change this to `constructor(match, handler, method)` or `constructor(match, handler, options)` where options would have `options.method`.

I'm still not crazy about the dangling method here. Would have liked either the match to filter on it OR registerRoute to have it.

packages/workbox-routing/lib/Route.mjs
```diff
@@ -0,0 +1,170 @@
...
+    this.handler = Route._normalizeHandler(handler);
```
gauntface: I know I pushed for the handler.handle() API but I'm regretting it now after the final code for: https://github.com/GoogleChrome/workbox/issues/757#issuecomment-326672407

Should we just require a function and simply this? Would need the strategies to change but I don't think it would be too much of a change.

packages/workbox-routing/lib/Route.mjs
```diff
@@ -0,0 +1,170 @@
...
+      core.assert.isType(match, 'function', {
+        moduleName: 'workbox-routing',
+        funcName: 'Route',
+        paramName: 'match',
+      });
+    }
+    this.match = match;
+
+    if (method) {
+      if (process.env.NODE_ENV !== 'production') {
```
gauntface: Nit: Could you move this in with the previous if statement? I know it'll require a second if on the method, but would at least group the assertions and make it clear what code is stripped.

packages/workbox-routing/lib/Route.mjs
```diff
@@ -0,0 +1,170 @@
...
+    this.handler = Route._normalizeHandler(handler);
+
+    if (process.env.NODE_ENV !== 'production') {
+      core.assert.isType(match, 'function', {
+        moduleName: 'workbox-routing',
+        funcName: 'Route',
+        paramName: 'match',
+      });
+    }
+    this.match = match;
```
gauntface: Are match, handler and method part of the public api? (i.e. do you expect developers to go route.match(), route.handler(), route.method?) If not then renaming these to _match, _handler, _method will allow uglify to minify it.

Ignore if they are public.

packages/workbox-routing/lib/Route.mjs
```diff
@@ -0,0 +1,170 @@
...
+    }
+  }
+
+  /**
+   * @param {function|Object} handler Either a function, or an object with a
+   * 'handle' method.
+   * @return {Object} An object with a handle method.
+   * @private
+   */
+  static _normalizeHandler(handler) {
```
gauntface: Nit: could you group the NODE_ENV + assertion checks together. It'll seperate the assertion from the logic of this method.

test/workbox-routing/node/route.mjs
```diff
@@ -0,0 +1,129 @@
+import {expect} from 'chai';
+
+import expectError from '../../../infra/utils/expectError.js';
+
+(async () => {
```
gauntface: why are you wrapping everything in an iife like this?

test/workbox-routing/node/route.mjs
```diff
@@ -0,0 +1,129 @@
...
+  const Route = (await import('../../../packages/workbox-routing/lib/Route.mjs')).default;
```
gauntface: This looks like it can be a top level import.

test/workbox-routing/node/route.mjs
```diff
@@ -0,0 +1,129 @@
...
+
+      it(`should not throw when called with a valid function handler and match parameters`, function() {
+        expect(() => new Route({handler: functionHandler, match})).not.to.throw();
+      });
+
+      it(`should throw when called with an invalid method`, async function() {
+        await expectError(
+          () => new Route({handler, match, method: invalidMethod}),
+          'invalid-value',
+          (error) => expect(error.details).to.have.property('paramName').that.equals('method')
```
gauntface: should this include module, class and func name?

package.json
jeffposnick: Good question. Looks like it's a [dependency](https://github.com/sinonjs/sinon/blob/master/package.json#L45) of `sinon` and my local `node_modules` must have been in a weird state, because it wasn't getting installed until I manually listed it. I'll remove it.

packages/workbox-core/models/messages/messages.mjs
jeffposnick: Done.

packages/workbox-routing/lib/Route.mjs
jeffposnick: Done.

packages/workbox-core/models/messages/messages.mjs
jeffposnick: Done.

packages/workbox-core/utils/assert.mjs
jeffposnick: Done.

packages/workbox-core/utils/assert.mjs
jeffposnick: Done.

packages/workbox-routing/lib/Route.mjs
jeffposnick: I think the idea is that someone might have a `match` function that wants to forward the parameter `false` to the handler. Pretty edge-casey, but I think most of the forwarding parameters stuff is not particularly used.

I'd rather just keep things as-is and match the v2 behavior, though.

packages/workbox-routing/lib/Route.mjs
jeffposnick: I'm going to go with `{@link module:workbox-routing.Route~handlerCallback}` instead, since that's what gets passed the parameter.

packages/workbox-routing/lib/Route.mjs
```diff
@@ -0,0 +1,170 @@
...
+ * @param {Object} [input.params] Parameters returned
+ * the Route's [match callback]{@link
+  *   module:workbox-routing.Route~matchCallback} function. This will be
+ * undefined if nothing was returned.
+ * @return {Promise<Response>} The response that will fulfill the request.
+ * @memberof module:workbox-routing
+ */
+
+/**
+ * A `Route` allows you to tell a service worker that it should handle
```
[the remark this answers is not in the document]
jeffposnick: I'll drop it.

packages/workbox-routing/lib/Route.mjs
```diff
@@ -0,0 +1,170 @@
...
+ * certain network requests using a specific response strategy.
+ *
+ * A consists or a matcher and a handler. A matcher needs to determine if a
```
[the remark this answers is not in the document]
jeffposnick: Added.

packages/workbox-routing/lib/Route.mjs
jeffposnick: I can drop the code samples. It is one more thing to keep up to date and that might drift as the API changes.

packages/workbox-routing/lib/Route.mjs
jeffposnick: The strategies all have state information relevant to their behavior and it might be messy to bind that all into a function. If we do make that change, I'd like to do it outside of the context of this PR and make sure that the corresponding change to the strategies accounts for that.

packages/workbox-routing/lib/Route.mjs
jeffposnick: I don't expect them to be used publicly. Thanks for the heads-up about minification.

test/workbox-routing/node/route.mjs
jeffposnick: I'll see if the changes we discussed on Slack will allow the dynamic `async import()` to work.

packages/workbox-routing/lib/Route.mjs
jeffposnick: Okay, I've rearranged things to group together the assertions.

packages/workbox-routing/lib/Route.mjs
jeffposnick: I don't think it makes sense for this set of `NODE_ENV` checks, as I'll end up repeating all of the logic in `if()` statements twice, and then we risk things getting out of sync. I'd rather keep it as-is.

packages/workbox-routing/lib/Route.mjs
jeffposnick: I've moved to positional parameters, with method last, as per the style guide decisions. We can revisit alternative approaches to specifying `method` in a later PR that includes the relevant `Router` code.

test/workbox-routing/node/route.mjs
jeffposnick: I am still having trouble without the `iife`, but things work with the top-level `async`, which is kind of standard for using `await`. I'm inclined to leave as-is and investigate further outside of this PR when I work on my next set of tests.

jeffposnick:
PTAL.

packages/workbox-routing/lib/Route.mjs
gauntface: But returning false isn't a truthy value, so which of the statements is correct, return a truthy value or return anything other than null?

gauntface:
@jeffposnick if there is any change you could dig into the code coverage to see any easy wins to make sure everything is covered off I'd really appreciate it. Run tests and then look at `coverage/lcov-report/index.html` has been super helpful for me.

packages/workbox-routing/lib/Route.mjs
jeffposnick: Oh, gotcha.

It should be "return anything other than `null`". I'll adjust the JSDoc.

jeffposnick:
Re: the code coverage, from what I could tell, it's due to not running all the tests when `NODE_ENV === 'production'`:

image

What's the best practice there? It seems like if our goal is to run all code paths under both `NODE_ENV` settings, then we should end up going back to the model we used in `v2`, where the tests themselves didn't set `NODE_ENV`, and it was set by the harness that kicked off the tests. (That also means that the test suite takes twice as long to complete.)

jeffposnick:
I didn't realize that I needed to add `workbox-routing` to `.travis.yml`, so the initial cuts of this PR wasn't running `gulp test --package='workbox-routing'`. Now that I've added it, I've found some test failures that only crop up when running the tests scoped to the `workbox-routing` package, and weren't cropping up when I ran all the tests locally. Stay tuned for a commit to fix that.

jeffposnick:
The code coverage is higher now.

I'm still not happy with the tests, but rather than delay this PR, I opened #814 to track some of my observations, and the tests will presumably evolve as we address those pain points.

PR #804, opened 2017-09-15
V3 caching

### LINKED ISSUE #800: V3 Cache Naming
Generally the cache naming situation in Workbox is a little confusing, so hopefully this can be tidied up a little in v3.

Current thinking is that defaults should be sane to begin with an easily altered.

```javascript
import core from 'workbox-core';

// Getter that returns {precache: 'precaching', runtime: 'runtime', prefix: 'workbox'}
console.log(core.cacheNames);
```

In the above example, the cache names would result in:
- **Precaching Cache**: workbox-precaching
- **Runtime Cache**: 'workbox-runtime

If the user wanted to customise these defaults, they could do something like:

```javascript
core.cacheNames = {precache: 'install', runtime: '', prefix: 'my-app'};
```

- **Precaching Cache**: my-app-install
- **Runtime Cache**: my-app

They could then use a custom cache name like:

```javascript
workboxSW.strategies.cacheFirst({
    cacheName: 'images'
});
```

Which would then use `my-app-images` for a change name.

The major risk with this is if someone alters this at some point after the caches have started being used. But for that scenario we should be able to detect when the service worker has installed and throw in the setter.

In terms of implementation, me and @jeffposnick discussed how we should have a single function that would handle getting the correct cachename, since users will be able to define custom cache names. We could do something along the lines of:

```javascript
core._private.cacheNameProvider.getPrecacheName(userCacheName);
core._private.cacheNameProvider.getRuntimeName(userCacheName);
```

In the above, if a user defined nothing (i.e. `undefined`) the above would grab the current prefix and precaching  or runtime name and return those joined with '-'. 

If a userCacheName was defined, it would return the current prefix joined with the cache name '-'.

@philipwalton @jeffposnick @addyosmani

gauntface:

Fixes #800

Implement cache names as discussed. I think it should make this story easy to follow for developers and offer all the flexibility we would need / want to offer.

One thing I enforced that we didn't discuss was that the `precache` and `runtime` cache must be at least one character long. The `prefix` and `suffix` can be empty if the developer wants.

There are some additional tests / changes to logic just to improve code coverage a little.

### DIFF
--- a/packages/workbox-core/index.mjs
+++ b/packages/workbox-core/index.mjs
@@ -22,6 +22,19 @@ class WorkboxCore {
   }
 
   /**
+   * This method returns the cache names used by Workbox. `cacheNames.precache`
+   * is used for the precached assets and `cacheNames.runtime` for everything
+   * else.
+   * @return {Object} An object with `precache` and `runtime` properies.
+   */
+  get cacheNames() {
+    return {
+      precache: _private.cacheNameProvider.getPrecacheName(),
+      runtime: _private.cacheNameProvider.getRuntimeName(),
+    };
+  }
+
+  /**
    * A setter for the logLevel allowing the developer to define
    * which messages should be printed to the console.
    * @param {number} newLevel the new logLevel to use.
--- a/packages/workbox-core/index.mjs
+++ b/packages/workbox-core/index.mjs
@@ -45,14 +45,14 @@ class WorkboxCore {
    * @param {Object} details
    */
   setCacheNameDetails(details) {
-    if (process.env.NODE_ENV === 'production') {
+    if (process.env.NODE_ENV !== 'production') {
       Object.keys(details).forEach((key) => {
         // TODO: Convert to Assertion
         if (typeof details[key] !== 'string') {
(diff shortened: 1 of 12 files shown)


packages/workbox-core/index.mjs
```diff
@@ -22,6 +22,60 @@ class WorkboxCore {
...
+   *
+   * Pass in an object containing any of the following properties,
+   * `prefix`, `suffix`, `precache` and `runtime`.
+   *
+   * Cache names are generated as `<prefix>-<precache or runtime>-<suffix>`.
+   * @param {Object} details
+   */
+  setCacheNameDetails(details) {
+    if (process.env.NODE_ENV !== 'production') {
+      Object.keys(details).forEach((key) => {
```
philipwalton: Style nit, should we be favoring `for (const {key, value} of Object.entires(...)` here?

gauntface: That throws for me with `Object.entries is not a function` on node version 8.2.1 Which means it'll probably have problems in the browser.

Issue #725, opened 2017-08-08
V3 Tracking Bug


gauntface:
This is a breakdown of the work in V3. This can and should be altered as work is realised if it's regarded as a core piece of work for V3.

I've done my best to break this down into sections of related work.

## Logging and Error Messages
Size has been a problem and one area where we could be smarter is logging messages to the console as well as logging text in error messages.

There have been a few suggestions but would be good someone to research and firm up a proposal with reasons for and against techniques:

**Research Tasks**
- [x] Errors use codes on prod and messages on dev. Logs are largely stripped.

**Actionable Tasks**
- [x] Create WorkboxError Class in workbox-core that extends Error and ensure it reads clearly as a WorkboxError in the console
- [x] Move LogHelper over to workbox-core and add tests for groups etc (I suspect the API could be tidied up)
- [x] Add a way for developers to set a log level and figure out how this should apply to a developer filter.
- [x] Write up article on how to filter logs by log level and custom filter plus any extra details related to the research above.

## Building / Releasing / Deploying [@gauntface]
Historically things have been a bit shaky in terms of how we build, package and release everything and there is some vital work we can do here and some *optional* work.

**Actionable / Required Tasks**
- [x] Figure out a good structure for release code on NPM.
    - [x] What is the name and location of the browser bundle file? `build/browser` and `build/module`.
    - [x] What is the location of the default ES2015 module file? It's just index.mjs. This is moved for the final build so versioning can be done as part of the release process.
    - [x] What is the location of additional classes, does it feel natural to use with ES2015 imports? Everything is exported at the top level.
    - [x] Should we minify the NPM module code (i.e. should we strip out the comments etc)? No it's staying as is - expect developers to do this themselves.
- [x] Add lerna back in to V3
    - [x] Should be able to use lerna bootstrap without concurrency flag being needed (We had a bad dependency chain before).
    - [x] Make use of prepublish to ensure any build steps are performed during lerna boostrap (We shouldn't be doing any building in the gulp file to make lerna work)
- [x] How do we upload the files from the builds to CDN?
    - [x] What should the URL structure be for the CDN releases (prod vs dev, modules, versions)?
    - [x] How should the files be published to the CDN?
- [x] How do publish the modules as Zips on Github?
- [x] Can we offer the zips as part of the workbox-cli to download them?
[...]

Issue #803, opened 2017-09-15
V3 Workbox-build


gauntface:
## workbox-build

- [x] Switch to CDN by default
- [x] Include option to not use CDN
- [x] Add function to copy workbox-build files over
- [x] Consider adding a customised Inject Manifest
- [ ] Consider user journeys and how they can  / should use Workbox-build
    - [ ] "I would like to manipulate the generate precache entries before precaching is done"
    - [ ] "I would like to create two sets of precache entries (one for es2015 modules, one for non-es2015 modules)"
    - [ ] "I would like to generate a service worker with just precaching"
    - [ ] "I would like to generate a service worker with runtime routes"
    - [ ] "I would like to create a manifest of entries"
    - [ ] "I would like to get an Array of entries"
- [ ] Write up docs on workbox-build user journeys.
- [x] Return the files and sizes included in the precache list
- [ ] Check APIs for WebPack works
- [x] Re-organize the codebase. Clear what the public interface is vs internal as helpers.
- [x] Create a base class for our shared options that are valid for all cases, and then subclasses for specific use cases where additional options can be used.
- [x] Move option validation (via joi, most likely).
- [x] Move warnings about unused/unsupported options to the class constructors, instead of spreading out validation across the codebase.

jeffposnick:
Just an update: I'm going to start by getting `workbox-build` moved over with more or less the same functionality, to unblock #808. The main work there seems to be to get the tests happy, as this is the first Node module that's being tested in the new infrastructure.

Substantive changes outlined above will follow once that is done.

jeffposnick:
Some additional goals I'm tracking for v3:

- Re-organize the codebase so that it's clear which modules are "entry points" as part of our public interface, and which are used internally as helpers.
- Create a base `class` for our shared options that are valid for all cases, and then subclasses for specific use cases where additional options can be used. Move option validation (via [`joi`](https://www.npmjs.com/package/joi), most likely) and warnings about unused/unsupported options to the `class` constructors, instead of spreading out validation across the codebase.

jeffposnick:
Here's the current state of things, with #833 merged in the `v3` branch, with special emphasis on changes for @goldhand's use case:

[...]

PR #808, opened 2017-09-17
Workbox Webpack Plugin v3

goldhand:
## tl;dr

This is a prototype for the `workbox-webpack-plugin` for the v3 release. This address some of the points mentioned in #696 . Not all of the configuration from the existing `workbox-webpack-plugin` has been migrated (or configuration from `sw-precache-webpack-plugin`) but that was not the focus of this prototype. The goal of this prototype is to solve for some of the difficult webpack-specific features, including:
* Supporting `webpack-dev-middleware` by generating the `service-worker`, `file-manifest`, and built version of `workbox-sw` **in-memory** (rather than writing to disk). I've tested this and confirmed support for `webpack-dev-middleware` / `webpack-dev-server`.
* Filtering files by chunk names.
* Copying`workbox-sw` to the webpack assets.

⚠️ *I haven't updated the tests yet, I was manually testing by using an ejected create-react-app instance that imports the plugin using a relative file path.*
⚠️  *One issue (mentioned in the [blocks](#blocks) section below, was that I wasn't able to import the `workbox-sw` package into the `workbox-webpack-plugin` package. See details in the blocks section for my work-around.*

# V3 Changes to workbox-webpack-plugin

Following are details of the design, proposed api changes, and issues I am having that I am hoping someone will provide help with.

## Requirements
*from the #725 v3 tracker*
* [ ] Switch to CDN by default
* [ ] Include option to not use CDN and copy files over
* [ ] Update API to implement v3 proposal
* [x] Use compilation assets to generate a manifest (instead of globPatterns)
* [x] Allow user to specify chunks to be used when creating the manifest
* [x] Generate service worker “in-memory” and make webpack responsible for writing to file system
* [ ] Support sw-precache config in workbox-webpack plugin (runtime-caching? drop staticFileIgnores in favor of “chunks” api)

Im not clear what the *CDN* related tasks are.

## Re-architecture
The objective is to provide a seamless experience for developers who want to use workbox in projects that rely on webpack for building.
Webpack contains information about the build files and versions of those files that workbox can leverage to manage service worker caching.

### Tenets
* Try to use as many existing workbox utilities to generate the service worker through webpack without compromising the developer experience.
* Ensure all required service worker assets are emitted through webpack so workbox can be used with `webpack-dev-middleware`.

### Design
* In most webpack build systems, webpack is aware of all the built static assets as well as if there are changes in those assets. Were going to use webpack to generate a file manifest that `workbox-sw` can use for precaching instead of using `workbox-build/generate-file-manifest`.
* Filter manifest files to include / exclude files by chunk name. This is provided by the `config.chunks` and `config.excludeChunks`, respectively (mirroring the `html-webpack-plugin` api).
* If we need to use the `compiler.inputFileSystem._readFile` method instead of `fs.readFile`, use the `workbox-webpack-plugin/utils/read-file.js` method that is configured to use `compiler.inputFileSystem._readFile` during the **run** phase of the webpack compilation lifecycle.
* `workbox-build` will copy a built `workbox-sw` into the service worker output path. This needs to done through the webpack. During the **make** phase of the webpack compilation, we use `workbox-webpack-plugin/utils/copy-workbox-sw` to add a built version of `workbox-sw` to the webpack compilation assets array. We can't use the `workbox-build/utils/copy-workbox-sw` utility because it writes to the filesystem.
* The version of `workbox-sw` is included in it's filename so we need that information to import it in the generated service worker. `workbox-webpack-plugin/utils/copy-workbox-sw` also returns the `workbox-sw` filename so we can pass it to `workbox-webpack-plugin/generate-webpack-sw`.
* During the **emit** phase of the webpack lifecycle, we:
  1. Get the manifest entries
  1. Use the entries to generate a file-manifest
  1. Generate a service worker with the file-manifest name and `workbox-sw` filename or copy a service worker from the `config.swSrc` then prepend it with the required `importScripts(workbox-sw.js, file-manifest.js)`.
  1. Add both the file-manifest and the service-worker to the webpack assets.

#### Todo
* I'm not passing any configuration into `sw-build.generateSW` as we would need to. This is because of the pattern where utilities *write* to the file system. If we don't change the utilities, one way to fix this would be to copy the *template populate* logic from `sw-build/generate-sw`. Ideally we could abstract the populating logic from `generate-sw` into a util that does not write to the file system.
* Honor `swDest` incase users want to separate the service worker from build directory.
* There may be cases where assets are not emitted by webpack (the `public/` dir in create-react-app, for example). We should provide a way to "merge" `workbox-build/get-file-manifest-entries` with `workbox-webpack-plugin/get-webpack-manifest-entries`.  `workbox-build/get-file-manifest-entries` supports globPatterns for finding assets, and can be used to generate manifest entries for files that do not go through webpack. Could be a simple `Object.assign(webpackEntries, workboxEntries)`.
* Supoort `sw-precache-webpack-plugin` configurations.
* Fix tests (once the `sw-build/generate-sw`) item is resolved.

### API changes
These are just proposals, we can stay within the exisiting API but I wanted to collect feedback for these changes.

* Running `workbox-build` in `injectManifest` mode will not copy `workbox-sw` over but this is a known pain point for developers. Instead, copy the `workbox-sw` over then prefix the existing service worker with an `importScripts(workbox-sw-V.xxx.js)` line.
* Always generate a file manifest (do not support `injectManifest` mode). We can effectively "inject" the manifest by taking a `config.manifestVarName` parameter that will be used in the `file-manifest.js` template: (`self.${config.manifestVarName} = ${fileManifest}`) and prepend an `importScripts(${manifestFileName})` line to the existing service worker.

### Blocks
* I wasn't able to import the `workbox-sw` package into the `workbox-webpack-plugin` package (in the `workbox-webpack-plugin/utils/copy-workbox-sw` utility). I kept getting a "module not found error" when I manually tested using the create-react-app approach). My work around was to add the `copy-workbox-sw` to the `workbox-build` package (that I was able to import) and manually test using that.
* A lot of the utilities use `fs` to read lodash templates. There may be problems using `fs` in webpack plugins (I removed it from `sw-precache-webpack-plugin` and @TheLarkInn mentioned using `compiler.outputFileSystem` instead).
* Utilities often *populate* then *write* the populated template to the filesystem. I'm having to duplicate and isolate the *populate* logic and am referencing the lodash templates across modules.


goldhand:
I signed it!

jeffposnick:
Thanks so much, @goldhand!

I'll give it a thorough review in a bit.

Some housekeeping: we'll want to target the `v3` branch with this PR, as we still might cut a few releases using the v2 major version number from `master` before we're ready to launch any breaking changes. At the top of my TODO list is to port over the `workbox-build` module to the `v3` branch, which should then make it easier for you to land your code in `v3` as well.

addyosmani:
>  At the top of my TODO list is to port over the workbox-build module to the v3 branch, which should then make it easier for you to land your code in v3 as well.

Was just talking to @gauntface about this. Exciting to see this slowly coming together, @goldhand. Thanks for getting this work started!

goldhand:
Cool, I'll move this to target the v3 branch. This is still very "WIP", just wanted to get something out there to see if anyone could help in areas.

gauntface:
Thanks for this work @goldhand.

I'm very keen to get all the changes to workbox-build known for V3 so this is very timely for @jeffposnick's work.

## Blocks

- workbox-sw would need to be added as a dependency to the webpack plugin to be able to copy the files over. This will likely change in V3 (but there will still be copying done).
    - Basically workbox-sw currently contains all the code from a number of modules. In V3 these will be seperate files and developers only need to pull in what they use.
- Why are there problems reading files in a webpack plugin (sorry if that's a stupid question)? When generating a service worker, we'll need to grab an example template and alter it. `compiler.outputFileSystem` sounds weird to me in that the reading is done before anything is outputted. Would it make send to alter workbox-build to use some other file reader API or would we need to switch to a template which is a string and manipulate that rather than read a file?
- What would an ideal API surface in workbox-build be to prevent you from needing to duplicate the logic? Would it make sense to add API's to workbox-build that would return a string (in the case of generating a SW) and an API that accepted and returns strings as taking in a service worker file and returning an altered service worker file.

# API Changes

I think it would be good to have a brainstorm on fixing this. There are a number of issues around this space and a number of possible solutions, all with varying degrees of magic / flexibility and usefulness. 

@jeffposnick Do you think you could work on a proposal for workbox-build that would address some of the concerns raised above? If not I'm more than happy to have a crack at it, but I know you are owning that module and I think having a fresh pair of eyes on it would be good. Also happy if you'd like to have a discussion of possible ideas and approaches before attempting a proposal. Basically don't want to block and don't want to waste your time.

I think a proposal might be helpful before starting work.

#### What is the CDN?

@goldhand In V3 we are planning on supporting a CDN. This should make it easy for developers to use Workbox on Glitch and also avoid the issue of needing to know where workbox-sw is in `node_modules` + it avoids having magic to inject stuff in a service worker script (as the cdn import can be copy and pasted).

packages/workbox-webpack-plugin/build.js
```diff
@@ -17,6 +17,6 @@ const fse = require('fs-extra');
 const path = require('path');
 
 module.exports = () => {
-  return fse.copy(path.join(__dirname, 'index.js'),
-    path.join(__dirname, 'build', 'index.js'));
+  return fse.copy(path.join(__dirname, 'src'),
```
gauntface: Is this needed?

I think this may have been a thing that was needed for older versions but the build model has changed in V3. At the moment lerna will just publish the directoy as is and won't use build at all.

packages/workbox-webpack-plugin/package.json
```diff
@@ -23,7 +23,8 @@
     "node": ">=4.0.0"
   },
   "dependencies": {
-    "workbox-build": "^2.0.1"
+    "workbox-build": "^2.0.1",
+    "workbox-sw": "^2.0.1"
```
gauntface: It's worrying if you added dep for workbox-sw and it can't be copied over :-S

packages/workbox-webpack-plugin/src/index.js
```diff
@@ -0,0 +1,108 @@
...
+        this._config[key] = value;
+      },
+    };
+  }
+
+  /**
+   * @private
+   * @param {Object} [compiler] default compiler object passed from webpack
+   */
+  apply(compiler) {
```
gauntface: Could you add some comments about the lifecycle of the compiler callbacks?

Reading through this code I'm guessing a lot about where things occur.

packages/workbox-webpack-plugin/src/index.js
```diff
@@ -0,0 +1,108 @@
...
+      setReadFile(readFile);
+      next();
+    });
+    compiler.plugin('make', (compilation, next) => {
+      copyWorkboxSW(compilation.options.output.path)
+      .then(({workboxSW, workboxSWMap, workboxSWName}) => {
+        // Add the workbox-sw file to compilation assets
+        compilation.assets[workboxSWName] = webpackAsset(workboxSW);
+        compilation.assets[`${workboxSWName}.map`] = webpackAsset(workboxSWMap);
+        this.config.set('workboxSWFilename', workboxSWName);
```
gauntface: Could this be this._config?

packages/workbox-webpack-plugin/src/index.js
```diff
@@ -0,0 +1,108 @@
...
+   */
+  constructor(config) {
+    this._config = config || {};
+  }
+
+  /**
+   * @private
+   * @return {Object} The configuration for a given compilation.
+   */
+  get config() {
```
gauntface: Why are you following this model of exporting an object that seems to read and write to the config option?

packages/workbox-webpack-plugin/src/index.js
```diff
@@ -0,0 +1,108 @@
...
+      });
+    });
+    compiler.plugin('emit', (compilation, next) => {
+      const config = this.config.get();
+      const {
+        manifestFilename = `file-manifest.${compilation.hash}.js`,
+        manifestVarName = '__file_manifest',
+        workboxSWFilename,
+        swSrc,
+      } = config;
```
gauntface: I've never seen this syntax before, what is this doing?

packages/workbox-webpack-plugin/src/lib/generate-webpack-manifest.js
```diff
@@ -0,0 +1,4 @@
+module.exports = (manifestEntries, manifestVarName) =>
```
gauntface: Nit: This file should probably be named something else. `generate-webpack-manifest` makes it sound like a webpack thing rather than a file manifest that is used in the service worker.

packages/workbox-webpack-plugin/src/lib/generate-webpack-sw.js
```diff
@@ -0,0 +1,29 @@
+const {readFile} = require('./utils/read-file');
+
+const importScripts = ({
```
gauntface: Not: I'd rename this function. importScripts is way to generic / confusing given the context. `wrapAsImport` or something in that direction would be helpful.

packages/workbox-webpack-plugin/src/lib/generate-webpack-sw.js
```diff
@@ -0,0 +1,29 @@
...
+  workboxSWFilename,
+  manifestFilename,
+}) => `importScripts('${workboxSWFilename}', '${manifestFilename}');`;
+
+const generateSW = ({
```
gauntface: Again, given that workbox-build has a generateSW method and workbox-cli has generate-sw, the naming is confusing and I'm not sure what this function is doing or why it's doing it.

packages/workbox-webpack-plugin/src/lib/generate-webpack-sw.js
```diff
@@ -0,0 +1,29 @@
+const {readFile} = require('./utils/read-file');
```
gauntface: Is this the webpack templating that you've used instead of workbox-builds templating?

If so, renaming this to `webpack-generate-sw-template` would help with context.

packages/workbox-webpack-plugin/src/lib/generate-webpack-sw.js
```diff
@@ -0,0 +1,29 @@
...
+}) => new Promise((resolve, reject) => resolve(`
+${importScripts({workboxSWFilename, manifestFilename})}
+
+const workboxSW = new self.WorkboxSW();
+workboxSW.precache(self.${manifestVarName});
+`));
+
+const generateOrCopySW = (config, swSrc) => new Promise((resolve, reject) => {
+  if (!swSrc) {
+    return generateSW(config).then((serviceWorker) => resolve(serviceWorker));
```
gauntface: Nit: Is `serviceWorker` a generated string or a file path?

packages/workbox-webpack-plugin/src/lib/generate-webpack-sw.js
```diff
@@ -0,0 +1,29 @@
...
+  } else {
+    return readFile(swSrc).then((serviceWorkerSource) =>
+      resolve(`${importScripts(config)}\n${serviceWorkerSource}`)
```
gauntface: I think the naming would help with this.

packages/workbox-webpack-plugin/src/lib/get-webpack-manifest-entries.js
```diff
@@ -0,0 +1,76 @@
...
+    // chunk has a name
+    Object.prototype.hasOwnProperty.call(chunk, 'name')
+      // did not specify include or chunk in include
+      && (!Array.isArray(include) || !!~include.indexOf(chunk.name))
+      // did not specify exclude or chunk not in exclude
+      && (!Array.isArray(exclude) || !~exclude.indexOf(chunk.name))
+  );
+
+/**
+ * Maps webpack asset filenames to their chunk hash
```
gauntface: Could you add a brief description of the input and outputs of this function. Would help read the code.

packages/workbox-webpack-plugin/src/lib/get-webpack-manifest-entries.js
```diff
@@ -0,0 +1,76 @@
...
+  const includeNames = config.chunks;
+  const excludeNames = config.excludeChunks;
+
+  // Array<{files: Array<(filename: string)>, name?: string}>
+  let {chunks} = compilation;
+
+  if (includeNames || excludeNames) {
+    // Only include chunks in config.chunks and
+    // exclude any chunks in config.excludeChunks
+    chunks = filterChunks(chunks, includeNames, excludeNames);
```
gauntface: Would it be accurate to do the following:

Rename `chunks` to `allChunks` or `allCompilationChunks`, renaming `includeNames` to `chunksToInclude` and `excludeNames` to `chunksToExclude` and then add a variables `chunksToPrecache` would would result in:

```javascript
let chunksToPrecache = allChunks;
if (chunksToInclude || chunksToExclude) {
    chunksToPrecache = filterChunks(allChunks, chunksToInclude, chunksToExclude);
}
```

packages/workbox-webpack-plugin/src/lib/get-webpack-manifest-entries.js
```diff
@@ -0,0 +1,76 @@
...
+
+  /**
+   * Files that will be used to generate the manifest entries
+   *
+   * If config.chunks is specifed, we only use files that belong to named chunks
+   * otherwise we use all of webpack's compilation.assets
+   *
+   * @type {Array<string>}
+   */
+  const manifestFiles = includeNames
```
gauntface: I'm kinda confused with all the alterations and this object keys here, would this change to:

```javascript
const assetsToChunkHash = mapAssetsToChunkHash(precacheChunks);
const manifestFiles = Object.keys(precacheChunks);
```

This being due to the check includeNames || excludeNames above rendering this if statement no longer needed.

packages/workbox-webpack-plugin/src/lib/get-webpack-manifest-entries.js
```diff
@@ -0,0 +1,76 @@
...
+ */
+const mapAssetsToChunkHash = (chunks) =>
+  chunks.reduce((assetMap, chunk) =>
+      Object.assign(
+        assetMap,
+        ...chunk.files.map((f) => ({[f]: chunk.renderedHash}))
+    ), {});
+
+const generateManifestEntries = (compiler, compilation, config) => {
+  const {hash} = compilation; // compilation hash
```
gauntface: You could inline the single use of hash to the bottom of this function as `compilation.hash`.

packages/workbox-webpack-plugin/src/lib/get-webpack-manifest-entries.js
```diff
@@ -0,0 +1,76 @@
+const url = require('url');
```
gauntface: To prevent developers from precaching huge files (i.e. videos / audio files) we have a size limit on the files we'll precache. Is that possible with WebPack?

packages/workbox-webpack-plugin/src/lib/utils/read-file.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * Use the `compiler.inputFileSystem._readFile` method instead of `fs.readFile`,
+ * `readFile` is configured to use `compiler.inputFileSystem._readFile` during
+ * the run phase of the webpack compilation lifecycle by passing the function
+ * to the `setReadFile` function.
+ */
+let readFileFn;
```
gauntface: I'm not crazy about this pattern but I can see why you've done it.

packages/workbox-webpack-plugin/src/lib/utils/webpack-asset.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Creates a webpack asset from a string that can be added to a compilation
```
gauntface: Not sure what this does or why it's needed, could you add some more context? Sorry I'm such a webpack noob :(

packages/workbox-webpack-plugin/src/index.js
addyosmani: I had a similar question. I thought that this might be trying to replicate a pattern used in the sw-precache webpack plugin, but I don't think this model in this form was used there (I could be wrong) :)

packages/workbox-webpack-plugin/src/index.js
addyosmani: @gauntface This is a form of loose destructuring assignment where properties can be unpacked from an object and assigned to a variable with a different name than the object property.

packages/workbox-webpack-plugin/src/lib/get-webpack-manifest-entries.js
addyosmani: Perhaps something like https://github.com/ai/size-limit/blob/master/index.js#L135 would do the trick?

packages/workbox-webpack-plugin/src/lib/get-webpack-manifest-entries.js
```diff
@@ -0,0 +1,76 @@
...
+  revision,
+  url: url.resolve(publicPath, filePath),
+});
+
+/**
+ * Filter chunks that have one of the provided names (from config.chunks)
+ * @param {Array<Object>} chunks Webpack chunks
+ * @param {Array<string>} include Chunk names to include
+ * @param {Array<string>} exclude Chunk names to exclude
+ * @return {Array<Object>} Filtered array of chunks
```
addyosmani: Once we have some docs together (for a more fleshed out version), it would be great for them to capture a few different examples of using chunks, include and exclude.

packages/workbox-webpack-plugin/src/lib/utils/webpack-asset.js
addyosmani: I understand the reason this is needed, but agree for future contributors it would be useful to expand on why this is needed.

packages/workbox-webpack-plugin/src/index.js
goldhand: No need for this! Will completely change this "config" when I clean up the api

packages/workbox-webpack-plugin/src/index.js
goldhand: This is destructuring with "defaults".

So `var manifestVarName = config.manifestVarName || '__file_manifest'`;

packages/workbox-webpack-plugin/src/lib/generate-webpack-sw.js
goldhand: @gauntface the problem with using `workbox-build`'s `generateSW` method is that it writes to the file system. The reason I have any of these `generate-template` utils is because I couldn't figure out how to use the existing code without writing to the filesystem.

This `generate-webpack-sw` util is a placeholder intended to highlight one of the issues described in the **blocks** section:

> Utilities often populate then write the populated template to the filesystem. I'm having to duplicate and isolate the populate logic and am referencing the lodash templates across modules.

Maybe I missed it, but does `workbox-cli` have a way to generate the template without writing to fs?

*sidenote: sorry for the single comments. Will answer the other points in a review later*

packages/workbox-webpack-plugin/build.js
jeffposnick: I've got a build process working in v3 for Node modules is this PR: https://github.com/GoogleChrome/workbox/pull/811

(As of right now, it determines that something is a Node module based on the presence of `engines.node` in `package.json`, which is a bit of magic.)

In summary, assuming #811 is merged first and this PR is rebased against the `v3` branch, you should be able to get rid of this `build.js` file entirely.

packages/workbox-webpack-plugin/package.json
jeffposnick: I wonder whether the issue you described was due to `lerna`-being-`lerna`.

If you do a `gulp lerna-bootstrap` in the top-level `workbox/` directory that might take care of ensuring that `workbox-build`'s dependency on `workbox-sw` gets satisfied, without you having to explicitly list it.

packages/workbox-webpack-plugin/package.json
```diff
@@ -23,7 +23,8 @@
...
   },
   "author": "Google's Web DevRel Team",
```
jeffposnick: Feel free to add your name!

packages/workbox-webpack-plugin/src/index.js
```diff
@@ -0,0 +1,108 @@
...
+ *   ]
+ * };
+ *
+ * @module workbox-webpack-plugin
+ */
+class WorkboxWebpackPlugin {
+  /**
+   * Creates an instance of WorkboxBuildWebpackPlugin.
+   *
+   * @param {Object} [config] All the options as passed to `workbox-build`.
```
jeffposnick: `{@link module:workbox-build~Configuration}` should, if I understand my JSDoc syntax, generate a link to the canonical description of all the `workbox-build` options.

packages/workbox-webpack-plugin/src/lib/generate-webpack-manifest.js
```diff
@@ -0,0 +1,4 @@
...
+  new Promise((resolve) => resolve(
```
jeffposnick: Nit: You can use `Promise.resolve(value)` as shorthand to create a promise that resolves with a given value. No need to explicitly construct a `Promise` and pass it a function.

packages/workbox-webpack-plugin/src/lib/generate-webpack-sw.js
jeffposnick: See https://github.com/GoogleChrome/workbox/pull/811/files#r140564059 for what we can do in v3 of`workbox-build`, if that would help.

packages/workbox-webpack-plugin/src/lib/get-webpack-manifest-entries.js
```diff
@@ -0,0 +1,76 @@
...
+ * @return {Object} {filename: hash} map
+ */
+const mapAssetsToChunkHash = (chunks) =>
+  chunks.reduce((assetMap, chunk) =>
+      Object.assign(
+        assetMap,
+        ...chunk.files.map((f) => ({[f]: chunk.renderedHash}))
+    ), {});
+
+const generateManifestEntries = (compiler, compilation, config) => {
```
jeffposnick: Something I noticed when running your code locally was that the `workbox-sw` (and `.map`) file were getting picked up here:

```
  {
    "revision": "9a0110a4f8c1796a1414",
    "url": "/workbox-sw.prod.v2.0.1.js"
  },
  {
    "revision": "9a0110a4f8c1796a1414",
    "url": "/workbox-sw.prod.v2.0.1.js.map"
  },
```

While including these files in the manifest won't break anything (the browser will ignore them when performing `importScripts()`), it means that there's some extra bytes transferred and stored on disk during SW installation, and it would be preferable to exclude them.

packages/workbox-webpack-plugin/src/lib/utils/copy-workbox-sw.js
```diff
@@ -0,0 +1,19 @@
+const path = require('path');
+const {readFile} = require('./read-file');
+const errors = require('workbox-build/src/lib/errors');
+
+module.exports = (readFileFn) => {
```
jeffposnick: FWIW, in the v3 branch, we're transpiling all the Node code, so feel free to use features like `async`/`await` if you'd prefer.

packages/workbox-webpack-plugin/src/lib/get-webpack-manifest-entries.js
```diff
@@ -0,0 +1,76 @@
+const url = require('url');
+
+const entry = (revision, filePath, publicPath) => ({
+  revision,
+  url: url.resolve(publicPath, filePath),
```
jeffposnick: I remember this coming up in a review of the earlier revision of the Webpack plugin: `publicPath` [is defined](https://webpack.js.org/configuration/output/#output-publicpath) as something that's prepended to every URL, not joined with it, so this should probably be `publicPath + filePath`;

packages/workbox-webpack-plugin/src/lib/get-webpack-manifest-entries.js
```diff
@@ -0,0 +1,76 @@
+const url = require('url');
+
+const entry = (revision, filePath, publicPath) => ({
+  revision,
```
jeffposnick: One thing that bears mentioning: as an optimization, the `precache()` function will not append a cache-busting query parameter when populating its caches if the `revision` field is not present.

That's the same behavior that in `sw-precache-webpack-plugin` was defined by the `dontCacheBustUrlsMatching` configuration option, for which we recommended using a RegExp that matches URLs that contained hashes in them already.

I'm not sure exactly how, but is there any indication at this point in the build process whether `filePath` contains a versioning hash in it? (I'd assume it wouldn't for HTML files, but would for most other assets.) If you can safely infer the presence of a hash, then you can leave out `revision` and get optimal behavior without developers needing to use `dontCacheBustUrlsMatching`.

goldhand: Makes sense, we can check if they are using a version hash in the chunk name and filter those out. Then just ignore the `dontCacheBustUrlsMatching` option altogether?

packages/workbox-webpack-plugin/src/lib/get-webpack-manifest-entries.js
goldhand: Oops, me too. Will change that

packages/workbox-webpack-plugin/src/lib/get-webpack-manifest-entries.js
goldhand: +1

packages/workbox-webpack-plugin/src/lib/get-webpack-manifest-entries.js
goldhand: yes

packages/workbox-webpack-plugin/src/lib/get-webpack-manifest-entries.js
goldhand: Oh, I thought we would want to cache these? I am curious, how come we don't want to cache these assets?

packages/workbox-webpack-plugin/src/lib/get-webpack-manifest-entries.js
goldhand: sure!

packages/workbox-webpack-plugin/src/lib/get-webpack-manifest-entries.js
goldhand: yeah it's kind of confusing. I'll update comments to clarify more.

`assetsToChunkHash` is a map of `filename`: `chunkhash` and only contains assets that belong to a chunk with a name property. It's also filtered out chunks according to `include` / `exclude`.

`compilation.assets` is all assets that belong to the webpack compilation, even assets that belong to chunks with no name.

packages/workbox-webpack-plugin/src/lib/utils/copy-workbox-sw.js
goldhand: YAY!

packages/workbox-webpack-plugin/src/lib/utils/read-file.js
goldhand: Yeah, me either. This is just incase we can't use `fs` or `fse` in the webpack compilation. Is there a better pattern for this you would suggest?

packages/workbox-webpack-plugin/src/lib/utils/webpack-asset.js
goldhand: I will elaborate what this is in the docs.

These are the minimum properties that webpack expects a single asset to have. Given a string (`asset`) we can structure it for webpack without any other arguments.

packages/workbox-webpack-plugin/build.js
jeffposnick: The goal is to get #811 reviewed and merged into `v3`, yes. So that would ultimately be what this PR would build on top of. I'll modify the code in that PR to address some of the things you've asked for.

packages/workbox-webpack-plugin/src/lib/get-webpack-manifest-entries.js
jeffposnick: Anything pulled in with `importScripts` (e.g. `workbox-sw.prod.v2.0.1.js`) is implicitly cached and available offline—this behavior is defined in the service worker specification ([section 7.3.2](https://w3c.github.io/ServiceWorker/#importscripts), item 3.1). Requests for those URLs don't trigger a service worker `fetch` event.

The `.map` isn't covered by this policy, but I don't actually know that the request for the `.map` made by DevTools while debugging actually triggers a `fetch` event either, since DevTools might not be considered an active client of the service worker. And even if the `.map` request does go through the service worker, needing to use the sourcemap while offline is an edge case that wouldn't warrant the extra bytes transferred and stored.

packages/workbox-webpack-plugin/src/lib/get-webpack-manifest-entries.js
jeffposnick: If someone goes out of their way to supply `dontCacheBustUrlsMatching` then perhaps it should still be honored, but being able to implicitly provide that optimization is definitely more developer-friendly.

I think it's more of a case of you not taking specific action one way or another in your Webpack-specific code.

packages/workbox-webpack-plugin/src/index.js
addyosmani: 👍

packages/workbox-webpack-plugin/src/lib/generate-webpack-manifest.js
addyosmani: > generate-manifest-with-webpack

This works!

packages/workbox-webpack-plugin/src/lib/get-webpack-manifest-entries.js
addyosmani: TBH I think it was static and something we internally enforced. @jeffposnick do you know if this was something we otherwise made configurable?

packages/workbox-webpack-plugin/src/lib/get-webpack-manifest-entries.js
addyosmani: Thanks!

jeffposnick:
@goldhand, as an FYI, #811 is now merged, so `workbox-build` is in the `v3` branch. The codebase at the moment is functionally equivalent to what was in `master`, so it doesn't have the modified interfaces we've been discussing (tracked in #803), but it does bring into `v3` testing setup and build steps for `node` projects, which were previously missing.

It should be possible to rebase against `v3` at this point, in advance of waiting for the `workbox-build` interfaces to be updated. Or you could wait for those interface changes, and do it all in one go. Let us know if you run into any rough patches.

packages/workbox-webpack-plugin/src/lib/get-webpack-manifest-entries.js
jeffposnick: It's configurable via the `maximumFileSizeToCacheInBytes` option.

packages/workbox-webpack-plugin/build.js
addyosmani: With #811 now merged into v3, now might be a good time to try rebuilding this PR on top of what is there. It's unclear to me how much churn we might see in the workbox-build interface changes but @jeffposnick seemed to be supportive of either rebasing against v3 or waiting a little longer. I think we're fine with whatever your preference is, @goldhand.

packages/workbox-webpack-plugin/package.json
addyosmani: Yeah. We definitely want you to feel like as much a part of the team as anyone else. Feel free to add yourself :)

@jeffposnick @gauntface at some point, we might want to change this to "Workbox Core Team" or reference an owners list we can also add folks like Will to.

packages/workbox-webpack-plugin/src/index.js
```diff
@@ -0,0 +1,108 @@
...
+ * @example
+ * const WorkboxBuildWebpackPlugin = require('workbox-webpack-plugin');
+ * .
+ * module.exports = {
+ *   ...
+ *   plugins: [
+ *    new WorkboxBuildWebpackPlugin({
+ *      chunks: ['main'],
+ *      excludeChunks: ['mutable'],
+ *      manifestFilename: string,
```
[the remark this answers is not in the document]
addyosmani: Thanks. Even including the description you provided above in comments would be more than sufficient to cover future contributors.

packages/workbox-webpack-plugin/src/lib/generate-webpack-sw.js
addyosmani: In favor of the naming Matt suggested above.

packages/workbox-webpack-plugin/src/lib/get-webpack-manifest-entries.js
addyosmani: > needing to use the sourcemap while offline is an edge case that wouldn't warrant the extra bytes transferred and stored.

I strongly agree with this. I've seen plenty of examples of folks also generating and accidentally caching `.map` files that were pretty large (1MB) causing a perf hit when it's easier to just not cache this asset.

jeffposnick:
https://github.com/GoogleChrome/workbox/issues/803#issuecomment-333951970 has an update on the current state of `workbox-build` in `v3`, with an eye on features that should simplify this PR.

goldhand:
Thanks @jeffposnick I will be working on this over the weekend! generate-sw.js is exactly what I needed and sweet `async/await`

anilanar:
I'm late to the party, I'd love to review this PR and test it against our complicated webpack configuration @trivago.

goldhand:
Thanks @anilanar ! Would appreciate that. Should have a more complete working version after this weekend. Here is the webpack-related work that will support running with webpack-dev-server. Need to bring in more of the workbox functionality now so if you want to add to this pr or use as a starting point for another pr, please do!

goldhand:
@jeffposnick I made a decent amount of progress this weekend:

* Migrated to `v3` branch
* Used `workbox-build` utilities where I could
* Updated documentation where requested in feedback
* Renamed files / functions so it was clearer what they do
* Noted missing todo's in code comments with links to the review comments for details (in case someone else wants to make these fixes).
* Added a method that returns all configuration params for `generateSWString`. I am highlighting this because the idea was we could explicitly specify whatever properties WorkboxWebpack plugin config should pass to `generateSWString`. This allow should allow us to easily customize the api / experience.

Going to call it a night but I'll try to update with more details / respond to any questions on this pr over this next week.

packages/workbox-webpack-plugin/src/index.js
```diff
@@ -0,0 +1,181 @@
...
+  constructor(config = {}) {
+    this.config = config;
+  }
+
+  /**
+   * @private
+   * @return {Object} All workbox configuration options that can be accepted
+   * by {@link module:workbox-build.generateSWString}
+   */
+  get generateSWStringOptions() {
```
goldhand: @jeffposnick this is where we should specify which generateSWString properties should be supported.

packages/workbox-webpack-plugin/src/lib/utils/resolve-webpack-url.js
```diff
@@ -0,0 +1,29 @@
...
+ * Use publicPath + filePath instead of url.resolve(publicPath, filePath) see:
+ * https://webpack.js.org/configuration/output/#output-publicpath
+ *
+ * @function resolveWebpackUrl
+ * @param {Array<string>} paths File paths to join
+ * @return {string} Joined file path
+ *
+ * @memberof module:workbox-webpack-plugin
+ */
+module.exports = (...paths) => paths.join('');
```
anilanar: This would get you in trouble, because all webpack plugins I've used so far are capable of handling public paths that do not end with a `/`, including webpack itself.

packages/workbox-webpack-plugin/src/lib/get-manifest-entries-with-webpack.js
```diff
@@ -0,0 +1,137 @@
...
+ */
+const mapAssetsToChunkHash = (chunks) =>
+  chunks.reduce((assetMap, chunk) =>
+      Object.assign(
+        assetMap,
+        ...chunk.files.map((f) => ({[f]: chunk.renderedHash}))
+    ), {});
+
+/**
+ * Generate an array of manifest entries using webpack's compilation data
```
anilanar: Question: For dev environments, should workbox filter out source maps? Also, some people may be creating source maps even for production builds to debug deployed code. Some users may need to filter out some assets depending on their use cases.

anilanar:
Great work! I'll retest after `HtmlWebpackPlugin` is handled.

packages/workbox-webpack-plugin/src/index.js
```diff
@@ -0,0 +1,181 @@
...
+     * system). If that is the case, it might be better to set different values
+     * for setReadFile using compiler.plugin('run') for production and
+     * compiler.plugin('watch-run') for development.
+     */
+    setReadFile(compiler.inputFileSystem._readFile);
+    /**
+     * During the make phase of the webpack compilation, we use
+     * workbox-webpack-plugin/utils/copy-workbox-sw to add a built version of
+     * workbox-sw to the webpack compilation assets array.
+     */
```
anilanar: Hmm, I think this is not what everyone would want. 
We build `service-worker` separately by another webpack build, using workbox from npm which is embed into service-worker file. We have bunch of custom loaders that modifies/optimizes/analyzes code, so we would not prefer to have it this way. This approach is also uncommon in webpack/js community.

packages/workbox-webpack-plugin/src/index.js
```diff
@@ -0,0 +1,181 @@
...
+ *      manifestFilename: 'file-manifest.js',
+ *      swSrc: './src/sw.js', // path to an existing service worker
+ *      filename: 'service-worker.js', // destination filename
+ *     });
+ *   ]
+ * };
+ *
+ * @module workbox-webpack-plugin
+ */
+class WorkboxWebpackPlugin {
```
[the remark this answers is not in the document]
goldhand: Makes sense, originally had it that way but I didn't see it in any other workbox api so I removed it. @jeffposnick what do you think about having `manifestVarName` as an option that defaults to `__precacheManifest`?

Only problem is that in the `copy-service-worker` method, we would need to be able to update that param when `swBuild.precache(<manifestVarName)` is run

packages/workbox-webpack-plugin/src/index.js
goldhand: Specifying the path to the service worker you create in `swSrc` would use that service worker then prepend it with `importScripts('workbox-sw.prod.v2.x.x.js', 'file-manifest.js')` so then you can just run:

`workboxBuild.precache(__precacheManifest)` in your service worker.

Not sure if this is the best flow...

see https://github.com/GoogleChrome/workbox/pull/808/files/a2dcaf4235f9f10456988a78759eb17dfedf6823#diff-5b99f4463e518de9c1c1cc612e8f6b59R73 for the swSrc parameter

packages/workbox-webpack-plugin/src/lib/get-manifest-entries-with-webpack.js
goldhand: See https://github.com/GoogleChrome/workbox/pull/808/files/a2dcaf4235f9f10456988a78759eb17dfedf6823#diff-dff2a0e4847723c3a1d299d74db40c0cR39

The idea was to provide a config option (`assetIgnoreRegex` or something) where users can provide an array of regular expressions to filter asset filenames out. By default we would use: [/\.map$/, /asset-manifest\.json$/]

It feels a little odd filtering by asset filenames. Normally (in my experience) webpack configs include/exclude assets based on chunk names, not actual filenames but for the case with `*.map$` files, they would need to be filtered by asset file names.

packages/workbox-webpack-plugin/src/lib/utils/resolve-webpack-url.js
goldhand: There was discussion about this. This is why I isolated the path logic too. Do you think we should do a `url.resolve` rather that `String.join`?

packages/workbox-webpack-plugin/src/lib/generate-or-copy-sw.js
```diff
@@ -0,0 +1,47 @@
...
+ * @return {Promise<string>} generated service worker string
+ *
+ * @memberof module:workbox-webpack-plugin
+ */
+module.exports = (config, swSrc) => new Promise((resolve, reject) => {
+  if (!swSrc) {
+    // use workbox-build to generate the service worker
+    return resolve(generateSWString(config));
+  } else {
+    return readFile(swSrc).then((serviceWorkerSource) =>
```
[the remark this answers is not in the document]
goldhand: I prefer absolute paths, how would we enforce that?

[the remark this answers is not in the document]
anilanar: `path.isAbsolute` looks useful. Plugin options would be validated in the constructor, similar to how webpack does it.

packages/workbox-webpack-plugin/src/lib/get-manifest-entries-with-webpack.js
anilanar: I guess source-maps can be filtered out by default. However I'm not sure how that would affect development experience when testing offline caching. Need to test it.

This problem is similar to how loaders handle file inclusion/exclusion using three properties: `{test, include, exclude}` and the code for that is in https://github.com/webpack/webpack/blob/master/lib/ModuleFilenameHelpers.js, look for `matchObject`. That's what `uglifyjs-webpack-plugin` is also using in  https://github.com/webpack-contrib/uglifyjs-webpack-plugin/blob/master/src/index.js.

packages/workbox-webpack-plugin/src/lib/utils/resolve-webpack-url.js
anilanar: Maybe keep it as it is. It is quite obvious what's wrong when you see a slash missing between publicPath and filename.

packages/workbox-webpack-plugin/src/index.js
anilanar: I assumed a raw `service-worker.js` would look as follows:

```
import WorkboxSW from 'workbox-sw';
const workboxSW = new WorkboxSW();
workboxSW.precache(self.__precacheManifest);
```

After building with webpack and processing with workbox:

```
importScripts('file-manifest.{hash}.js');
(function () {
      /* webpack generated code here */
});
```

Ideally, users can opt to use `workbox-precaching` package instead of the whole `workbox-sw` when they don't need anything else than simple precaching.

jeffposnick:
Thanks for your work on this, @goldhand, and thanks for your feedback, @anilanar!

I plan on giving it a thorough review and address some of the `workbox-build` comments. As discussed, given the amount of work in this PR, we're open to the idea of merging into `v3` as long as the core functionality is there, and track some incremental changes or outstanding feature requests in issues to be addressed in follow-up pRs.

packages/workbox-webpack-plugin/src/index.js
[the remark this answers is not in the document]
jeffposnick: The format is an array of [`ManifestEntry` objects](https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-build#.ManifestEntry).

One pattern that you could adopt for the standalone manifest imports is

```js
self.__precacheManifest = [entriesDefinedInThisFile].concat(self.__precacheManifest || []);
```

That's basically what we're already doing in the `v3` version of the SW template file, to merge anything passed in via the `manifestEntries` template variable with anything defined already via `self.__precacheManifest`.

packages/workbox-webpack-plugin/src/lib/utils/resolve-webpack-url.js
jeffposnick: Yes, I'm the one who keeps making an issue out of this 😄 

It seems like we should be following the conventions from the Webpack docs, which says that `publicPath` is a string prefix, but I can stop making a big deal about it if folks think that in practice, it's always joined with a path separator.

packages/workbox-webpack-plugin/src/index.js
jeffposnick: FWIW, we're expecting it to be common for developers to refer to `workbox-sw` from our new CDN URL, support for which is baked in to `workbox-build` v3. The CDN `workbox-sw` basically acts as a lightweight loader + API surface, and will use `importScripts` at runtime to pull in additional dependencies for things like routing, caching, etc.

That being said, we know that not every developer will want to rely on the CDN copy, and Webpack users in particular are in a good position to bundle an optimized copy of `workbox-sw` + the additional routing, caching, etc. code and serve it locally.

CC: @philipwalton who's also exploring the ergonomics of bundling Workbox code.

jeffposnick:
I'm going to merge, and have filed separate issues to track some unresolved questions/feedback to be addressed in follow-up PRs:

- https://github.com/GoogleChrome/workbox/issues/933
- https://github.com/GoogleChrome/workbox/issues/934
- https://github.com/GoogleChrome/workbox/issues/935

There will also be some general code style and `workbox-build` integration PRs post-merge.

packages/workbox-webpack-plugin/src/lib/get-manifest-entries-with-webpack.js
goldhand: So considering we filter both **chunkname** and **filename** what should the api look like?

We could provide explicit chunk / filename filtering eg:
```js
WorkboxPlugin({
  ...
  filterChunks: {
    test: /\.js($&#124;\?)/i,
  },
  filterFilenames: {
    exclude: /\/excludes/,
  },
});
```

Or we can just filter both using the same properties, eg:
```js
WorkboxPlugin({
  ...
  test: /\.js($&#124;\?)/i,
  exclude: /\/excludes/,
});
```
and just filter both **chunknames** and **filenames** by those provided values?

Something else to consider is [html-webpack-plugin's config](https://github.com/jantimon/html-webpack-plugin#configuration) the `{chunks, excludeChunks}` options are very nice. Maybe use those and apply the `{test, include, exclude}` options *only* to **filenames** (**chunknames** are handled by the `chunks` option...

goldhand: @anilanar ^ (meant to tag you in that comment)

packages/workbox-webpack-plugin/src/lib/utils/resolve-webpack-url.js
goldhand: I'll just make note to make sure the documentation is clear on the path logic incase there is any trouble. We can revisit using `url.resolve()` if this is an issue.

packages/workbox-webpack-plugin/src/index.js
goldhand: This will need some more consideration. I think we need to decide if `workbox-sw.js` should be copied into the webpack assets array, use the CDN url or expect the developer to import the `workbox-sw` file into a `service-worker` webpack entry point. Maybe provide an option in the config to specify what option we should use and select a smart default... not sure what *smart* would be ;)

packages/workbox-webpack-plugin/src/lib/get-manifest-entries-with-webpack.js
jeffposnick: FYI: I've opened https://github.com/GoogleChrome/workbox/issues/935 to track this, and moved over some relevant conversation. I don't want too much of the conversation to be lost in a closed PR.

dingquantracy:
When will you guys publish the V3 to npm, so I can use the latest version to solve problems in webpack-dev-middle, thx

PR #811, opened 2017-09-20
V3: workbox-build initial checkin

jeffposnick:
R: @addyosmani @gauntface @philipwalton @prateekbh 
CC: @goldhand 

Apologies for the size of this PR, but there's a lot of code needed to get `workbox-build` happy in v3.

Some things to keep in mind:
- Functionality-wise, this is in keeping with what `workbox-build` supported in v2. There are plans to update the interfaces/clean some things up, but I wanted to get a mostly-identical version migrated first to unblock #808.
  - One exception: All of the runtime code, and some of the tests, have moved from Promises to `async`/`await` (transpiled with Babel to maintain Node 4 support).
  - Another exception: The `dynamicUrlToDependencies` and `staticFileGlobs` options printed warnings when used in v2, and I've fully deprecated them in v3.
- This is the first package that targets the Node environment, rather than the Browser. I had to make some hacks to the testing setup, and some worse hacks to the build process, to get that to work with the way things are set up in v3. Longer-term, it might make sense to revisit the v3 build setup to allow individual packages to specify their own build process, like we allowed in v2.
- I had to move over a bunch of test/infrastructure code from v2 to get the `workbox-build` tests working. For the most part, I stuck them in `infra/`, and that's already getting a bit messy.


packages/workbox-build/src/lib/write-sw.js
```diff
@@ -0,0 +1,85 @@
...
+      workboxSWOptionsString,
+      runtimeCaching,
+    }).trim() + '\n';
+  } catch (error) {
+    throw new Error(
+      `${errors['populating-sw-tmpl-failed']}. '${error.message}'`);
+  }
+
+  try {
+    await fse.writeFile(swDest, populatedTemplate);
```
jeffposnick: TODO: (Find a clean way to) Change the interface of this module so that it could just generate the SW string and return that, instead of always writing to disk.

This would unblock the Webpack usage, described at https://github.com/GoogleChrome/workbox/pull/808/files#r139604544

goldhand: Yes, if there is a way to import a `createSWTemplate` function or something that returns the value of `populatedTemplate` (given a `config` object), this would really help eliminate some code duplication in the webpack plugin.

gulp-tasks/build-packages.js
```diff
@@ -142,22 +143,59 @@ const buildPackage = (packagePath, buildType) => {
...
 const cleanPackages = (packagePath) => {
   const outputDirectory = path.join(packagePath,
-    constants.PACKAGE_BUILD_DIRNAME, constants.BROWSER_BUILD_DIRNAME);
+    constants.PACKAGE_BUILD_DIRNAME);
   return fs.remove(outputDirectory);
 };
 
 gulp.task('build-packages:clean', gulp.series(
-  packageRunnner('build-packages:clean', cleanPackages)
+  packageRunnner('build-packages:clean', {
```
gauntface: I'm struggling to understand what this is going with the NODE and BROWSER variants (Splitting up relevant stuff into two seperate tasks would help).

gulp-tasks/test.js
```diff
@@ -14,7 +14,7 @@ const runNodeTests = () => {
     options.push('--grep', global.cliOptions.grep);
   }
   return spawn(getNpmCmd(), ['run', 'test', '--',
-    `./test/${global.packageOrStar}/node/**/*.mjs`,
+    `./test/${global.packageOrStar}/node/**/*.{js,mjs}`,
```
gauntface: Why not write tests in `mjs`?

gulp-tasks/utils/constants.js
```diff
@@ -23,4 +23,11 @@ module.exports = {
 
   // This is the environments that we should use for NODE_ENV.
   BUILD_TYPES: ['dev', 'production'],
+
+  // Used to identify whether a given package should be built as a Node or
+  // Browser bundle.
+  PROJECT_TYPES: {
```
gauntface: Can we just add this information to the package.json under "workbox". Reading through the above code, this has added a lot of boilerplate and I'm still not sure how packages are determined as node or browser packages.

gulp-tasks/utils/package-runner.js
```diff
@@ -2,12 +2,26 @@ const path = require('path');
 const glob = require('glob');
 const oneLine = require('common-tags').oneLine;
 
+const constants = require('./constants');
```
gauntface: Sorry I'm not a fan of this approach.

Could we just have two variants of package runner. One for node and one for browser, and depending on the package runner called, the func is called with the relevant packages.

```js
browserPackageRunner(displayName, func)
nodePackageRunner(displayName, func)
```

This change along with seperating browser builds from node builds in gulp tasks should make things a little cleaner.

infra/utils/e2e-sw-validator.js
```diff
@@ -0,0 +1,16 @@
+const fakeRunAndCompare = require('./validator/fake-run-and-compare');
```
gauntface: I'd kill this test for now.

It caused a great deal of headaches and needs to be re-considered.

infra/utils/validator/fake-run-and-compare.js
```diff
@@ -0,0 +1,115 @@
+const vm = require('vm');
```
gauntface: Kill this as well.

infra/utils/validator/test-in-browser.js
```diff
@@ -0,0 +1,135 @@
+const querystring = require('querystring');
```
gauntface: Kill this as well.

gulp-tasks/test.js
jeffposnick: - There's a large corpus of existing `workbox-build` tests written using CommonJS imports that I want to import as part of this PR to ensure that the modified build steps didn't break anything that worked in v2.
- The Node-targeted packages, like `workbox-build`, are sticking with CommonJS imports and exports internally. (We decided that we wanted the source to work with a modern version of Node without transpilation, which means no ESM, yet.) There's no need to write the tests as `.mjs` since they'll be consuming CommonJS modules.

gulp-tasks/utils/package-runner.js
jeffposnick: There's still an underlying question in v3 of how we determine whether a given package targets the browser or Node—in v2, when we shipped build rules and tests in each package's directory, it wasn't ambiguous.

I'm cool with renaming the tasks and breaking them up into separate files, but I think the lingering question is whether you're on board with checking `engines.node` in `package.json` and using that to determine whether we're in Browser or Node "mode".

infra/utils/e2e-sw-validator.js
jeffposnick: I'd like to maintain parity with the v2 test suite in this PR. I know the test is problematic and the tests may change in the future, but I ended up finding some broken things that I wouldn't have otherwise noticed during the move by virtue of ensuring the existing tests worked.

infra/utils/validator/fake-run-and-compare.js
jeffposnick: See my previous comment about wanting to maintain testing parity in this PR.

infra/utils/validator/test-in-browser.js
jeffposnick: See my previous comment.

gulp-tasks/utils/constants.js
jeffposnick: Sure, that sounds like a good approach. I'll update the existing `package.json` files to reflect that they're targeted to the browser.

gulp-tasks/utils/build-browser-package.js
```diff
@@ -0,0 +1,178 @@
...
+ *   example: {
+ *     foo,
+ *   },
+ * }"
+ * @param {*} exportObject This is the object that needs to be converted
+ * to a string.
+ * @param {*} levels This value is just used for indenting the code to
+ * ensure the final output is human readable and easy to understand.
+ */
+const convertExportObject = (exportObject, levels = 0) => {
```
jeffposnick: @gauntface I couldn't tell if/where this was used, but I carried it over from your original build code. Happy to delete it if it's not needed.

gulp-tasks/build-packages.js
```diff
@@ -1,204 +1,59 @@
...
-  const joinString = `,\n${padding}`;
-  return `{\n${padding}${outputStrings.join(joinString)}\n${closePadding}}`;
-};
+  }
 
-gulp.task('build-packages', gulp.series([
+  // Our various builds all return gulp streams. Kick them off and merge the
+  // result into a single stream that will emit 'end' when they have completed.
+  // Use that to signal the overall task completion.
+  return eventStream.merge(buildStreams.map((func) => func()))
```
jeffposnick: It seems like there *should* be a more straightforward way doing this in `gulp` 4, but after banging my head for a while, this was the only thing that I could come up with that would wait for the end of all the streams that are returned by the build functions.

jeffposnick:
PTAL at the revised approach to determining the appropriate build process.

gulp-tasks/utils/build-browser-package.js
gauntface: Please delete - self reference meant it wasn't caught by eslint.

gulp-tasks/build-packages.js
```diff
@@ -1,204 +1,59 @@
...
-    constants.PACKAGE_BUILD_DIRNAME, constants.BROWSER_BUILD_DIRNAME);
-  return fs.remove(outputDirectory);
+    constants.PACKAGE_BUILD_DIRNAME);
+  return fse.remove(outputDirectory);
 };
 
 gulp.task('build-packages:clean', gulp.series(
-  packageRunnner('build-packages:clean', cleanPackages)
+    packageRunnner.wrapFunction(
+      'Cleaning Package',
```
gauntface: I've been trying to follow the pattern of using the gulp task name so it's clear from logs where this log is from: in this case ''build-packages:clean''

gulp-tasks/build-packages.js
```diff
@@ -1,204 +1,59 @@
...
-// i.e. we'll have a browser build for no NODE_ENV and one for 'production'
-// NODE_ENV and the same for sw and node tests.
-const packageBuilds = constants.BUILD_TYPES.map((buildType) => {
-  return packageRunnner('build-package', buildPackage, buildType);
-});
+gulp.task('build-packages:build', (done) => {
+  const buildStreams = [];
 
-gulp.task('build-packages:build', gulp.series(packageBuilds));
+  const nodePackages = packageRunnner.getPackages('node');
```
gauntface: It feels weird requiring us to get the packages and pass that around vs, vs `packageRunner.runOnBrowserPackages('<task>', buildNodePackage);`

gulp-tasks/build-packages.js
gauntface: If you seperate this out into seperate tasks it'll be cleaner.

```js
gulp.task('build-packages:build', gulp.parallel('build-browser-packages', 'build-node-packages'));
```

Then create `gulp-tasks/build-browser-packages.js` and `gulp-tasks/build-node-packages.js`

gauntface:
@jeffposnick Still some comments. I'm not sure why getting and passing around the packages is preferable over just saying runOnNodePackages and runOnBrowserPackages. Regarding the event stream stuff - seperating out into files and running tasks should simplify it.

If you want me have PR a change into your branch to get gulp happy I'm more than happy to.

gulp-tasks/build-packages.js
jeffposnick: There are tasks which need to be run regardless of which `packageType` we're dealing with, like `build-packages:clean`.

Having a method that could potentially return everything, but also could be filtered down when needed, made the most sense to me.

With your approach, we'd either end up with

```js
packageRunner.runOnBrowserPackages('build-packages:clean', cleanPackage);`
packageRunner.runOnNodePackages('build-packages:clean', cleanPackage);`
```

or creating a new `packageRunner.runOnNodeAndBrowserPackages()` method, neither of which I find appealing.

gauntface: How about `packagerunner.run('<task name>', 'node'|'browser'|'all', func)`

jeffposnick: If you're happy with that, we can go with it.

gulp-tasks/build-packages.js
jeffposnick: That seems to work!

jeffposnick:
I had to put in another change to get the revised `gulp.series()` to be happy when there's nothing to run, but I'm going to assume the LGTM stands.

PR #815, opened 2017-09-22
Uses babel-preset-env to transpile our bundled code

### LINKED ISSUE #812: Using `async`, `await` makes workbox without further transpilation break on Samsung Internet
See twitter discussion here: https://twitter.com/addyosmani/status/910631449753141253

@poshaughnessy noticed the starbucks PWA's SW isn't installing properly on Samsung Internet.

Turns out this is because of usage of `async` and `await`. 

**Library Affected**:
all

**Browser & Platform**:
Samsung Internet

Probably best to not use it in distributed versions if you expect people to be able to use it as is in Samsung Internet.

To repro, open https://preview.starbucks.com in inspect devices on samsung internet. It fails on line 80 of our SW which is the included Workbox code: https://preview.starbucks.com/serviceworker.js

WHich is the first usage of async/await.

jeffposnick:
R: @philipwalton  @addyosmani @gauntface
CC: @poshaughnessy @HenrikJoreteg 

Fixes #812 

I'm hopeful that this should lead to compatibility with Samsung Browser 5, as it's based on Chrome 51. Assuming this is merged, we can publish a pre-release tag of `v2.0.2-alpha` if folks want to test against that in the real world and confirm compatibility prior to the official release.

While we're touching the Babel configs, I took the opportunity to switch from the deprecated `babili` package to its current incarnation, `babel-minify`.

The final, gzipped payload size of the `workbox-sw.prod.js` bundle is 14,560 bytes after applying the transpilation, which is an increase of only ~260 bytes vs. the previous `2.0.1` bundle.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,16 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "JSONStream": {
+      "version": "1.3.1",
+      "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-1.3.1.tgz",
+      "integrity": "sha1-cH92HgHa6eFvG8+TcDt4xwlmV5o=",
+      "dev": true,
+      "requires": {
+        "jsonparse": "1.3.1",
+        "through": "2.3.8"
+      }
+    },
     "accepts": {
       "version": "1.3.3",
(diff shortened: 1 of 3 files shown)


HenrikJoreteg:
Thanks for getting this sorted out folks 🎉

PR #823, opened 2017-09-27
V3 precaching model

gauntface:
This adds checks to pre-caching to ensure that on a second install, only the updated entries are cached / updated.

Also updated the tests for the revision details model to be more accurate (Using actual PrecacheEntry models rather than a fake object mimicking the model

### DIFF
--- a/test/workbox-precaching/node/controllers/test-PrecacheController.mjs
+++ b/test/workbox-precaching/node/controllers/test-PrecacheController.mjs
@@ -326,8 +326,6 @@ describe(`[workbox-precaching] PrecacheController`, function() {
         expect(cachedResponse).to.exist;
       }
 
-      // TODO Check indexedDB entries
-
       // Make sure we print some debug info.
       expect(logStub.callCount).to.be.gt(0);
     });
@@ -446,8 +444,6 @@ describe(`[workbox-precaching] PrecacheController`, function() {
         expect(cachedResponse).to.exist;
       }
 
-      // TODO Check indexedDB entries
-
       // Make sure we print some debug info.
       expect(logStub.callCount).to.be.gt(0);
 
@@ -486,8 +482,6 @@ describe(`[workbox-precaching] PrecacheController`, function() {
         let cachedResponse = await cache.match(inputUrl);
         expect(cachedResponse).to.exist;
       }
-
-      // TODO Check indexedDB entries
 
       // Make sure we print some debug info.
       expect(logStub.callCount).to.be.gt(0);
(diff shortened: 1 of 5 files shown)


test/workbox-precaching/node/controllers/test-PrecacheController.mjs
```diff
@@ -398,5 +401,90 @@ describe(`[workbox-precaching] PrecacheController`, function() {
...
+
+      const updateInfo = await precacheControllerOne.install();
+      expect(updateInfo.updatedEntries.length).to.equal(cacheListOne.length);
+      expect(updateInfo.notUpdatedEntries.length).to.equal(0);
+
+      const cache = await caches.open(cacheNameProvider.getPrecacheName());
+      const keysOne = await cache.keys();
+      expect(keysOne.length).to.equal(cacheListOne.length);
+
+      for (let i = 0; i < cacheListOne.length; i++) {
```
jeffposnick: Here and later on, you could simplify to:

```js
const urls = cacheListOne.map((entry) => entry.url || entry);
await Promise.all(urls.map(async (url) => {
  const cachedResponse = await cache.match(url);
  expect(cachedResponse).to.exist;
}));

File: test/workbox-precaching/node/controllers/test-PrecacheController.mjs [hunk 2]
```diff
@@ -398,5 +401,90 @@ describe(`[workbox-precaching] PrecacheController`, function() {
...
+
+      // Reset as addToCacheList will log.
+      logStub.reset();
+
+      const updateInfoTwo = await precacheControllerTwo.install();
+      expect(updateInfoTwo.updatedEntries.length).to.equal(2);
+      expect(updateInfoTwo.notUpdatedEntries.length).to.equal(2);
+
+      const keysTwo = await cache.keys();
+      // +1 allows cache keys 'index.1234.html' and 'index.4321.html'
```
jeffposnick: I don't follow what this comment means—is this test potentially flaky, depending on whether `activate` has fired?

gauntface: The test is isn't flaky.

This test is to check install() method behavior, not activate() method behavior.

So at this stage both 'index.1234.html' and 'index.4321.html' would exist in the cache. 'index.1234.html' from the previous precache (which may still be used by current clients) and 'index.4321.html' from the current precache (which won't be used by anything until activate event).

The activate method (Needs to be added after this PR) will perform this cleanup.

PR #827, opened 2017-09-28
Doc sweep

gauntface:
Moving to the new template of devsite there were a few details that shouldn't be exposed.

This is just a PR to add a simple way to check with `npm run docs` and makes the problem methods / classes private.


gauntface:
Sigh. The number of times NPM run script behaves differently to just straight command line is ever increasing.

No on a JSDoc config. If complains of fs.stat, delete your node_modules and build files and try again (i.e. run `git clean -f -d`)

jeffposnick:
👀 👍

PR #828, opened 2017-09-28
V3 precaching minification

gauntface:
R: @jeffposnick @philipwalton 

This shouldn't be any functional changes, but should reduce the file size (I've tried to hold of off anything too crazy). But if I've over stepped the mark, please let me know.


addyosmani:
> packages/workbox-precaching/builds/browser/workbox-precaching.prod.js

Nice work on the 10% reduction in size!

packages/workbox-precaching/models/PrecachedDetailsModel.mjs
```diff
@@ -46,25 +48,25 @@ export default class PrecachedDetailsModel {
...
   /**
    * Add an entry to the details model.
    * @param {PrecacheEntry} precacheEntry
    */
-  async addEntry(precacheEntry) {
+  async _addEntry(precacheEntry) {
     const db = await this._getDb();
     await db.put(precacheEntry._entryId, {
-      revision: precacheEntry._revision,
+      [REVISON_FIELD]: precacheEntry._revision,
```
addyosmani: Is the switch to an array-keyed revision field necessary for the minification wins here?

gauntface: It's a small win, means 'revision' string previously entered twice or so can be replaced by a variable name and the variable name can be mangled to something like `a`.

gauntface:
@jeffposnick sadly there is not rule of thumb. It's very much just considering the output and seeing what (if any) changes can be made to enable uglify to mangle names.

PR #832, opened 2017-09-29
V3 precaching docs

gauntface:

Apologies for the number of changes.

- Moved `builds` back to `build` I shouldn't of changed the build directory from previous build
- Adding in JSDoc base template with command `gulp docs` that will build, start server (with browsersync) and watch for changes. This is largely a convenience for us. The DevSite template seems to fairly well represent the JSDoc theme. Hopefully this will be ok
- Added doc building to travis just to make sure we don't so anything that breaks docs (We had a subtle issue in a 2.0.X release. Smoke test like this would have caught it out).
- Done some re working of some of the existing docs. Some tidying up the output / module association, some just trimming of the explanations (We should be expanding in docs rather than reference docs).


gauntface:
@jeffposnick do you know if babel-runtime is needed in workbox-build? I don't think I touched any of that stuff, so not sure why Travis is suddenly upset :(

jeffposnick:
`babel-runtime` is needed for the transpiled code in `build/` when run in Node 4.

PR #833, opened 2017-10-01
First stage of v3 workbox build refactor

jeffposnick:
CC: @goldhand 

Hey All—This is the first stage of the `workbox-build` refactor for v3. The code as it is should work, and I've modified several of the interfaces to account for the expected use cases, including @goldhand's Webpack usage.

The PR is quite large, and I've gotten to a point where the majority is done and reviewable, so I wanted to file a PR now. Look for inline comments about anything that I plan on changing or points of interest.

The major missing pieces that I'm saving for a follow-up PR are:
- Revamped JSDocs.
- Reworking the way [`joi` option validations](https://www.npmjs.com/package/joi) cascade—I'm using some `class` inheritance right now, but it's clunky and doesn't support multiple parents in a chain.
- End-to-end tests for `entry-points/generare-sw.js`, which I just haven't gotten to yet.


infra/utils/validator/service-worker-runtime.js
```diff
@@ -0,0 +1,57 @@
+const assert = require('assert');
```
jeffposnick: I'm using this in the end-to-end tests to validate all of the service worker functionality—it can check individual method calls on `WorkboxSW` with specific parameters, etc.

It borrows a bit from @gauntface's original validation work using a `vm` instance and a fake context to run the actual code.

packages/workbox-build/src/entry-points/generate-sw-no-fs.js
```diff
@@ -0,0 +1,32 @@
...
+ * The ideal use case for this method is when you have a larger build system
+ * that has its own way of interacting with the filesystem (such as a Webpack).
+ *
+ * @param {Object} input
+ * @return {Promise<String>} A populated service worker template, based on the
+ * other configuration options provided.
+ *
+ * @memberof module:workbox-build
+ */
+async function generateSWNoFS(input) {
```
jeffposnick: This is one of the new interfaces, catered to @goldhand's use case in the Webpack plugin where both the template used as input and the final output be passed as strings, rather than use the filesystem.

The `generateSWNoFS()` reflects that it is similar to `generateSW()`, without using the filesystem for anything, but I'd imagine that name should change. (Any ideas?)

Apologies re: the lack of JSDocs right now, but you can see usage within the corresponding end to end test.

packages/workbox-build/src/entry-points/generate-sw.js
```diff
@@ -0,0 +1,74 @@
...
+ * contains the aggregate size of all the precached entries, in bytes, and the
+ * `count` property contains the total number of precached entries.
+ *
+ * @memberof module:workbox-build
+ */
+async function generateSW(input) {
+  const options = new GenerateSWOptions(input);
+
+  const destDirectory = path.dirname(options.swDest);
+  const buildType = (process.env.NODE_ENV &&
```
jeffposnick: This logic will cause the `dev` build of WorkboxSW (either from the CDN or locally copying it) to be used iff `NODE_ENV` starts with `'dev'`. It was very difficult for folks who were using `generateSW()` in v2 to opt-in to using the `dev` build, so hopefully this will make it easier for folks to benefit from the extra debugging in their own dev server.

If folks are concerned about this approach, then I can make it so that developers explicitly pass in an option to accomplish the same thing instead of inferring it.

packages/workbox-build/src/entry-points/generate-sw.js
```diff
@@ -0,0 +1,74 @@
...
+      pathToWorkboxSWFileRelativeToDest);
+  }
+
+  const {count, size, manifestEntries} = await getFileManifestEntries(options);
+
+  await writeServiceWorkerUsingDefaultTemplate(Object.assign({
+    manifestEntries,
+  }, options));
+
+  return {count, size};
```
jeffposnick: We're now returning the count and size of the manifest, so that builds tools using `generateSW()` can display it to their users.

packages/workbox-build/src/entry-points/get-manifest.js
```diff
@@ -0,0 +1,26 @@
...
+ * @return {Promise<{manifestEntries: Array<ManifestEntry>,
+ * count: Number, size: Number}>} A promise that resolves once the precache
+ * manifest is determined. The `size` property contains the aggregate size of
+ * all the precached entries, in bytes, the `count` property contains the total
+ * number of precached entries, and the `manifestEntries` property contains all
+ * the `ManifestEntry` items.
+ *
+ * @memberof module:workbox-build
+ */
+async function getManifest(input) {
```
jeffposnick: I've removed the old [`generateFileManifest()`](https://workboxjs.org/reference-docs/latest/module-workbox-build.html#.generateFileManifest) method in favor of this. It runs the same logic, but instead of writing the results to disk (and making the user choose between CJS and ESM format), it just returns the manifest as a variable, for added flexibility.

Folks who want the old behavior can still serialize that variable to disk in the format they prefer.

packages/workbox-build/src/entry-points/inject-manifest.js
```diff
@@ -0,0 +1,61 @@
...
+ * the service worker file has been written to `swDest`. The `size` property
+ * contains the aggregate size of all the precached entries, in bytes, and the
+ * `count` property contains the total number of precached entries.
+ *
+ * @memberof module:workbox-build
+ */
+async function injectManifest(input) {
+  const options = new InjectManifestOptions(input);
+
+  const globalRegexp = new RegExp(options.injectionPointRegexp, 'g');
```
jeffposnick: The injection point regexp is now configurable.

packages/workbox-build/src/entry-points/options/base-options.js
```diff
@@ -0,0 +1,56 @@
+const joi = require('joi');
```
jeffposnick: I'm offloading all parameter validation to `joi`, and running, it whenever one of the `entry-point/` modules are used, upfront, before any of the `lib/` modules are called.

As mentioned, the current approach of cascading/composed options is a little chunky, relying on class inheritance, so expect that to change in a follow-up PR. Custom error messages for specific validation failures are also to come in a follow-up.

If you have any comments about the overall approach to validation, usage of `joi`, etc. then that's fair game in this PR!

packages/workbox-build/src/index.js
```diff
@@ -1,11 +1,9 @@
...
-const injectManifest = require('./lib/inject-manifest');
+const generateSW = require('./entry-points/generate-sw');
+const generateSWNoFS = require('./entry-points/generate-sw-no-fs');
+const getManifest = require('./entry-points/get-manifest');
+const injectManifest = require('./entry-points/inject-manifest');
 
 /**
- * # workbox-build
- *
  * This Node module can be used to generate a list of assets that should be
```
jeffposnick: Revamped JSDocs will come in a future PR.

packages/workbox-build/src/lib/constants.js
```diff
@@ -1,10 +0,0 @@
-module.exports = {
```
jeffposnick: These have moved to `joi` defaults.

packages/workbox-build/src/lib/errors.js
```diff
@@ -32,7 +32,7 @@ module.exports = {
   'unable-to-copy-workbox-sw': ol`workbox-sw is needed by the service worker
```
jeffposnick: Some of these messages need to be pruned, as they're handled implicitly by `joi` validation.

packages/workbox-build/src/lib/use-build-type.js
```diff
@@ -0,0 +1,5 @@
+const DEFAULT_BUILD_TYPE = 'prod';
```
jeffposnick: This module doesn't necessarily have the most robust logic, and would need to change if we change our naming conventions.

packages/workbox-build/src/templates/sw.js.tmpl
```diff
@@ -0,0 +1,33 @@
+<% if (importScripts) { %>
+importScripts(<%= importScripts.map(JSON.stringify).join(',') %>);
```
jeffposnick: I've added support for passing in arbitrary `importScripts: Array<string>`, which I think should help @goldhand.

packages/workbox-build/src/templates/sw.js.tmpl
```diff
@@ -0,0 +1,33 @@
...
+ *    updated, old cache entries are left as is.
+ *
+ * By changing the file manifest manually, your users may end up not receiving
+ * new versions of files because the revision hasn't changed.
+ *
+ * Please use workbox-build, workbox-cli, workbox-webpack-plugin, or some other
+ * equivalent build tool to generate the precache manifest for you as part of
+ * your pre-deployment build process.
+ */
+self.__precacheManifest = <%= JSON.stringify(manifestEntries, null, 2) %>.concat(self.__precacheManifest || []);
```
jeffposnick: Another change to help @goldhand is that we now check for `self.__precacheManifest`, which can be set inside of a script that is pulled in via `importScripts()`.

`self.__precacheManifest` and `manifestEntries` can be used together if you want some entries set via `importScripts()`, and others set via a parameter to the template.

test/workbox-build/node/entry-points/get-manifest.js
```diff
@@ -0,0 +1,263 @@
...
+        }
+      });
+    }
+  });
+
+  describe('invalid parameter values', function() {
+    for (const param of SUPPORTED_PARAMS) {
+      it(`should reject with a ValidationError when '${param}' is null`, async function() {
+        const options = Object.assign({}, BASE_OPTIONS);
+        options[param] = null;
```
jeffposnick: This is not quite as long as @gauntface's original test cases, where most parameters were passed a variety of bad inputs, but I figure that `joi` validation should handle the nuance, and we can just test for a single failure.

test/workbox-build/node/lib/copy-workbox-sw.js
```diff
@@ -0,0 +1,46 @@
+const expect = require('chai').expect;
```
jeffposnick: Most of the logic in the unit tests have stayed the same, with the exception of removing the tests for bad parameters.

I'm assuming that all usage will go through one of the `entry-point/` modules, and they handle validation.

test/workbox-build/node/entry-points/inject-manifest.js
```diff
@@ -0,0 +1,247 @@
+const expect = require('chai').expect;
+const path = require('path');
+const tempy = require('tempy');
```
jeffposnick: I'm just using this whenever we need to write to disk, with the assumption that the operating system knows how to clean up temporary files.

infra/utils/validator/service-worker-runtime.js
```diff
@@ -0,0 +1,57 @@
...
+const expect = require('chai').expect;
+const fse = require('fs-extra');
+const makeServiceWorkerEnv = require('service-worker-mock');
+const sinon = require('sinon');
+const vm = require('vm');
+
+module.exports = async ({swFile, swCode, expectedMethodCalls}) => {
```
gauntface: This code will pretty much destroy the testing changes coming in because it'll override the SW env stuff. Not something actionable on this PR, but more an FYI.

packages/workbox-build/src/lib/populate-sw-template.js
```diff
@@ -0,0 +1,56 @@
...
+  directoryIndex,
+  handleFetch,
+  ignoreUrlParametersMatching,
+  importScripts,
+  manifestEntries,
+  navigateFallback,
+  navigateFallbackWhitelist,
+  runtimeCaching,
+  skipWaiting,
+  swTemplate,
```
gauntface: Why is this being passed in?

Couldn't this be a static string template in workbox-build that would be passed to lodash?

packages/workbox-build/src/entry-points/generate-sw-no-fs.js
jeffposnick: Sure, I'll go with `generateSWString()`.

test/workbox-build/node/entry-points/inject-manifest.js
jeffposnick: I wasn't, with that devil-may-care attitude inspired by https://github.com/sindresorhus/tempy#why-doesnt-it-have-a-cleanup-method:

> Why doesn't it have a cleanup method?
>The operating system will clean up when needed. No point in us wasting resources and adding complexity.

I could add it a cleanup step, but the one thing that makes me hesitate is that all that AppVeyor flakiness seemed to be tied to failed attempts to clean up what could have been treated as temporary files

infra/utils/validator/service-worker-runtime.js
jeffposnick: This code uses a handcrafted `context` object that's passed to  `vm.runInNewContext()` and I don't anticipate interference due to the `global` state from other tests.

The forthcoming `generate-sw.js` e2e test may make use of `NODE_ENV`, but it will be specifically set to the values expected prior to running the tests.

I can refactor the spy setup and validation into separate functions.

packages/workbox-build/src/lib/populate-sw-template.js
jeffposnick: My initial motivation for taking this in as a string was to avoid having to make a `fse.readFile()` call to read it in to memory, since that wouldn't work nicely with Webpack. And including

```js
const SW_TEMPLATE = `really long template string`;
```

inline within another module seems kind of messy.

After thinking about it some more, what I could do is change things so that the `swTemplate` is standalone in its own module file, that consists of

```js
module.exports = `really long template string`;
```

I think that would play nicely with Webpack and still let it be a standalone thing, and theoretically easy to test via `proxyquire()`. I'll make that change.

test/workbox-build/node/entry-points/get-manifest.js
```diff
@@ -0,0 +1,263 @@
...
+          expect(error.name).to.eql('ValidationError');
+          expect(error.details[0].context.key).to.eql(param);
+        }
+      });
+    }
+  });
+
+  describe('should generate a valid manifest when properly configured', function() {
+    it(`should use defaults when all the required parameters are present`, async function() {
+      const options = Object.assign({}, BASE_OPTIONS);
```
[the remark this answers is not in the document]
jeffposnick: The idea behind `BASE_OPTIONS` was to set up the minimum set of required options that could then be tweaked via `Object.assign()` in each test variation. Only `globDirectory` is a required option, and all the other options for this test case come from the defaults.

test/workbox-build/node/lib/populate-sw-template.js
```diff
@@ -0,0 +1,100 @@
...
+  });
+
+  it(`should pass the expected options to the template using mostly defaults`, function() {
+    const runtimeCachingPlaceholder = 'runtime-caching-placeholder';
+    const swTemplate = 'template';
+    const workboxOptionsString = '{}';
+
+    const innerStub = sinon.stub().returns('');
+    const outerStub = sinon.stub().returns(innerStub);
+    const populateSWTemplate = proxyquire(MODULE_PATH, {
```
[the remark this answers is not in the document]
jeffposnick: Sorry, just some hackery to account for the fact that there are two calls being made as part of the `lodash.template` API—one to convert the initial string into a template, which is itself a function that takes in parameters and returns the final string.

I'll clarify in a comment.

test/workbox-build/node/entry-points/get-manifest.js
```diff
@@ -0,0 +1,263 @@
...
+      }, {
+        url: 'url2',
+        revision: 'c154bc7cdfbfbfb73e23f853bd8fcec0',
+      }]);
+      expect(count).to.eql(8);
+      expect(size).to.eql(4790);
+    });
+
+    it(`should use defaults when all the required parameters, and 'manifestTransforms' are present`, async function() {
+      const transform1 = (entries) => entries.filter((entry) => {
```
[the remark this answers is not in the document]
jeffposnick: Done.

test/workbox-build/node/entry-points/inject-manifest.js
```diff
@@ -0,0 +1,247 @@
...
+          throw new Error('Unexpected success.');
+        } catch (error) {
+          expect(error.name).to.eql('ValidationError');
+          expect(error.details[0].context.key).to.eql(param);
+        }
+      });
+    }
+  });
+
+  describe(`should handle various runtime errors`, function() {
```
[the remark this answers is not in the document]
jeffposnick: Done.

jeffposnick:
Merging after addressing Matt's comments, as this should hopefully make @goldhand's work easier.

I'll update https://github.com/GoogleChrome/workbox/issues/803 with the current status and remaining TODOs, and I'm happy to address any additional feedback in follow-up PRs.

PR #838, opened 2017-10-03
Moving tests to individual processes. Coverage is broke and output is…

gauntface:
R: @jeffposnick @philipwalton 

Each test runs in it's own process now.


gulp-tasks/test.js
```diff
@@ -1,47 +1,69 @@
...
-  runNodeTestsWithEnv('dev'),
+  () => runNodeTestsWithEnv('dev'),
 ));
 
+gulp.task('test:node:clean', () => {
+  return fse.remove(path.join(__dirname, '..', '.nyc_output'));
+});
+
+gulp.task('test:node:coverage', () => {
+  return spawn(getNpmCmd(), ['run', 'coverage-report']);
```
jeffposnick: When reading up on `gulp` v4, I learned that there is native support for using a `spawn`ed process as a return value from a task: https://www.joezimjs.com/javascript/complete-guide-upgrading-gulp-4/#4-Return-a-Child-Process

I think we need to keep `./utils/spawn-promise-wrapper` around for use cases where there are multiple subprocesses, or cleanup needs to be done after the subprocess runs, but it should be usable here.

gulp-tasks/test.js
```diff
@@ -1,47 +1,69 @@
...
+    process.env.NODE_ENV = originalNodeEnv;
+
+    logHelper.error(err);
+    throw new Error(`[Workbox Error Msg] 'gulp test' discovered errors.`);
+  }
+};
+
+const runNodeTestsWithEnv = (nodeEnv) => {
+  const packagesToTest = glob.sync(`test/${global.packageOrStar}`);
+  return packagesToTest.reduce(async (promiseChain, packageToTest) => {
```
jeffposnick: One of the nicer parts of `async`/`await` [you don't have to use `reduce()` chains anymore](https://developers.google.com/web/fundamentals/primers/async-functions#example_outputting_fetches_in_order):

```js
const runNodeTestsWithEnv = async (nodeEnv) => {
  const packagesToTest = glob.sync(`test/${global.packageOrStar}`);
  for (const packageToTest of packagesToTest) {
    await runNodeTestSuite(packageToTest, nodeEnv);
  }
};
```

gulp-tasks/test.js
gauntface: Good to know.

gulp-tasks/test.js
```diff
@@ -1,47 +1,68 @@
...
-      options.push('--grep', global.cliOptions.grep);
-    }
-    const originalNodeEnv = process.env.NODE_ENV;
-
-    process.env.NODE_ENV = nodeEnv;
-    return spawn(getNpmCmd(), ['run', 'test', '--',
-      `./test/${global.packageOrStar}/node/**/*.{js,mjs}`,
+const runNodeTestSuite = async (testPath, nodeEnv) => {
+  logHelper.log(oneLine`
+  Running Node test on ${logHelper.highlight(testPath)}
```
philipwalton: The indentation looks messed up here.

PR #840, opened 2017-10-03
V3 precaching default

gauntface:

This is the last piece of precaching. Small helper API for precaching to manage the install and activate step.

Also fixed a small JSDoc issue where it wasn't using my local copy.

### DIFF
--- a/packages/workbox-precaching/controllers/PrecacheController.mjs
+++ b/packages/workbox-precaching/controllers/PrecacheController.mjs
@@ -32,7 +32,7 @@ class PrecacheController {
    * This method will add items to the precache list, removing duplicates
    * and ensuring the information is valid.
    *
-   * @param {Array<Object|String>} entries Array of entries to
+   * @param {Array<Object|string>} entries Array of entries to
    * precache.
    */
   addToCacheList(entries) {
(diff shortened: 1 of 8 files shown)


packages/workbox-precaching/controllers/PrecacheController.mjs
```diff
@@ -32,20 +32,20 @@ class PrecacheController {
    * This method will add items to the precache list, removing duplicates
    * and ensuring the information is valid.
    *
-   * @param {Array<Object|String>} userEntries Array of entries to
+   * @param {Array<Object|String>} entries Array of entries to
```
philipwalton: s/String/string/

packages/workbox-precaching/default-precaching-export.mjs
```diff
@@ -0,0 +1,30 @@
+import PrecacheController from './controllers/PrecacheController.mjs';
+
+const precacheController = new PrecacheController();
+let listenersAdded = false;
+
+const addListeners = () => {
+  listenersAdded = true;
+  self.addEventListener('install', () => precacheController.install());
```
jeffposnick: For both `install` and `activate`, you should either:
- pass in the corresponding `event` parameter to the `precacheController` method and call `event.waitUntil()` within that method
- or wrap the call to the `precacheController` methods in `event.waitUntil()` right here

Since both `precacheController.install()` and `.cleanup()` are `async` methods, it probably makes more sense to take the second approach:

```js
  event.waitUntil(precacheController.install());
});
```

gauntface: Such a noob mistake - thanks Jeff.

PR #842, opened 2017-10-03
V3 workbox-sw

gauntface:
R: @jeffposnick @philipwalton 

Start of workbox-sw.

### DIFF
--- a/packages/workbox-sw/index.mjs
+++ b/packages/workbox-sw/index.mjs
@@ -19,24 +19,30 @@ class WorkboxSW {
    */
   constructor(options) {
     this._options = Object.assign({
-      debug: location.hostname === 'localhost',
+      debug: self.location.hostname === 'localhost',
       pathPrefix: null,
     }, options);
   }
 
   /**
-   * Force a service worker to become active, instead of waiting, and
-   * to claim any available clients.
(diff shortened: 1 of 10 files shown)


packages/workbox-sw/index.mjs
```diff
@@ -0,0 +1,48 @@
...
+      debug: self.location.hostname === 'localhost',
+      pathPrefix: null,
+    }, options);
+  }
+
+  /**
+   * Force a service worker to become active, instead of waiting. This is
+   * normally used in conjunction with `clientsClaim()`.
+   */
+  skipWaiting() {
```
jeffposnick: Couple of points re: `skipWaiting()` and `clientsClaim()`:

- In v2 they were booleans passed to the `WorkboxSW` constructor. I'm curious what's behind the change to making them methods.
- Neither of them need to be wrapped in `event.waitUntil()`, since they just return promises that resolve immediately.

gauntface: @jeffposnick main reason was that the config was largely pulled in from sw-precache config and I didn't see any valid in reason in parsing the config, then using the config to call an internal function that did this vs, just getting the dev to call the function.

PR #846, opened 2017-10-04
Give a precache-specific Router precedence

### LINKED ISSUE #839: Runtime caching conflicts with precache
When runtime caching is applied to a page that is precached, that page won't work offline anymore.

This affects the workbox-sw library. Tested on Chrome 61 on MacOS.

The Progressive Web AMP codelab at https://codelabs.developers.google.com/codelabs/amp-pwa-workbox/ illustrates the issue. The homepage is precached, but if you navigate from any of the articles to the homepage while offline the homepage won't be displayed and the console shows the following error:
```
The FetchEvent for "http://localhost:8081/" resulted in a network error response: the promise was rejected.
Promise rejected (async)
_isListenerRegistered.self.addEventListener @ workbox-sw.dev.v2.0.0.js:71
```
If you visit the homepage after the SW is installed and before going offline, so the page is added to the runtime cache, the page will be displayed.

I also included a more basic test case without app shell:

[workbox-issue.zip](https://github.com/GoogleChrome/workbox/files/1352954/workbox-issue.zip)

Is this indeed a bug or is this expected behaviour and is it up to the developer to keep precached files in mind when adding routes?

jeffposnick:

Fixes #839

This change should lead to the cache-first route for precached assets always having precedence over runtime caching routes.

I think this sort of change is safe to roll out as part of a minor release, since while we are changing existing behavior, we're changing it in a way that fixes a previously introduced bug and brings us in line with our documented behavior.

### DIFF
--- a/packages/workbox-sw/src/lib/workbox-sw.js
+++ b/packages/workbox-sw/src/lib/workbox-sw.js
@@ -128,11 +128,20 @@ class WorkboxSW {
       cacheId,
     });
 
+    // Create a Router instance that's used by the `Route` for precached assets.
+    // See https://github.com/GoogleChrome/workbox/issues/839
+    this._precacheRouter = new Router(
+      this._revisionedCacheManager.getCacheName(),
+    );
+
     this._router = new Router(
       this._revisionedCacheManager.getCacheName(),
     );
 
     if (handleFetch) {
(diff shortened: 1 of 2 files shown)


gauntface:
Nasty but makes sense. :+1:

PR #847, opened 2017-10-05
V3 integration test start

gauntface:
R: @jeffposnick @philipwalton 

Far from perfect, but a start. Just wanted early feedback at the moment.

### DIFF
--- a/gulp-tasks/test-integration.js
+++ b/gulp-tasks/test-integration.js
@@ -1,0 +1,116 @@
+const gulp = require('gulp');
+const oneLine = require('common-tags').oneLine;
+const glob = require('glob');
+const seleniumAssistant = require('selenium-assistant');
+const path = require('path');
+const clearRequire = require('clear-require');
+
+// const spawn = require('./utils/spawn-promise-wrapper');
+// const getNpmCmd = require('./utils/get-npm-cmd');
+const logHelper = require('../infra/utils/log-helper');
+const testServer = require('../infra/testing/test-server');
+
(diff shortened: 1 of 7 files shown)


gulp-tasks/test-integration.js
```diff
@@ -0,0 +1,116 @@
...
+
+const runIntegrationForBrowser = async (browser) => {
+  const webdriver = await browser.getSeleniumDriver();
+
+  try {
+    const packagesToTest =
+      glob.sync(`test/${global.packageOrStar}/integration`);
+
+    for (const packageToTest of packagesToTest) {
+      for (const nodeEnv of ['dev', 'production']) {
```
jeffposnick: Could you read this from https://github.com/GoogleChrome/workbox/blob/v3/gulp-tasks/utils/constants.js#L24?

Otherwise I think we're going to end up changing `dev` to `development` at some point and everything is going to get out of sync.

test/workbox-precaching/integration/precache-and-update.js
```diff
@@ -0,0 +1,13 @@
+describe(`[workbox-precaching] Precache and Update`, function() {
```
jeffposnick: Do you want to put something about this being an Integration test in the `describe()` comment?

gauntface: It should be somewhat clear it's an integration test because of the gulp logs surrounding the mocha results.

If you think it'll be helpful happy to add a notion of `[workbox-precaching] {<Test Type>} <Test Title>` to our top level describe calls.

PR #849, opened 2017-10-05
V3 publishing tidy

gauntface:

I tried getting CDN and github publishing working and noticed a number of quirks and issues with publishing (was a bit of a black box).

This is a rejig of the code, number of issues getting fixed.


gulp-tasks/utils/publish-helpers.js
```diff
@@ -0,0 +1,179 @@
...
+
+const SOURCE_CODE_DIR = 'source-code';
+const GROUPED_BUILD_FILES = 'grouped-build-files';
+
+const doesDirectoryExist = async (directoryPath) => {
+  let stats = null;
+  try {
+    stats = await fs.stat(directoryPath);
+  } catch (err) {
+    // NOOP
```
jeffposnick:
```js
return false;
```

(So that you can safely call `stats.isDirectory()` later on.)

gulp-tasks/utils/publish-helpers.js
```diff
@@ -0,0 +1,179 @@
...
+  }
+  return stats !== null;
```
jeffposnick: Since you presumably want this to fail if `directoryPath` is a file, how about:

```js
return stats.isDirectory();
```

gulp-tasks/utils/publish-helpers.js
```diff
@@ -0,0 +1,179 @@
...
+};
+
+const getBuildPath = (tagName) => {
+  const tempReleasePath = path.join(
+    __dirname, '..', '..', constants.GENERATED_RELEASE_FILES_DIRNAME);
+  return path.join(tempReleasePath, tagName);
+};
+
+const downloadGitCommit = async (tagName, gitBranch) => {
+  if (!tagName) {
```
jeffposnick: Completely optional, but from Node code, I've just been using the built-in [`assert` module](https://nodejs.org/api/assert.html) for this sort of thing.

gulp-tasks/utils/publish-helpers.js
```diff
@@ -0,0 +1,179 @@
...
+    throw new Error(`You must provide a gitBranch to 'downloadGitCommit()`);
+  }
+
+  const sourceCodePath = path.join(getBuildPath(tagName), SOURCE_CODE_DIR);
+
+  logHelper.log(oneLine`
+    Download Git Commit ${logHelper.highlight(gitBranch)}.
+  `);
+
+  const dirExists = await doesDirectoryExist(sourceCodePath);
```
jeffposnick: I'm a little fuzzy on the logic here—`doesDirectoryExist()` takes in `sourceCodePath`, and `sourceCodePath` uniquely depends on `tagName`. But it doesn't look like `gitBranch` is being checked.

So it seems like you'd skip `git clone` when the directory for the `tagName` exists, even if it might not have `gitBranch` checked out. Could you run into issues if the directory doesn't have the branch checked out?

gulp-tasks/utils/publish-helpers.js
```diff
@@ -0,0 +1,179 @@
...
+
+    const pattern = path.posix.join(
+      sourceCodePath, 'packages', '**',
+      constants.PACKAGE_BUILD_DIRNAME, constants.BROWSER_BUILD_DIRNAME,
+      '*.{js,map}');
+
+    // Copy files from the source code and move into the grouped build
+    // directory. In others, have a flat file structure of just the built files.
+    const filesToIncludeInBundle = glob.sync(pattern);
+    for (let fileToInclude of filesToIncludeInBundle) {
```
jeffposnick: Switch this to use `await Promise.all()` so that the copies [aren't sequential](https://developers.google.com/web/fundamentals/primers/async-functions#careful_avoid_going_too_sequential):

```js
const copyPromises = files.map((file) => fs.copy(file, path.join(groupedBuildFiles, path.basename(file)));
await Promise.all(copyPromises);
```

gulp-tasks/utils/publish-helpers.js
```diff
@@ -0,0 +1,179 @@
...
+
+  const groupedBuildFiles =
+    path.join(getBuildPath(tagName), GROUPED_BUILD_FILES);
+
+  await new Promise((resolve, reject) => {
+    const writeStream = fs.createWriteStream(outputFilePath);
+    writeStream.on('close', () => resolve(outputFilePath));
+
+    const archive = archiver(format, options);
+    archive.on('error', (err) => {
```
jeffposnick: Nit: this is equivalent to

```js
archive.on('error', reject);
```

gulp-tasks/utils/publish-helpers.js
```diff
@@ -0,0 +1,179 @@
...
+    archive.pipe(writeStream);
+    // Adds the directory contents to the zip.
+    archive.directory(groupedBuildFiles, false);
+    archive.finalize();
+  });
+
+  return outputFilePath;
+};
+
+const createTarGz = async (tagName) => {
```
jeffposnick: This doesn't have to be marked `async`.

gulp-tasks/utils/publish-helpers.js
```diff
@@ -0,0 +1,179 @@
...
+const createZip = (tagName) => {
+  return createArchive(tagName, 'zip', 'zip', {
+    zlib: {level: 9},
+  });
+};
+
+const createBundles = async (tagName, gitBranch) => {
+  await groupBuildFiles(tagName, gitBranch);
+
+  const tarPath = await createTarGz(tagName);
```
jeffposnick: If you want to run both archive creations [in parallel](https://developers.google.com/web/fundamentals/primers/async-functions#careful_avoid_going_too_sequential):

```js
const [tarPath, zipPath] = await Promise.all([createTarGz(tagName), createZip(tagName)]);
return {tarPath, zipPath};
```

gulp-tasks/utils/publish-helpers.js
gauntface: This directory should only ever be created by this code. The only scenario where what you've mentioned has occurred if someone / something created the directory in the expected location.

gulp-tasks/utils/publish-helpers.js
gauntface: I did this originally and hit some weird issues.

gulp-tasks/utils/publish-helpers.js
gauntface: To be honest, I want the publish process to be as sane and reduce any errors as possible so would prefer to keep this sequential just in that vain (plus it's relatively quick anyway).

PR #850, opened 2017-10-06
V3 Precaching (Steps to integration test)

gauntface:

PR started to grow in the number of places I was touching so figured I'd stop at this point to get review.

- Adds `gulp test-server` which runs the test server used for integration.
- Precaching tidy up to make messages a little easier to parse (run `gulp build && gulp test-server` and go to [http://localhost:3004/test/workbox-precaching/static/precache-and-update/](http://localhost:3004/test/workbox-precaching/static/precache-and-update/) Then play with clear data, refreshing, unregister sw and refresh.
- Adds a method to test server that makes it possible to reliably get module build code and behind the scenes gets dev or prod builds based on NODE_ENV. This also works with source maps. 
- Restricts tests to Chrome and Firefox.


infra/testing/test-server.js
```diff
@@ -1,19 +1,51 @@
...
 const app = express();
 
-app.get('/', (req, res) => {
-  res.send('Hello World!');
+app.get(/\/__WORKBOX\/buildFile\/(workbox-[A-z]*)(\.(?:dev|prod)\.(.*))*/, (req, res) => {
+  const moduleName = req.params[0];
+  const fileExtension = req.params[2];
+
+  const modulePath = path.join(__dirname, '..', '..', 'packages', moduleName);
+  const pkg = require(
```
jeffposnick: As far as I can tell, `require()` always uses paths separated with `/`, even under Windows. (Correct me if I'm wrong.)

So you can either just hardcode the `/` or use `path.posix.join()`.

infra/testing/test-server.js
```diff
@@ -1,19 +1,51 @@
...
+    path.join(modulePath, 'package.json')
+  );
+
+  const browserFile = pkg.browser;
+  const libraryPath = path.dirname(path.join(modulePath, browserFile));
+  let libraryFileName = path.basename(browserFile);
+  switch (process.env.NODE_ENV) {
+    case 'dev':
```
jeffposnick: Can you read this from https://github.com/GoogleChrome/workbox/blob/v3/gulp-tasks/utils/constants.js#L24 instead of hardcoding the possible values?

gauntface: Nope. it's `dev` and `prod` for file names instead of `dev` and `production` and even then not idea how I'd reference that in a case statement like this unless I changed structure to an object in constants which might fix both of these issues.

jeffposnick: Fine to keep as-is for now.

Let's chat about it in person, because I think we need to expose more info about the environment constants and library build names somewhere. I'm bumping up against this with `workbox-build` tests, too.

jeffposnick: See https://github.com/GoogleChrome/workbox/issues/851

Issue #848, opened 2017-10-05
Only accept functions as handlers when creating Routes


jeffposnick:
**Library Affected**:
`workbox-routing`

A brief history of the `handler` parameter that's passed to the `Route` constructor:
- Initially, it had to be an Object that had a property named `handle`, and that property was a function. This played nicely with our `workbox-runtime-caching` classes, but was a pain for anyone who wanted to write their own handler logic.
- We switched things up to support either Objects with `handle` properties **or** functions, but that led to some [messy and hard to explain code](https://github.com/GoogleChrome/workbox/blob/8dfda82f4ba8fbd5973f33425360ee3c4e250570/packages/workbox-routing/src/lib/normalize-handler.js) that normalized the input.

I'm thinking that for v3, we can just drop support for Objects with `handle` properties and only take in functions.

We could retrofit the various `workbox-runtime-caching` subclasses by adding in a factory method on the parent [`Handler` class](https://github.com/GoogleChrome/workbox/blob/8dfda82f4ba8fbd5973f33425360ee3c4e250570/packages/workbox-runtime-caching/src/lib/handler.js) that would return a function that uses [`bind()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind) to wrap the `Handler.handle()` method, which will ensure that `this` can still be used.

Thoughts?

gauntface:
I've been thinking the same thing after working on the runtime-caching so would love for this to be the case, would also make the `Route` a little nicer passing in two functions and seeing two functions on the prototype.

gauntface:
Also gonna steal assignment if you don't mind ;)

jeffposnick:
Any reason to try to sneak this into the v3 release? It still could be worthwhile, but we can save some breaking changes for v4 😄

gauntface:
Yeah I'm pro that - precaching felt a bit close to add those changes, so I'm down for that.

jeffposnick:
We've decided not to make this change.

PR #854, opened 2017-10-06
v3 Router class

jeffposnick:

This adds in the `Router` class to v3's `workbox-routing`, and uses a freshly initialized `Router` instance as the default export for the module, along with unit tests.

(https://github.com/GoogleChrome/workbox/issues/848 has some discussions about what kind of parameters we want to accept when creating a `Route`. That has some tangential bearing on this PR, but it's mainly focused on `Router` behavior, so I think this can be dealt with independently.)


infra/utils/sw-env.js
```diff
@@ -18,6 +18,17 @@ const swEnv = makeServiceWorkerEnv();
 // This is needed to ensure new URL('/', location), works.
 swEnv.location = 'https://example.com';
 
+// See https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent/FetchEvent
```
jeffposnick: I know @philipwalton bumped into this as well, but while the `FetchEvent` is mocked in https://github.com/pinterest/service-workers/blob/master/packages/service-worker-mock/utils/events.js it isn't exported from there, and I needed it to be defined in the global scope in order to properly run the tests.

If this isn't the best practice for getting it defined globally, I'm happy to adjust.

packages/workbox-routing/default-export.mjs
```diff
@@ -0,0 +1,6 @@
+import Router from './lib/Router.mjs';
```
jeffposnick: @gauntface is this how default exports are supposed to work? I was a little unsure about how this should be set up.

infra/testing/sw-env.js
```diff
@@ -24,6 +24,17 @@ const swEnv = makeServiceWorkerEnv();
 // This is needed to ensure new URL('/', location), works.
 swEnv.location = 'https://example.com';
 
+// See https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent/FetchEvent
+class FetchEvent {
```
gauntface: Nit: could you move this out into mock-fetch-event.js and then import the FetchEvent class and add it to global or swEnv. Just helps seperate the mock logic from what mocks are actually being used.

packages/workbox-core/models/messages/messages.mjs
```diff
@@ -42,6 +42,17 @@ export default {
       `${expectedType}.`;
   },
 
+  'incorrect-class': ({expectedClass, paramName, moduleName, className,
+                       funcName}) => {
+    if (!expectedClass || !paramName || !moduleName || !className ||
+      !funcName) {
+      throw new Error(`Unexpected input to 'not-of-type' error.`);
```
gauntface: Tweak the code to 'incorrect-class'

packages/workbox-core/models/messages/messages.mjs
```diff
@@ -93,4 +104,25 @@ export default {
       `setCacheDeatils({${cacheNameId}: '...'}). Received a value of ` +
       `'${JSON.stringify(value)}'`;
   },
+
+  'unregister-routes-no-routes-found-for-method': ({method}) => {
```
gauntface: This error's code is a little confusing.

`unregister-route-but-not-found-with-method` would be better IMO.

packages/workbox-core/models/messages/messages.mjs
```diff
@@ -93,4 +104,25 @@ export default {
...
+    if (!method) {
+      throw new Error(`Expected a 'method' for error ` +
+        `'unregister-routes-no-routes-found-for-method'`);
+    }
+
+    return `One of the routes you're trying to unregister was not previously ` +
+      `registered. It has a method of ${method}, and there are no routes ` +
+      `registered with that method.`;
```
gauntface: > One of the routes you're trying to unregister was not previously registered for the method type '${method}'.

infra/utils/sw-env.js
zackargyle: This seems like it should be handled by the mock lib. Just put this up for you. Thoughts?
[https://github.com/pinterest/service-workers/pull/50](https://github.com/pinterest/service-workers/pull/50)

packages/workbox-routing/default-export.mjs
gauntface: Yeah this looks like it should would.

I assume you'll want the Router publicly available too so it should be exported with `export {Router};`.

Developers could then:
```javascript
const {Router} from 'workbox-routing';
const routing from 'workbox-routing';
```

packages/workbox-routing/default-export.mjs
```diff
@@ -0,0 +1,6 @@
...
+
+const router = new Router();
+router.addFetchListener();
```
gauntface: We mentioned in the last meeting that no lower level classes would register event listeners instead the default exports would do that and the classes would have API's that we make it simple to use.

So for this I'd expect something along the lines of:

```javascript
self.addEventListener('fetch', (event) => router.handleFetch(event));
```
For the most part this is fine, the only down side is that you may only want to add the event listener if the user uses Router, rather than if they just import the module (which would likely be the case if they wanted to use the Router and not the default class).

packages/workbox-routing/default-export.mjs
gauntface: Yes, this is what I've done for precaching and I think it works. But obviously open to alternatives as long as the end result for developers is sane.

infra/testing/sw-env.js
zackargyle: Just put this together for you all. Adds all of the constructors to the global scope which should help for any instanceof checks you do.
[https://github.com/pinterest/service-workers/pull/50](https://github.com/pinterest/service-workers/pull/50)

packages/workbox-routing/lib/Router.mjs
```diff
@@ -0,0 +1,328 @@
...
+
+import core from 'workbox-core';
+import {_private} from 'workbox-core';
+
+import normalizeHandler from './normalizeHandler.mjs';
+
+/**
+ * The Router takes one or more [Routes]{@link Route} and allows you to apply
+ * that routing logic to determine the appropriate way of handling requests
+ * inside of a service worker.
```
gauntface: Alternative simplification: 
> The Router takes one or more Routes and passes each fetch event through it's routing logic to determine the appropriate way to respond with a Request.

packages/workbox-routing/lib/Router.mjs
```diff
@@ -0,0 +1,328 @@
...
+ *
+ * It also allows you to define a "default" handler that applies to any requests
+ * that don't explicitly match a `Route`, and a "catch" handler that responds
+ * to any requests that throw an exception while being routed.
```
gauntface: > If no route matches a given a request, the Router will use a "default" handler if one is defined.
>
> Should any of the Route's throw an error, you can define a "catch" handler to gracefully deal with these issues and respond with a Request.

packages/workbox-routing/lib/Router.mjs
```diff
@@ -0,0 +1,328 @@
...
+ *
+ * You can use the `handleRequest()` method to pass a `FetchEvent` through the
+ * router and ultimately get a "routed response" back.
+ * If you'd like this to be handled automatically for you, calling
+ * `addFetchListener()` will cause the `Router` to respond to `fetch` events.
```
gauntface: Two cents: I'd drop this in favor of explaining in the respective method descriptions.

packages/workbox-routing/lib/Router.mjs
```diff
@@ -0,0 +1,328 @@
...
+ *
+ * If a request matches multiple routes, precedence will be given to the
```
gauntface: 👍

packages/workbox-routing/lib/Router.mjs
```diff
@@ -0,0 +1,328 @@
...
+  /**
+   * This will register a `fetch` event listener on your behalf which will check
+   * the incoming request to see if there's a matching route, and only respond
+   * if there is a match.
+   *
+   * @return {boolean} Returns `true` if this is the first time the method is
+   * called and the listener was registered. Returns `false` if called again,
+   * as the listener will only be registered once.
+   */
+  addFetchListener() {
```
gauntface: As I mentioned above - in the last meeting we were going to keep event listeners in the default export classes and keep classes free of these methods.

packages/workbox-routing/lib/Router.mjs
```diff
@@ -0,0 +1,328 @@
...
+      this._isListenerRegistered = true;
+      self.addEventListener('fetch', (event) => {
+        const responsePromise = this.handleRequest(event);
+        if (responsePromise) {
+          event.respondWith(responsePromise);
+        }
+      });
+      return true;
+    } else {
+      _private.logger.warn(
```
gauntface: This log feels optional and could be stripped from prod.

packages/workbox-routing/lib/Router.mjs
```diff
@@ -0,0 +1,328 @@
...
+   *
+   * @return {boolean} Returns `true` if this is the first time the method is
+   * called and the listener was registered. Returns `false` if called again,
+   * as the listener will only be registered once.
+   */
+  addFetchListener() {
+    if (!this._isListenerRegistered) {
+      this._isListenerRegistered = true;
+      self.addEventListener('fetch', (event) => {
+        const responsePromise = this.handleRequest(event);
```
gauntface: Is handleRequest synchronous? i.e. we know if a Route is going to handle this and so we don't need to call event.waitUntil?

packages/workbox-routing/lib/Router.mjs
```diff
@@ -0,0 +1,328 @@
...
+    } else {
+      _private.logger.warn(
+        `addFetchListener() has already been called for this Router.`);
+      return false;
+    }
+  }
+
+  /**
+   * This can be used to apply the routing rules to generate a response for a
+   * given request inside your own `fetch` event handler.
```
gauntface: > Apply the routing rules to a FetchEvent object to get a Response from an appropriate Route or handler.

packages/workbox-routing/lib/Router.mjs
```diff
@@ -0,0 +1,328 @@
...
+        moduleName: 'workbox-routing',
+        className: 'Router',
+        funcName: 'handleRequest',
+        paramName: 'event',
+      });
+    }
+
+    const url = new URL(event.request.url);
+    if (!url.protocol.startsWith('http')) {
+      _private.logger.log(`The URL '${url}' does not start with 'http', so ` +
```
gauntface: Do you think this should be in prod? I'm on the fence.

packages/workbox-routing/lib/Router.mjs
```diff
@@ -0,0 +1,328 @@
...
+
+    let {handler, params} = this._findHandlerAndParams(event, url);
+
+    // If we don't have a handler because there was no matching route, then
+    // fall back to defaultHandler if that's defined.
+    if (!handler && this.defaultHandler) {
+      handler = this.defaultHandler;
+    }
+
+    if (handler) {
```
gauntface: Nit: I'd invert this to a !handler check and return undefined early, then un-nest the following logic.

infra/testing/sw-env.js
zackargyle: Landed. If you update your service-worker-mock dep to @1.5.0 you can remove this and the manual `swEnv.ServiceWorkerGlobalScope` below.

packages/workbox-routing/lib/Router.mjs
```diff
@@ -0,0 +1,328 @@
...
+  /**
+   * An optional `handler` that's called by default when no routes
+   * explicitly match the incoming request.
+   *
+   * If the default is not provided, unmatched requests will go against the
+   * network as if there were no service worker present.
+   *
+   * @param {function|module:workbox-runtime-caching.Handler} handler
+   * This parameter can be either a function or an object which is a subclass
+   * of `Handler`.
```
gauntface: I'd drop the subclass `Handler` bit and just say "or an object with a `handle` function. See [Handler Interface for details](<link>)."

packages/workbox-routing/lib/Router.mjs
```diff
@@ -0,0 +1,328 @@
...
+   *
+   * Either option should result in a `Response` that the `Route` can use to
+   * handle the `fetch` event.
```
gauntface: Nit: "Either option should return a `Response` object, which will be used to respond to the `fetch` event."

packages/workbox-routing/lib/Router.mjs
```diff
@@ -0,0 +1,328 @@
...
+   *
+   * See [handlerCallback]{@link module:workbox-routing.Route~handlerCallback}
+   * for full details on using a callback function as the `handler`.
```
gauntface: I would drop that and either add it as the type of the handler or link it up early on in the doc.

packages/workbox-routing/lib/Router.mjs
```diff
@@ -0,0 +1,328 @@
...
+    this.catchHandler = normalizeHandler(handler);
+  }
+
+  /**
+   * Registers an array of routes with the router.
+   *
+   * @param {Array<module:workbox-routing.Route>} routes An array of routes to
+   * register.
+   */
+  registerRoutes(routes) {
```
gauntface: Is there any way we could see if anyone is using the registerRoutes method vs just using the singular registerRoute method.

Just think of code size.

test/workbox-routing/node/Route.mjs
```diff
@@ -91,43 +91,43 @@ describe(`workbox-routing: Route`, function() {
   });
 
   it(`should use the method provided when called with a valid method in dev`, function() {
-    if (process.env.NODE_ENV == 'production') return this.skip();
+    if (process.env.NODE_ENV === 'production') return this.skip();
 
     const route = new Route(match, handler, method);
-    expect(route._method).to.equal(method);
+    expect(route.method).to.equal(method);
```
gauntface: If you use `.method` the minifier won't be able to manipulate the value. Do we need it to be `.method` and public facing?

test/workbox-routing/node/Router.mjs
```diff
@@ -0,0 +1,302 @@
...
+describe(`[workbox-routing] Router`, function() {
+  const sandbox = sinon.sandbox.create();
+  const {logger} = _private;
+  const MATCH = () => {};
+  const HANDLER = () => {};
+  const METHOD = 'POST';
+  const EXPECTED_RESPONSE_BODY = 'test body';
+
+  beforeEach(async function() {
+    clearRequire.all();
```
gauntface: Is this needed?

test/workbox-routing/node/Router.mjs
```diff
@@ -0,0 +1,302 @@
...
+    // Run this in the `beforeEach` hook as well as the afterEach hook due to
+    // a mocha bug where `afterEach` hooks aren't run for skipped tests.
+    // https://github.com/mochajs/mocha/issues/2546
+    sandbox.restore();
+    // Prevent logs in the mocha output.
+    sandbox.stub(logger, 'warn');
+    sandbox.stub(logger, 'log');
+  });
+
+  afterEach(function() {
```
gauntface: Change this to `after()` . I realised in some of my recent tests afterEach didn't run after the test suite had finished meaning stubs were lingering. and afterEach isn't needed if beforeEach alway cleans stubs.

test/workbox-routing/node/Router.mjs
```diff
@@ -0,0 +1,302 @@
...
+
+  describe(`constructor`, function() {
+    it(`should construct without any inputs`, function() {
+      expect(() => {
+        new Router();
+      }).to.not.throw();
+    });
+  });
+
+  describe(`calling addFetchListener()`, function() {
```
gauntface: Can we stick the format of:

- top level describe [<module name>] <Class name>
    - describe <function name>
        - it tests
    - describe <function name>
        - it tests

It something Phil started calling out in reviews and I think it's a nice / good pattern to follow.

test/workbox-routing/node/Router.mjs
```diff
@@ -0,0 +1,302 @@
...
+      const router = new Router();
+
+      await expectError(
+        () => router.registerRoute({}),
+        'missing-a-method',
+        (error) => {
+          expect(error.details).to.have.property('moduleName').that.eql('workbox-routing');
+          expect(error.details).to.have.property('className').that.eql('Router');
+          expect(error.details).to.have.property('funcName').that.eql('registerRoute');
+          expect(error.details).to.have.property('paramName').that.eql('route');
```
gauntface: Is there any way to expect a specific method name to be missing (Just thinking if I got this error and it didn't call out the method that was missing, it could be frustrating).

test/workbox-routing/node/Router.mjs
```diff
@@ -0,0 +1,302 @@
...
+        }
+      );
+
+      await expectError(
```
gauntface: Nit: Could you break these out into seperate tests.

test/workbox-routing/node/Router.mjs
```diff
@@ -0,0 +1,302 @@
...
+        () => router.registerRoute({match: MATCH, method: METHOD}),
+        'incorrect-type',
```
gauntface: Not sure why this is an incorrect type. I was expecting a missing method because there is no handle method.

test/workbox-routing/node/Router.mjs
```diff
@@ -0,0 +1,302 @@
...
+          expect(error.details).to.have.property('moduleName').that.eql('workbox-routing');
+          expect(error.details).to.have.property('className').that.eql('Router');
+          expect(error.details).to.have.property('funcName').that.eql('registerRoute');
+          expect(error.details).to.have.property('paramName').that.eql('route');
+        }
+      );
+
+      await expectError(
+        () => router.registerRoute({match: MATCH, handle: HANDLER}),
+        'incorrect-type',
```
gauntface: Maybe pull these tests apart into missing-a-method and incorrect-type groups and use the generate-test-variants helper to reduce boilerplate.

test/workbox-routing/node/Router.mjs
```diff
@@ -0,0 +1,302 @@
...
+
+      // Routes without an explicit method will default to GET.
+      const getRoute1 = new Route(MATCH, HANDLER);
+      const getRoute2 = new Route(MATCH, HANDLER, 'GET');
+      const putRoute1 = new Route(MATCH, HANDLER, 'PUT');
+      const putRoute2 = new Route(MATCH, HANDLER, 'PUT');
+      // We support passing in Objects that match the expected interface in addition to Routes.
+      const postRoute = {
+        match: MATCH,
+        handler: {handle: HANDLER},
```
gauntface: Can handler just be a function here or does it *have* to be wrapped like this?

Would love it if it didn't need to be wrapped.

test/workbox-routing/node/Router.mjs
```diff
@@ -0,0 +1,302 @@
...
+
+      router.unregisterRoutes([getRoute2]);
+      router.unregisterRoute(putRoute2);
+
+      expect(router._routes.get('GET')).to.have.members([getRoute1]);
+      expect(router._routes.get('PUT')).to.have.members([putRoute1]);
+    });
+  });
+
+  describe(`setting default and catch handlers`, function() {
```
gauntface: Could we seperate out setDefaultHandler() and setCatchHandler and inline the handling requests piece into these relevant function calls.

test/workbox-routing/node/Router.mjs
```diff
@@ -0,0 +1,302 @@
...
+      const router = new Router();
+      const route = new Route(
+        () => false,
+        () => new Response(),
+      );
+      router.registerRoute(route);
+
+      // route.match() always returns false, so the Request details don't matter.
+      const event = new FetchEvent('fetch', {request: new Request(self.location)});
+      const response = await router.handleRequest(event);
```
gauntface: Here you are await'ing but I don't think you were in the fetch event listener.

infra/testing/sw-env.js
jeffposnick: I've switched to using https://github.com/pinterest/service-workers/pull/50

packages/workbox-routing/default-export.mjs
jeffposnick: I'm confused. This is how things are implemented in this PR:

- The lower-level `Router` class doesn't register any listeners by default. Instead, it exposes an `addFetchListener()` method that needs to be explicitly called.
- The default export initializes a `Router` and calls `addFetchListener()`.

Is your comment about there being an unintended side effect of importing the module? I.e. even if you don't plan on using the `default` export, it ends up unconditionally executing the code that creates the `Router` and establishes an event listener?

Assuming that's the case, that's inadvertent on my part, and mainly has to do with me not understanding how the default exports end up working in all of the module usage scenarios.

Let me know and I'll rethink the logic here.

packages/workbox-routing/lib/Router.mjs
jeffposnick: Oh... okay. Ignore my earlier comment.

packages/workbox-routing/default-export.mjs
jeffposnick: Sorry, just realized what you meant from reading your other comment. I'll update the code.

packages/workbox-routing/lib/Router.mjs
jeffposnick: Yup, `handleRequest` is synchronous.

It will return a `Promise<Response>` if there's a matching route, or a default route, but it doesn't do anything like `await` or call `.then()` on the promise.

packages/workbox-routing/lib/Router.mjs
jeffposnick: Removed from prod.

packages/workbox-routing/lib/Router.mjs
jeffposnick: Removed from prod.

packages/workbox-routing/lib/Router.mjs
jeffposnick: I'm going to keep the logic as-is, as it reflects there being a few possible states:
- `handler` is set by a matching route
- `handler` is set by `defaultHandler`
  - `catchHandler` is set
  - `catchhandler` isn't set
- `handler` isn't set

test/workbox-routing/node/Route.mjs
jeffposnick: I can switch.

packages/workbox-routing/lib/Router.mjs
jeffposnick: If you'd like to drop the multiple use case, we can do it now that we're making breaking changes.

jeffposnick: https://github.com/GoogleChrome/workbox/issues/856

test/workbox-routing/node/Router.mjs
jeffposnick: I could see how that would be confusing. Thanks for pointing that out.

test/workbox-routing/node/Router.mjs
jeffposnick: See https://github.com/GoogleChrome/workbox/issues/848

I'd like to leave as-is for now and then come to a decision on that issue, as changing things will involve a bunch of other cascading changes that will further muddy this PR.

test/workbox-routing/node/Router.mjs
jeffposnick: Yeah, I'll switch.

test/workbox-routing/node/Router.mjs
jeffposnick: I've switched to generate-test-variants

packages/workbox-routing/lib/Router.mjs
gauntface: ok. It's a personal preference. I always favor early return than needless nesting.

packages/workbox-core/models/messages/messages.mjs
```diff
@@ -93,4 +104,24 @@ export default {
...
       `'${JSON.stringify(value)}'`;
   },
+
+  'unregister-route-but-not-found-with-method': ({route, method}) => {
+    if (!method || !route) {
+      throw new Error(`Unexpected input to ` +
+        `'unregister-route-but-not-found-with-method' error.`);
+    }
+
+    return `The route you're trying to unregister, ${route}, was not ` +
```
gauntface: Not blocking, but it's unclear with `route` is so I'm unsure what this would print. Does the string value of `route` always produce something meaningful?

packages/workbox-routing/lib/Route.mjs
```diff
@@ -103,8 +103,8 @@ class Route {
       }
     }
 
-    this._handler = normalizeHandler(handler);
-    this._match = match;
+    this.handler = normalizeHandler(handler);
```
gauntface: Why can't this be `_handler` and `_match` sorry if you answered in previous PR

packages/workbox-routing/lib/Router.mjs
```diff
@@ -0,0 +1,248 @@
...
+        className: 'Router',
+        funcName: 'handleRequest',
+        paramName: 'event',
+      });
+    }
+
+    const url = new URL(event.request.url);
+    if (!url.protocol.startsWith('http')) {
+      if (process.env.NODE_ENV !== 'production') {
+        _private.logger.log(`The URL '${url}' does not start with 'http', so ` +
```
gauntface: Should this be an error / warning?

packages/workbox-routing/lib/Router.mjs
```diff
@@ -0,0 +1,248 @@
...
+          `it can't be handled.`);
+      }
+      return;
+    }
+
+    let {handler, params} = this._findHandlerAndParams(event, url);
```
gauntface: Not blocking. I expected this to be a method more like, `_routeRequest` that could find a Route, call it's handler and return a Response rather than break apart the Route to pass around it's callback like this.

packages/workbox-routing/lib/Router.mjs
```diff
@@ -0,0 +1,248 @@
...
+   * This parameter can be either a function or an object with a `handle`
+   * function. See the
+   * [Handler interface]{@link module:workbox-routing.Route~handlerCallback}
+   * for details.
+   *
+   * Either option should result in a `Response` object, which will be used to
+   * respond to the `fetch` event.
+   */
+  setDefaultHandler(handler) {
+    this.defaultHandler = normalizeHandler(handler);
```
gauntface: Why not make this `_defaultHandler`, otherwise the minification can't take affect.

packages/workbox-routing/lib/Router.mjs
```diff
@@ -0,0 +1,248 @@
...
+        responsePromise = responsePromise.catch((error) => {
+          return this.catchHandler.handle({url, event, error});
+        });
+      }
+      return responsePromise;
+    }
+  }
+
+  /**
+   * Checks the incoming event.request against the registered routes, and if
```
gauntface: nit: Add back ticks to `event.request`

packages/workbox-routing/lib/Router.mjs
```diff
@@ -0,0 +1,248 @@
...
+   * @return {Object} Returns an object with `handler` and `params` properties.
+   * They are populated if a matching route was found or `undefined` otherwise.
+   * @private
+   */
+  _findHandlerAndParams(event, url) {
+    const routes = this._routes.get(event.request.method) || [];
+    for (const route of routes) {
+      let matchResult = route.match({url, event});
+      if (matchResult) {
+        _private.logger.log(`The router found a matching route.`, {
```
gauntface: Could you wrap this in NODE_ENV.

packages/workbox-routing/lib/Router.mjs
```diff
@@ -0,0 +1,248 @@
...
+          route,
+          request: event.request,
+        });
+
+        if (Array.isArray(matchResult) && matchResult.length === 0) {
+          // Instead of passing an empty array in as params, use undefined.
+          matchResult = undefined;
+        } else if (matchResult.constructor === Object &&
```
gauntface: Is there a reason to check `matchResult.constructor` vs typeof matchResult === 'Object' ? The latter is a bit clearer in terms of intention.

packages/workbox-routing/lib/Router.mjs
```diff
@@ -0,0 +1,248 @@
...
+        };
+      }
+    }
+
+    // If we didn't have a match, then return undefined values.
+    return {handler: undefined, params: undefined};
+  }
+
+  /**
+   * An optional `handler` that's called by default when no routes
```
gauntface: Nit: Drop the 'default' text: "...that's called when no routes..."

packages/workbox-routing/lib/Router.mjs
jeffposnick: We let the match function return whatever it wants, and `typeof SomeClassInstance` will also return `'object'`. This special-case stuff is needed due to ExpressRoute commonly returning these messy values.

packages/workbox-routing/lib/Router.mjs
jeffposnick: The possibility of there being a `defaultHandler` and/or a `catchHandler` complicates what you describe.

packages/workbox-core/models/messages/messages.mjs
jeffposnick: Good point.

jeffposnick:
Thanks for the reviews. I'll get it hammered into my head that we need `_` for all instance variables.

PR #858, opened 2017-10-09
V3 sw load mod

gauntface:

- Adds ability to load modules in workbox-sw
    - Three options around this:
        1. `modulePathPrefix` allows you to customise the path workbox-sw uses to import modules.
        2. `modulePathCb` allows a developers to completely control where a module is looked up.
        3. `disableModuleImports` force the library to only use globals and throw an error otherwise.
- Adds new line after description lint rule
- Allows deploying the v3 branch to the CDN regardless of whether it's already deployed or not (Will be helpful for testing)

Biggest question I have with this PR is around how we manage versions on CDN vs NPM.

At the moment we version modules separately so we avoid needless update churn. The CDN updates whenever a new git tag is released.

In this PR, I'm assuming the version in the `lerna.json` file is a safe way to get the latest tag being pushed to the CDN and using that inject the version to use into workbox-sw. (See VERSION = `WORKBOX_VERSION_TAG`). I'm not sure if this will be enough when publishing with lerna (i.e. will it bump this version before cutting the release or is it done after).

### DIFF
--- a/packages/workbox-sw/index.mjs
+++ b/packages/workbox-sw/index.mjs
@@ -80,7 +80,14 @@ class WorkboxSW {
       );
     }
 
-    importScripts(this._getImportPath(moduleName));
+    let modulePath = this._getImportPath(moduleName);
+    try {
+      importScripts(modulePath);
+    } catch (err) {
+      console.error(
+        `Unable to import module '${moduleName}' with path '${modulePath}'.`);
+      throw err;
+    }
   }
 
   /**
--- a/packages/workbox-sw/index.mjs
+++ b/packages/workbox-sw/index.mjs
@@ -75,7 +75,7 @@ class WorkboxSW {
       // This can't be a WorkboxError as we can't rely on workbox-core being
       // loaded.
       throw new Error(
-        `Attempted to load module while disableModuleImports is true.`
+        `Attempted to load '${moduleName}' while disableModuleImports is true.`
       );
     }
 
--- a/packages/workbox-sw/index.mjs
+++ b/packages/workbox-sw/index.mjs
@@ -5,6 +5,8 @@
  * @module workbox-sw
  */
 
+const CDN_PATH = `https://storage.googleapis.com/workbox-cdn/releases`;
+
 /**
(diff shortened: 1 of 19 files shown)


jeffposnick:
(I've opened https://github.com/GoogleChrome/workbox/issues/859 to discuss some of the publishing-related changes in v3.)

packages/workbox-sw/index.mjs
```diff
@@ -20,8 +23,31 @@ class WorkboxSW {
...
+      disableModuleImports: false,
     }, options);
+    this._env = this._options.debug ? 'dev' : 'prod';
+
+    if (!this._options.disableModuleImports) {
+      this.loadModule('workbox-core');
+    }
+  }
+
+  /**
```
jeffposnick: How would you feel about using proxies here to do the loading? The code would be more concise without having to list out all the different possible submodules, and there's theoretically some benefit from being able to release new modules that could be loaded from the CDN without having to explicitly update `workbox-sw` to let it know about them.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy

(Doesn't have to be in this PR—just something to think about.)

packages/workbox-sw/index.mjs
```diff
@@ -39,6 +65,58 @@ class WorkboxSW {
...
+  loadModule(moduleName) {
+    if (this._options.disableModuleImports) {
+      // This can't be a WorkboxError as we can't rely on workbox-core being
+      // loaded.
+      throw new Error(
+        `Attempted to load '${moduleName}' while disableModuleImports is true.`
+      );
+    }
+
+    importScripts(this._getImportPath(moduleName));
```
jeffposnick: As discussed, this could be something to do wrap in a `try`/`catch` and give the user some additional context about, but we can wait until we play around with dynamic loading further and explore the different failure possibilities.

QingLeiLi:
hey, i have some question. Is there a way to set the modulePathCb  option? I`m in China, as you know, we cann`t visit google CDN. so, is there a plan expose options to WorkboxSW Class?

PR #860, opened 2017-10-09
_private Fix and Precaching Integration Test

gauntface:
R: @jeffposnick @philipwalton 

Found a bug in precaching / core (The `_private` value was being mangled but needed to be reliable in all modules). This integration reproduced the bug and this PR contains the fix.

Also had to move to passing the seleniumAssistant browser (instead of the webdriver instance) so tests can close / reopen browsers - useful to start with clean profile / service worker state.

### DIFF
--- a/infra/testing/test-server.js
+++ b/infra/testing/test-server.js
@@ -9,6 +9,10 @@ const logHelper = require('../utils/log-helper');
 const app = express();
 
 app.get(/\/__WORKBOX\/buildFile\/(workbox-[A-z]*)(\.(?:dev|prod)\.(.*))*/, (req, res) => {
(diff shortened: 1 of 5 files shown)


infra/testing/test-server.js
```diff
@@ -9,6 +9,10 @@ const logHelper = require('../utils/log-helper');
 const app = express();
 
 app.get(/\/__WORKBOX\/buildFile\/(workbox-[A-z]*)(\.(?:dev|prod)\.(.*))*/, (req, res) => {
+  res.header('Cache-Control', 'private, no-cache, no-store, must-revalidate');
```
jeffposnick: For when you **really** want to avoid the browser cache...

PR #864, opened 2017-10-09
V3 CacheFirst

gauntface:

Adds CacheFirst Strategy.

- Uses the default runtime cache name
- Can be override with options.cacheName in constructor
- plugins can be added to override caching logic etc
- Removed waitOnCache in favor of using event.waitUntil and fixing tests to support this.

### DIFF
--- a/packages/workbox-runtime-caching/CacheFirst.mjs
+++ b/packages/workbox-runtime-caching/CacheFirst.mjs
@@ -32,6 +32,8 @@ class CacheFirst {
    * @param {Object} options
    * @param {string} options.cacheName Cache name to store and retrieve
    * requests. Defaults to cache names provided by `workbox-core`.
+   * @param {string} options.plugins Workbox plugins you may want to use in
+   * conjunction with this caching strategy.
    */
   constructor(options = {}) {
     this._cacheName =
--- a/packages/workbox-runtime-caching/CacheFirst.mjs
+++ b/packages/workbox-runtime-caching/CacheFirst.mjs
@@ -1,0 +1,83 @@
+/*
+ Copyright 2016 Google Inc. All Rights Reserved.
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


packages/workbox-runtime-caching/CacheFirst.mjs
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * @memberof module:workbox-runtime-caching
+ */
+class CacheFirst {
+  /**
+   * @param {Object} options
+   * @param {string} options.cacheName Cache name to store and retrieve
+   * requests. Defaults to cache names provided by `workbox-core`.
+   */
+  constructor({cacheName, plugins} = {}) {
```
jeffposnick: Just checking: are you deliberately sticking with destructured object parameters rather than positional because either of those two parameters might be optional?

packages/workbox-runtime-caching/CacheFirst.mjs
```diff
@@ -0,0 +1,83 @@
...
+      event.request,
+      this._plugins
+    );
+
+    // Keep the service worker while we put the request to the cache
+    event.waitUntil(
+      _private.cacheWrapper.put(
+        this._cacheName,
+        event.request,
+        response.clone(),
```
jeffposnick: I am not 100% sure about the order in which code inside of an `event.waitUntil()` executes, and to be on the safe side, I'd recommend structuring things so that you do the `clone()` prior to calling `event.waitUntil()`.

E.g.

```js
const responseClonse = request.clone();
event.waitUntil(..., responseClone, ...)

return response;
```

test/workbox-runtime-caching/node/test-CacheFirst.mjs
```diff
@@ -0,0 +1,169 @@
...
+          cacheWillUpdate: ({request, response}) => {
+            return response;
+          },
+        },
+      ],
+    });
+    const firstHandleResponse = await cacheFirst.handle(event);
+
+    // Wait until cache.put is finished.
+    await cachePromise;
```
jeffposnick: Nice use of the `event.waitUntil()` stub to detect when the caching is complete!

packages/workbox-runtime-caching/CacheFirst.mjs
gauntface: Yeah it's because they are optional. I'll change it to just be `options`

PR #869, opened 2017-10-10
V3 publishing

gauntface:

- Adds lerna publish (currently doesn't push to git or npm but does push to CDN)
- Adds a build step for ESM code
- Adds version number to all build files of formation workbox:<package name>:<package version>
- Fixes tests that broke during this PR
- Removes some mocks that are no longer needed


package.json
```diff
@@ -102,5 +103,7 @@
     "vinyl-buffer": "^1.0.0",
     "vinyl-source-stream": "^1.1.0"
   },
-  "dependencies": {}
+  "dependencies": {
+    "gulp-debug": "^3.1.0"
```
jeffposnick: This looks like it snuck through from testing.

infra/utils/versioning.js
```diff
@@ -0,0 +1,5 @@
+module.exports = (pkgName) => {
+  const pkgJson = require(`../../packages/${pkgName}/package.json`);
+  const details = ['workbox', pkgName, pkgJson.version];
+  return `function v(){v._version='${details.join(':')}'};v();`;
```
jeffposnick: Quick sanity check: are there any concerns about the symbol `v` ending up clashing with anything else in the same scope? Does `v` get added in before or after the uglification takes place/

gauntface:
@jeffposnick that should be the publishing stuff c'est finis.

jeffposnick:
It all seems reasonable. Thanks for fighting though this.

Issue #865, opened 2017-10-09
Changes to the workbox-cli interface in v3


jeffposnick:
**Library Affected**:
*workbox-cli*

Here are few ideas for changing the workbox-cli interface. Some might be controversial, so feel free to push back.

1. Move from `generate:sw` and `inject:manifest` to `generateSW` and `injectManifest`.
2. No longer infer whether we should be in "wizard" mode or not. Developers can use `wizard` as the command to enter wizard mode.
3. Get rid of the `--config-file` optional flag in favor of making it a required argument for all the modes. For "wizard" mode, it would be the file to write to, and for the other modes, it would be the file to read from.
4. No longer take in build configuration via command-line flags. It's always been confusing understanding whether the flags or the config file options took precedence, and it's also not ideal that CLI flags can only be strings, while the underlying config options might be arbitrary JS types. Developers who need to use different configurations for different modes can specify different config files rather than using different flags.

Here's the help text that I was thinking of, which should make the changes described above a bit more concrete:

```sh
$ ./bin.js --help

  The command line interface for the workbox-build module.

  Usage:
  $ workbox command path/to/workbox-config.js
  
  Arguments:
    command  Can be one of:
             - wizard: Runs the configuration wizard, which will generate a config
               file based on your local build environment.
             - generateSW: Creates a new service worker file based on the
               options in the config file. See
               https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-build#.generateSW
             - injectManifest: Takes an existing service worker file as a
               starting point and creates a service worker file with precaching
               manifest "injected", based on the options in the config file. See
               https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-build#.injectManifest

    path/to/workbox-config.js  In 'wizard' mode, this will be used as the
                               destination for saving your configuration.
                               In 'generateSW' or 'injectManifest' mode, it should
                               be an existing file, in CommonJS module format.
                               The exported object's properties should follow
                               https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-build#.Configuration
                 
  Examples:
    $ workbox wizard new-workbox-config.js
    $ workbox generateSW workbox-config.js
    $ workbox injectManifest config/workbox-config.js
```

gauntface:
I'm in favor of all of the above.

1. Will wizard mode still read the default values from the config file?
[...]

PR #870, opened 2017-10-10
v3: initial workbox-cli code

jeffposnick:

This includes the changes described in https://github.com/GoogleChrome/workbox/issues/865

The `wizard` code isn't ported over yet, but since the other commands are just light wrappers on top of `workbox-build.generateSW` and `workbox-build.injectManifest()`, they should work.

I plan on making some follow-up improvements to the validation code in `workbox-build` that will end up filtering through to this usage.


packages/workbox-cli/src/lib/help-text.js
```diff
@@ -0,0 +1,42 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+**/
+
+module.exports = `Usage:
+$ workbox command path/to/workbox-config.js
+  
+Arguments:
+  command  Can be one of:
+           - wizard: Runs the configuration wizard, which will generate a config
```
gauntface: Take bits as you want / don't want:

```
Usage:
$ workbox <command> [options]
  
Commands:

wizard                    Runs the configuration wizard, which will generate a 
                          config file.

generateSW <config>       Creates a new service worker file based on the
                          options in the config file.
                          See https://goo.gl/zQz4By.

injectManifest <config>   Takes an existing service worker file and creates a
                          copy of it with a precaching manifest "injected" into
                          it. The precaching manifest is generated based on the
                          options in the config file.
                          See https://goo.gl/yB6KZL

Config File:

In 'generateSW' or 'injectManifest' mode, the config file should be an 
existing JavaScript file, in CommonJS module format.
The exported object's properties should follow https://goo.gl/YYPcyY.

Examples:
  $ workbox wizard
  $ workbox generateSW workbox-config.js
  $ workbox injectManifest config/workbox-config.js
```

packages/workbox-cli/src/lib/logger.js
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+**/
+
+const chalk = require('chalk');
+
+module.exports = {
+  debug: (...args) => console.log(chalk.gray(...args)),
+  log: (...args) => console.log(...args),
+  warn: (...args) => console.warn(chalk.red(...args)),
```
gauntface: Yellow is a more comment color choice for warning and would match what we've done in the browser.

test/workbox-cli/node/lib/logger.js
```diff
@@ -0,0 +1,40 @@
+const expect = require('chai').expect;
+const sinon = require('sinon');
+
+const logger = require('../../../../packages/workbox-cli/src/lib/logger');
+
+describe(`[workbox-cli] lib/copy-workbox-sw.js`, function() {
+  const sandbox = sinon.sandbox.create();
+  // Instead of using before/after hooks, the sandbox is explicitly restored
```
gauntface: I would still add beforeEach and after calls to restore just in case of errors, plug you can stub out each console function and then do `console.log.calledOnce` etc.

packages/workbox-cli/src/lib/help-text.js
gauntface: Would also be good to add help and version flags.

jeffposnick:
Thanks for the review. I might have gone a little crazy with some of the stack trace reporting, and I'd like to sit down with Matt & Phil and go over what the current CLI experience is before merging.

packages/workbox-cli/src/lib/cleanup-stack-trace.js
```diff
@@ -0,0 +1,31 @@
+/**
```
jeffposnick: This might be overkill, but if we keep it, I'll add a test.

jeffposnick: @gauntface was in favor of the cleaner stack traces, so I've kept it and added a test for this module.

jeffposnick:
Matt gave his verbal 👍 for the new commits.

PR #872, opened 2017-10-10
Linting rule for license header + adds missing headers.

### LINKED ISSUE #867: eslint rule for for license headers
**Library Affected**:
*all*

We mentioned this in a team meeting once, but I wanted to track it in an issue: we should add in a linting rule for the required Apache 2.0 license header via https://github.com/Stuk/eslint-plugin-header

jeffposnick:

Fixes #867 

I'm just enforcing it for `packages/**/*.{mjs,js}` and I'm only checking for the Copyright portion (since we've used copyright headers that vary in formatting subtly), but that might be sufficient. Let me know if you think otherwise and I'll adjust the PR.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -18,7 +18,7 @@ module.exports = {
     "jsdoc/newline-after-description": 2,
   },
   plugins: [
-      'jsdoc',
+    'jsdoc',
   ],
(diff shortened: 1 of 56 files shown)


gauntface:
Thanks for taking the time to do this.

PR #877, opened 2017-10-11
Adding network first strategy

gauntface:
R: @jeffposnick @philipwalton 

Adds NetworkFirst caching strategy

### DIFF
--- a/packages/workbox-runtime-caching/NetworkFirst.mjs
+++ b/packages/workbox-runtime-caching/NetworkFirst.mjs
@@ -54,17 +54,12 @@ class NetworkFirst {
       _private.cacheNames.getRuntimeName(options.cacheName);
 
     if (options.plugins) {
-      let isUsingCacheWillUpdate = false;
-      for (let plugin of options.plugins) {
-        if (plugin.cacheWillUpdate) {
-          isUsingCacheWillUpdate = true;
-          break;
-        }
-      }
+      let isUsingCacheWillUpdate =
+        options.plugins.some((plugin) => !!plugin.cacheWillUpdate);
       this._plugins = isUsingCacheWillUpdate ?
         options.plugins : [defaultPlugin, ...options.plugins];
     } else {
-      // No plugins passed in so just use default plug.
+      // No plugins passed in, use the default plugin.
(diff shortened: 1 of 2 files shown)


packages/workbox-runtime-caching/NetworkFirst.mjs
```diff
@@ -0,0 +1,162 @@
...
+   * request and the request will fallback to the cache. This option is meant
+   * to combat "[lie-fi](https://developers.google.com/web/fundamentals/performance/poor-connectivity/#lie-fi)"
+   * scenarios.
+   */
+  constructor(options = {}) {
+    this._cacheName =
+      _private.cacheNames.getRuntimeName(options.cacheName);
+
+    if (options.plugins) {
+      let isUsingCacheWillUpdate = false;
```
jeffposnick: Nit; you can rewrite as:

```js
const isUsingCacheWillUpdate = options.plugins.some((plugin) => plugin.cacheWillUpdate);
```

packages/workbox-runtime-caching/NetworkFirst.mjs
```diff
@@ -0,0 +1,162 @@
...
+      for (let plugin of options.plugins) {
+        if (plugin.cacheWillUpdate) {
+          isUsingCacheWillUpdate = true;
+          break;
+        }
+      }
+      this._plugins = isUsingCacheWillUpdate ?
+        options.plugins : [defaultPlugin, ...options.plugins];
+    } else {
+      // No plugins passed in so just use default plug.
```
jeffposnick: "plug" -> "plugin"

packages/workbox-runtime-caching/NetworkFirst.mjs
```diff
@@ -0,0 +1,162 @@
...
+        _private.cacheWrapper.put(
+          this._cacheName,
+          event.request,
+          responseClone,
+          this._plugins
+        )
+      );
+
+      return response;
+    }, () => this._respondFromCache(event.request));
```
jeffposnick: My preference is for explicit `.catch()` rather than the two-argument `.then()`.

PR #878, opened 2017-10-11
Adding helpers to workbox-build

gauntface:
R: @jeffposnick 

Hopefully this would allow you to use CDN in workbox-build. Let me know if not.

### DIFF
--- a/gulp-tasks/build-node-packages.js
+++ b/gulp-tasks/build-node-packages.js
@@ -1,32 +1,9 @@
 const gulp = require('gulp');
-const fs = require('fs-extra');
-const path = require('path');
 
 const buildNodePackage = require('./utils/build-node-package');
 const packageRunnner = require('./utils/package-runner');
 
-// This is needed for workbox-build
-gulp.task('build-node-packages:update-cdn-details', function() {
-  const cdnDetails = fs.readJSONSync(path.join(
-    __dirname, '..', 'cdn-details.json'
-  ));
-
-  const lernaPkg = fs.readJSONSync(path.join(
-    __dirname, '..', 'lerna.json'
-  ));
-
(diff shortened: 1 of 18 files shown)


gulp-tasks/build-node-packages.js
```diff
@@ -1,9 +1,34 @@
 const gulp = require('gulp');
+const fs = require('fs-extra');
+const path = require('path');
 
 const buildNodePackage = require('./utils/build-node-package');
 const packageRunnner = require('./utils/package-runner');
 
+// This is needed for workbox-build
+gulp.task('build-node-packages:update-cdn-details', function() {
```
jeffposnick: Just checking on the release flow when this task is included:

1. It gets run as part of the `build-node-packages` task.
1. It will potentially update `packages/workbox-build/src/cdn-details.json`.
1. If it does update the file, the modification won't be committed to `git`.

Could that lead to problems when `build-node-packages` is being run as part of a release? I.e. will the release only publish whatever is currently committed to `git`, and not include the pending change?

gauntface: When lerna is publishing it'll call `npm run version` which is called *after* the version has been bumped but before anything is committed.

workbox-build will build because of the `npm run version` script calling `npm run build` This means that the cdn version will be updated in workbox-build before it's commited.

Realistically, we only need to build after the version bump, but we run build in prepare because that is what lerna's bootstrapping will call.

gulp-tasks/build-node-packages.js
```diff
@@ -1,9 +1,34 @@
...
+  const cdnDetails = JSON.parse(
```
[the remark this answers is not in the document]
gauntface: noice!

PR #881, opened 2017-10-11
V3 core sweep

gauntface:
Tidies docs, minor size wins.

Changes .logLevel setter to setLogLevel as it makes it a lot cleaner in the resulting docs.

### DIFF
--- a/test/workbox-core/node/controllers/test-WorkboxCore.mjs
+++ b/test/workbox-core/node/controllers/test-WorkboxCore.mjs
@@ -55,10 +55,14 @@ describe(`workbox-core WorkboxCore`, function() {
       '',
       [],
       {},
-    ], (variant) => {
+    ], function(variant) {
+      if (process.env.NODE_ENV !== constants.BUILD_TYPES.dev) {
+        return this.skip();
+      }
+
       return expectError(() => {
         core.logLevel = variant;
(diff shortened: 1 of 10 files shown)


test/workbox-core/node/controllers/test-WorkboxCore.mjs
```diff
@@ -55,10 +55,14 @@ describe(`workbox-core WorkboxCore`, function() {
       '',
       [],
       {},
-    ], (variant) => {
+    ], function(variant) {
+      if (process.env.NODE_ENV !== constants.BUILD_TYPES.dev) {
```
addyosmani: Earlier we check against `process.env.NODE_ENV !== 'production'` (direct string comparison) while here we check on `dev` from the set of build type constants. Minor nit, but should the former one be ~ `process.env.NODE_ENV ~== constants.BUILD_TYPES.production` for pattern consistency?

gauntface: No. In the code, these checks *MUST* be able to equate immediately, meaning that rollup replace we result in `if('production' !== 'production')` which rollup can convert to `if(false)` and ultimately strip as a result. In the latter, we are using this just for testing and builds.

PR #882, opened 2017-10-12
V3 runtime caching docs

gauntface:
R: @jeffposnick @philipwalton 

Minor tweaks to docs

### DIFF
--- a/packages/workbox-runtime-caching/CacheOnly.mjs
+++ b/packages/workbox-runtime-caching/CacheOnly.mjs
@@ -14,13 +14,14 @@
 */
 
 import {_private} from 'workbox-core';
+import core from 'workbox-core';
 
 /**
- * An implementation of a [cache-only](https://developers.google.com/web/fundamentals/instant-and-offline/offline-cookbook/#cache-only)
+ * An implementation of a
+ * [cache-only]{@link https://developers.google.com/web/fundamentals/instant-and-offline/offline-cookbook/#cache-only}
  * request strategy.
  *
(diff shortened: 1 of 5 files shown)


packages/workbox-runtime-caching/CacheOnly.mjs
```diff
@@ -14,13 +14,14 @@
...
 
 /**
- * An implementation of a [cache-only](https://developers.google.com/web/fundamentals/instant-and-offline/offline-cookbook/#cache-only)
+ * An implementation of a
+ * [cache-only]{@link https://developers.google.com/web/fundamentals/instant-and-offline/offline-cookbook/#cache-only}
  * request strategy.
  *
- * This class is useful if you want to take advantage of Workbox cache naming
- * and / or plugins.
+ * This class is useful if you want to take advantage of the Workbox plugins.
```
addyosmani: It reads like the previous text here (cache naming/or plugins) was more specific. Was there a reason we wanted the wording more general?

gauntface: IMO the cache naming wasnt a sell to be honest. its trivial to access now.

gauntface: @jeffposnick @philipwalton do you have a preference here?

jeffposnick: Leaving out mention of the cache names is fine. I'm 👍 with the text in your PR.

PR #883, opened 2017-10-12
Use await/async inside of build:update-cdn-details

jeffposnick:
R: @addyosmani  @addyosmani @gauntface

Switching to the promise-based `readJSON()`/`writeJSON()` and using `await`/`async` inside the task seems to have cleared up the task completion issues I was seeing.

PR #884, opened 2017-10-12
Fix a Windows path issue in the workbox-cli test.

jeffposnick:
R: @philipwalton @gauntface

This should fix https://ci.appveyor.com/project/gauntface/sw-helpers/build/1.0.772#L353

Unfortunately, AppVeyor didn't re-run with my latest commits in the PR that introduced the test, so it wasn't picked up prior to merging.


test/workbox-cli/node/app.js
```diff
@@ -10,7 +10,7 @@ describe(`[workbox-cli] app.js`, function() {
   const PROXIED_CONFIG_FILE = path.resolve(process.cwd(), '/will/be/proxied');
   const PROXIED_ERROR = new Error('proxied error message');
   const PROXIED_CONFIG = {};
-  const INVALID_CONFIG_FILE = '/does/not/exist';
+  const INVALID_CONFIG_FILE = path.resolve(process.cwd(), '/does/not/exist');
```
gauntface: Will windows be ok with the forward slashes?

jeffposnick: `path.resolve()` converts them to whatever the OS uses, so it will be okay with this change in place.

jeffposnick: ...at least it should, since I'm doing the same thing earlier, but this PR's AppVeyor run didn't complete due to #883

jeffposnick:
Re-running with #883 merged to confirm that it passes AppVeyor.

PR #887, opened 2017-10-12
v3 generateSW() end-to-end test

jeffposnick:

This adds in a missing end-to-end test for `workbox-build.generateSW()`. I also updated the underlying code to use the new `cdn-utils` module instead of a hardcoded URL.


packages/workbox-build/src/entry-points/generate-sw.js
```diff
@@ -55,7 +51,7 @@ async function generateSW(input) {
     process.env.NODE_ENV.startsWith('dev')) ? 'dev' : 'prod';
 
   if (options.importWorkboxFromCDN) {
-    const cdnUrl = useBuildType(CDN_URL, buildType);
+    const cdnUrl = cdnUtils.getModuleUrl('workbox-sw', buildType);
```
gauntface: I don't know if there is ever a need to use the dev build of workbox-sw.

`new WorkboxSW({debug: true | false})` is the intended path to switch between dev and prod. workbox-sw is just a thin wrapper.

test/workbox-build/node/entry-points/generate-sw.js
```diff
@@ -0,0 +1,462 @@
...
+          url: 'styles/stylesheet-2.css',
+          revision: '884f6853a4fc655e4c2dc0c0f27a227c',
+        }, {
+          url: 'webpackEntry.js',
+          revision: 'd41d8cd98f00b204e9800998ecf8427e',
+        }]]],
+      }});
+    });
+
+    it(`should use defaults and call importScripts() with a copy of workbox-sw when importWorkboxFromCDN is false`, async function() {
```
gauntface: This will probably need something a little smarter along the lines of copying over all the individual modules and using the workbox-sw pathPrefix to say where it's stored.

test/workbox-build/node/entry-points/generate-sw.js
```diff
@@ -0,0 +1,462 @@
...
+
+    it(`should use defaults when all the required parameters are present, with additional WorkboxSW() configuration`, async function() {
+      const swDest = tempy.file();
+      const workboxSWOptions = {
+        cacheId: 'test',
+        clientsClaim: true,
+        directoryIndex: 'test.html',
+        handleFetch: false,
+        ignoreUrlParametersMatching: [/test1/, /test2/],
+        skipWaiting: true,
```
gauntface: This will likely need moving around in the templates.

jeffposnick:
Thanks—`workbox-build` still lists `workbox-sw` 2.x as a dependency. I need to get things updated to pull in 3.0 and deal with the new syntax. I'll continue tracking things in https://github.com/GoogleChrome/workbox/issues/803

PR #889, opened 2017-10-12
V3 workbox cache expiration

gauntface:

I stopped at this point because I altered the DB and touched a mess of stuff. Woohoo.


packages/workbox-cache-expiration/CacheExpirationManager.mjs
```diff
@@ -0,0 +1,160 @@
...
+
+    // Use a Set to remove any duplicates following the concatenation, then
+    // convert back into an array.
+    const urls = [...new Set(oldEntries.concat(extraEntries))];
+    await Promise.all([
+      this._deleteFromCache(urls),
+      this._deleteFromIDB(urls),
+    ]);
+
+    if (process.env.NODE_ENV) {
```
jeffposnick: Is this meant to be `process.env.NODE_ENV !== 'production'`? If not, when would this be false?

gauntface: Good spot - fixed hopefully.

jeffposnick:
LGTM, but the automated tests are now unhappy.

Issue #886, opened 2017-10-12
workbox-precaching - precache, addRoute, precacheAndRoute()


gauntface:
Changing default behavior of precaching which will be a breaking change for developers.

gauntface:
Note: the workbox-sw config will move to these methods.

gauntface:
Implemented

PR #893, opened 2017-10-13
Adding the route logic

gauntface:
R: @jeffposnick @addyosmani

This implements #886 minus tests. Wanted to get a gut check on the approach and will add tests in a follow up PR.

### DIFF
--- a/packages/workbox-precaching/controllers/PrecacheController.mjs
+++ b/packages/workbox-precaching/controllers/PrecacheController.mjs
@@ -308,12 +308,8 @@ class PrecacheController {
    * @return {Array<string>} An array of URLs.
    */
   async getCachedUrls() {
-    if (!this._parsedCacheUrls) {
-      this._parsedCacheUrls = Array.from(this._entriesToCacheMap.keys())
-        .map((url) => new URL(url, location).href);
-    }
-
-    return this._parsedCacheUrls;
+    return Array.from(this._entriesToCacheMap.keys())
+    .map((url) => new URL(url, location).href);
   }
 }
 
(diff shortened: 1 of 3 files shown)


packages/workbox-precaching/controllers/PrecacheController.mjs
```diff
@@ -301,6 +301,20 @@ class PrecacheController {
       };
     });
   }
+
+  /**
+   * Returns an array of fully qualified URL's that will be precached.
+   *
+   * @return {Array<string>} An array of URLs.
+   */
+  async getCachedUrls() {
```
jeffposnick: I'm thinking that it would be safer to always go through the steps of creating a fresh array of URLs and not saving and reusing the previous `this._parsedCacheUrls` value, since the underlying `this._entriesToCacheMap` value might be changed at any time.

Alternatively, you could set `this._parsedCacheUrls` to `null` whenever `this._entriesToCacheMap` is modified, but that's probably overkill.

(This might have been an issue with v2, too.)

jeffposnick: I'm thinking that it would be safer to always go through the steps of creating a fresh array of URLs and not saving and reusing the previous `this._parsedCacheUrls` value, since the underlying `this._entriesToCacheMap` value might be changed at any time.

Alternatively, you could set `this._parsedCacheUrls` to `null` whenever `this._entriesToCacheMap` is modified, but that's probably overkill.

(This might have been an issue with v2, too.)

gauntface: Just changes it to always return a fresh array.

addyosmani: 👍

addyosmani:
LGTM2

PR #894, opened 2017-10-13
For the temporary github and CDN stuff, remove the generated files

gauntface:
R: @jeffposnick 

This is only useful for us testing stuff out between full publishes, but I just hit it with one of the demos so would be good to land.

### DIFF
--- a/gulp-tasks/publish-cdn.js
+++ b/gulp-tasks/publish-cdn.js
@@ -6,11 +6,16 @@ const cdnUploadHelper = require('./utils/cdn-helper');
 const publishHelpers = require('./utils/publish-helpers');
 const githubHelper = require('./utils/github-helper');
 const logHelper = require('../infra/utils/log-helper');
+const constants = require('./utils/constants');
 
 const findMissingCDNTags = async (tagsData) => {
   const missingTags = [];
   for (let tagData of tagsData) {
     let exists = await cdnUploadHelper.tagExists(tagData.name);
+
+    if (tagData.name.indexOf('3.0.0-alpha') !== -1) {
+      exists = false;
+    }
 
     if (!exists) {
       missingTags.push(tagData);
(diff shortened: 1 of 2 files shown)


gulp-tasks/publish-cdn.js
```diff
@@ -6,12 +6,17 @@ const cdnUploadHelper = require('./utils/cdn-helper');
...
 const githubHelper = require('./utils/github-helper');
 const logHelper = require('../infra/utils/log-helper');
+const constants = require('./utils/constants');
 
 const findMissingCDNTags = async (tagsData) => {
   const missingTags = [];
   for (let tagData of tagsData) {
     let exists = await cdnUploadHelper.tagExists(tagData.name);
 
+    if (tagData.name.indexOf('3.0.0-alpha') !== -1) {
```
jeffposnick: Nit:

```js
exists = tagDate.name.includes('3.0.0-alpha');
```

gulp-tasks/publish-cdn.js
```diff
@@ -54,6 +59,11 @@ gulp.task('publish-cdn:generate-from-tags', async () => {
 });
 
 gulp.task('publish-cdn:temp-v3', async () => {
+  // Let's force this to always be fresh - in case we run it outside of
+  // gulp publish
+  fs.removeSync(
```
jeffposnick: Since this is already an `async` function, could you switch this to:

```js
await fs.remove(...);
```

And switch the other `fs` method calls in this function, too? I know it doesn't make a difference in the current order of operations, but getting into the habit of preferring `await fs.something()` should guard against recurrences of https://github.com/GoogleChrome/workbox/pull/883

gulp-tasks/publish-github.js
```diff
@@ -71,6 +73,11 @@ gulp.task('publish-github:generate-from-tags', async () => {
 // TODO: Delete this task when v3 is about to launch.
 // This is just to publish v3 branch as v3.0.0-alpha
 gulp.task('publish-github:temp-v3', async () => {
+  // Let's force this to always be fresh - in case we run it outside of
+  // gulp publish
+  fs.removeSync(
```
jeffposnick: Same comment re: `await fs.remove()`.

PR #896, opened 2017-10-13
V3 routing docs

gauntface:

This is very rough, but it at least surfaces the available top level functions on the workbox-routing default export.


jeffposnick:
Looks like `index.mjs` needs to be updated to point to `./Route.mjs` instead of `./lib/Route.mjs`.

Otherwise, 👍

PR #897, opened 2017-10-13
RegExpRoute + unit tests

jeffposnick:

The unit tests are ported from the v2 test suite.


packages/workbox-routing/lib/RegExpRoute.mjs
```diff
@@ -0,0 +1,119 @@
...
+ * requests, then going back to our styles example, you could use
+ * `new RegExp('https://third-party-origin\.com/styles/')`, which will trigger
+ * a match for URLs like
+ * <code><strong>https://third-party-origin.com/styles/example.css</strong></code>.
+ *
+ * If you want to use a single `RegExpRoute` that triggers a match for both
+ * same- and cross-origin requests, then use a `RegExp` that includes a wildcard
+ * that will match the start of any URL, like new `RegExp('.+/styles/')`.
+ *
+ * @memberof module:workbox-routing
```
gauntface: Please trim down the this description to a few short paragraphs. We can elaborate a lot more in the modules page, but for reference docs something alone the lines of:

> RegExpRoute is a helper class to make defining regular expression based
> [Routes]{@link module:workbox-routing.Route} easy.
> 
> Please note that this only needs to match part of the URL for a match on requests on the 
> same origin. For third party urls, you must define a RegExp that matches the start of the URL.
>
> new RegExpRoute(new RegExp('/styles/'), ...); // Matches all local styles
> 
> new RegexpRoute(new RegExp('https://example\.com/styles/'), ...); // Matches third party styles

packages/workbox-routing/lib/RegExpRoute.mjs
```diff
@@ -0,0 +1,119 @@
...
+ * @extends Route
+ */
+class RegExpRoute extends Route {
+  /**
+   * Constructor for `RegExpRoute`.
```
gauntface: Can we just drop this description? Templates for JSDoc should highlight it's a constructor.

packages/workbox-routing/lib/RegExpRoute.mjs
```diff
@@ -0,0 +1,119 @@
...
+   *
+   * @param {RegExp} regExp The regular expression to match against URLs.
+   * If the `RegExp` contains [capture groups](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#grouping-back-references),
```
gauntface: Maybe move this into the constructor description or into the class description.

packages/workbox-routing/lib/RegExpRoute.mjs
```diff
@@ -0,0 +1,119 @@
...
+   * then the array of captured values will be passed to `handler` as `params`.
+   * @param {
+   *   module:workbox-routing.Route~handlerCallback |
+   *   module:workbox-runtime-caching.Handler
```
gauntface: Handler is no more, so probably worth just referencing the handlerCallback.

packages/workbox-routing/lib/RegExpRoute.mjs
```diff
@@ -0,0 +1,119 @@
...
+        return null;
+      }
+
+      // Require that the match start at the first character in the URL string
+      // if it's a cross-origin request.
+      // See https://github.com/GoogleChrome/workbox/issues/281 for the context
+      // behind this behavior.
+      if ((url.origin !== location.origin) && (result.index !== 0)) {
+        if (process.env.NODE_ENV !== 'production') {
+          _private.logger.debug(`Skipping route, because while ${regExp} matched
```
gauntface: Nothing to action on for this PR, but I suspect this could be nice to explain in the title more succinctly and then add drop for more details.

test/workbox-routing/node/RegExpRoute.mjs
```diff
@@ -0,0 +1,89 @@
...
+import {devOnly} from '../../../infra/testing/env-it.js';
+
+describe(`[workbox-routing] RegExpRoute`, function() {
+  const SAME_ORIGIN_URL = new URL('https://example.com');
+  const CROSS_ORIGIN_URL = new URL('https://cross-origin-example.com');
+  const PATH = '/test/path';
+  const HANDLER = {handle: () => {}};
+
+  const sandbox = sinon.sandbox.create();
+  sandbox.stub(global, 'location').value(SAME_ORIGIN_URL);
```
gauntface: I would probably add a beforeEach with sandbox restore and re-setting of the location stub - just to be safe (I suspect that if anyone adds a test to this class, this will get missed and cause confusion).

jeffposnick:
Thanks—comments should be all addressed.

jeffposnick:
I had originally left this out of the `index.js`, so it wasn't exported anywhere public. I'm pretty sure that it's meant to be, so I've added it post-review.

If it isn't meant to be included due to size considerations or whatnot, I could remove it, @gauntface

PR #898, opened 2017-10-13
V3 sw proxy

gauntface:

This uses the proxy stuff to make it possible to dynamically pull in workbox modules via importScripts and via the need to know about the `workbox` namespace and 'default' export name.

See what you think. I am slightly concerned that the list of modules is manually curated and not tested, so I will likely change the list to come from a auto-generated build step that will look for modules that have namespaces different to their module name and add only those to the mapping.

I also am concerned that any value will result in an attempted network request - this will likely just lead to some super unnecessary hits on storage that could be prevented. BUT for now I'll leave as is.

### DIFF
--- a/packages/workbox-sw/index.mjs
+++ b/packages/workbox-sw/index.mjs
@@ -49,22 +49,36 @@ class WorkboxSW {
     if (!this._options.disableModuleImports) {
       this.loadModule('workbox-core');
     }
+
+    this._setupProxy();
   }
 
   /**
-   * Get workbox-core.
+   * This method adds a proxy to the WorkboxSW object.
+   *
+   * @private
    */
-  get core() {
-    return workbox.core.default;
-  }
+  _setupProxy() {
+    const moduleNameMapping = {
+      'expiration': 'cache-expiration',
+      'strategies': 'runtime-caching',
+    };
 
-  /**
-   * Get workbox-precaching.
-   */
-  get precaching() {
-    this.loadModule('workbox-precaching');
+    const handler = {
+      get(target, key) {
+        let moduleName = `workbox-${key}`;
+        if (moduleNameMapping[key]) {
+           moduleName = moduleNameMapping[key];
+        }
+        console.info(`importing: '${moduleName}'`);
 
-    return workbox.precaching.default;
+        this.loadModule(moduleName);
+
+        return workbox[key].default;
+      },
+    };
+
+    new Proxy(this, handler);
   }
 
   /**
--- a/packages/workbox-sw/index.mjs
+++ b/packages/workbox-sw/index.mjs
@@ -24,14 +24,18 @@ import './_version.mjs';
  */
 
 const CDN_PATH = `WORKBOX_CDN_ROOT_URL`;
-const MODULE_KEY_TO_NAME_MAPPING = {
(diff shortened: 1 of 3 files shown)


packages/workbox-sw/index.mjs
```diff
@@ -24,6 +24,14 @@ import './_version.mjs';
  */
 
 const CDN_PATH = `WORKBOX_CDN_ROOT_URL`;
+const MODULE_KEY_TO_NAME_MAPPING = {
+  'expiration': 'workbox-cache-expiration',
+  'strategies': 'workbox-runtime-caching',
+};
+const MODULE_NAME_TO_KEY_MAPPING = {
```
jeffposnick: It might be nice to maintain one copy of this mapping, and then automate the reversal:

```js
const MODULE_NAME_TO_KEY_MAPPING = {};
for (const key of Object.keys(MODULE_KEY_TO_NAME_MAPPING)) {
  const name = MODULE_KEY_TO_NAME_MAPPING[key];
  MODULE_NAME_TO_KEY_MAPPING[name] = key;
}
```

And then optionally you could run a sanity check to make sure there's really a 1:1 mapping:

```js
  if (Objects.keys(MODULE_KEY_TO_NAME_MAPPING).length !== Object.keys(MODULE_NAME_TO_KEY_MAPPING).length) {
    throw WorkboxError();
  }
}
```

packages/workbox-sw/index.mjs
```diff
@@ -89,20 +97,56 @@ class WorkboxSW {
...
-        `Attempted to load '${moduleName}' while disableModuleImports is true.`
-      );
+    if (!this._options.disableModuleImports) {
+      let modulePath = this._getImportPath(moduleName);
+      try {
+        importScripts(modulePath);
+      } catch (err) {
+        // We can't rely on workbox-core being loaded so using console
+        console.error(
+          `Unable to import module '${moduleName}' from '${modulePath}'.`);
```
jeffposnick: Maybe add a `TODO` to provide some more context about this error. Like, if it's a scenario where we know they're using local scripts rather than a CDN copy, then this failure might be due to only deploying `workbox-sw.js` and not the helper modules. Or if it's a CDN path that fails, it's probably due to just using a bad `moduleName`.

packages/workbox-sw/index.mjs
addyosmani: I like the idea of minimizing copies of the mapping here. +1 to Jeff's suggestion.

addyosmani:
"I will likely change the list to come from a auto-generated build step that will look for modules that have namespaces different to their module name and add only those to the mapping." I had concerns about the manual curation aspect - was the above change something you were planning to follow up with in this PR or as a TODO/issue for a future one?

packages/workbox-sw/index.mjs
gauntface: I've automated the switch around. Would be good to just automate the creation of these altogether so it's in-sync with the package.json namespaces.

gauntface:
I've added a TODO to move over to the generated list. I'm feeling less and less comfortable with us allowing `workboxSw.anything.foo()` resulting in a load for `workbox-anything` which we *know* doesn't exist.

PR #899, opened 2017-10-14
Adding some tests and getting expiration working

gauntface:
R: @jeffposnick 

Another baby steps PR for cache expiration. This finishes the expiration for max age, but doesn't clear the cache yet. Will do that along with the maximum entries clean up.

### DIFF
--- a/packages/workbox-cache-expiration/CacheExpirationManager.mjs
+++ b/packages/workbox-cache-expiration/CacheExpirationManager.mjs
@@ -76,6 +76,9 @@ class CacheExpirationManager {
 
   /**
    * Expires entries for the given cache and given criteria.
+   *
+   * @return {Promise<Array<string>>} Returns an array of URLs that were
+   * removed.
    */
   async expireEntries() {
     const now = Date.now();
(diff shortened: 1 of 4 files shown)


packages/workbox-cache-expiration/CacheExpirationManager.mjs
```diff
@@ -88,46 +91,52 @@ class CacheExpirationManager {
...
-      this._deleteFromIDB(urls),
+      this._deleteFromCache(allUrls),
+      this._deleteFromIDB(allUrls),
     ]);
 
     if (process.env.NODE_ENV !== 'production') {
+      // TODO break apart entries deleted due to expiration vs size restraints
       _private.logger.groupCollapsed(
-        `Expired ${urls.length}entries have been removed from the cache.`);
+        `Expired ${allUrls.length}entries have been removed from the cache.`);
```
jeffposnick: Nit: This log message formatting looks a bit weird. Can you confirm the spacing?

packages/workbox-cache-expiration/CacheExpirationManager.mjs
```diff
@@ -149,12 +158,14 @@ class CacheExpirationManager {
...
   /**
-   * @return {Promise}
+   * @param {Array<string>} urls Array of URLs to delete from IDB
    *
    * @private
    */
-  _deleteFromIDB() {
-    return Promise.resolve();
+  async _deleteFromIDB(urls) {
+    for (let url of urls) {
```
jeffposnick: Nit: You can use `const` instead of `let` here. (Which, I found that confusing when I realized that...)

packages/workbox-cache-expiration/CacheExpirationManager.mjs
gauntface: Argh - that needs a space. will add in.

PR #905, opened 2017-10-16
V3 welcome move

gauntface:

Was slightly scared that the logging from core was being a pain in the butt, so now moving to when Core is constructed rather than just when it's imported. To do this I needed to adjust the logger to not rely on importing core to get the log level.


packages/workbox-core/utils/logger.mjs
```diff
@@ -54,5 +59,6 @@ const groupEnd = () => {
     console.groupEnd();
   }
 };
-
+export {setLoggerLevel};
```
jeffposnick: Is there any effective difference when using two individual exports of single-property objects vs. combining this and `getLoggerLevel` into a two-property object?

This is not blocking if this is just a style preference. If there *is* a difference, it would be good to add a comment explaining what the difference is.

gauntface: Just preference.

PR #906, opened 2017-10-16
V3 demos

gauntface:
R: @jeffposnick @addyosmani 

First PR for add demo's. This is just precaching (largely to reduce the size of the PR to focus on the approach.

https://workbox-demos.firebaseapp.com

In future PR I want to update the CDN URL based on the latest (make it part of the publish step) and I would like to be able to test these files using local build files somehow.


addyosmani:
Fyi requires a rebase against master.

PR #907, opened 2017-10-16
Cache Expiration Remove extra entries

gauntface:
R: @jeffposnick @addyosmani 

This implements the remove extra entries.

Return sorted timestamps in the models - makes everything a little cleaner in the Manager.

### DIFF
--- a/packages/workbox-cache-expiration/CacheExpirationManager.mjs
+++ b/packages/workbox-cache-expiration/CacheExpirationManager.mjs
@@ -130,10 +130,14 @@ class CacheExpirationManager {
     if (!this._maxAgeSeconds) {
       return [];
     }
+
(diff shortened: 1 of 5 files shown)


packages/workbox-cache-expiration/CacheExpirationManager.mjs
```diff
@@ -145,7 +149,19 @@ class CacheExpirationManager {
...
   async _findExtraEntries() {
-    return [];
+    const extraUrls = [];
+
+    if (!this._maxEntries) {
+      return [];
+    }
+
+    const timestamps = await this._timestampModel.getAllTimestamps();
+    while (timestamps.length > this._maxEntries) {
```
jeffposnick: This would loop infinitely if a developer inadvertantly sets `this._maxEntries` to a negative number.

PR #908, opened 2017-10-16
Expires cache entries

gauntface:
R: @jeffposnick @addyosmani 

Last piece of the CacheExpirationManager for now - will work on the plugin once this looks ok.

### DIFF
--- a/packages/workbox-cache-expiration/CacheExpirationManager.mjs
+++ b/packages/workbox-cache-expiration/CacheExpirationManager.mjs
@@ -71,6 +71,7 @@ class CacheExpirationManager {
 
     this._maxEntries = config.maxEntries;
     this._maxAgeSeconds = config.maxAgeSeconds;
+    this._cacheName = cacheName;
     this._timestampModel = new CacheTimestampsModel(cacheName);
   }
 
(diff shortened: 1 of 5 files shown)


packages/workbox-cache-expiration/CacheExpirationManager.mjs
```diff
@@ -187,6 +192,47 @@ class CacheExpirationManager {
...
+  /**
+   * Can be used to check if a URL has expired or not before it's used.
+   *
+   * Note: This method will not remove the cached entry, call
+   * `expireEntries()` to remove indexedDB and Cache entries.
+   *
+   * @param {string} url
+   * @return {boolean}
+   */
+  async isURLExpired(url) {
```
jeffposnick: Is this currently being used? If not, how do you envision it being used?

(This effectively moves some of our discussion about `isResponseFresh()` from email to GitHub.)

gauntface: It will be a function developers of the module can use if they want.

The plugin I'm just go to add in the current behavior and leave it as it.

jeffposnick: 👍

PR #921, opened 2017-10-18
Adding expiration plugin

gauntface:

- Adds plugin
- Moves CacheExpiration class back to original name
- Adds error if method `isURLExpired()`  is used when no maxAgeSecond is used.

### DIFF
--- a/packages/workbox-cache-expiration/CacheExpirationPlugin.mjs
+++ b/packages/workbox-cache-expiration/CacheExpirationPlugin.mjs
@@ -111,7 +111,6 @@ class CacheExpirationPlugin {
    */
   async cacheDidUpdate({cacheName, newResponse, url} = {}) {
     if (process.env.NODE_ENV !== 'production') {
-      // TODO Check info
       core.assert.isType(cacheName, 'string', {
         moduleName: 'workbox-cache-expiration',
         className: 'CacheExpirationPlugin',
--- a/packages/workbox-cache-expiration/CacheExpirationPlugin.mjs
+++ b/packages/workbox-cache-expiration/CacheExpirationPlugin.mjs
@@ -15,6 +15,7 @@ import {_private} from 'workbox-core';
 import core from 'workbox-core';
 
 import CacheExpiration from './CacheExpiration.mjs';
+import './_version.mjs';
 
 /**
  *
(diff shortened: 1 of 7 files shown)


packages/workbox-cache-expiration/CacheExpirationPlugin.mjs
```diff
@@ -0,0 +1,211 @@
...
+
+        // If there is no Date header (i.e. if this is a cross-origin response),
+        // then we don't know for sure whether the response is fresh or not.
+        // One thing we can do is trigger cache expiration, which will clean up
+        // any old responses based on IDB timestamps, and ensure that when a
+        // cache-first handler is used, stale responses will eventually be
+        // replaced (though not until the *next* request is made).
+        // See https://github.com/GoogleChrome/workbox/issues/691
+
+        const cacheExpiration = this._getCacheExpiration(cacheName);
```
jeffposnick: Even though it's not going to block the `respondWith()`, there's extra asynchronous code that's going to run as the result of this, and if there is a burst of requests, those asynchronous tasks will all be scheduled in short succession.

This was less of a concern in v2, since we used a mutex to ensure that only one expiration job happened at any given time; see https://github.com/GoogleChrome/workbox/issues/922

packages/workbox-cache-expiration/CacheExpirationPlugin.mjs
```diff
@@ -0,0 +1,211 @@
...
+   *
+   * @private
+   * @param {Object} input
+   * @param {string} input.cacheName Name of the cache the responses belong to.
+   * @param {Response} input.newResponse The new value in the cache.
+   * @param {string} input.url The URL for the cache entry.
+   */
+  async cacheDidUpdate({cacheName, newResponse, url} = {}) {
+    if (process.env.NODE_ENV !== 'production') {
+      // TODO Check info
```
jeffposnick: Is this still a real TODO, or is it just a leftover comment?

packages/workbox-cache-expiration/CacheExpirationPlugin.mjs
gauntface: will add it in during the next PR.

This code as a whole is just full of edge cases and unpredictability so the mutex is needed.

gauntface:
@jeffposnick Added the mutex back in.

PR #923, opened 2017-10-18
Demos deploy

gauntface:

Not the prettiest solution to this, but it works.

- Adds a CDN JSON file that be updated when we publish a new release.
- Opens the door to use the demo site along with the local build files by swapping out the CDN details
- Fixes up demos that I added the view files for but didn't mean to.


gauntface:
I'll add a todo on my list to get you all access.

PR #927, opened 2017-10-19
Add workbox-google-analytics

philipwalton:

The code for this PR is functionally identical to PR #770, with additional v3 updates.

Note that I had to update `workbox-background-sync` code slightly to make the callbacks async since `workbox-google-analytics` now has to asynchronously access the request text stored in a `Blob`.


infra/testing/sw-env.js
```diff
@@ -46,3 +49,4 @@ global.location = 'https://example.com';
 global.registration.sync = new SyncManager();
 global.Request = Request;
 global.SyncEvent = SyncEvent;
+global.URLSearchParams = URLSearchParams;
```
gauntface: May need to double check Edge support for this. It's not noted on MDN.

packages/workbox-google-analytics/index.mjs
```diff
@@ -0,0 +1,172 @@
...
+    } else {
+      params = new URL(url).searchParams;
+    }
+
+    // Set the qt param prior to apply the hitFilter or parameterOverrides.
+    const queueTime = Date.now() - timestamp;
+    params.set('qt', queueTime);
+
+    if (config.parameterOverrides) {
+      for (const param of Object.keys(config.parameterOverrides)) {
```
gauntface: I'm assuming Object.assign has a different behavior here.

packages/workbox-google-analytics/index.mjs
```diff
@@ -0,0 +1,172 @@
...
+        const value = config.parameterOverrides[param];
+        params.set(param, value);
+      }
+    }
+
+    if (typeof config.hitFilter == 'function') {
```
gauntface: '==='

packages/workbox-google-analytics/index.mjs
```diff
@@ -0,0 +1,172 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+
+/**
+ * @module workbox-google-analytics
+ */
+
+import {Queue, QueuePlugin} from 'workbox-background-sync';
```
gauntface: It's a shame that to use this module a developer will need to import, core, backgroundsync, routing and runtime-caching to work.

packages/workbox-google-analytics/lib/constants.mjs
```diff
@@ -0,0 +1,23 @@
...
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+import '../_version.mjs';
+
+export const CACHE_NAME = 'workbox-google-analytics';
```
gauntface: Is it worth moving this to workbox-core just to make sure it follows the same pattern as runtime and precaching?

test/workbox-google-analytics/node/test-index.mjs
```diff
@@ -0,0 +1,336 @@
...
+
+describe(`[workbox-google-analytics] initialize`, function() {
+  const sandbox = sinon.sandbox.create();
+  const reset = () => {
+    Queue._queueNames.clear();
+    clearObjectStore();
+    resetEventListeners();
+    sandbox.restore();
+  };
+  process.on('unhandledRejection', catchUnhandledPromiseRejections);
```
gauntface: Nit: Add a comment explaining what this is for.

infra/testing/sw-env.js
philipwalton: I can try to ask some of the Edge folks if they're planning to implement this with service worker. I wouldn't want to add code to do this just out of fear that it wouldn't work once Edge ships.

packages/workbox-google-analytics/index.mjs
philipwalton: I think it's only a shame if it means someone who only wants this module gets all the extra stuff in routing and runtime caching. That's why I was advocating for the ability to do imports like this:

```js
import Route from 'workbox-routing/Route.mjs';
import Router from 'workbox-routing/Router.mjs';
```

instead of this:

```js
import {Route, Router} from 'workbox-routing';
``

Assuming we can get it to work only importing what it needs then I don't consider it a shame, I consider it a win.

I mean, I could easily implement this myself in this package, but isn't the point that users who want multiple workbox packages now get a minimal footprint? And they get things like consistent logging out of the box?

File: infra/testing/sw-env.js (hunk 1 above)
Comment (philipwalton) [reply infra/testing/sw-env.js]: Looks like Jake already made this point: https://wpdev.uservoice.com/forums/257854-microsoft-edge-developer/suggestions/16856668-support-the-searchparams-property-of-url-objects

File: packages/workbox-google-analytics/lib/constants.mjs (hunk 5 above)
Comment (philipwalton) [reply packages/workbox-google-analytics/lib/constants.mjs]: Hmmm, yeah probably.

File: packages/workbox-google-analytics/index.mjs [hunk 7]
```diff
@@ -0,0 +1,172 @@
...
+import './_version.mjs';
+
+/**
+ * Promisifies the FileReader API to await a text response from a Blob.
+ *
+ * @private
+ * @param {Blob} blob
+ * @return {Promise<string>}
+ */
+const getTextFromBlob = (blob) => {
```
jeffposnick: What do you think about putting this into a separate `.mjs` file? That would make it easier to write a unit test for it.

infra/testing/sw-env.js
philipwalton: Based on https://twitter.com/patrickkettner/status/921144430299512832, let's keep this as-is and re-evaluate how to patch/polyfill for Edge in the event it ships SW without URLSearchParams.

test/workbox-google-analytics/node/test-index.mjs
```diff
@@ -0,0 +1,338 @@
...
+
+    self.dispatchEvent(new FetchEvent('fetch', {
+      request: new Request(`https://${GOOGLE_ANALYTICS_HOST}` +
+          `${COLLECT_PATH}`, {
+        method: 'POST',
+        body: PAYLOAD,
+      }),
+    }));
+
+    await waitUntil(() => Queue.prototype.addRequest.callCount === 2);
```
jeffposnick: I think there might be a creative solution to this that doesn't involve polling. Untested, but how about:

```js
const addRequestCalledTwice = new Promise((resolve) => {
  let callCount = 0;
  const addRequestStub = sinon.stub(Queue.prototype, 'addRequest').callsFake((args) => {
    addRequestStub.wrappedMethod(args);
    if (++callCount >= 2) {
      resolve();
    }
  });
});

// Then, later on:
await addRequestCalledTwice;
```

You'd have to have some way of creating new promises/stubs if you need to wait on it a few different times.

infra/testing/sw-env.js
philipwalton: And this: https://twitter.com/dstorey/status/921211253623701505

packages/workbox-google-analytics/index.mjs
philipwalton: It's hard to test as is since it'd basically just be testing our stubs, but it's tested pretty well implicitly (as an implementation detail) via the rest of the tests in this package.

I think it's fine for now, but if another package needs this functionality then it makes sense to move to workbox-core and add explicit tests for it.

PR #929, opened 2017-10-19
Precaching coverage

gauntface:

- Adds tests for the default export logic
- Fixes a few bugs
- Adds an auto-logger stub that basically silences a big chunk of the output from the library (I need to find a solution for the workbox-core welcome message that is clogging up tests)

### DIFF
--- a/test/workbox-core/node/utils/test-logger.mjs
+++ b/test/workbox-core/node/utils/test-logger.mjs
@@ -6,15 +6,18 @@ import logger from '../../../../packages/workbox-core/utils/logger.mjs';
 
 
 describe(`workbox-core logger`, function() {
-  let sandbox;
-
-  before(async function() {
-    sandbox = sinon.sandbox.create();
-  });
+  const sandbox = sinon.sandbox.create();
 
   beforeEach(function() {
     // Reset between runs
     core.setLogLevel(LOG_LEVELS.debug);
+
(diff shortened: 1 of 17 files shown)


test/workbox-core/node/utils/test-logger.mjs
```diff
@@ -6,15 +6,18 @@ import logger from '../../../../packages/workbox-core/utils/logger.mjs';
...
-  before(async function() {
-    sandbox = sinon.sandbox.create();
-  });
+  const sandbox = sinon.sandbox.create();
 
   beforeEach(function() {
     // Reset between runs
     core.setLogLevel(LOG_LEVELS.debug);
+
+    // Undo the logger stubs setup in infra/testing/auto-stub-logger.mjs
```
jeffposnick: Do the stubs need to be re-applied, or does it rely on the fact that this test is always going to run after the rest of the tests that use `logger`?

gauntface: So the auto stub runs before every test case and stubs out logger, preventing the methods from actually console.*. In these tests, I want to make sure they call the expect console.* method with the right args, so it's only in this case that I don't want them stubbed and instead I stub out the specific console.* method.

It's not totally great, but I think it'll prevent noisy logs going forward and I'll move the current console.log() calls to some flavor of logger (just want to exclude the workbox prefix for grouped logs).

PR #931, opened 2017-10-19
V3 sw mock update

gauntface:

Updates service worker mock.

There are a few issues that crop up when we switch everything over that can be switched over.

The most worrying thing is the headers and background sync - The mock just inherits Map, I don't know if this is correct or not, but when passing this into background sync indexed DB started to complain about needing JSON friendly objects.

Routing also had some issues that appears with the FetchEvent, but that was because the mock checks if the request input is of type 'Request' which we override so it doesn't actually work :(

### DIFF
--- a/infra/testing/sw-env-mocks/FetchEvent.js
+++ b/infra/testing/sw-env-mocks/FetchEvent.js
@@ -13,9 +13,6 @@
  limitations under the License.
 */
 
-
-const ExtendableEvent = require('./ExtendableEvent');
-
 // FetchEvent
 // https://www.w3.org/TR/service-workers-1/#fetch-event-section
 class FetchEvent extends ExtendableEvent {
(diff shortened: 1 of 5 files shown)


infra/testing/sw-env.js
```diff
@@ -12,37 +12,38 @@
...
 global.importScripts = () => {};
 global.location = 'https://example.com';
 global.registration.sync = new SyncManager();
-global.Request = Request;
 global.SyncEvent = SyncEvent;
+
+// Because we override Request, we can use the serivce-worker-mock because it
+// checks for type of Request, that will no longer match in our tests.
+// We need to use Request until https://github.com/pinterest/service-workers/issues/65
+global.Request = Request;
```
philipwalton: Does order matter here?

It's not clear to me if this re-ordering is due to the comment or a change in the `service-worker-mock` library.

infra/testing/sw-env-mocks/FetchEvent.js
```diff
@@ -13,9 +13,6 @@
  limitations under the License.
 */
 
-
-const ExtendableEvent = require('./ExtendableEvent');
```
philipwalton: I think this needs to stay otherwise order of imports will matter.

philipwalton:
Maybe we should hold off on upgrading until https://github.com/pinterest/service-workers/issues/65 is addressed since our current mock is working (unless it's not?).

PR #936, opened 2017-10-20
Precaching warning tweaks

gauntface:

- This swaps out the hella long precaching warnig with a link to docs (This will need to be substituted for a d.g.c link but for now it gets the point across
- Also added the ability to surpess this warning in case developers who are revisioning assets that are revisioned want to shut it up. I doubt it'll get much use but I know I'd want it if I used this.

This is the last bit of precaching I want to touch. I'll add some integration tests for workbox routing and upgrading / changing service workers, but that'll be it apart from docs.

### DIFF
--- a/packages/workbox-precaching/utils/showWarningsIfNeeded.mjs
+++ b/packages/workbox-precaching/utils/showWarningsIfNeeded.mjs
@@ -56,9 +56,8 @@ export default (entriesMap) => {
     `\n\n`
   );
 
-  // TODO (gauntface) Before launch swap for d.g.c/web URL and use
-  // logger.unprefixed.warn for these messages.
-  logger.warn(`You can learn more about why this might be a ` +
+  // TODO (gauntface) Before launch swap for d.g.c/web URL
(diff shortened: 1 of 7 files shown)


packages/workbox-precaching/utils/showWarningsIfNeeded.mjs
```diff
@@ -50,39 +50,17 @@ export default (entriesMap) => {
...
-    `precaching to make sure your users don't end up in a broken state.\n`
-  );
-
-  // TODO Add link to docs here.....
-  _private.logger.debug(`You can learn more about this issue and possible ` +
-    `solutions here...`);
+  // TODO (gauntface) Before launch swap for d.g.c/web URL and use
+  // logger.unprefixed.warn for these messages.
+  _private.logger.warn(`You can learn more about why this might be a ` +
+    `problem here: https://docs.google.com/a/google.com/document/d/` +
```
jeffposnick: You need to switch this to a non-`@google.com` document, and ideally do File->Publish to the Web... to get a nice, no-login-required, URL.

Issue #940, opened 2017-10-23
WorkboxSW doesn't work unless instantiated on the workbox namespace


philipwalton:
The loader mechanism we have in `WorkboxSW` currently loads `workbox-core` when instantiated, and then it loads any other packages on demand when the namespace is requested.

For most packages this works fine, but it fails if the requested package depends on other workbox packages (that aren't `workbox-core`).

For example, the following code fails to load and run `workbox-google-analytics`:

```js
importScripts('path/to/workbox-sw.prod.js');

const workboxSW = new WorkboxSW();
workboxSW.googleAnalytics.initialize();
```

<img width="1440" alt="screen shot 2017-10-22 at 3 37 39 pm" src="https://user-images.githubusercontent.com/326742/31903980-c13c9088-b7de-11e7-92f3-83c988e35752.png">

But this works:

```js
importScripts('path/to/workbox-sw.prod.js');

self.workbox = new WorkboxSW();
workboxSW.googleAnalytics.initialize();
```

<img width="1440" alt="screen shot 2017-10-22 at 3 37 00 pm" src="https://user-images.githubusercontent.com/326742/31903991-c9d6c268-b7de-11e7-8a94-97260755a165.png">

The reason is the generated `workbox-google-analytics` bundle looks like this:

```js
this.workbox = this.workbox || {};
this.workbox.googleAnalytics = (function (
  exports,
  workboxBackgroundSync,
  workboxCore,
  workboxRouting,
  workboxRuntimeCaching) {
  // ...
}(
  {},
  workbox.backgroundSync,
  workbox.core,
  workbox.routing,
  workbox.strategies
));
//# sourceMappingURL=workbox-google-analytics.dev.js.map
```

As you can see towards the bottom, when this code is run it expects `workbox.backgroundSync`, `workbox.routing`, and `workbox.strategies` to already be populated, and unless these namespaces can *also* be loaded on demand (i.e. via a `WorkboxSW` instance on `self.workbox`) it errors.

Since the generated code for these bundles only uses the namespace `workbox.*` and since `importScripts` doesn't return a module (it just runs the code), I don't see any solution other than making `self.workbox` *be* the loader.

philipwalton:
#941 is a possible solution to this issue.

gauntface:
I'm pretty confused at how you landed on all of the above.

The problem is that routing, background-sync and runtime-strategies aren't loaded when you try to use google analytics.

[...]

PR #941, opened 2017-10-23
Make WorkboxSW a singleton on self.workbox

philipwalton:

To address #940, this PR updates `workbox-sw` to instantiate and only expose a singleton instance of `WorkboxSW` on `self.workbox`. To handle custom options, it adds a new `.setConfig()` method.

Here's an example of what the usage would look like with these changes:

```js
importScripts('path/to/workbox-sw.prod.js');

workbox.setConfig({
  // ...
});

// Now it's safe to run the following code since `workbox` knows to
// dynamcially load all dependencies.
workbox.googleAnalytics.initialize();
```

* * *

Breaking changes:

* This no longer supports the idea of a *default* export handling the majority use cases on the namespace -- instead it's exposed on `workbox.packageName.default`.
* This also removes the `disableModuleImports` option, as I don't really see a use case for `workbox-sw` without the loader features.


gulp-tasks/utils/build-browser-bundle.js
```diff
@@ -36,6 +36,8 @@ const globals = (moduleId) => {
   }
 
   const packageName = splitModuleId.shift();
+
+  // TODO(philipwalton): remove this restriction.
```
gauntface: This needs a little nuance other than just remove. If this is removed, how do you plan on linking the browser namespace to the file path in the ES2015 import?

gulp-tasks/utils/build-browser-bundle.js
```diff
@@ -50,7 +52,16 @@ const globals = (moduleId) => {
   const packagePath = path.join(__dirname, '..', '..', 'packages', packageName);
   try {
     const pkg = require(path.join(packagePath, 'package.json'));
-    return `${constants.NAMESPACE_PREFIX}.${pkg.workbox.browserNamespace}`;
+    const pkgNamespace = pkg.workbox.browserNamespace;
```
gauntface: I'm not opposed to any of these changes but it feels seperate from the PR you are trying to solve.

test/workbox-sw/node/test-WorkboxSW.mjs
```diff
@@ -32,37 +29,33 @@ describe(`[workbox-sw] WorkboxSW`, function() {
...
-        if (url.includes('workbox-core')) {
-          global.workbox.core = global.workbox.core || fakeWorkbox.core;
-        }
-        if (url.includes('workbox-precaching')) {
-          global.workbox.precaching = global.workbox.precaching || fakeWorkbox.precaching;
-        }
-        if (url.includes('workbox-runtime-caching')) {
-          global.workbox.strategies = global.workbox.strategies || fakeWorkbox.strategies;
-        }
+    self.workbox = {};
```
gauntface: Shouldn't this be a delete?

packages/workbox-sw/WorkboxSW.mjs
```diff
@@ -0,0 +1,166 @@
...
+      },
+    });
+  }
+
+  /**
+   * Updates the configuration options.
+   *
+   * @param {Object=} [options]
+   */
+  config(options = {}) {
```
gauntface: `config()` feels vague to me, would prefer `setConfig()` or `configure()` just so it has an action associated with it.

gulp-tasks/utils/build-browser-bundle.js
philipwalton: Updated the comment based on our discussion to add future support for importing things like `workbox-routing/Route.mjs`

gulp-tasks/utils/build-browser-bundle.js
philipwalton: Yep, removed the comments and reverted to the original version. This was just experimentation I forgot about.

test/workbox-sw/node/test-WorkboxSW.mjs
philipwalton: Yep, changed.

packages/workbox-sw/WorkboxSW.mjs
philipwalton: Agreed, changed to `setConfig()`.

gulp-tasks/utils/build-browser-bundle.js
```diff
@@ -36,6 +36,9 @@ const globals = (moduleId) => {
   }
 
   const packageName = splitModuleId.shift();
+
+  // TODO(philipwalton): remove this restriction and allow for
+  // one-level-below imports, e.g. `workbox-routing/Router.mjs`.
```
gauntface: This has changed with latest v3 so will need rebasing.

gulp-tasks/utils/build-browser-bundle.js
```diff
@@ -85,9 +90,9 @@ module.exports = (packagePath, buildType) => {
     prefix = '';
   }
 
-  const exports = pkgJson.workbox.disabledNamedExports ? 'default' : 'named';
-  const namespace =
-    `${prefix}${pkgJson.workbox.browserNamespace}`;
+  const namespace = pkgJson.workbox.browserNamespace ?
+      `${prefix}${pkgJson.workbox.browserNamespace}` : null;
+  const exports = namespace ? 'named' : 'none';
```
gauntface: This shouldn't be needed now since everything for browser is under the default export which means rollup will export without any naming (i.e. none).

packages/workbox-sw/index.mjs
```diff
@@ -14,183 +14,7 @@
...
-    }
-
-    pathParts.push(fileName);
-
-    return pathParts.join('/');
-  }
-}
-
-export default WorkboxSW;
+self.workbox = new WorkboxSW();
```
gauntface: Why not just use rollup browser namespace? I.e. set the namespace to workbox?

philipwalton:
I made a few updates to incorporate the latest changes and feedback. The primary thing I modified from the previous PR is now all packages have to include `workbox.*` in their namespace value in `package.json`. The reason is so `workbox-sw` can use just `workbox` without having to have an additional setting to ignore the prefix.

philipwalton:
My preference is being explicit about it is better. It also seems a bit counterintuitive to set the namespace field to an empty string and *still* get a namespace set. But if folks disagree I'm happy to revert.

PR #945, opened 2017-10-26
Module rethink

### LINKED ISSUE #943: Module Format / Set Up
This is a summary from a discussion I had with Phil that will hopefully solve a number of issues:

- Webpack + treeshaking #912
- WorkboxSW dependency tree #940
- WorkboxSW usage vs Module usage (covered at the bottom of this issue)

# Proposal

Now that others have had time to play with and test out the current structure it would be good to finalize project structure and move all modules over to this structure. This will affect workbox-sw but will mean usage of workbox-sw vs the individual workbox modules will be the same.

## Public and Private Modules

- /
    - PublicClassFoo.mjs
    - PublicClassBar.mjs
    - _private/
        - PrivateModuleClassFoo.mjs
        - PrivateModuleClassBar.mjs

To help tree shaking in Webpack we need to access the files directly. This introduces some difficulty in ensuring that the conversion from an import to a browser namespace is possible. To highlight the issue, consider:

```javascript
import ClassName from 'workbox-*/directory/something/ClassName.mjs';
```
This is perfectly valid, but to convert to a browser namespace, we would want Rollup to swap to:

```javascript
const ClassName = workbox.*.directory.something.ClassName;
```

This means that the entry point file for Rollup *MUST* have an export of `directory.something.ClassName`. This is difficult to automate building or ensure.

To support the above - all public classes must be defined in the top level (i.e. where `PublicClassFoo.mjs` and `PublicClassBar.mjs` is defined). Nested directories will not be allowed.

All classes that need to be exposed but aren't meant for public consumption (this is largely specific to workbox-core), the `_private` directory will be used and must have a flat structure.

This means we will convert code like:

```javascript
import {private} from 'workbox-core';
import {Router} from 'workbox-routing';
``` 
to

```javascript
import logger from 'workbox-core/_private/logger.mjs';
import Router from 'workbox-routing/Router.mjs';
```

The rollup build step will throw if a file is referenced that is nested in anything other than `_private`, this allows us to have a strict import to browser namespace conversion.

We can then either automate building the export class for this OR a simple test to ensure all public classes (i.e. top level classes) are exported via the index.mjs file (or what ever is referenced in package.json).

## Default Exports

We have the notion of the "Default Export" or more specifically, an export that is pre-configured for the most common use case.

At the moment Rollup is configured to have "named" exports that mean if we have a module with:

```javascript
export {
  Router,
  Route,
}
export default defaultExport
```

The result module code (for a third party) is:

```javascript
import {Router} from 'workbox-routing';
import {Route} from 'workbox-routing';
import routing from 'workbox-routing';
```

The resulting browser namespace code is:

```javascript
workbox.routing.Router
workbox.routing.Route
workbox.routing.default
```

With the suggestion from Phil in #941 means that developers would need to reference default, which is non-ideal. Instead, with the module structure above, have two files for entry points, one that is exclusively used by Rollup to build the browser bundle, and one that is the source code for the module exports.

The browser bundle file will join the default export objects and methods with the exported classes, meaning we no longer need to used "named"exports and have a single default export:

```javascript
// This is the browser bundle file
import Router from './Router.mjs';
import Router from './Route.mjs';
import moduleDefaultExport from './routing-default.mjs';

const finalBrowserExport = Object.assign({
  Router,
  Route,
}, moduleDefaultExport);

export default finalBrowserExport;
```
The module file would be:
```javascript
// This is the module file (i.e. what is used as "module" field in package.json
import Router from './Router.mjs';
import Router from './Route.mjs';
import moduleDefaultExport from './routing-default.mjs';

export {
  Router,
  Route,
}

export default moduleDefaultExport;
```

This would enable us to keep the same module usage above, but the browser API would drop the `default` property and instead methods would be on the default object. The only risk is if a default export overlaps with an exported value. This is a minimal risk.

### WorkboxSW vs Module Usage

This default problem mentioned above is currently papered over by 'WorkboxSW' where referencing a module name would result in importScripts() being called and the default export being returned by the proxy getter.

If we do the above and opt for the workbox-sw singlton in #941, the approach of using workbox-sw vs a module directly is identical.

Originally Workbox-sw would be used as:

```javascript
importScript('...../workbox-sw');

const wb = new WorkboxSW();
wb.precaching.preacheAndRoute();
```

To do the same with just the workbox-precaching module, we'd do

```javascript
importScript('...../workbox-core');
importScript('...../workbox-precaching');

workbox.precaching.default.precacheAndRoute();
```

If we move to a singleton for workbox-sw, we can do the following:

```javascript
importScript('...../workbox-sw');

workbox.precaching.preacheAndRoute();
```

gauntface:

Fixes #943

This is just for `workbox-routing` at the moment, but wanted to get early feedback before proceeding to all modules.


gulp-tasks/utils/build-browser-bundle.js
```diff
@@ -36,21 +36,53 @@ const globals = (moduleId) => {
...
+      // Tried to pull in default export of module - this isn't allowed.
+      // A specific file must be referenced
+      throw new Error(oneLine`
+        You cannot use a module directly - you must specify a file, this is to
+        encourage best practice for tree shaking (i.e. only pulling in what you
+        use). Please fix '${moduleId}'
+      `);
+    }
+    case 1: {
+      // Pulling in a specific file is allowed as long as the exprts have
```
philipwalton: Are you thinking there's a TODO here, or is the check somewhere else?

gauntface: I couldn't think of a way to import and test in the browser bundle (i.e. we don't have @std/esm loaded for this to work. At the moment there are tests to check for _private and I would duplicate that for public classes. This sadly doesn't capture examples where an import for an external module is requested, but that module doesn't actually exist - we could check for the actual file, but other than that, I'm kinda stumped.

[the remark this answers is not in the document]
philipwalton: Oh, I see. You're doing it in a test below.

gulp-tasks/utils/build-browser-bundle.js
```diff
@@ -96,15 +127,13 @@ module.exports = (packagePath, buildType) => {
...
-  logHelper.log(`    Exports: ${logHelper.highlight(exports)}`);
   logHelper.log(`    Namespace: ${logHelper.highlight(namespace)}`);
   logHelper.log(`    Filename: ${logHelper.highlight(outputFilename)}`);
 
   return rollupStream({
-    input: moduleIndexPath,
+    input: moduleBrowserPath,
     rollup,
     format: 'iife',
-    exports,
```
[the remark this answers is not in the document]
gauntface: I can re-add when that happens - this is largely a mental note of - browser builds (at least for current changes) require default only export - (i.e. named exports aren't allowed). workbox-sw can be special cased if needs be (i.e. invert the current default behavior).

philipwalton:
LGTM, thanks for doing all this!

gulp-tasks/utils/build-browser-bundle.js
```diff
@@ -36,21 +36,31 @@ const globals = (moduleId) => {
...
+  const packagePath = path.join(__dirname, '..', '..', 'packages', packageName);
+  const namespacePath = splitModuleId.map((modulePiece) => {
+    // The browser namespace will need the file extension removed
+    return path.basename(modulePiece, path.extname(modulePiece));
+  }).join('.');
+
+  if (splitModuleId.length === 0 || splitModuleId.length > 1) {
+    // Tried to pull in default export of module - this isn't allowed.
+      // A specific file must be referenced
+      throw new Error(oneLine`
```
jeffposnick: Nit: This indentation seems off.

PR #948, opened 2017-10-27
Initial WorkboxWebpackPlugin e2e test

jeffposnick:
This adds an initial end-to-end test for the Webpack plugin.

It's pretty basic, but I wanted to make sure the Webpack code had at least one complete test before I made any changes to it.

A general TODO for me is to put together a variety of test cases, both for different Webpack build configurations and for different `WorkboxWebpackPlugin` configurations. Once more test cases are added, I'll likely refactor some of the validation.


packages/workbox-webpack-plugin/src/lib/utils/copy-workbox-sw.js
```diff
@@ -23,7 +23,9 @@ module.exports = (readFileFn) => {
   // TODO: this should also support importing workbox-sw from a CDN
   const buildType = (process.env.NODE_ENV &&
     process.env.NODE_ENV.startsWith('dev')) ? 'dev' : 'prod';
-  const defaultWorkboxSourcePath = require.resolve('workbox-sw');
+  // workbox-sw is installed as a dependency of workbox-build.
+  const defaultWorkboxSourcePath = require.resolve(
```
jeffposnick: This is a stop-gap, as this code is likely to change to a) support CDNs and b) use the methods in `workbox-build` to copy over not on the `workbox-sw` file, but also any other dependencies that are required.

test/workbox-webpack-plugin/static/example-project-1/index.html
```diff
@@ -0,0 +1,74 @@
...
+  .catch((err) => {
+    document.body.style.color = 'red';
+    document.body.innerHTML = `<h1>${err.message}</h1>
+
+    <pre>${err.stack}</pre>`;
+
+    console.error(err.message)
+    console.error(err.stack);
+
+    // This is just to help with local development so we know there is an error
```
gauntface: This will result in tests taking an extra 5 seconds to error on Travis.

Could you add a check for `process.env.TRAVIS` and timeout when it's not defined, otherwise error immediately?

jeffposnick: The HTML/JS/etc. underneath `example-project-1` aren't actually loaded and run. I'm just carrying over the same set of "real world sample files" that we used in v2 to test out the precaching behavior.

(The files could easily just consist of a single byte so that they each corresponded to a different hash, but it's nice having files that are a bit larger, since we do take file size into account in a few places.)

jeffposnick:
(The test failures are due to https://github.com/pinterest/service-workers/issues/68)

PR #951, opened 2017-10-30
Core coverage

gauntface:

I don't think I've made many (if any) large refactors of Core in a while so just wanted to up the coverage in an effort to move core closer to the "done" category.


philipwalton:
I'm confused at to what the idbhelper test is testing. Is this just to get the coverage up? And BTW, once I finish my IDB stuff we won't need these tests, so I wouldn't spend much time on them.

gauntface:
Yes it's just to increase coverage - i.e. it doesn't through and it puts and gets as expected.

PR #954, opened 2017-10-30
Logging changes

gauntface:

I noticed that after updating the demos that logging from fetch and cache wrappers were very verbose in precaching so I took another stab at them.

- precaching and runtime caching modules will create groups that will "group" the relevant logs from fetch and cache wrappers under the appropriate request heading
- fetch and cache wrapper now print logs to debug. This means by default they won't be seen. This means developers / bug traces can include all of the fetch and cache wrapper info if needed.
- runtime caching has some more specific logs for each strategy that they didn't have before.

### DIFF
--- a/demos/functions/views/demo/workbox-cache-expiration/sw.hbs
+++ b/demos/functions/views/demo/workbox-cache-expiration/sw.hbs
@@ -1,8 +1,5 @@
-importScripts('{{CDN_URL}}/workbox-sw.dev.js');
-
-workbox.setConfig({
-  debug: true
-});
+importScripts('{{CDN_URL}}/workbox-core.dev.js');
+importScripts('{{CDN_URL}}/workbox-cache-expiration.dev.js');
 
 const expirationManager = new workbox.expiration.CacheExpiration(
   'cache-expiration', {
--- a/packages/workbox-precaching/controllers/PrecacheController.mjs
+++ b/packages/workbox-precaching/controllers/PrecacheController.mjs
@@ -156,7 +156,7 @@ class PrecacheController {
     const entriesToPrecache = [];
     const entriesAlreadyPrecached = [];
 
-    for (let precacheEntry of this._entriesToCacheMap.values()) {
+    for (const precacheEntry of this._entriesToCacheMap.values()) {
       if (await this._precacheDetailsModel._isEntryCached(precacheEntry)) {
         entriesAlreadyPrecached.push(precacheEntry);
       } else {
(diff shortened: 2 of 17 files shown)


demos/functions/views/demo/workbox-cache-expiration/sw.hbs
```diff
@@ -1,8 +1,5 @@
-importScripts('{{CDN_URL}}/workbox-sw.dev.js');
-
-workbox.setConfig({
-  debug: true
-});
+importScripts('{{CDN_URL}}/workbox-core.dev.js');
+importScripts('{{CDN_URL}}/workbox-cache-expiration.dev.js');
```
philipwalton: Out of curiosity, why not use `workbox-sw` here?

packages/workbox-precaching/controllers/PrecacheController.mjs
```diff
@@ -152,33 +153,33 @@ class PrecacheController {
...
-        } else {
-          notUpdatedEntries.push(precacheEntry);
-        }
-      });
+    const entriesToPrecache = [];
+    const entriesAlreadyPrecached = [];
 
-      cachePromises.push(promiseChain);
-    });
+    for (let precacheEntry of this._entriesToCacheMap.values()) {
```
philipwalton: Nit: you can use `const` in for-of loops. Google style says to use `const` everywhere unless you have to use `let` due to reassignment.

packages/workbox-precaching/controllers/PrecacheController.mjs
```diff
@@ -152,33 +153,33 @@ class PrecacheController {
...
 
     if (process.env.NODE_ENV !== 'production') {
-      printInstallDetails(updatedEntries, notUpdatedEntries);
+      logger.groupEnd();
     }
 
     return {
-      'updatedEntries': updatedEntries,
-      'notUpdatedEntries': notUpdatedEntries,
+      'updatedEntries': entriesToPrecache,
```
philipwalton: Nit: are quotes needed around these keys?

demos/functions/views/demo/workbox-cache-expiration/sw.hbs
gauntface: The cdn_url only applies to top level so Workbox-sw would need the different path passed in which might look a tad weird for published demos. Need to find a better to switch better CDN and local builds

gauntface: Opened an issue for changing how local and cdn builds are used.

https://github.com/GoogleChrome/workbox/issues/959

PR #956, opened 2017-10-30
Ensure expireEntries is called

gauntface:

If I've understood the logic for CacheExpirationPlugin the only time expireEntries is called is when there is no date or a cache entry is updated

This isn't such a bad problem for maxEntries, but for maxAgeSeconds it has the following negative side effects:

- If there is a date header but it can't be parsed, it'll be treated as fresh, but never expired.
- If the date header can be parsed, we'll mark the response as fresh or not, but we won't remove it from the cache because we never expire it (meaning Cache and iDB will grow).

### DIFF
--- a/packages/workbox-cache-expiration/CacheExpirationPlugin.mjs
+++ b/packages/workbox-cache-expiration/CacheExpirationPlugin.mjs
@@ -114,6 +114,9 @@ class CacheExpirationPlugin {
       return true;
     }
 
+    // Check if the 'date' header will suffice a quick expiration check.
+    // See https://github.com/GoogleChromeLabs/sw-toolbox/issues/164 for
+    // discussion.
     const dateHeaderTimestamp = this._getDateHeaderTimestamp(cachedResponse);
     if (dateHeaderTimestamp === null) {
       // Unable to parse date, so assume it's fresh.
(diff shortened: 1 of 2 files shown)


packages/workbox-cache-expiration/CacheExpirationPlugin.mjs
```diff
@@ -92,11 +92,64 @@ class CacheExpirationPlugin {
...
    */
   cachedResponseWillBeUsed({cacheName, cachedResponse} = {}) {
-    if (this.isResponseFresh({cacheName, cachedResponse})) {
-      return cachedResponse;
+    let isFresh = this._isResponseDateFresh(cachedResponse);
+
+    // Expire entries to ensure that even if the expiration date has
+    // expired, it'll only be used once.
+    const cacheExpiration = this._getCacheExpiration(cacheName);
+    cacheExpiration.expireEntries();
```
jeffposnick: There's a bunch of lines of code changed, but as far as I can tell from the PR's description, this is the only one that actually matters—can you confirm that?

gauntface: Yeah this is the important bit.

PR #958, opened 2017-10-31
Webpack plugin style tweaks

jeffposnick:

Some light refactoring of the Webpack code in v3 to adhere to our code style.

There's now (one) e2e test, so hopefully that will pick up whether there are any functional changes. (There shouldn't be in this PR.)


packages/workbox-webpack-plugin/src/lib/utils/copy-workbox-sw.js
```diff
@@ -19,24 +19,27 @@ const {readFile} = require('./read-file');
 const errors = require('workbox-build/src/lib/errors');
 const useBuildType = require('workbox-build/src/lib/use-build-type');
 
-module.exports = (readFileFn) => {
+module.exports = async () => {
```
jeffposnick: `readFileFn` was never used in the original code.

packages/workbox-webpack-plugin/src/index.js
```diff
@@ -23,9 +23,9 @@ const copyWorkboxSW = require('./lib/utils/copy-workbox-sw');
 const {setReadFile} = require('./lib/utils/read-file');
 
 /**
- * Use the instance of this in the plugins array of the webpack config.
+ * Use the instance of this in the `plugins` array of the Webpack config.
  *
- * @example <caption>Zero-configuration</caption>
+ * @example <caption>Zero-configuration usage:</caption>
```
gauntface: Can we remove these examples and cover them in the webpack module docs: 

https://docs.google.com/document/d/1w53nZZCAT765IE3eAzbMg1qP4z72BJ3Ors3vz6QQP4A/edit?usp=sharing

packages/workbox-webpack-plugin/src/index.js
```diff
@@ -115,34 +115,39 @@ class WorkboxWebpackPlugin {
      * compiler.plugin('watch-run') for development.
      */
     setReadFile(compiler.inputFileSystem._readFile);
+
     /**
-     * During the make phase of the webpack compilation, we use
+     * During the make phase of the Webpack compilation, we use
```
gauntface: I think webpack is meant to be all lower case as weird as it may seem.

From: http://webpack.github.io/docs/

> If you are new to webpack, you may want

jeffposnick: Go figure.

PR #962, opened 2017-10-31
Adding workbox-sw demo + moving demos over to using workbox-sw

### LINKED ISSUE #959: Demos + workbox-sw and CDN_URL
In the demos, change the way the service worker files are built such that instead of using the individual modules, just append the workbox-sw config to the start of the file along the lines of:

```javascript
import('<Local or CDN path>/workbox-sw.js');
workbox.setConfig({
  debug: true,
  <IF Local Path>modulePathPrefix: <Local Path></IF>
})

<SW File Contents Here>
```

This will be slightly weird for demo authors but we can include a comment that gets replaced with the above string.

gauntface:

Fixes #959 

1. Moves demos over to using workbox-sw with local builds for testing.
1. Adds workbox-sw demo (which I'm treating as the grab bag demo for illustrating a more complete example).


jeffposnick:
👍

PR #965, opened 2017-11-01
v3: workbox-cli wizard

jeffposnick:

Here's the code for the v3 `workbox-cli` wizard.

It's a modified version of the v2 wizard, based on some of the discussions at https://github.com/GoogleChrome/workbox/issues/865

It asks the questions necessary to set you up for a basic `generateSW` build process, saves the config file you'd need to use, and prints out the command you need to run as part of your build. v2 had some questions around the `injectManifest` build process as well (it asked for `swSrc`), but I feel like it's less likely that folks who have an existing service worker file with the required injection point need to go through the wizard process, so I'd rather simplify and optimize for `generateSW`.

I wanted to get some feedback on the code before I add tests, so I'm just looking for a 👍  on the question flow for now, and will push a commit to this PR with tests once we're good.

Here's the basic experience:

```sh
$ ./bin.js wizard
? What is the root of your web app? src
? Which file types would you like to precache? js
? Where would you like to save the service worker file created by this configuration? build/sw.js
? Finally, where would you like to save these configuration options so that they can be used by workbox? workbox-config.js
All done! To build your service worker, run

  workbox generateSW workbox-config.js

as the last step in your web app's build process.
You can further customize your service worker by making changes to workbox-config.js. See https://goo.gl/YYPcyY for details.
```


packages/workbox-cli/src/lib/questions/ask-extensions-to-cache.js
```diff
@@ -0,0 +1,81 @@
...
+// The key used for the question/answer.
+const name = 'globPatterns';
+
+/**
+ * @param {string} globDirectory The directory used for the root of globbing.
+ * @return {Promise<[string]>} The unique file extensions corresponding to all
+ * of the files under globDirectory.
+ */
+async function getAllFileExtensions(globDirectory) {
+  const substringsToIgnore = ignoredDirectories.map(
```
jeffposnick: TODO: Put in a spinner since this might take some time for directories with lots and lots of files.

jeffposnick: I've added it in https://github.com/GoogleChrome/workbox/pull/965/commits/469ca6f1aead8d1f1b1852e6fe87ded444e34642

packages/workbox-cli/src/lib/errors.js
```diff
@@ -24,4 +24,14 @@ module.exports = {
   'config-validation-failed': `Your configuration is invalid:`,
   'workbox-build-runtime-error': `Service worker generation failed:`,
   'unknown-command': `Unknown command:`,
+  'no-file-extensions-found': ol`No files could be found that are suitable for
+    caching.`,
+  'no-file-extensions-selected': `Please select at least one file extension.`,
+  'sw-src-does-not-exist': ol`The source service worker file you provided does
```
gauntface: Nit: Feels a little awkward "The source service worker file you provided does not exist. Please provide the path to an existing file."

Maybe change to something like:

"Could not find the service worker file at '<Insert absolute file path here>'. Please make sure it exists."

packages/workbox-cli/src/lib/errors.js
```diff
@@ -24,4 +24,14 @@ module.exports = {
...
+    not exist. Please provide the path to an existing file.`,
+  'invalid-sw-dest': ol`Please enter a valid path to use for the service worker
```
gauntface: Since this is the destination path, what exactly needs to the exist, the directory? Or something else? The error message makes it sound like I need to create an empty file.

packages/workbox-cli/src/lib/errors.js
```diff
@@ -24,4 +24,14 @@ module.exports = {
...
+    file that's created.`,
+  'glob-directory-invalid': ol`The path you entered isn't a valid directory.`,
```
gauntface: By invalid - do you mean doesn't exist or is a file or ...?

packages/workbox-cli/src/lib/errors.js
```diff
@@ -24,4 +24,14 @@ module.exports = {
...
+  'invalid-config-location': ol`Please enter a valid path to use for the saved
```
gauntface: Does this require a file structure to exist to work? or does this error occur in other situations?

packages/workbox-cli/src/lib/questions/ask-config-location.js
```diff
@@ -0,0 +1,46 @@
...
+const name = 'configLocation';
+
+/**
+ * @return {Promise<Object>} The answers from inquirer.
+ */
+function askQuestion() {
+  return inquirer.prompt([{
+    name,
+    message: ol`Finally, where would you like to save these configuration
+      options so that they can be used by workbox?`,
```
gauntface: Nit: Rephrase to: "Where would you like to save these configuration options?" I think given the context, it would be used by workbox and by adding finally we must always put this at the end of the questioning, which may not be the case.

packages/workbox-cli/src/lib/questions/ask-extensions-to-cache.js
```diff
@@ -0,0 +1,87 @@
...
+
+  return [...extensions];
+}
+
+/**
+ * @param {string} globDirectory The directory used for the root of globbing.
+ * @return {Promise<Object>} The answers from inquirer.
+ */
+async function askQuestion(globDirectory) {
+  const spinner = ora({
```
gauntface: Maybe add a comment above this explaining that to show this question, we need to know what file extensions are in the globDirectory before it can be asked.

Initially reading this - it was really unclear why the spinner was being shown before anything was asked.

packages/workbox-cli/src/lib/questions/ask-extensions-to-cache.js
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * @param {string} globDirectory The directory used for the root of globbing.
+ * @return {Promise<[string]>} The unique file extensions corresponding to all
+ * of the files under globDirectory.
+ */
+async function getAllFileExtensions(globDirectory) {
+  const substringsToIgnore = ignoredDirectories.map(
+    (directory) => `${path.sep}${directory}${path.sep}`);
+
+  const allFiles = await nodeDir.promiseFiles(globDirectory, 'file',
```
gauntface: Is there a reason to to use node-dir over glob here?

packages/workbox-cli/src/lib/questions/ask-extensions-to-cache.js
```diff
@@ -0,0 +1,87 @@
...
+    {shortName: false, recursive: true});
+
+  const filteredFiles = allFiles.filter(
```
gauntface: You don't need to act on this, but this code is unreadable to me, I assume it'll work but: filter > every > includes just results in in something that is very hard to parse.

packages/workbox-cli/src/lib/questions/ask-extensions-to-cache.js
```diff
@@ -0,0 +1,87 @@
...
+ */
+async function askQuestion(globDirectory) {
+  const spinner = ora({
+    text: `Examining files in ${globDirectory}...`,
+    stream: process.stdout,
+  }).start();
+  const fileExtensions = await getAllFileExtensions(globDirectory);
+  spinner.stop();
+
+  assert(fileExtensions.length > 0, errors['no-file-extensions-found']);
```
gauntface: Just tested assert on node and the output from this isn't the most friendly thing:

> assert(false, 'test')
AssertionError [ERR_ASSERTION]: test
    at repl:1:1
    at ContextifyScript.Script.runInThisContext (vm.js:50:33)
    at REPLServer.defaultEval (repl.js:240:29)
    at bound (domain.js:301:14)
    at REPLServer.runBound [as eval] (domain.js:314:12)
    at REPLServer.onLine (repl.js:441:10)
    at emitOne (events.js:120:20)
    at REPLServer.emit (events.js:210:7)
    at REPLServer.Interface._onLine (readline.js:282:10)
    at REPLServer.Interface._line (readline.js:631:8)

packages/workbox-cli/src/lib/questions/ask-extensions-to-cache.js
```diff
@@ -0,0 +1,87 @@
...
+    default: fileExtensions,
+  }]);
+}
+
+module.exports = async (globDirectory) => {
+  const answers = await askQuestion(globDirectory);
+  const extensions = answers[name];
+  assert(extensions.length > 0, errors['no-file-extensions-selected']);
+
+  return [`**/*.{${extensions.join(',')}}`];
```
gauntface: Why return this as an array?

packages/workbox-cli/src/lib/questions/ask-root-of-web-app.js
```diff
@@ -0,0 +1,86 @@
...
+      type: 'list',
+      message: 'What is the root of your web app?',
+      choices: subdirectories.concat([
+        new inquirer.Separator(),
+        manualEntryChoice,
+      ]),
+    }, {
+      name,
+      when: (answers) => answers[name] === manualEntryChoice,
+      message: `Path to your web app's root:`,
```
gauntface: Nit: Does it make sense to change this to more of a request?

"Please enter the path to the root of your web app"

packages/workbox-cli/src/lib/questions/ask-root-of-web-app.js
```diff
@@ -0,0 +1,86 @@
...
+    }]);
+  } else {
+    return inquirer.prompt([{
+      name,
+      message: `Path to your web app's root:`,
```
gauntface: Pull this and the string above out into a const in this file?

packages/workbox-cli/src/lib/questions/ask-root-of-web-app.js
```diff
@@ -0,0 +1,86 @@
...
+    }]);
+  }
+}
+
+module.exports = async () => {
+  const answers = await askQuestion();
+  const globDirectory = answers[name];
+
+  try {
+    const stat = await fse.stat(globDirectory);
```
gauntface: I would expect this to make a directory if it doesn't exist, but that might just be me.

packages/workbox-cli/src/lib/questions/ask-sw-dest.js
```diff
@@ -0,0 +1,46 @@
...
+// The key used for the question/answer.
+const name = 'swDest';
+
+/**
+ * @return {Promise<Object>} The answers from inquirer.
+ */
+function askQuestion() {
+  return inquirer.prompt([{
+    name,
+    message: ol`Where would you like to save the service worker file created
```
gauntface: Nit: Reword to "Where you like your service worker file to be saved?"

packages/workbox-cli/src/lib/questions/ask-sw-dest.js
```diff
@@ -0,0 +1,46 @@
...
+      by this configuration?`,
+    type: 'input',
+    default: 'build/sw.js',
+  }]);
+}
+
+module.exports = async () => {
+  const answers = await askQuestion();
+  const swDest = answers[name].trim();
```
gauntface: What happens if a directory is passed in instead of path with a filename?

packages/workbox-cli/src/lib/run-wizard.js
```diff
@@ -0,0 +1,39 @@
...
+const fse = require('fs-extra');
+const ol = require('common-tags').oneLine;
+
+const askQuestions = require('./questions/ask-questions');
+const logger = require('./logger');
+
+module.exports = async () => {
+  const config = await askQuestions();
+  const {configLocation} = config;
+  delete config.configLocation;
```
gauntface: please add a comment on why this is needed.

packages/workbox-cli/src/lib/run-wizard.js
```diff
@@ -0,0 +1,39 @@
...
+
+  const contents = `module.exports = ${JSON.stringify(config, null, 2)};`;
+  await fse.writeFile(configLocation, contents);
+
+  logger.log(`All done! To build your service worker, run
```
gauntface: Nit: "to rebuild"

packages/workbox-cli/src/lib/errors.js
jeffposnick: Actually, turns out that error string isn't being used—I think I carried it over from v2? I've deleted it.

packages/workbox-cli/src/lib/errors.js
jeffposnick: It's an error message that's displayed when you provide an empty string as the `swDest` value.

If you provide any other value, you won't see that error message, but you may run into runtime errors when you generate a SW if the path can't be used.

packages/workbox-cli/src/lib/errors.js
jeffposnick: Either. If it's not a valid directory for any reason—either it doesn't exist at all, or it exists but it's a file, you'll see that.

packages/workbox-cli/src/lib/errors.js
jeffposnick: Similar to the `invalid-sw-dest` error, it's something that gets triggered when you enter an empty string.

Also similar, if it's a path that's not valid, you'll end up with a runtime error when the write operation fails.

packages/workbox-cli/src/lib/questions/ask-extensions-to-cache.js
jeffposnick: That's a good point—I'll switch.

packages/workbox-cli/src/lib/questions/ask-extensions-to-cache.js
jeffposnick: Will do.

packages/workbox-cli/src/lib/questions/ask-extensions-to-cache.js
jeffposnick: We [clean up any exceptions](https://github.com/GoogleChrome/workbox/blob/v3/packages/workbox-cli/src/lib/cleanup-stack-trace.js) prior to showing them to the user. The failure here looks like:

```sh
$ ./bin.js wizard
? What is the root of your web app? Manually enter path
? Path to your web app's root: /tmp/a

No files could be found that are suitable for caching.
```

(with the last line in red).

packages/workbox-cli/src/lib/questions/ask-extensions-to-cache.js
jeffposnick: `globPatterns` expects an array of strings, and it's in keeping with the approach used in the other questions to return values that correspond directly to the config options.

packages/workbox-cli/src/lib/questions/ask-root-of-web-app.js
jeffposnick: If we make a new directory then it definitely won't contain any files with extensions to cache, and the next question (which deals with file extensions) will always fail.

packages/workbox-cli/src/lib/questions/ask-root-of-web-app.js
jeffposnick: Done.

packages/workbox-cli/src/lib/questions/ask-root-of-web-app.js
jeffposnick: Done.

packages/workbox-cli/src/lib/questions/ask-sw-dest.js
jeffposnick: Done.

packages/workbox-cli/src/lib/questions/ask-sw-dest.js
jeffposnick: The write operation will fail during the build process, with an error explaining why.

packages/workbox-cli/src/lib/run-wizard.js
jeffposnick: "Rebuild" isn't accurate, since I am not triggering the build process following the completion of the wizard's questions.

Instead, I'm displaying this message because I want to push developers in the direction of modifying their build scripts to add in the `workbox generateSW` step, which ensures that they get a fresh precache payload each time they rebuild their web app.

My concern is that running the build following the wizard completion process makes it seem like it's a one-off thing that's now done, rather than something that needs to be done repeatedly.

packages/workbox-cli/src/lib/questions/ask-config-location.js
jeffposnick: I've updated the wording.

jeffposnick:
Thanks for the review, @gauntface! Doesn't seem like there's anything too contentious about this approach, so I've addressed the feedback and am writing some tests for this PR now.

packages/workbox-cli/src/lib/questions/ask-extensions-to-cache.js
gauntface: So you are removing the `AssertionError [ERR_ASSERTION]:` in the log as well somehow?

packages/workbox-cli/src/lib/run-wizard.js
gauntface: So if I run the wizard, I don't get a service worker? In which case we aren't 'All Done', there is next step that needs to be done. if I'm following.

gauntface:
You caught me off guard slightly with one comment: https://github.com/GoogleChrome/workbox/pull/965#discussion_r148426818

Could we address that before carrying on

packages/workbox-cli/src/lib/questions/ask-extensions-to-cache.js
jeffposnick: Yes, when there's an exception, the [CLI logs the `message` property along with the cleaned-up stacktrace](https://github.com/GoogleChrome/workbox/blob/v3/packages/workbox-cli/src/bin.js#L38-L39). The `message` property of an assertion failure just uses the string that was passed in:

```sh
> try { assert(false, 'this is a test'); } catch(e) { console.error(e.message); }
this is a test
```

When the CLI is used with `--debug`, we log the raw stracktrace, which would lead to the `AssertionError [ERR_ASSERTION]:` text, but that's not what users would see by default.

packages/workbox-cli/src/lib/run-wizard.js
jeffposnick: If you run the wizard, you get a configuration file that could then be used with `workbox generateSW path/to/config-file.js`.

"All done" refers to being all done with questions and having a usable configuration that's been written to disk. I'm happy to remove that if it comes across as confusing.

You still need to run workbox to generate a SW, ideally as part of your build process, which is what I am attempting to get across in that message.

In v3, `workbox --help` displays:

```sh
  Commands:
    wizard
      Runs the configuration wizard, which will generate a
      config file based on answers to questions.
```

which conveys that same message: run the wizard to create a config file.

gauntface: Yeah we need to change the wording of this.

Maybe something like "You are now ready to build your service worker, just run 'workbox generate-sw'.'".

Sorry I completely missed this.

jeffposnick: I've further tweaked the message after your initial feedback. It now looks like:

```sh
Your config has been saved. To build your service worker, run

  workbox generateSW ${configLocation}

as part of a build process. See https://goo.gl/8zZy1P for details.
You can further customize your service worker by making changes to ${configLocation}.
See https://goo.gl/YYPcyY for details.
```

With the first shortlink going to our npm scripts guide, and the second going to our configuration JSDocs. (Both can be updated if/when we get more relevant docs at different URLs, but those were the best places to point to as of right now.)

I really want to hit home the fact that making `workbox generateSW` part of your build is the recommended course of action. I also want to leave folks with guidance about how to modify the basic configuration, since confusion about what configuration options are supported seems to come up frequently.

jeffposnick:
GitHub's hidden this in its UI, so here's my response to https://github.com/GoogleChrome/workbox/pull/965#discussion_r148853217:

I've further tweaked the message after your initial feedback. It now looks like:

```sh
Your config has been saved. To build your service worker, run

  workbox generateSW ${configLocation}

as part of a build process. See https://goo.gl/8zZy1P for details.
You can further customize your service worker by making changes to ${configLocation}.
See https://goo.gl/YYPcyY for details.
```

With the first shortlink going to our npm scripts guide, and the second going to our configuration JSDocs. (Both can be updated if/when we get more relevant docs at different URLs, but those were the best places to point to as of right now.)

I really want to hit home the fact that making `workbox generateSW` part of your build is the recommended course of action. I also want to leave folks with guidance about how to modify the basic configuration, since confusion about what configuration options are supported seems to come up frequently.

gauntface:
@jeffposnick I still feel like this is going to get overlooked. You need like a next step kind of lead on this message.

Unless I see something in the first sentence that says more work is needed, I reckon people will ignore the wall of text and go looking to test their new service worker.

jeffposnick:
I can drop `Your config has been saved.` and then the first sentence becomes `To build your service worker, run...`

PR #968, opened 2017-11-01
Update indexedDBHelper and DBWrapper

philipwalton:

This PR updates our DBWrapper implementation to use a WIP general-purpose IDB wrapper library I'm developing. Note that in this PR I've just updated the code to the new API; I haven't taken advantage of any of the new features. We can do that in another PR.

Notable changes:

1) I've removed the `indexedDBHelper` export and now just export a `DBWrapper` class, which packages can consume directly.

2) The `DBWrapper` class handles opening the connection behind the scenes (prior to making any transactions), so our code doesn't have to worry about it. This allows us to create instances in constructor functions since it happens sync.

```js
// No connection is made by doing this.
const db = new DBWrapper('my-db', 1);

// This calls `await db.open()` in the background
const entries = await db.getAll('my-store');
```

3) `DBWrapper` has a general purpose query function called `getAllMatching()`, which is similar to `getAll()` but it supports specifying an [index](https://developer.mozilla.org/en-US/docs/Web/API/IDBIndex), [direction](https://developer.mozilla.org/en-US/docs/Web/API/IDBCursor/direction), [key range](https://developer.mozilla.org/en-US/docs/Web/API/IDBKeyRange), and count (like "limit" in SQL) as well as optionally returning the keys (useful for `autoIncrement:true`).

There are places in our code where we call `getAll()` and then filter the results. This is inefficient since in many cases we can do the filter via the `getAllMatching()`. For example, `workbox-cache-expiration` can request only entries with timestamps greater than X by doing something like this:

```js
const expiredEntries = await db.getAllMatching('store-name', {
  index: 'timestamp',
  query: IDBKeyRange.lowerBound(X),
});
```

Or if you wanted to get just the entry with the newest timestamp, you can do something like this:

```js
const [oldestEntry] = await db.getAllMatching('store-name', {
  index: 'timestamp',
  direction: 'prev',
  count: 1,
});
```

Both of these are much more efficient than doing `getAll()` since they only search/return a subset of the object store.


philipwalton:
BTW, this PR has a total prod size increase of 0.75K. If I remove some of the class methods that we're not using, I can shave 0.58K off of that (making the delta only +0.17K). However, my concern is that if I release this as a standalone library, I'd almost certainly want to keep those methods included.

What do folks think about the trade-offs here?

packages/workbox-core/_private/DBWrapper.mjs
```diff
@@ -0,0 +1,351 @@
...
+   *   1. It's usually better to keep a connection open since opening
+   *      a new connection is somewhat slow.
+   *   2. Connections are automatically closed when the reference is
+   *      garbage collected.
+   * The primary use case for needing to close a connection is when another
+   * reference (typically in another tab) needs to upgrade it and would be
+   * blocked by the current, open connection.
+   */
+  close() {
+    if (this._db) this._db.close();
```
jeffposnick: I guess it's not something linting picks up, but I'm under the impression that we try to avoid single-line `if()`s.

packages/workbox-core/_private/DBWrapper.mjs
```diff
@@ -0,0 +1,351 @@
...
+
+    const checkIfDone = () => {
+      if (called) return;
+
+      let successfulRequests = 0;
+      for (const request of requests) {
+        if (request.readyState == 'done') {
+          // Exit early if there's any errors. Errors will bubble to the
+          // transaction, so we don't have to report with them here.
+          if (request.error) return;
```
jeffposnick: See earlier comment about single-line `if()`s.

packages/workbox-core/_private/DBWrapper.mjs
```diff
@@ -0,0 +1,351 @@
...
+   * `transaction()` so any errors are automatically handled.
+   *
+   * @param {Array<IDBRequest>} requests
+   * @param {function(Array<*>):undefined} callback
+   */
+  static onsuccessAll(requests, callback) {
+    let called = false;
+
+    const checkIfDone = () => {
+      if (called) return;
```
jeffposnick: See earlier comment about single-line if()s.

packages/workbox-core/_private/DBWrapper.mjs
philipwalton: This is allowed in Google style, see [section 4.1.1](https://google.github.io/styleguide/jsguide.html#formatting-braces-all).

I prefer the single-line format for things that are conceptually simple like this. Personally I find that when these lines take up too much room in a function, the logic is harder to follow.

I'm happy to change if folks all disagree, but I wanted to point out that it's not against Google JS style.

packages/workbox-core/_private/DBWrapper.mjs
```diff
@@ -0,0 +1,351 @@
...
+   * blocked by the current, open connection.
+   */
+  close() {
+    if (this._db) this._db.close();
+  }
+
+  /**
+   * Deletes the database.
+   */
+  async deleteDatabase() {
```
gauntface: This method looks like it's only being used in the tests, if this is the case it should probably be moved out of the source code and into `infra/testing/`.

packages/workbox-background-sync/models/QueueStore.mjs
```diff
@@ -60,55 +58,19 @@ export class QueueStore {
    * value as a StorableRequest instance. If no entry exists, it returns
    * undefined.
    *
-   * @return {Promise<StorableRequest|undefined>}
+   * @return {StorableRequest|undefined}
```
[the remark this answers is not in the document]
philipwalton: Yeah, I was thinking about bringing this up to the group as a style guide issue.

Initially I was all for being explicit about the fact that async functions return promises, but now I'm starting to lean the other way (partially since it plays better with current tools).

The other thing I've been doing is not mixing async and promises. I.e. even if I have a variable that I know is a promise, I `return await` it, since my understanding is it leads to [better stack traces](https://mathiasbynens.be/notes/async-stack-traces).

Anyway, I guess this is a larger discussion...

packages/workbox-core/_private/DBWrapper.mjs
```diff
@@ -0,0 +1,351 @@
...
+        } else if (this._onupgradeneeded) {
+          this._onupgradeneeded(evt);
+        }
+      };
+      openRequest.onsuccess = (evt) => {
+        const db = evt.target.result;
+        if (openRequestTimedOut) {
+          db.close();
+        } else {
+          db.onversionchange = this._onversionchange;
```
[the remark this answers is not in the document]
philipwalton: Not quite. I was confused by this as well, but after looking into it I realized that the purpose of `onversionchange` is so *other* instances can "relinquish" (or optionally not relinquish) their control over the database.

The IDB spec ensures only one version of a database is ever running at a time, so if an `open()` call requests a new version, all existing open references to that database will get the `onversionchange` event (with the assumption that they will call `close()`, so the requester can successfully upgrade). If any of the other open instance don't call `close()` immediately, the requestor gets the `blocked` event (the idea being that it can then show a message to the user asking them to close other open tabs to this site).

The timeout is implemented because there are [cases when an open request will never get any event](https://github.com/w3c/IndexedDB/issues/223), so we have to handle that case too.

The logic in [these tests](https://github.com/GoogleChrome/workbox/blob/194b512aa3b654194ebd344c2cc227f7809d4fda/test/workbox-core/node/_private/test-DBWrapper.mjs#L186-L223) might help make this more clear.

packages/workbox-core/_private/DBWrapper.mjs
philipwalton: Yeah, I mention this is https://github.com/GoogleChrome/workbox/pull/968#issuecomment-341247613, but after thinking about it a little more, I realized when I release this as a standalone package I can separate out the `deleteDatabase` and `onsuccessAll` methods into standalone helpers.

That should cut down on the size a bit.

packages/workbox-core/_private/DBWrapper.mjs
```diff
@@ -0,0 +1,351 @@
...
+  /**
+   * A helper that makes it easier to run IDBRequests in parallel. This
+   * method takes an array of IDBRequests and runs the callback once all of
+   * them are successful. This method should be used along with the
+   * `transaction()` so any errors are automatically handled.
+   *
+   * @param {Array<IDBRequest>} requests
+   * @param {function(Array<*>):undefined} callback
+   */
+  static onsuccessAll(requests, callback) {
```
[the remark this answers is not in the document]
philipwalton: This is one of these cases where I can't use promises at all because in Safari/FF the transaction will complete before the promise resolves.

But as I say in https://github.com/GoogleChrome/workbox/pull/968#discussion_r148584683, since this is just a helper (and not actually used in any of our code) I will remove it.

philipwalton:
I've removed some of those unused helpers, and now the total size has actually gone down 150B \o/.

packages/workbox-background-sync/models/QueueStore.mjs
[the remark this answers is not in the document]
jeffposnick: (As an aside, https://eslint.org/docs/rules/no-return-await is not enabled, so there's nothing from the linting perspective stopping up from using `return await...`.)

packages/workbox-core/_private/DBWrapper.mjs
jeffposnick: Okay, thanks for clarifying. I'm cool if our style guide is cool (though I don't know that I'm going to start doing it).

packages/workbox-background-sync/models/QueueStore.mjs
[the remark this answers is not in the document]
gauntface: For internal functions it's not an issue but for developer facing docs it needs to be returning a promise regardless of anything internally

gauntface:
LGTM from me - can't approve cos mobile site is Le pants.

packages/workbox-background-sync/models/QueueStore.mjs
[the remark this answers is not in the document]
philipwalton: Agreed for public facing functions, though It would be great if the docs could automatically detect that it's an async function and indicate that accordingly.

Issue #969, opened 2017-11-01
Stop creating dev + prod builds of workbox-sw


jeffposnick:
**Library Affected**:
*workbox-sw*

The code in both builds is currently the same. The only different is that the `dev` build is unminified (but we have a sourcemap available).

PR #973, opened 2017-11-02
Make generateSW() copy a larger set of libraries

jeffposnick:
R: @addyosmani @gauntface

This updates the `workbox-build.generateSW()` behavior so that when `importWorkboxFromCDN` is false, all(*) of the Workbox libraries (both `dev` and `prod` builds, including the sourcemaps) are copied over to the output directory, instead of just the `workbox-sw` library.

There are some pending TODOs to update the logic when we move to a single `workbox-sw.js` file instead of `workbox-sw.prod.js` (see #969) and also to deal with the eventual migration of the library path metadata from the `browser` field of each module's `package.json` to the `main` field.

I'll have to update the generated service worker file to make proper use of these libraries, but that's going to be part of a separate PR to update the service worker template v3-compatible.

_(*) I'm copying all the libraries that might actually be used by the service worker created by `generateSW()`. If a user needs, e.g., the background sync library, my assumption is that they're using `injectManifest()` and we don't need to deal with that in this PR. See https://github.com/GoogleChrome/workbox/issues/976_


packages/workbox-build/src/entry-points/generate-sw.js
```diff
@@ -17,13 +17,16 @@
...
-const copyWorkboxSW = require('../lib/copy-workbox-sw');
+const copyWorkboxLibraries = require('../lib/copy-workbox-libraries');
 const generateSWSchema = require('./options/generate-sw-schema');
 const getFileManifestEntries = require('../lib/get-file-manifest-entries');
 const validate = require('./options/validate');
 const writeServiceWorkerUsingDefaultTemplate =
   require('../lib/write-sw-using-default-template');
 
+// TODO: Update for https://github.com/GoogleChrome/workbox/issues/969.
+const WORKBOX_SW_FILENAME = 'workbox-sw.prod.js';
```
gauntface: Why does this need to be hardcoded? Code it use the file name in the workbox-sw package.json browser (soon to be main) field?

packages/workbox-build/src/entry-points/generate-sw.js
```diff
@@ -47,35 +50,29 @@ async function generateSW(input) {
...
 
-    // Add a few extra ignore rules to whatever might be specified to avoid
-    // picking up the generated service worker or the workbox-sw files.
+    // The Workbox library files should not be precached, since they're cached
+    // automatically by virtue of being used with importScripts().
     options.globIgnores = [
-      path.basename(pathToWorkboxSWFileRelativeToDest),
-      `${path.basename(pathToWorkboxSWFileRelativeToDest)}.map`,
-    ].map((file) => `**/${file}`).concat(options.globIgnores);
+      `**/${workboxDirectoryName}/*.js*`,
```
gauntface: Could you make sure there are tests that cover the workbox directory being at the root of the glob directory as well as being nested.

Just unsure if glob patterns will match if there is no nesting with this pattern. (Only mentioning because of the number of issues people raised after my first attempt at workbox-build)

packages/workbox-build/src/entry-points/generate-sw.js
```diff
@@ -47,35 +50,29 @@ async function generateSW(input) {
...
+    ].concat(options.globIgnores || []);
 
     // importScripts may or may not already be an array containing other URLs.
-    options.importScripts = (options.importScripts || []).concat(
-      pathToWorkboxSWFileRelativeToDest);
+    // Either way, list WORKBOX_SW_FILENAME first.
+    options.importScripts = [
+      `${workboxDirectoryName}/${WORKBOX_SW_FILENAME}`,
```
gauntface: Few comments:

1. If WORKBOX_SW_FILENAME isn't used anywhere else - just inline it here and move the comment.
1. This feels a little error prone as you'll need to update the `modulePathPrefix` for this to use the local files.

packages/workbox-build/src/lib/errors.js
```diff
@@ -45,8 +45,8 @@ module.exports = {
   'sw-write-failure': 'Unable to write the service worker file.',
   'sw-write-failure-directory': ol`Unable to write the service worker file;
     'swDest' should be a full path to the file, not a path to a directory.`,
-  'unable-to-copy-workbox-sw': ol`workbox-sw is needed by the service worker
-    and could not be copied over to your new site.`,
+  'unable-to-copy-workbox-libraries': ol`One or more libraries needed by your
```
gauntface: I might change the wording of this to make it sound more like a workbox problem. This sounds like it could be the developers problem. 

"One of more of the Workbox librares could not be copied over to '${destDirectory}'"

test/workbox-build/node/lib/copy-workbox-libraries.js
```diff
@@ -0,0 +1,49 @@
...
+    const ensureDirStub = sinon.stub().resolves();
+
+    const copyWorkboxLibraries = proxyquire(MODULE_PATH, {
+      'fs-extra': {
+        copy: copyStub,
+        ensureDir: ensureDirStub,
+      },
+    });
+
+    const workboxDirectory = await copyWorkboxLibraries(DEST_DIRECTORY);
```
gauntface: This is writing to `/test`, what happens if you write to './test'?

test/workbox-build/node/lib/copy-workbox-libraries.js
```diff
@@ -0,0 +1,49 @@
...
+    // The workboxDirectory value is a relative path from DEST_DIRECTORY to the
+    // new directory. We check if ensureDir was called with the full path.
+    expect(path.join(DEST_DIRECTORY, workboxDirectory)).to.eql(ensureDirStub.args[0][0]);
+
+    // This reflects the number of library files that were copied over:
+    // - both prod and dev builds
+    // - source maps for each
+    expect(copyStub.callCount).to.eql(20);
+  });
```
gauntface: It might be good to add a test that actually copies the files over and ensures the appropriate files are copied over.

packages/workbox-build/src/entry-points/generate-sw.js
```diff
@@ -17,13 +17,16 @@
 const path = require('path');
 
 const cdnUtils = require('../lib/cdn-utils');
-const copyWorkboxSW = require('../lib/copy-workbox-sw');
+const copyWorkboxLibraries = require('../lib/copy-workbox-libraries');
```
gauntface: Can you export this method as part of the `workbox-build` API such that anyone using the injectManifest can easily get all the files.

jeffposnick: Yup. I'll add some JSDocs as well.

packages/workbox-build/src/entry-points/generate-sw.js
jeffposnick: Good point.

packages/workbox-build/src/entry-points/generate-sw.js
jeffposnick: I've switched to inferring it from `workbox-sw`'s `package.json`.

packages/workbox-build/src/lib/errors.js
jeffposnick: Updated.

test/workbox-build/node/lib/copy-workbox-libraries.js
jeffposnick: I've added a test case for that.

test/workbox-build/node/lib/copy-workbox-libraries.js
jeffposnick: The e2e test actually copies the files over, but is currently just checking for the file count as well. I'll update that test to check the filenames.

packages/workbox-build/src/entry-points/generate-sw.js
jeffposnick: Sure, I've added an end-to-end test that specifically exercises the `globIgnores` effect for both top- and sub-directory matches.

jeffposnick:
Thanks for the feedback—I think everything's been address.

Regarding cleanup: the new method doesn't attempt to clean up copies of older versions of the libraries. If you'd like to see that functionality, we can track it via a separate issue. (My inclination is not to make what could possibly be the wrong guess about what to delete.)

PR #974, opened 2017-11-02
Runtime logging

gauntface:

After testing out the Router + Runtime Strategies I noticed the logging was getting grouped incorrectly because the asynchronous nature of fetching and responding - groups were getting nested.

This PR undoes these issues and at least makes it possible to follow what is happening at the top level.

I only tidied up the Router slightly, just using early returns instead of nesting if statements.


packages/workbox-runtime-caching/CacheFirst.mjs
```diff
@@ -93,19 +92,28 @@ class CacheFirst {
...
+        logs.push(
+          `Found a cached response in the '${this._cacheName}' cache.`);
       }
     }
 
     if (process.env.NODE_ENV !== 'production') {
+      logger.groupCollapsed(
+        messages.strategyStart('CacheFirst', event));
+      for (let log of logs) {
+        if (Array.isArray(log)) {
```
jeffposnick: Is `log` ever going to be an array? I'm not seeing any instances where `logs.push()` is called with anything other than a single string.

gauntface: Removed.

PR #982, opened 2017-11-04
Adding default export with simple wrappers for runtime-caching

gauntface:
R: @jeffposnick @addyosmani 

`workbox-sw` has a wrapper around the strategies to make it more inline with sw-toolbox.

The wrapper simply consists of a function that returns an instance of a caching strategy configured with plugins.


packages/workbox-runtime-caching/utils/pluginBuilder.mjs
```diff
@@ -0,0 +1,44 @@
...
+
+  // TODO: Assert plugins is an array
+
+  const pluginParamsToClass = {
+    cacheExpiration: CacheExpirationPlugin,
+    // TODO: Add support for 'broadcastCacheUpdate': BroadcastCacheUpdatePlugin,
+    // TODO: Add support for 'cacheableResponse': CacheableResponsePlugin,
+  };
+
+  // Iterate over known plugins and add them to Request Wrapper options.
```
jeffposnick: Instead of iterating over known plugins, how about iterating over the `options`, and then logging in `dev` mode when there's an unsupported option used? That sounds more in keeping with our v3 approach to reporting problematic usage that silently ignoring the unsupported config.

Something like:

```js
for (const [pluginName, config] of Object.entries(options)) {
  // Special case for the `cacheName` config.
  if (pluginName === 'cacheName') {
    continue;
  }

  const PluginClass = pluginParamsToClass[pluginName];
  if (PluginClass) {
    plugins.push(new PluginClass(config));
  } else if (process.env.NODE_ENV !== 'production') {
    // Log something about pluginName not being supported.
  }
}
```

PR #989, opened 2017-11-06
Adding some more context to the precaching docs

### LINKED ISSUE #984: [workbox-precaching] JSDoc: Do I need to call addRoute()?
Looking at the JSDoc for `precaching` I can’t quite tell if it is enough to just call `precache()` or if I *must* to call `addRoute()` as well to make the module handle requests.

gauntface:
R: @jeffposnick @addyosmani @surma 

Fixes #984 

Docs on precaching and `addRoute`


surma:
LGTM

jeffposnick:
👍

PR #1003, opened 2017-11-07
Publishing fixes

gauntface:
I noticed a few issues with the publishing of CDN / demos.

`workbox-sw` was loading older versions of modules even though the released version of `workbox-sw` was newer and the files were available.

The issue was that `workbox-build` was being used to determine the CDN URL but there was no depedency on `workbox-sw` and `workbox-build` meaning that Lerna can build them in order they wanted.

I've removed the versioned function from `workbox-build` and instead switched demos and rollup reference to a function in the build tools that will use the latest lerna version directly.

### DIFF
--- a/gulp-tasks/publish-demos.js
+++ b/gulp-tasks/publish-demos.js
@@ -2,14 +2,15 @@ const gulp = require('gulp');
 const path = require('path');
 const fs = require('fs-extra');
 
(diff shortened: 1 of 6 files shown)


jeffposnick:
`workbox-build` should have a dependency on `workbox-sw` now that https://github.com/GoogleChrome/workbox/pull/973 is merged, but keeping the code separate from `workbox-build` sounds cleaner in general.

PR #1009, opened 2017-11-08
Move workbox-build's template over to v3 syntax

jeffposnick:
R: @addyosmani @gauntface

This PR includes the changes required to get the auto-generated service worker to use the v3 syntax.

Along with the new syntax, I've tweaked the comments in the generated service worker to point folks in the right direction for registration and serving best practices. I'm not sure if folks will read the comments, but it can't hurt.


infra/testing/validator/service-worker-runtime.js
```diff
@@ -6,49 +6,56 @@ const sinon = require('sinon');
...
+
   return {context, methodsToSpies};
 }
 
 function validateMethodCalls({methodsToSpies, expectedMethodCalls}) {
-  for (const method of Object.keys(expectedMethodCalls)) {
-    const spy = methodsToSpies[method];
-    expect(spy.args).to.deep.equal(expectedMethodCalls[method],
-      `while testing method calls for ${method}`);
+  for (const [method, spy] of Object.entries(methodsToSpies)) {
```
jeffposnick: This will now not only check that the expected methods were called, but also check that the known methods which *shouldn't* have been called actually *weren't*, which is an improvement.

packages/workbox-build/src/lib/populate-sw-template.js
```diff
@@ -47,22 +44,27 @@ module.exports = ({
       undefined,
   };
 
-  let workboxOptionsString = JSON.stringify(workboxOptions, null, 2);
+  let precacheOptionsString = JSON.stringify(precacheOptions);
```
gauntface: Personally I liked the indented version, especially as it meant it was more human readable.

packages/workbox-build/src/templates/sw-template.js
```diff
@@ -14,23 +14,41 @@
...
+ *
+ * Here are some next steps:
+ *
+ * - Your web app needs to register this file.
+ *   See https://goo.gl/DNGzMp
+ *
+ * - Disable HTTP caching for this file.
+ *   See https://goo.gl/rWuKgq
+ *
+ * The rest of the code in this file is auto-generated. Please don't update this
```
gauntface: Nit: drop "in this file" on this line, reading from top to bottom there is "this file" referenced in previous bullet points and sounds repitive here.

packages/workbox-build/src/templates/sw-template.js
```diff
@@ -14,23 +14,41 @@
...
 /**
- * This service worker file needs to be registered by your web application.
- * Please see https://developers.google.com/web/fundamentals/primers/service-workers/registration
- */
-<% if(manifestEntries) {%>
-/**
- * DO NOT EDIT __precacheManifest MANUALLY!
- *
- * The workboxSW.precache() method does the following:
+ * The workboxSW.precacheAndRoute() method does the following:
```
gauntface: In Workbox v3 branch there is a pretty thorough explanation of what precaching is and how it works. Could we shorten this comment to:

"// Efficiently cache and server requests
 // Learn more https://developers.google.com/web/tools/Workbox/modules/workbox-precaching"

PR #1016, opened 2017-11-09
Create a single (prod) build of workbox-sw.

### LINKED ISSUE #969: Stop creating dev + prod builds of workbox-sw

jeffposnick:
R: @addyosmani @gauntface

Fixes #969 

Assuming this is merged, I'll go through the CDN publishing steps from start to finish to confirm that everything works for me, and to get the new `workbox-sw.js` pushed out there.

### DIFF
--- a/packages/workbox-build/src/lib/cdn-utils.js
+++ b/packages/workbox-build/src/lib/cdn-utils.js
@@ -24,7 +24,10 @@ const getVersionedCDNUrl = () => {
 };
 
 const getModuleUrl = (moduleName, buildType) => {
-  return `${getVersionedCDNUrl()}/${moduleName}.${buildType.slice(0, 4)}.js`;
+  if (buildType) {
+    return `${getVersionedCDNUrl()}/${moduleName}.${buildType.slice(0, 4)}.js`;
+  }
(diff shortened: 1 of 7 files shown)


packages/workbox-build/src/lib/cdn-utils.js
```diff
@@ -24,7 +24,10 @@ const getVersionedCDNUrl = () => {
 };
 
 const getModuleUrl = (moduleName, buildType) => {
-  return `${getVersionedCDNUrl()}/${moduleName}.${buildType.slice(0, 4)}.js`;
+  if (buildType) {
```
gauntface: Warning: this is pedantic.

Is there any chance you'd be willing to get the modules package.json and make sure the build type is valid (i.e. if pkg.workbox.prodOnly && buildtype && buildType === 'dev' then throw).

Only reason is that I'd love for this function to be as fool proof as possible just so further down the line we don't get caught out.

The inverse is also possible when it should be if !pkg.workbox.prodOnly && !buildType throw buildType required.

PR #1027, opened 2017-11-10
Fixing browser namespace exports etc.

### LINKED ISSUE #1021: Imports and Browser Namespace can be incorrect
Noticed this error in the final browser builds.

The expected browser namespace for `workbox-cache-expiration` is:

```javascript
workbox.expiration.CacheExpiration;
workbox.expiration.CacheExpirationPlugin;
```

In `workbox-runtime-caching` we have imports like this:

```javascript
import {CacheExpirationPlugin} from 'workbox-cache-expiration/CacheExpirationPlugin.mjs';
```

This is completely valid. The reason for the named export is to avoid a "hasOwnProperty" that rollup adds if we use `export default`.

The current build process will swap the `workbox-cache-expiration/CacheExpirationPlugin.mjs` import with the browser namespace `workbox.expiration.CacheExpirationPlugin`, which is the expected API above.

Because of the named export however, the expectation is that `workbox.expiration.CacheExpirationPlugin` will then export a Class `CacheExpirationPlugin` meaning a complete browser namespace of `workbox.expiration.CacheExpirationPlugin.CacheExpirationPlugin`.

Possible fixes:

- Instead of converting imports `<module name>/<file name>.mjs` to `workbox.<module namespace>.<file name>` we can just swap out to `workbox.<module namespace>`.
    - **Cons**: This will break the current usage of `_private.mjs` in core, but we could just move those files to the top level and label them underscore and balloon the workbox.core namespace.
- Instead of importing the files, just import named exports from `_public.mjs` (same as `_private.mjs`).
    - **Cons**: This will balloon webpack builds.(NOTE: This is my understanding from what Phil has raised issues about)

gauntface:

Fixes #1021 

Previously an `import {SomeFile} from 'workbox-*/SomeFile.mjs'` would convert to 
`workbox.*.SomeFile`. This was incorrect in the runtime code since the named exports would look in the namespace `workbox.*.SomeFile.SomeFile`. To fix this I switched the browser bundle to not include the file name in the name space i.e. `import {SomeFile} from 'workbox-*/SomeFile.mjs'` => `workbox.*`. This broke the pattern for `_private.mjs` which was a hack to get around the top level file restriction. I special cased the `_private` directory so that you can do `import {SomeFile} from 'workbox-*/_private/SomeFile.mjs'` which will change to `workbox.*._private`.

I then found an issue where the context for plugins (i.e. in the plugin `this` was undefined). So I ended up cleaning that up with `.call` like Phil is doing when he calls callbacks. This should prevent any edge cases for developers using this as well.


jeffposnick:
This all seems reasonable to me.

I'm not sure if you always want a 👍 from @philipwalton given that he's had a deeper involvement with the custom build use cases.

PR #1028, opened 2017-11-10
Support for registerRoute(<string>)

### LINKED ISSUE #1012: Treat registerRoute('/foo/:bar', ...) as exact match in v3
**Library Affected**:
*workbox-routing*

Following up from https://github.com/GoogleChrome/workbox/issues/1011, in v3 we are going to stop creating `ExpressRoute`s implicitly whenever there's a string argument passed to the `registerRoute()` helper.

Instead, strings will be tested for an exact match, either against the `url.path` (if the string begins with `'/'`) or against the `url.href` (if the string begins with `http`).

We'll continue the logging that https://github.com/GoogleChrome/workbox/issues/1011 introduces, and ensure that we direct developers to documentation about constructing an equivalent `RegExpRoute` and/or creating their own custom `matchHandler` as alternatives.

jeffposnick:
R: @addyosmani @gauntface

Fixes #1012

### DIFF
--- a/packages/workbox-routing/_default.mjs
+++ b/packages/workbox-routing/_default.mjs
@@ -14,7 +14,12 @@
   limitations under the License.
 */
 
-import {assert, WorkboxError, cacheNames} from 'workbox-core/_private.mjs';
+import {
+  WorkboxError,
+  assert,
+  cacheNames,
+  logger,
+} from 'workbox-core/_private.mjs';
 import {Router} from './Router.mjs';
 import {Route} from './Route.mjs';
 import {RegExpRoute} from './RegExpRoute.mjs';
@@ -30,7 +35,7 @@ if (process.env.NODE_ENV !== 'production') {
  */
 class DefaultRouter extends Router {
   /**
-   * Easily register a Regular Expression or function with a caching
+   * Easily register a RegExp, string, or function with a caching
    * strategy to the Router.
    *
    * This method will generate a Route for you if needed and
@@ -39,6 +44,7 @@ class DefaultRouter extends Router {
    *
    * @param {
    * RegExp|
+   * string|
    * workbox.routing.Route~matchCallback|
    * workbox.routing.Route
    * } capture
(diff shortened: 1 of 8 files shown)


packages/workbox-routing/_default.mjs
```diff
@@ -14,7 +14,12 @@
   limitations under the License.
 */
 
-import {assert, WorkboxError, cacheNames} from 'workbox-core/_private.mjs';
+import {
+  WorkboxError,
+  assert,
+  cacheNames,
+  logger,
+} from 'workbox-core/_private.mjs';
```
gauntface: This will need to be changed to:

```javascript
import {WorkboxError} from 'workbox-core/_private/WorkboxError.mjs';
import {assert} from 'workbox-core/_private/assert.mjs';
...
```

packages/workbox-routing/_default.mjs
```diff
@@ -54,20 +60,65 @@ class DefaultRouter extends Router {
...
+        if (!(capture.startsWith('/') || capture.startsWith('http'))) {
+          throw new WorkboxError('invalid-string', {
+            moduleName: 'workbox-routing',
+            className: 'DefaultRouter',
+            funcName: 'registerRoute',
+            paramName: 'capture',
+          });
+        }
+
+        // We want to only prohibit these characters in the pathname portion of
```
gauntface: Can you change this to "We want to check if express style characters are in the pathname only."

packages/workbox-routing/_default.mjs
```diff
@@ -54,20 +60,65 @@ class DefaultRouter extends Router {
...
+        // the URL.
+        const valueToCheck = capture.startsWith('http') ?
+          new URL(capture).pathname :
+          capture;
+        // See https://github.com/pillarjs/path-to-regexp#parameters
+        const wildcards = '[*:?+]';
+        if (valueToCheck.match(new RegExp(`${wildcards}`))) {
+          throw new WorkboxError('invalid-wildcards', {
```
gauntface: I think you can have these characters in a URL (i.e. search parameters) so this really shouldn't be an error but should just be a warning.

packages/workbox-routing/_default.mjs
```diff
@@ -54,20 +60,65 @@ class DefaultRouter extends Router {
...
+          });
+        }
       }
-      route = new ExpressRoute(capture, handler, method);
-    } **/
-    if (capture instanceof RegExp) {
+
+      const matchCallback = ({url}) => {
+        // If we have a path-only capture pattern...
+        if (capture.startsWith('/')) {
```
gauntface: This if statement feels error prone.

Wouldn't this have the same result:
```javascript
// If capture is full URL, new URL will do nothing, if it's '/' it'll have same origin
return (url.href === new URL(capture, location).href);
```

You could nest a debug message to call out "the pathname matches but the origin is different so if you intended a match, please use the full URL".

jeffposnick:
PTAL. It should have pr-bot properly configured now as well.

PR #1029, opened 2017-11-10
Bot fix

gauntface:
Trying to fix le bot.


jeffposnick:
I could imagine that we are legitimately over aggregate budget now, since we previously weren't picking up `workbox-sw` or `workbox-runtime-caching`.

I'm not sure why there are only 7 items in the `All File Sizes` table, though? I don't *think* that I touched that code, unless it's something that got wonky when the `browser/` subpath was removed from `build/`...

jeffposnick:
👍 Thanks for cleaning this up.

PR #1035, opened 2017-11-10
Adding log for opaque responses

gauntface:
R: @jeffposnick @addyosmani 

@jeffposnick is there anywhere else you think it would be helpful to log when we are working with opaque responses and drop them?

### DIFF
--- a/packages/workbox-core/_private/cacheWrapper.mjs
+++ b/packages/workbox-core/_private/cacheWrapper.mjs
@@ -137,8 +137,9 @@ const _isResponseSafeToCache = async (request, response, plugins) => {
     if (process.env.NODE_ENV !== 'production') {
       if (!responseToCache.ok) {
         if (responseToCache.status === 0) {
+          // TODO: Add a link to guide on third-party request handling
           logger.warn(`The response for '${request.url}' is an opaque ` +
-            `response. Workbox will not cache this by default.`);
+            `response. Workbox does not cache opaque responses by default.`);
         } else {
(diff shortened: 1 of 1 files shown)


packages/workbox-core/_private/cacheWrapper.mjs
```diff
@@ -134,6 +134,19 @@ const _isResponseSafeToCache = async (request, response, plugins) => {
   }
 
   if (!pluginsUsed) {
+    if (process.env.NODE_ENV !== 'production') {
+      if (!responseToCache.ok) {
+        if (responseToCache.status === 0) {
+          // TODO: Add a link to guide on third-party request handling
+          logger.warn(`The response for '${request.url}' is an opaque ` +
+            `response. Workbox does not cache opaque responses by default.`);
```
jeffposnick: It muddies the waters a bit, but a more accurate second sentence would be:

"The caching strategy that you're using will not cache opaque responses by default."

This accounts for the fact that some strategies will cache those responses. I don't think there's any way to get the name of the underlying caching strategy from the context of the `cacheWrapper`, which limits how specific the message could get.

gauntface: I left it vague in that way because I'm not 100% sure it's safe to assume they are using a caching strategy. (i.e.  a plugin could alter it to say it's unsafe despite a normally friendly caching strategy i.e. stalewhilereval or network first)

I'm gonna change it since it's an edge case but can also change back if you agree.

PR #1041, opened 2017-11-13
Fix env-it to pass through done(), and fix up webpack plugin test

jeffposnick:
R: @addyosmani @gauntface

In the course of looking into some of the webpack issues, I realized that the existing webpack plugin test wasn't failing when it should be.

The underlying issue was that the `env-it` module didn't pass through the `done` callback to the wrapped test case.

I've fixed that bug in `env-it`, and also cleaned up in the webpack plugin test so that it's legitimately working with the new v3 service worker format.

I've also modified the test so that it includes an asset with a `[chunkhash]`, in preparation for some further work.

### DIFF
--- a/infra/testing/env-it.js
+++ b/infra/testing/env-it.js
@@ -3,24 +3,49 @@ const constants = require('../../gulp-tasks/utils/constants');
 module.exports = {
   devOnly: {
     it: function(title, cb) {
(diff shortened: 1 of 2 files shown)


jeffposnick:
Hold off on reviewing this—I need to modify it to make it happy with wrapped functions which return promises.

PR #1042, opened 2017-11-13
Exclude revision info when the URL includes a known hash.

### LINKED ISSUE #939: Don’t cache bust urls that already have [chunkhash/hash] specified in webpack builds
Related to https://github.com/GoogleChrome/workbox/pull/808#pullrequestreview-64684918 We should omit the `revision` property of webpack assets if the asset is already using it's own filename cache busting.

> One thing that bears mentioning: as an optimization, the precache() function will not append a cache-busting query parameter when populating its caches if the revision field is not present.

> That's the same behavior that in sw-precache-webpack-plugin was defined by the dontCacheBustUrlsMatching configuration option, for which we recommended using a RegExp that matches URLs that contained hashes in them already.

> I'm not sure exactly how, but is there any indication at this point in the build process whether filePath contains a versioning hash in it? (I'd assume it wouldn't for HTML files, but would for most other assets.) If you can safely infer the presence of a hash, then you can leave out revision and get optimal behavior without developers needing to use dontCacheBustUrlsMatching.

*Note: we should sitill honor `dontCacheBustUrlsMatching` if specified.*

jeffposnick:

Fixes #939, perhaps somewhat naively.

If a URL contains either a `hash`/`fullHash`/`renderedHash` that's associated with the Webpack build, then I *think* it's safe to assume that it is being versioned by webpack, and we don't need to maintain a separate revision field for it.

There's a very small chance of strings that are not actually hashes being present in both the filename and being used internally by webpack, but I don't think that's worth worrying about. This approach seems less likely to lead to false positives than what we were doing with `sw-precache`, where a [basic RegExp](https://github.com/GoogleChromeLabs/sw-precache#dontcachebusturlsmatching-regex) was used to detect hashes and omit revisioning info.

This isn't ready to be merged yet, as I need to update the tests once #1041 is merged, but I wanted to solicit some feedback.

One area that this doesn't address are folks who use `[chunkhash:8]`or `[hash:6]` or some substring of the hash in filenames. We could address that by assuming that, e.g., `:8` is the most common substring and looking for the first 8 characters of each hash in the URLs, but my inclination is hold off on that due to the number of assumptions we'd have to make.

### DIFF
--- a/infra/testing/validator/compare-manifest-entries.js
+++ b/infra/testing/validator/compare-manifest-entries.js
@@ -1,0 +1,15 @@
+const expect = require('chai').expect;
+
+module.exports = (expectedEntries, actualEntries) => {
+  const urlToIndex = new Map(expectedEntries.map((entry, index) => {
+    return [entry.url, index];
+  }));
+
+  expect(actualEntries).to.have.lengthOf(expectedEntries.length,
+    'The expected and actual number of manifest entries do not match.');
+  for (const entry of actualEntries) {
+    expect(urlToIndex.has(entry.url), entry.url).to.be.true;
+    const expectedEntry = expectedEntries[urlToIndex.get(entry.url)];
+    expect(entry).to.eql(expectedEntry);
+  }
+};
--- a/package-lock.json
+++ b/package-lock.json
@@ -2238,6 +2238,12 @@
         }
       }
     },
+    "boolbase": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/boolbase/-/boolbase-1.0.0.tgz",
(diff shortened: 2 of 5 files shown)


jeffposnick:
This is ready for review.

jeffposnick:
Thanks for the feedback, @gauntface.

Re: the first point, you very easily use `[hash]-[name].js` or `[name]-[hash].js` or anything else for your file naming, and I have seen that type of pattern in the wild, so it's not safe to assume that hashes will always be surrounded by `.`.

Re: the second, I've filed https://github.com/GoogleChrome/workbox/issues/1047 since that sounds like something we should do in general, outside of this PR.

jeffposnick:
I'm going to merge; @goldhand, if you have any questions/concerns post-merge, we can address them in a follow-up PR.

PR #1048, opened 2017-11-15
v3: babel-preset-env for browser bundles

### LINKED ISSUE #1043: Use babel-preset-env in v3
We're not currently transpiling the browser-targeted runtime code.

We should start doing that, using the same target level as v2.

The wrinkle is that we only want the polyfills + async-to-generator code included in the `workbox-core` bundle.

jeffposnick:

Fixes #1043 

The code to get `babelHelpers.asyncToGenerator` added to the global scope in `workbox-core` isn't pretty, but after struggling a bit with various Rollup configs, just forcibly injecting it with the `banner` option ended up the most viable approach.

One thing that I'm not 100% sure about is whether there are use cases in which folks might end up consuming the browser bundles of a `workbox-*` module directly, without going through `workbox-sw` and/or without having `workbox-core` loaded.

My understanding is that whenever you use a `workbox-*` browser bundle you need to have at least `workbox-core` loaded already or many other things won't work, so I think we're safe, but I'd love to have confirmation.


gulp-tasks/utils/build-browser-bundle.js
```diff
@@ -127,16 +128,23 @@ module.exports = (packagePath, buildType) => {
   logHelper.log(`    Namespace: ${logHelper.highlight(namespace)}`);
   logHelper.log(`    Filename: ${logHelper.highlight(outputFilename)}`);
 
+  const plugins = rollupHelper.getDefaultPlugins(buildType);
+
+  const banner = pkgJson.workbox.includeBabelHelpers ?
+    fs.readFileSync(path.join(__dirname, 'async-to-generator.js')):
```
philipwalton: Should be a space before the `:` at the end of this line.

gulp-tasks/utils/async-to-generator.js
```diff
@@ -0,0 +1,33 @@
+/* eslint-disable */
+// This is a light wrapper on top of the babel-helpers runtime code, meant to
+// be included in the workbox-core bundle.
+var babelHelpers = {
```
philipwalton: I'd suggest assigning this to `self.babelHelpers`. That makes it more clear that this is intended to be used outside of this module.

gulp-tasks/utils/build-browser-bundle.js
```diff
@@ -127,16 +128,23 @@ module.exports = (packagePath, buildType) => {
   logHelper.log(`    Namespace: ${logHelper.highlight(namespace)}`);
   logHelper.log(`    Filename: ${logHelper.highlight(outputFilename)}`);
 
+  const plugins = rollupHelper.getDefaultPlugins(buildType);
+
+  const banner = pkgJson.workbox.includeBabelHelpers ?
```
philipwalton: I wonder if adding the `includeBabelHelpers` flag makes this more confusing than just hard coding workbox-core. My concern is that someone might start a new module and copy and paste workbox-core's package.json as a starter template and forget to remove this.

If we only want the helpers added in one place, I think it's fine to hard code when that should happen.

gulp-tasks/utils/rollup-helper.js
```diff
@@ -12,6 +13,18 @@ module.exports = {
   getDefaultPlugins: (buildType) => {
     const plugins = [];
 
+    const babelConfig = {
+      presets: [['env', {
+        targets: {
+          browsers: ['chrome >= 51'],
```
philipwalton: Maybe add a comment here as to why we've chosen Chrome 51.

gulp-tasks/utils/async-to-generator.js
philipwalton: I'd also suggest linking to the source of this file on Github.

If there's no easy way to dynamically include this (I looked for a way and couldn't find an obvious solution), then I think the second best thing is to add a comment in the source as to where it came from and suggest periodically looking to see if it's been updated.

philipwalton: Oh, one more thing...

I think we should rename this to `external-helpers.js` or something like that and add a comment explaining why this is the only helper we're including.

I can see a situation where we need to add more helpers in the future, and we might not remember why we only included this one.

philipwalton:
> One thing that I'm not 100% sure about is whether there are use cases in which folks might end up consuming the browser bundles of a workbox-* module directly, without going through workbox-sw and/or without having workbox-core loaded.

I don't think so, at least not for the browser bundles we're providing in the CDN, and that's the primary use case for generating these files, so I think it's OK.

> My understanding is that whenever you use a workbox-* browser bundle you need to have at least workbox-core loaded already or many other things won't work, so I think we're safe, but I'd love to have confirmation.

That's true today AFAIK. The only potential issue is if we introduce a new module with no dependencies, then it potentially wouldn't get these helpers. Personally, I'm not super concerned about that, and I suppose it could be solved by adding a test with an explanation as to why it failed.

jeffposnick:
I've made most of the changes that @philipwalton suggested.

I am keeping the definition as `var babelHelpers =` instead of `self.babelHelpers =` in the hopes that it might pave the way for better minification. Right now, the final prod bundles all include references to `babelHelpers.asyncToGenerator` and if those references could be renamed to something shorter, that would cut down on some bytes. I don't fully understand the wizardry used to minify the v3 codebase, so I'm punting on that and we could track it in a follow-up issue if this PR is merged.

gulp-tasks/utils/build-browser-bundle.js
jeffposnick: You already need to edit the `package.json` `workbox` metadata if you're creating a new module, or else you'll end up reusing namespaces. So copy-and-pasting without making changes is already a non-starter.

I like the model of keeping build-specific metadata for a given package in a central place. Navigating through the various `gulp` tasks and their dependent modules is already a bit daunting without having to also check for hardcoded package-specific logic there.

philipwalton:
> Right now, the final prod bundles all include references to babelHelpers.asyncToGenerator and if those references could be renamed to something shorter, that would cut down on some bytes.

Hmmm, you wouldn't want to rename that unless you could guarantee it was always deterministically renamed to the same thing in all bundles (regardless of other code on those bundles). I'm not sure if uglify supports that.

philipwalton:
Actually, it looks like it is supported: https://github.com/mishoo/UglifyJS2#cli-mangling-property-names---mangle-props

> When you compress multiple files using this option, in order for them to work together in the end we need to ensure somehow that one property gets mangled to the same name in all of them. For this, pass --name-cache filename.json and UglifyJS will maintain these mappings in a file which can then be reused. It should be initially empty. Example:

gulp-tasks/utils/build-browser-bundle.js
philipwalton: SGTM

gulp-tasks/utils/external-helpers.js
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable */
```
gauntface: why disable eslint?

gulp-tasks/utils/external-helpers.js
```diff
@@ -0,0 +1,35 @@
...
+// This is extracted from the Babel runtime, meant to be included once in
+// workbox-core and reused throughout all of the other modules as a way of
+// avoiding code duplication.
+// See https://github.com/babel/babel/blob/9e0f5235b1ca5167c368a576ad7c5af62d20b0e3/packages/babel-helpers/src/helpers.js#L240
```
gauntface: This link doesn't explain much.

What is this code doing, why is it in the  repo and why isn't babel preset env not adding this for us?

gulp-tasks/utils/rollup-helper.js
```diff
@@ -12,6 +13,22 @@ module.exports = {
...
+      presets: [['env', {
+        targets: {
+          // This corresponds to the version of Chroumium used by
+          // Samsung Internet 5.x, which is the minimum non-evergreen browser
+          // we currently support.
+          browsers: ['chrome >= 51'],
+        },
+        modules: false,
+      }]],
+      // The babelHelpers code will be injected into the workbox-core bundle.
```
gauntface: This message doesn't follow the next line referencing a file not in workbox-core.

I'm not following how this is all glued together.

gauntface:
I'm also unsure why modules are going in size quite so much - are they actually sharing pollyfills or are they just needing lots of seperate polyfills resulting and we are just allowing them to be in the respective modules?

jeffposnick:
If we used `babel-preset-env` + Rollup "out of the box" we'd end up injecting identical copies of that `asyncToGenerator` function into each of the browser bundles that use `async`/`await`.

The `babel-preset-env` + Rollup configuration used in this PR does not inject the `asyncToGenerator` function into any of the browser bundles—it outputs code that assumes `babelHelpers.asyncToGenerator` exists in the global scope.

To make sure that there's exactly one definition of `babelHelpers.asyncToGenerator` in the global scope, I'm adding it separately from the Babel transpilation—it's manually inserted as the `banner` used when using Rollup to build the `workbox-core` bundle.

Regarding bundle size, the only polyfill being used is `asyncToGenerator`, and as mentioned, there's only one copy of that. The primary driver in change in bundle size has to do with transpilation required to use the polyfill vs. using native `await`. (Online diff tools are all kind of ugly, but here's a link showing an example of what's changed: https://www.diffnow.com/?report=c85ke)

Mangling the `babelHelpers.asyncToGenerator` function name could help with bundle size, but I'm not sure how to properly do it—see https://github.com/GoogleChrome/workbox/pull/1048#issuecomment-344626396

gauntface:
I don't think you can mangle this with the current setup.

Each module is mangled seperately and anything shared should have a constant name.

But if you can find a way to globally mangle - that would be cool, but could also be error prone.

jeffposnick:
I had played around a bit with using Rollup's `replace` plugin to rename `babelHelpers.asyncToGenerator` to something shorter but bumped into issue, and assumed that there would be an Uglify way of doing things. But if that's not the case, I'll spend more time investigating `replace`.

I'll also beef up the supporting comments.

philipwalton:
@gauntface the uglify option I linked to above says that it's possible to do across bundles, and I think we should look into that in a future PR, as I imagine it'd lead to lots of savings.

@jeffposnick personally I'd still use `self.babelHelpers` since the intent is to assign it to the global object. And we definitely *don't* want that name being minified until we get multi-file minification support added since it's referenced by that name in other files.

addyosmani:
> assumed that there would be an Uglify way of doing things

I wasn't aware that there was a reliable Uglify way of accomplishing this, but as Phil pointed out I guess you can get away with it using mangle props: https://github.com/mishoo/UglifyJS2#cli-mangling-property-names---mangle-props

> Mangling the babelHelpers.asyncToGenerator function name could help with bundle size, but I'm not sure how to properly do it

It doesn't appear (need to take a second look) that mangling asyncToGenerator would give us that much of a bundle size saving even if we were to figure out how to properly do it. 

> @jeffposnick personally I'd still use self.babelHelpers since the intent is to assign it to the global object. And we definitely don't want that name being minified until we get multi-file minification support added since it's referenced by that name in other files.

This sounds reasonable.

gulp-tasks/utils/external-helpers.js
addyosmani: From the comments at least, it isn't clear why we're extracting this from the Babel runtime. If this is for a workaround because babel-preset-env is unable to inject for us, it would be good to clarify that in the comments.

gulp-tasks/utils/external-helpers.js
addyosmani: I didn't follow why that was done here either.

gauntface:
I'm +1 landing this PR as is.

Regarding mangling - we can do that as a final build step is required but should use a specific regex to only mangle known fields and could be very useful to minifiy the workbox-core pieces, but definitely feels risky without integration tests to make sure all the glue works.

jeffposnick:
I've filed https://github.com/GoogleChrome/workbox/issues/1061 to track the mangling of `babelHelpers.asyncToGenerator`.

PR #1049, opened 2017-11-15
v2: Warn on usage of Express-style wildcard characters

### LINKED ISSUE #1011: Add deprecation warning when using ExpressRoutes w/wildcards in v2
**Library Affected**:
*workbox-routing*

As discussed on Slack, we're unlikely to support `ExpressRoute` in v3 (at least not as part of `workbox-routing`). We will instead treat string paths passed in to `registerRoute()` as being exact matches against the `path` or full `href` of the incoming URL.

To pave the way for this new behavior, we should cut a new v2 patch release that adds in a log message explaining the upcoming change in v3. This message should be triggered whenever a developer creates a route using a string that has special meaning in `ExpressRoute`.

Based on https://github.com/pillarjs/path-to-regexp#parameters, the following characters should trigger the message:
- `:`
- `?`
- `*`
- `+`

jeffposnick:
R: @addyosmani @gauntface

Fixes #1011 

Using `warn()` vs. a lower log level is up for negotiation.

I was planning on writing a test to make sure that `logHelper.warn()` was getting called, but then I realized we had issues in v2 with stubbing out code that is internal to the bundle. I can't find any examples of stubbing out `logHelper` methods, but if someone knows of some prior art, I'll copy it and add in some tests.

### DIFF
--- a/packages/workbox-sw/src/lib/router.js
+++ b/packages/workbox-sw/src/lib/router.js
@@ -107,17 +107,18 @@ class Router extends SWRoutingRouter {
       const valueToCheck = capture.startsWith('http') ?
         new URL(capture, location).pathname :
         capture;
-      const match = valueToCheck.match(new RegExp(`${wildcards}`));
-      if (match) {
+      const possibleExpressString = valueToCheck.match(
+        new RegExp(`${wildcards}`));
+      if (possibleExpressString) {
         logHelper.warn({
           message: `registerRoute() was called with a string containing an ` +
-            `Express-style wildcard character. While this is currently ` +
-            `supported, it will no longer be treated as a wildcard match in ` +
-            `an upcoming release of Workbox. For equivalent behavior, please ` +
-            `switch to using a regular expression instead.`,
+            `Express-style wildcard character. In the next version of `+
+            `Workbox, Express-style wildcards won't be supported, and ` +
+            `strings will be treated a exact matches. Please switch to ` +
+            `regular expressions for equivalent behavior.`,
           data: {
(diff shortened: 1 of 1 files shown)


packages/workbox-sw/src/lib/router.js
```diff
@@ -101,6 +102,26 @@ class Router extends SWRoutingRouter {
       if (capture.length === 0) {
         throw ErrorFactory.createError('empty-express-string');
       }
+      // See https://github.com/pillarjs/path-to-regexp#parameters
+      const wildcards = '[*:?+]';
+      const valueToCheck = capture.startsWith('http') ?
+        new URL(capture, location).pathname :
+        capture;
+      const match = valueToCheck.match(new RegExp(`${wildcards}`));
```
gauntface: Nit: Change `match` to `possibleExpressString`

packages/workbox-sw/src/lib/router.js
```diff
@@ -101,6 +102,26 @@ class Router extends SWRoutingRouter {
...
+      if (match) {
+        logHelper.warn({
+          message: `registerRoute() was called with a string containing an ` +
+            `Express-style wildcard character. While this is currently ` +
+            `supported, it will no longer be treated as a wildcard match in ` +
```
gauntface: Simplify the message here.

Suggestion: 
`registerRoute() was called with a string containing an express-style wild card character. In the next version of Workbox express style routes won't be supported by default, instead strings will have to be exact matches. Please switch to regulare expressions`.

PR #1057, opened 2017-11-16
v3: workbox-cacheable-response

jeffposnick:

This re-adds `workbox-cacheable-response` for v3.

An open question I had is whether [`cacheOkAndOpaquePlugin.mjs`](https://github.com/GoogleChrome/workbox/blob/v3/packages/workbox-strategies/plugins/cacheOkAndOpaquePlugin.mjs) should stick around, or whether it should be replaced by an instance of `CacheableResponsePlugin`.

### DIFF
--- a/packages/workbox-cacheable-response/CacheableResponse.mjs
+++ b/packages/workbox-cacheable-response/CacheableResponse.mjs
@@ -21,7 +21,7 @@ import {logger} from 'workbox-core/_private/logger.mjs';
 import './_version.mjs';
 
 /**
- * The `CacheableResponse` class allows you to set up rules determining what
+ * This class allows you to set up rules determining what
  * status codes and/or headers need to be present in order for a
  * [`Response`](https://developer.mozilla.org/en-US/docs/Web/API/Response)
  * to be considered cacheable.
@@ -39,9 +39,9 @@ class CacheableResponse {
    * @param {Object} config
    * @param {Array<number>} [config.statuses] One or more status codes that a
    * `Response` can have and be considered cacheable.
-   * @param {Object<string,string>} [config.headers] One or more headers that a
-   * `Response` can have and be considered cacheable. If multiple headers are
-   * provided, only one needs to be present.
+   * @param {Object<string,string>} [config.headers] A mapping of header names
+   * and expected values that a `Response` can have and be considered cacheable.
+   * If multiple headers are provided, only one needs to be present.
    */
   constructor(config = {}) {
     if (process.env.NODE_ENV !== 'production') {
--- a/packages/workbox-cacheable-response/CacheableResponse.mjs
+++ b/packages/workbox-cacheable-response/CacheableResponse.mjs
@@ -1,0 +1,136 @@
+/*
+  Copyright 2017 Google Inc.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+      https://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+
+import {WorkboxError} from 'workbox-core/_private/WorkboxError.mjs';
+import {assert} from 'workbox-core/_private/assert.mjs';
+import {getFriendlyURL} from 'workbox-core/_private/getFriendlyURL.mjs';
+import {logger} from 'workbox-core/_private/logger.mjs';
+import './_version.mjs';
+
+/**
+ * The `CacheableResponse` class allows you to set up rules determining what
+ * status codes and/or headers need to be present in order for a
+ * [`Response`](https://developer.mozilla.org/en-US/docs/Web/API/Response)
+ * to be considered cacheable.
+ *
+ * @memberof workbox.cacheableResponse
+ */
+class CacheableResponse {
+  /**
+   * To construct a new CacheableResponse instance you must provide at least
+   * one of the `config` properties.
+   *
+   * If both `statuses` and `headers` are specified, then both conditions must
+   * be met for the `Response` to be considered cacheable.
+   *
+   * @param {Object} config
+   * @param {Array<number>} [config.statuses] One or more status codes that a
+   * `Response` can have and be considered cacheable.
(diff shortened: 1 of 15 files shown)


packages/workbox-core/models/messages/messages.mjs
```diff
@@ -59,7 +59,7 @@ export default {
     }
     return `The parameter '${paramName}' passed into ` +
       `'${moduleName}.${className}.${funcName}()' must be an instance of ` +
-      `class ${expectedClass}.`;
+      `class ${expectedClass.name}.`;
```
jeffposnick: This was just something random I ran into when testing out the new code—you need the `.name` or else it doesn't log anything meaningful.

test/workbox-cacheable-response/node/CacheableResponse.mjs
```diff
@@ -0,0 +1,162 @@
...
+        expect(err.details).to.have.property('className').that.eql('CacheableResponse');
+        expect(err.details).to.have.property('funcName').that.eql('isResponseCacheable');
+        expect(err.details).to.have.property('paramName').that.eql('response');
+      });
+    });
+
+    it(`should return true when one of the statuses match the response`, function() {
+      const cacheableResponse = new CacheableResponse({statuses: VALID_STATUSES});
+      const response = new Response('', {status: VALID_STATUS});
+      // TODO: Remove after there's a proper Response mock.
```
jeffposnick: There are incompatibilities with the `.headers` implementation in the default `Response` mock provided by `service-worker-mocks`.

I was going to resolve those incompatibilities by switching over to our own `Response` mock as part of this PR, but unfortunately we have other tests and some module code that relies on the incorrect behavior.

Stay tuned for a separate PR to resolve those issues. Once that's in place, I can take out this manual override.

jeffposnick: https://github.com/GoogleChrome/workbox/pull/1059

packages/workbox-cacheable-response/CacheableResponse.mjs
```diff
@@ -0,0 +1,136 @@
...
+*/
+
+import {WorkboxError} from 'workbox-core/_private/WorkboxError.mjs';
+import {assert} from 'workbox-core/_private/assert.mjs';
+import {getFriendlyURL} from 'workbox-core/_private/getFriendlyURL.mjs';
+import {logger} from 'workbox-core/_private/logger.mjs';
+import './_version.mjs';
+
+/**
+ * The `CacheableResponse` class allows you to set up rules determining what
```
gauntface: Nit: "This Class" will suffice for docs.

packages/workbox-cacheable-response/CacheableResponse.mjs
```diff
@@ -0,0 +1,136 @@
...
+   * To construct a new CacheableResponse instance you must provide at least
+   * one of the `config` properties.
+   *
+   * If both `statuses` and `headers` are specified, then both conditions must
+   * be met for the `Response` to be considered cacheable.
+   *
+   * @param {Object} config
+   * @param {Array<number>} [config.statuses] One or more status codes that a
+   * `Response` can have and be considered cacheable.
+   * @param {Object<string,string>} [config.headers] One or more headers that a
```
gauntface: Nit: "One or more" makes this sounds like an Array is expected. Maybe change to "A mapping of headers and expected values"

packages/workbox-cacheable-response/CacheableResponsePlugin.mjs
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * A class implementing the `cacheWillUpdate` lifecycle callback. This makes it
+ * easier to add in cacheability checks to requests made via Workbox's built-in
+ * strategies.
+ *
+ * @memberof workbox.cacheableResponse
+ */
+class CacheableResponsePlugin {
+  /**
+   * @param {CacheableResponse} cacheableResponse A configured
```
gauntface: Why is this being passed in? This should just either extend or construct the value.

Will be a horrible API otherwise:

```javascript
workbox.cacheable.CacheableResponsePlugin(
  new workbox.cacheable.CacheableResponse({...})
);
```

packages/workbox-cacheable-response/CacheableResponsePlugin.mjs
```diff
@@ -0,0 +1,44 @@
...
+   * `CacheableResponse` instance.
+   */
+  constructor(cacheableResponse) {
+    this._cacheableResponse = cacheableResponse;
+    this.cacheWillUpdate = this.cacheWillUpdate.bind(this);
```
gauntface: This shouldn't be needed due to the way the plugins are called (The context is explicitly set to the plugin itself.

packages/workbox-cacheable-response/CacheableResponsePlugin.mjs
jeffposnick: Right, that's why I was asking about v3 best practices for this in the Slack channel.

I'm happy to change it to take the same parameters as the `CacheableResponse` class and construct an internal instance under the hood. That seems cleaner to me, too.

I based this code on what's being done in https://github.com/GoogleChrome/workbox/blob/v3/packages/workbox-background-sync/QueuePlugin.mjs#L28, and maybe the same change should be considered there? CC: @philipwalton

jeffposnick:
@gauntface, PTAL.

packages/workbox-cacheable-response/CacheableResponsePlugin.mjs
philipwalton: I'm fine with changing the API to have the plugins class take the same params as the underlying class and just expose the plugin methods. The main downside of that is there's code and documentation duplication.

My primary concern with the old model (using class inheritance) was it made it easy for accidental naming collisions, and I don't think we should go back to that.

packages/workbox-cacheable-response/CacheableResponsePlugin.mjs
```diff
@@ -22,12 +23,21 @@ import './_version.mjs';
...
+   * @param {Array<number>} [config.statuses] One or more status codes that a
+   * `Response` can have and be considered cacheable.
+   * @param {Object<string,string>} [config.headers] A mapping of header names
+   * and expected values that a `Response` can have and be considered cacheable.
+   * If multiple headers are provided, only one needs to be present.
    */
-  constructor(cacheableResponse) {
-    this._cacheableResponse = cacheableResponse;
-    this.cacheWillUpdate = this.cacheWillUpdate.bind(this);
+  constructor(config = {}) {
```
addyosmani: Plugin constructor appears to have been changed to not require the cacheable response as input per Matt's request. This change LGTM.

PR #1060, opened 2017-11-17
Broadcast Cache Update

gauntface:
R: @jeffposnick @philipwalton 

Broadcast Cache Update Porting Over.


packages/workbox-broadcast-cache-update/MESSAGE_TYPES.mjs
```diff
@@ -0,0 +1,20 @@
+/*
```
jeffposnick: Instead of `MESSAGE_TYPES.mjs`, could you name this file `constants.mjs` and use a named `MESSAGE_TYPES` export?

Or just name this file `messageTypes.mjs` and keep the default export?

packages/workbox-broadcast-cache-update/BroadcastCacheUpdatePlugin.mjs
```diff
@@ -0,0 +1,64 @@
...
+import {BroadcastCacheUpdate} from './BroadcastCacheUpdate.mjs';
+import './_version.mjs';
+
+/**
+ * This plugin will automatically broadcast a message whenever a cached response
+ * is updated.
+ *
+ * @memberof workbox.broadcastUpdate
+ */
+class BroadcastCacheUpdatePlugin extends BroadcastCacheUpdate {
```
jeffposnick: Could you match what I ended up doing with `CacheableResponsePlugin`: don't extend the base class, but instead take in identical constructor parameters and create an internal instance that you then use in `cacheDidUpdate`?

https://github.com/GoogleChrome/workbox/pull/1057/files#diff-7ecb07c7bc862719cacd71553537e137R39

packages/workbox-broadcast-cache-update/broadcastUpdate.mjs
```diff
@@ -0,0 +1,96 @@
...
+ * @param {BroadcastChannel} input.channel The `BroadcastChannel` to use.
+ * @param {string} input.cacheName The name of the cache in which the updated
+ *        `Response` was stored.
+ * @param {string} input.url The URL associated with the updated `Response`.
+ * @param {string} input.source A string identifying this library as the source
+ *        of the update message.
+ *
+ * @memberof workbox.broadcastUpdate
+ */
+const broadcastUpdate = ({channel, cacheName, url, source} = {}) => {
```
jeffposnick: As per the v3 style guide, I think this should be switched over to positional rather than named parameters. (None of them are optional.)

packages/workbox-broadcast-cache-update/BroadcastCacheUpdatePlugin.mjs
gauntface: Is there a reason for this preference?

PR #1062, opened 2017-11-17
Update the QueuePlugin class to take Queue params

philipwalton:

Based on the discussion in #1057, this PR updates the `QueuePlugin` class to not take a `Queue` instance, but rather take the same parameters as `Queue` and instantiate one internally.

### DIFF
--- a/packages/workbox-background-sync/QueuePlugin.mjs
+++ b/packages/workbox-background-sync/QueuePlugin.mjs
@@ -13,6 +13,7 @@
  limitations under the License.
 */
 
+import {Queue} from './Queue.mjs';
 import './_version.mjs';
 
 /**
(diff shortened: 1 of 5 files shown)


packages/workbox-background-sync/QueuePlugin.mjs
```diff
@@ -23,10 +24,12 @@ import './_version.mjs';
...
-   * @param {Queue} queue The Queue instance to add failed requests to.
+   * @param {...*} queueArgs Args to forward to the composed Queue instance.
+   *    See the [Queue]{@link workbox.backgroundSync.Queue} documentation for
+   *    parameter details.
    */
-  constructor(queue) {
-    this._queue = queue;
+  constructor(...queueArgs) {
+    this._queue = new Queue(...queueArgs);
     this.fetchDidFail = this.fetchDidFail.bind(this);
```
jeffposnick: @gauntface mentioned that `bind()` shouldn't be necessary, as the code in `workbox-core` that invokes plugins properly sets the `this` context to the plugin.

PR #1064, opened 2017-11-21
Use a hash of asset.source() for non-chunk assets

### LINKED ISSUE #1037: Use per-asset revisions/hashes in the Webpack plugin output
**Library Affected**:
*workbox-webpack-plugin*

Copied over from https://github.com/GoogleChrome/workbox/issues/934#issuecomment-343935191:

> I've just noticed that revision numbers are the same for both twig files although files themselves are different. Also some files in your manifest have identical revision numbers. I thought revisions were calculated from file hashes?

jeffposnick:
CC: @anilanar

Fixes #1037

The high-level logic used in this PR:
- If a filepath is present as part of a chunk, continue using the hash associated with that chunk for the `revision` field when creating a `ManifestEntry`.
- For everything in `compilation.assets` that isn't also part of a chunk, use a hash of the `asset.source()` for the `revision` field. (Prior to this PR, the hash associated with the overall compilation would be used.)
- Use a hash of the precache manifest's (sorted) contents as part of the name for the manifest file, instead of the overall compilation hash.

The changes mean that it's less likely that users will end up redownloading individual assets that haven't changed in between deployments.

I also refactored some of the `.map()s` to use `for ... of` when I thought readability would improve.

The precache manifest is now stable-sorted, which makes the test logic simpler as well.


packages/workbox-webpack-plugin/src/index.js
```diff
@@ -128,33 +131,26 @@ class WorkboxWebpackPlugin {
...
-        filename = (swSrc && path.basename(swSrc)) || 'sw.js',
-      } = this.config;
+      const {swSrc} = this.config;
+      const serviceWorkerFilename = swSrc ? path.basename(swSrc) : 'sw.js';
 
-      const entries = getEntries(compiler, compilation, this.config);
+      const entries = getEntries(compilation, this.config);
+
+      const fileManifest = generateManifest(entries);
+      const fileManifestAsset = webpackAsset(fileManifest);
```
gauntface: Nit: Would is be possible to rename this? "webpackAsset" doesn't really explain what it's doing.

packages/workbox-webpack-plugin/src/lib/generate-manifest-with-webpack.js
```diff
@@ -27,11 +29,12 @@ const PRECACHE_MANIFEST_VAR = '__precacheManifest';
...
- * @return {Promise<string>} service worker manifest file string
+ * @return {string} service worker manifest file string
  *
  * @private
  */
 module.exports = (manifestEntries) => {
-  const entriesJson = JSON.stringify(manifestEntries, null, 2);
-  return Promise.resolve(`self.${PRECACHE_MANIFEST_VAR} = ${entriesJson};`);
+  const sortedEntries = manifestEntries.sort((a, b) => a.url < b.url);
+  const entriesJson = stringify(sortedEntries, {space: 2});
```
gauntface: You might want to add a comment explaining that stringify is used here to remove any randomness from JSON.stringify making testing easier.

packages/workbox-webpack-plugin/src/index.js
addyosmani: I understand what Jeff's referring to here, but agree that either clearer naming for contributors who aren't as familiar with webpack or a comment to clarify would be useful.

packages/workbox-webpack-plugin/src/lib/generate-manifest-with-webpack.js
jeffposnick: I've added a comment. (The main motivation was to remove randomness so that the hashes were consistent, and the fact that testing also got easier was a nice side-effect.)

addyosmani:
Manually resolved the conflict in package.json. Landing.

PR #1074, opened 2017-11-28
workbox-cli copyLibraries

### LINKED ISSUE #1066: Add a new workbox copyLibraries <path> command
**Library Affected**:
*workbox-cli*

We should expose a new command in `workbox-cli` that will run https://github.com/GoogleChrome/workbox/blob/v3/packages/workbox-build/src/lib/copy-workbox-libraries.js using a directory specified.

Example usage:

```sh
workbox copyLibraries ./path/to/parentDir
```

jeffposnick:
R: @addyosmani @gauntface

Fixes #1066

### DIFF
--- a/packages/workbox-cli/src/app.js
+++ b/packages/workbox-cli/src/app.js
@@ -24,7 +24,7 @@ const logger = require('./lib/logger');
 const readConfig = require('./lib/read-config');
 const runWizard = require('./lib/run-wizard');
 
-module.exports = async (command, configFile) => {
+module.exports = async (command, options) => {
   assert(command, errors['missing-command-param']);
 
   switch (command) {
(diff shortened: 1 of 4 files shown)


packages/workbox-cli/src/app.js
```diff
@@ -33,12 +33,23 @@ module.exports = async (command, configFile) => {
       break;
     }
 
+    case 'copyLibraries': {
+      assert(options, errors['missing-dest-dir-param']);
+      const parentDirectory = path.resolve(process.cwd(), options);
+
+      const dirName = await workboxBuild.copyWorkboxLibraries(parentDirectory);
```
gauntface: This is a bit of a weird one, but would it make sense for `copyWorkboxLibraries` to return the full path? That seems more useful than the dirname especially given `path.dirname(...)` would provide this.

PR #1076, opened 2017-11-28
Add workbox.precaching.suppressWarnings() to the SW template.

### LINKED ISSUE #1047: Call `suppressWarnings()` in the service worker template
**Library Affected**:
*workbox-build*

Moved over from https://github.com/GoogleChrome/workbox/pull/1042#pullrequestreview-76543644

I think it's safe to unconditionally call [`workbox.precaching.suppressWarnings()`](https://github.com/GoogleChrome/workbox/blob/v3/packages/workbox-precaching/_default.mjs#L253) inside the [`sw-template.js` template](https://github.com/GoogleChrome/workbox/blob/v3/packages/workbox-build/src/templates/sw-template.js), since the template is used in conjunction with a build tool, and we can assume that if a build tool created a `ManifestEntry` without a `revision`, it's intentional.

jeffposnick:
R: @addyosmani @gauntface

Fixes #1047

### DIFF
--- a/infra/testing/validator/service-worker-runtime.js
+++ b/infra/testing/validator/service-worker-runtime.js
@@ -11,6 +11,7 @@ function setupSpiesAndContext() {
     clientsClaim: sinon.spy(),
     precaching: {
       precacheAndRoute: sinon.spy(),
+      suppressWarnings: sinon.spy(),
     },
     routing: {
       registerNavigationRoute: sinon.spy(),
@@ -42,6 +43,7 @@ function setupSpiesAndContext() {
     setCacheNameDetails: workbox.core.setCacheNameDetails,
     setConfig: workbox.setConfig,
     skipWaiting: workbox.skipWaiting,
+    suppressWarnings: workbox.precaching.suppressWarnings,
   };
 
   return {context, methodsToSpies};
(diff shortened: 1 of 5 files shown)


jeffposnick:
`suppressWarnings()` will be called whenever both are true:
- The templated SW is used, which is to say, either via the `generateSW` entry point using `workbox-build`/`workbox-cli`, or the `generateSWString` entry point using `workbox-webpack-plugin` **and**
- There's one or more entries that are being precached, regardless of whether those entries have just a `url` field or whether it has a `url` and `revision` field.

In other words, if you're using `generateSW` or `generateSWString` we can be confident that whatever set of entries you're passing in to be precached comes from our build tools, and there's no need to show the warnings.

gauntface:
Only downsides are:

1. We are adding code that isn't needed and we know it isn't needed
1. We are encouraging it's use, which if a developer copies and pastes the SW to then make further changes, they'll suppress the warning even though that may not be revisioning assets.

It's nothing big and is a nit.

addyosmani:
Discussed this with Matt offline. Acknowledging the downsides, we're fine with this change landing given the (mostly) certain that entries passing in to be precached coming from our build tools will not need the warnings shown with generateSW/generateSWString.

PR #1079, opened 2017-11-29
Throw a WorkboxError when cacheExpiration is used without cacheName

### LINKED ISSUE #1014: `cacheExpiration` on the `runtime` cache - should we prevent?
At the moment this is valid:

```javascript
workbox.routing.registerRoute(/\/images\//, workbox.strategies.cacheFirst({
  cacheExpiration: {
    maxEntries: 4
  }
}));
```

Reading this code for the first time you'd expect the rule to only apply to images, but that's not the case, it applies to all of the `runtime` cache. Should we require developers to specify a cache to enforce this behavior?

jeffposnick:
R: @addyosmani @gauntface

Fixes #1014

I don't know what the criteria is for doing this sort of thing in `NODE_ENV !== 'production'` vs. unconditionally, but I went with unconditionally.


addyosmani:
> Should we add a check to: https://github.com/GoogleChrome/workbox/blob/v3/packages/workbox-cache-expiration/CacheExpirationPlugin.mjs#L114 before treating the issue as fixed?

I think adding the check there before we call this fully good would be useful.

jeffposnick:
I'm going to revisit this PR once https://github.com/GoogleChrome/workbox/pull/1084 lands.

addyosmani:
Feel free to re-explore now that https://github.com/GoogleChrome/workbox/pull/1084 has landed

jeffposnick:
> One thing that we may still face is that if the developer uses the plugin outside of this builder, the error will not throw.

I do not believe that the `Plugin` constructor would have the necessarily state information to detect whether or not there's a cache name configured. The `Plugin` only finds out about the associated cache name once the runtime caching takes place:

https://github.com/GoogleChrome/workbox/blob/fb016edeeee503652e09cc8809fe2245bdc874d0/packages/workbox-cache-expiration/Plugin.mjs#L88

My inclination is that if we were to detect this sort of thing at runtime, it should be a warning rather than a fatal error.

But my preference would be to keep the original approach of throwing a fatal error when the strategy is misconfigured ahead of time vs. a warning at runtime.

How do other folks feel about those options?

gauntface:
For dev we could a check of: 'If no cache name and a plugin of type "expiration.Plugin", throw an error' and for prod, throw at runtime.

jeffposnick:
I know that this was already approved, but I've made changes along the lines of https://github.com/GoogleChrome/workbox/pull/1079#issuecomment-349064473 and it would be great to have it looked over again prior to merging.

packages/workbox-strategies/_default.mjs
```diff
@@ -56,6 +58,15 @@ const mapping = {
 const defaultExport = {};
 Object.keys(mapping).forEach((keyName) => {
   defaultExport[keyName] = (options = {}) => {
+    if (process.env.NODE_ENV !== 'production') {
+      if (options.cacheExpiration && !options.cacheName) {
```
gauntface: This isn't needed anymore as the cacheExpiration option doesn't get used - developers have to explicitly create the Plugin.

test/workbox-strategies/node/test-default.mjs
```diff
@@ -17,6 +20,21 @@ describe(`[workbox-strategies] Default Export`, function() {
       expect(strategy._plugins.length).to.equal(1);
       expect(strategy._plugins[0]).to.equal(CUSTOM_PLUGIN);
     });
+
+    devOnly.it(`should throw when cacheExpiration is used without cacheName`, async function() {
+      await expectError(
+        () => strategies.cacheFirst({
+          cacheExpiration: {
```
gauntface: Like wise - this test should be failing since this option does nothing now.

packages/workbox-cache-expiration/Plugin.mjs
```diff
@@ -86,6 +88,13 @@ class Plugin {
    * @private
    */
   _getCacheExpiration(cacheName) {
+    if (cacheName === cacheNames.getRuntimeName()) {
+      throw new WorkboxError('cache-expiration-requires-cache-name', {
```
gauntface: NIT: Maybe this error code should be 'expire-custom-caches-only'

packages/workbox-core/models/messages/messages.mjs
```diff
@@ -180,4 +180,12 @@ export default {
     return `The arguments passed into responsesAreSame() appear to be ` +
       `invalid. Please ensure valid Responses are used.`;
   },
+  'cache-expiration-requires-cache-name': ({moduleName, funcName}) => {
+    if (!moduleName || !funcName) {
+      throw new Error(`Unexpected input to 'cache-expiration-requires-cache` +
+        `-name' error.`);
+    }
+    return `You must also configure 'cacheName' when using 'cacheExpiration' ` +
```
gauntface: This message should be changed to something like `You must provide a 'cacheName' property when using the expiration Plugin with a strategy.`.

jeffposnick:
Unless I've missed something else obvious, this should have the required changes now, @gauntface. Thanks for your patience.

Issue #1071, opened 2017-11-27
workbox-strategies will "auto" import cachable-response and broadcast-cache-update


gauntface:
Due to the way strategies will attempt to convert configuration into a plugin and how Rollup builds the final bundle, using a strategy will auto import the workbox plugins regardless of whether they are used or not.

Only way around this is to require developers to use the plugins directly, which would unless make it a bit clearer that Workbox has some notion of plugin support.

Thoughts?

gauntface:
To be specific, these are the versions:

Today, developers can do:
```js
workbox.strategies.staleWhileRevalidate({
  cacheName: 'hello-world',
  cacheExpiration: {
    ...
  },
  broadcastCacheUpdate: {
    ...
  },
  cacheableResponse: {
    ...
  }
});
```

Regardless of the configuration, this will pull in the `workbox-cache-expiration`, `workbox-cacheable-response` and `workbox-broadcast-cache-update` modules, meaning the first time a developer runs `workbox.strategies` it'll actually be pulling three modules.

I'd like to undo this and instead require explicit plugins:

```js
workbox.strategies.staleWhileRevalidate({
  cacheName: 'example',
  plugins: [
    new workbox.expiration.CacheExpirationPlugin({ ... }),
    new workbox.broadcastUpdate.BroadcastCacheUpdatePlugin({ ... }),
    new workbox.cacheableResponse.CacheableResponsePlugin({ ... }),
  ]
});
```

This removes the dependency between strategies and the three other modules and makes it more explicit on the developer.

We can do more to naming if we felt it would make the API more pallet-able:

```js
workbox.strategies.staleWhileRevalidate({
  cacheName: 'example',
  plugins: [
    new workbox.expiration.Plugin({ ... }),
    new workbox.broadcastUpdate.Plugin({ ... }),
    new workbox.cacheableResponse.Plugin({ ... }),
  ]
});
```

or:

```js
workbox.strategies.staleWhileRevalidate({
  cacheName: 'example',
  plugins: [
    workbox.expiration.plugin({ ... }),
    workbox.broadcastUpdate.plugin({ ... }),
    workbox.cacheableResponse.plugin({ ... }),
  ]
});
```

jeffposnick:
Thanks for investigating the options.

I am on board with the second proposal,

```js
workbox.strategies.staleWhileRevalidate({
  cacheName: 'example',
  plugins: [
    new workbox.cacheExpiration.Plugin({ ... }),
[...]

PR #1082, opened 2017-11-30
Adds Broadcast plugin to helper function

gauntface:
This is somewhat dependent on whether we want this vs a different approach noted in #1071


gauntface:
Going to go for alternative to this mentioned in issue.

PR #1084, opened 2017-11-30
Plugins changes

### LINKED ISSUE #1071: workbox-strategies will "auto" import cachable-response and broadcast-cache-update

gauntface:

Fixes #1071

Changes plugins to `Plugin` for relevant modules and removes the config to plugin builder.


jeffposnick:
Changes look good, modulo whatever testing issues you were seeing.

gauntface:
Raised issue for testing errors.: https://github.com/GoogleChrome/workbox/issues/1086

PR #1085, opened 2017-11-30
Pass through all webpack config options to generateSWString()

### LINKED ISSUE #934: Pass-through config from Webpack plugin to workbox-build
Adapted from @goldhand's comment at https://github.com/GoogleChrome/workbox/pull/808#discussion_r144745346 in reference to the ` get generateSWStringOptions()` getter:

> @jeffposnick this is where we should specify which generateSWString properties should be supported.

jeffposnick:
CC: @anilanar 

Fixes #934 

As an initial implementation, it should provide flexibility to handle both webpack-specific precaching as well as take advantage of `runtimeCaching`, `templatedUrls`, and the other options exposed via `workbox-build`.

(This might not be the 100% final interface, as folks who start using the webpack plugin might have feedback about certain options that feel awkward when used in this context.)

### DIFF
--- a/packages/workbox-build/src/entry-points/options/defaults.js
+++ b/packages/workbox-build/src/entry-points/options/defaults.js
@@ -17,6 +17,8 @@
 module.exports = {
   globIgnores: ['node_modules/**/*'],
   globPatterns: ['**/*.{js,css,html}'],
+  // Use a different default for generateSWString.
+  generateSWStringGlobPatterns: [],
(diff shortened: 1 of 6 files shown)


jeffposnick:
(I'm taking a look at the tests...)

gauntface:
Running with `gulp docs --pretty` shows that `sanitizeConfig` is being defined but should be private or should be made private.

PR #1089, opened 2017-11-30
Adding cacheable response demo

gauntface:

- Adds cacheable response to workbox-sw (I need to auto-generate the list of namespaces to module names)
- Adds cachable response demo
- Prints out status and headers for a response for easier debugging

### DIFF
--- a/packages/workbox-sw/controllers/WorkboxSW.mjs
+++ b/packages/workbox-sw/controllers/WorkboxSW.mjs
@@ -28,6 +28,7 @@ const MODULE_KEY_TO_NAME_MAPPING = {
   strategies: 'strategies',
   precaching: 'precaching',
   routing: 'routing',
+  cacheableResponse: 'cacheable-response',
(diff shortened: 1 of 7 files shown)


packages/workbox-sw/controllers/WorkboxSW.mjs
```diff
@@ -28,6 +28,7 @@ const MODULE_KEY_TO_NAME_MAPPING = {
   strategies: 'strategies',
   precaching: 'precaching',
   routing: 'routing',
+  cacheableResponse: 'cacheable-response',
```
jeffposnick: FWIW, I realize that `broadcastCacheUpdate` is missing from there, too. Do you want to sneak it in as part of this PR?

gauntface: Nah I'll do it along with a PR for the broadcast cache update demo

PR #1091, opened 2017-11-30
Adding broadcast channel demo

gauntface:

Adds Broadcast channel demo.


demos/functions/views/demo/workbox-broadcast-cache-update/sw.hbs
```diff
@@ -0,0 +1,30 @@
...
+const broadcastUpdate = new workbox.broadcastUpdate.BroadcastCacheUpdate(
+  'broadcast-demo');
+
+const initialResponse = new Response('Response 1', {
+  headers: {
+    'last-modified': Date.now().toString(),
+  }
+});
+
+self.addEventListener('message', function(event) {
```
philipwalton: Is there a reason you're not using an arrow function here?

PR #1095, opened 2017-12-01
Revamp runtimeCaching.options to account for new plugins usage.

### LINKED ISSUE #1088: Change runtime-caching-converter's ouput
**Library Affected**:
*workbox-build*

The code in https://github.com/GoogleChrome/workbox/blob/v3/packages/workbox-build/src/lib/runtime-caching-converter.js needs to be changed to account for https://github.com/GoogleChrome/workbox/pull/1084.

jeffposnick:
R: @addyosmani @gauntface

Fixes #1088

### DIFF
--- a/infra/testing/validator/service-worker-runtime.js
+++ b/infra/testing/validator/service-worker-runtime.js
@@ -8,7 +8,15 @@ const vm = require('vm');
 function setupSpiesAndContext() {
   const importScripts = sinon.spy();
   const workbox = {
+    // To make testing easier, return the name of the plugin.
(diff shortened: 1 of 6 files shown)


jeffposnick:
This PR takes the current functionality in v3 and pretty directly migrates it to support the new `Plugin`-based approach to configuring everything.

I also filed https://github.com/GoogleChrome/workbox/issues/1096 to track a larger discussion about `runtimeCaching.options` in v3, and that might lead to follow-up PRs depending on what the resolution is.

PR #1098, opened 2017-12-01
alpha.2 release

gauntface:

Just doing this to ensure I don't break anything by mistake

### DIFF
--- a/gulp-tasks/publish-cdn.js
+++ b/gulp-tasks/publish-cdn.js
@@ -26,10 +26,13 @@ const handleCDNUpload = async (tagName, gitBranch) => {
   const buildFilesDir =
     await publishHelpers.groupBuildFiles(tagName, gitBranch);
 
+  // Git adds 'v' to tag name, lerna does not in package.json version.
+  // We are going to publish to CDN *without* the 'v'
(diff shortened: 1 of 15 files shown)


gulp-tasks/publish-cdn.js
```diff
@@ -26,8 +26,13 @@ const handleCDNUpload = async (tagName, gitBranch) => {
   const buildFilesDir =
     await publishHelpers.groupBuildFiles(tagName, gitBranch);
 
-  logHelper.log(`Uploading '${tagName}' to CDN.`);
-  const publishUrls = await cdnUploadHelper.upload(tagName, buildFilesDir);
+  let friendlyTagName = tagName;
+  if (friendlyTagName.indexOf('v') === 0) {
```
addyosmani: Changes here LGTM on the whole. I would add in a comment explaining why we're doing the index check against `v`as the hangouts history for that convo is likely to not be accessible to future contributors.

jeffposnick:
👍

gulp-tasks/publish-cdn.js
addyosmani: 👍 with change

PR #1104, opened 2017-12-03
Default the glob options follow & strict to true, with new options to override

### LINKED ISSUE #1093: Follow symbolic links when performing glob operations
**Library Affected**:
workbox-build

**Issue**:
Hi, 
We are trying to wrap workbox with ember build using the package workbox-build. 
On post build hook ember-cli has some symbolic-links. The problem is that when workbox-build tries to list the files, those files that are symbolic are ignored and then not included on the manifest. In fact, we get no error at all.

We want to raise an error when an unusual error is encountered when attempting to read a directory and follow symlinked directories when expanding ** patterns.

For that,[node-glob](https://github.com/isaacs/node-glob) by default dont follow symlinks if its not specified. So we want an available config to add these glob options:

```JavaScript
// config
runtimeCaching: [ ...],
globDirectory: ...
globPatterns: ..
globFollows: true, // new one
globStrict: true // new one
```

On [get-file-details.js](https://github.com/BBVAEngineering/workbox/blob/v3/packages/workbox-build/src/lib/get-file-details.js)

```JavaScript
// 
globbedFiles = glob.sync(globPattern, {
      cwd: globDirectory,
      ignore: globIgnores,
      strict: globStrict,
      follows: globFollows,
    });
```

Thanks!

beatrizdemiguelperez:

Fixes #1093 


packages/workbox-build/src/_types.js
```diff
@@ -132,6 +132,19 @@ import './_version.mjs';
  *
  * E.g. `['**\/ignored.html']`
  *
+ * @property {Boolean} [globFollow=true] Follow symlinked directories when
+ * expanding ** patterns. Note that this can result in a lot of duplicate
+ * references in the presence of cyclic links.
+ *
+ * E.g. `true`
```
gauntface: Can you remove the e.g. `true` piece.

packages/workbox-build/src/_types.js
```diff
@@ -132,6 +132,19 @@ import './_version.mjs';
  *
  * E.g. `['**\/ignored.html']`
  *
+ * @property {Boolean} [globFollow=true] Follow symlinked directories when
+ * expanding ** patterns. Note that this can result in a lot of duplicate
```
gauntface: Could you add single quotes around the patterns like: '**'

packages/workbox-build/src/_types.js
```diff
@@ -132,6 +132,19 @@ import './_version.mjs';
...
+ * references in the presence of cyclic links.
+ *
+ * E.g. `true`
+ *
+ * @property {Boolean} [globStrict=true] When an unusual error is encountered
+ * when attempting to read a directory, the process will just continue on in
+ * search of other matches. Set the strict option to raise an error in these
+ * cases.
+ *
+ * E.g. `true`
```
gauntface: Please remove this - Our JSDocs will call out the default value.

packages/workbox-build/src/lib/get-file-details.js
```diff
@@ -21,12 +21,22 @@ const errors = require('./errors');
...
+    globPattern,
+    globFollow,
+    globStrict,
+  } = globOptions;
   let globbedFiles;
   try {
     globbedFiles = glob.sync(globPattern, {
       cwd: globDirectory,
       ignore: globIgnores,
+      follow: typeof globFollow !== 'undefined'? globFollow : true,
```
gauntface: @jeffposnick you've been doing some work on the defaults, is this the best approach or should this be moved to a defaults value.

gauntface:
cc @jeffposnick 

Thanks for this @beatrizdemiguelperez, I've added some nit comments if you could take a look. One comment is more asking Jeff for comment who has been working on this a lot more closely than I have lately.

@adrigzr could you confirm you are all ok with this landing in Workbox (Just cos the weird Google license I have to worry about).

Out of interest, so it make sense to change the input to something like:

```
{
    globDirectory: ...,
    globPattern: ...,
    globOptions: {
        follow: ...,
        strict: ...,
        ignore: ...,
    }
}
```

or even: 

```
{
    globPattern: ...,
    globOptions: {
        cwd: ...,
        follow: ...,
        strict: ...,
        ignore: ...,
    }
}
```

I'm just thinking that if we did this, we could arguably just pass in the glob options and support anything in the future without changes.

adrigzr:
Hi @gauntface,

Thanks for the review. Yes, I'm working with @beatrizdemiguelperez on this feature. It's all ok. I've already signed the Google agreement.

About the change you suggest (`globOptions`), it was our first approach, but given that it wouldn't be backward compatible, we didn't make it that way.

On the other hand, as documentation suggests, setting `follow: true` on `node-glob` can lead to duplicate references in the presence of cyclic links. Are we sure we want this by default?

Thanks again for your time.

beatrizdemiguelperez:
Yes, as @adrigzr said we want to be compatible with the previous version. 
Workbox-build has already those globOptions {globDirectory, globIgnores, globPattern} separately so we think that is better add two more for backward compatible

Again thanks :)

jeffposnick:
Thanks for the contribution!

I'm of the opinion that the [`options`](https://github.com/isaacs/node-glob#options) which could be passed to `glob.sync()` are long enough and have the potential to change the output to the extent that we should continue validating them and only setting the ones we explicitly want to. So I'm fine with creating new top-level options for this functionality instead of just passing through a bag of `options` object directly.

I left a few inline comments, but my one request would be some additional changes to the tests under [`workbox-build/node/entry-points/`](https://github.com/GoogleChrome/workbox/tree/cf372f08294296c4f91e43607682b7e89ed80da6/test/workbox-build/node/entry-points).
- There are tests there that check the behavior of the defaults set by [`joi`](https://www.npmjs.com/package/joi)'s validation, which would need to be updated.
- I'd feel better about the test coverage if there were additions to the end-to-end tests in that directory that used something like [`ensureSymlink`](https://github.com/jprichardson/node-fs-extra/blob/master/docs/ensureSymlink.md) to create a symlinked source directory, and tested the output with `globFollow` enabled and disabled.

If you're not comfortable or it's not clear how to best make changes to the test suite let me know, and I can push a commit with additional test cases.

adrigzr:
Hi @jeffposnick,

Thanks for your answer. I've submitted the requested changes. Could you check the PR code again?

jeffposnick:
Thanks, @beatrizdemiguelperez & @adrigzr!

jeffposnick:
I'll give @gauntface a little time to chime in in case he has other feedback, and if not, merge.

PR #1108, opened 2017-12-06
v3: workbox-range-requests module

jeffposnick:
R: @addyosmani @gauntface

Fairly straightforward port of the `workbox-range-requests` code from v2, with some test changes to accommodate the mocked interfaces, and some additional error messages.

I updated the `workbox-sw` mapping, but I might be missing some other "meta" things that need to be changed when we add a new library.

### DIFF
--- a/packages/workbox-range-requests/Plugin.mjs
+++ b/packages/workbox-range-requests/Plugin.mjs
@@ -1,0 +1,55 @@
+/*
+ Copyright 2016 Google Inc. All Rights Reserved.
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+     http://www.apache.org/licenses/LICENSE-2.0
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+import {createPartialResponse} from './createPartialResponse.mjs';
+
+import './_version.mjs';
+
+/**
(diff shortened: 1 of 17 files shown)


packages/workbox-range-requests/Plugin.mjs
```diff
@@ -0,0 +1,55 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+import {createPartialResponse} from './createPartialResponse.mjs';
+
+import './_version.mjs';
+
+/**
+ * A class implementing the `cachedResponseWillBeUsed` lifecycle callback.
```
gauntface: This description should cover the purpose of the class, not the implementation details.

> The range request plugin makes it easy for a Request with a 'range' header to be forfilled by a cached response.
> It does this by intercepting the `cachedResponseWillBeUsed` plugin callback and returning the appropriate subset of the cached Response body.

packages/workbox-range-requests/Plugin.mjs
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @memberof workbox.rangeRequests
+ */
+class Plugin {
+  /**
+   * @param {Object} options
+   * @param {Request} options.request The original request, which may or may not
+   * contain a Range: header.
+   * @param {Response} options.cachedResponse The complete cached response.
+   * @return {Promise<Response>} If request contains a Range: header, then
```
gauntface: Nit, could you shorten this to: "If request contains a 'Range' header, then a new Response with status 206 whose body is a subset of `cachedResponse` is returned, otherwise `cachedResponse` is returned as-is."

jeffposnick: Done.

packages/workbox-range-requests/Plugin.mjs
jeffposnick: Done.

PR #1138, opened 2017-12-15
Adds ability to add plugins to precaching lifecycle

gauntface:
R: @jeffposnick 

This adds plugins to precaching so a developer could add in the broadcast cache update behavior if they want it.

```javascript
workbox.precaching.addPlugins([
    new workbox.broadcastUpdate.Plugin({ channelName: 'example-channel' })
]);
```

### DIFF
--- a/packages/workbox-precaching/_default.mjs
+++ b/packages/workbox-precaching/_default.mjs
@@ -28,6 +28,7 @@ if (process.env.NODE_ENV !== 'production') {
 let installActivateListenersAdded = false;
 let fetchListenersAdded = false;
 let suppressWarnings = false;
+let plugins = [];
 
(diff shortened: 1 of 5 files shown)


packages/workbox-precaching/_default.mjs
```diff
@@ -254,4 +258,15 @@ moduleExports.suppressWarnings = (suppress) => {
   suppressWarnings = suppress;
 };
 
+/**
+ * Add plugins to precaching.
+ *
+ * @param {Array<Object>} newPlugins
+ *
+ * @alias workbox.precaching.suppressWarnings
```
jeffposnick: suppressWarnings  => addPlugins

PR #1139, opened 2017-12-15
Precaching idb fix

gauntface:
R: @jeffposnick @philipwalton 

More changes to ensure idb works.

In IndexedDB version 1, we used two object stores 'workbox-precaching' and 'precaching-details-models'. In IndexedDB version 2 we only use 'precaching-details-models'. To ensure we are upgrading to a clean working state, this PR removes any old object stores and creates a new one.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10243,7 +10243,7 @@
     },
     "jsdoc-baseline": {
       "version": "https://github.com/hegemonic/baseline/tarball/master",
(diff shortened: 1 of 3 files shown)


gauntface:
So I fixed the integration tests, but the underlyinig indexedDB mock isn't correct, so I've raised a PR against phil's version of the repo here: https://github.com/philipwalton/shelving-mock-indexeddb/pull/1

Issue #933, opened 2017-10-19
Make it easier to use a custom Webpack SW bundle with the plugin


jeffposnick:
@anilanar brought up the following in https://github.com/GoogleChrome/workbox/pull/808#discussion_r144961184. (Please correct me if there's any nuance missing.)

> I assumed a raw `service-worker.js` could look as follows:
> 
> ```
> import WorkboxSW from 'workbox-sw';
> const workboxSW = new WorkboxSW();
> 
> const modifyManifest = manifest => { /* do something with manifest */ }
> workboxSW.precache(modifyManifest(self.__precacheManifest));
> ```
> 
> After building with webpack and processing with workbox-webpack-plugin:
> 
> ```
> importScripts('file-manifest.{hash}.js');
> (function () {
>       /* webpack generated code here */
> })();
> ```
> 
> Ideally, users can opt to use `workbox-precaching` package instead of the whole `workbox-sw` when they don't need anything else than simple precaching.

gauntface:
@jeffposnick this can be closed given the workbox-sw changes no just being a loader

jeffposnick:
I'd like to leave this open, as there are still some ergonomic issues around automatically opting-out of using the CDN copy of Workbox if we see a signal that a custom build is being used.

gauntface:
@jeffposnick could you provide a little more info.

My understanding is that to opt out of the CDN from webpack the following would happen:

- Developer would inject the manifest meaning the plugin would just inject and do nothing with respect to using the CDN vs local files.
    - If the developer decides NOT to use the CDN, they'd get a copy of the files from Github, workbox-cli or workbox-build and then set up workbox-sw accordingly (https://developers.google.com/web/tools/workbox/modules/workbox-sw#using_local_workbox_files_instead_of_cdn)

jeffposnick:
There's not a huge amount of code that needs to be changed, but there's currently some awkwardness due to assumptions being made in the webpack plugin, roughly around

https://github.com/GoogleChrome/workbox/blob/f743aa9c817f6c20003c2711c14c71000c0f30f7/packages/workbox-webpack-plugin/src/index.js#L123-L128

The webpack plugin currently uses the presence/absence of `swSrc` to toggle between the equivalents of `injectManifest` and `genereateSW` modes. My rough summary of where things stands/what should probably change:

- When in `injectManifest` mode, we need to stop automatically import `workbox-sw.js` from the CDN.
[...]

PR #1141, opened 2017-12-18
Support for importWorkboxFrom: <chunkname> and associated fixes

jeffposnick:
R: @prateekbh @gauntface @goldhand 

This is another follow-up to #933, but doesn't close it yet, as there's still a TODO: to implement `importWorkboxFrom: 'local'`. This PR was large enough already so I wanted to break it up further.

I included a refactoring of a decent chunk of `get-manifest-entries-with-webpack.js` in this PR, because I discovered that it wasn't properly filtering out all assets based on the `chunks`/`excludeChunks` whitelist/blacklist. That became an issue since `importWorkboxFrom: <chunkname>` is supposed to implicitly add `<chunkname>` to the blacklist.

This also includes a few more relevant tests, both to ensure that `importWorkboxFrom: <chunkname>` works as expected, as well as ensure that whitelist/blacklist behavior works in general.

Finally, I discovered that errors which occurred in the previous codebase would lead to a promise being rejected within the plugin's execution, but that failure wouldn't be bubbled up as a webpack error. This refactors the error detection so that asynchronous failures will be properly reported via the webpack callback (and includes a test for that condition).

### DIFF
--- a/packages/workbox-webpack-plugin/src/index.js
+++ b/packages/workbox-webpack-plugin/src/index.js
@@ -59,8 +59,7 @@ class WorkboxWebpackPlugin {
   }
 
   /**
-   * @param compilation
-   * @return {Promise.<void>}
+   * @param {Object} compilation The webpack compilation.
    * @private
    */
   async handleEmit(compilation) {
--- a/packages/workbox-webpack-plugin/src/index.js
+++ b/packages/workbox-webpack-plugin/src/index.js
@@ -59,24 +59,69 @@ class WorkboxWebpackPlugin {
   }
 
   /**
-   * @return {Object} All workbox configuration options that can be accepted
-   * by {@link module:workbox-build.generateSWString}
-   *
+   * @param compilation
+   * @return {Promise.<void>}
    * @private
    */
-  get generateSWStringOptions() {
+  async handleEmit(compilation) {
     const {
-      importScripts = [],
-      // TODO: decide on other options that should be passed to
-      // `generateSWString`. What operations should be possible for webpack
-      // users? eg, should webpack users also be able to specify a
-      // `globDirectory` that allows manifestEntries to be added that are
-      // outside webpack's scope?
+      importWorkboxFrom = 'cdn',
+      swSrc,
     } = this.config;
+    const serviceWorkerFilename = swSrc ? path.basename(swSrc) : 'sw.js';
 
-    return {
-      importScripts,
-    };
+    let workboxSWImport;
+    switch (importWorkboxFrom) {
+      case 'cdn': {
+        workboxSWImport = getModuleUrl('workbox-sw');
+        break;
+      }
+
+      case 'local': {
+        // TODO: Implement.
+        break;
+      }
+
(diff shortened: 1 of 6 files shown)


packages/workbox-webpack-plugin/src/index.js
```diff
@@ -59,24 +59,68 @@ class WorkboxWebpackPlugin {
...
+        break;
+      }
+
+      case 'disabled': {
+        // No-op.
+        break;
+      }
+
+      default: {
+        for (const chunk of compilation.chunks) {
```
gauntface: can you add a comment here on what the intended behavior is. My initial thoughts from the bug was that importWorkboxFrom *had* to be a specific value but it looks like `importWorkboxFrom: 'foo'` is valid, but I'm not sure what happens.

packages/workbox-webpack-plugin/src/index.js
```diff
@@ -59,24 +59,68 @@ class WorkboxWebpackPlugin {
...
+          if (chunk.name === importWorkboxFrom) {
+            workboxSWImport = chunk.files;
+            this.config.excludeChunks = (this.config.excludeChunks || [])
```
gauntface: could you move this initialisation out of this switch statement just to seperate the initialisation and the addition of the chunk name.

packages/workbox-webpack-plugin/src/index.js
```diff
@@ -59,24 +59,68 @@ class WorkboxWebpackPlugin {
...
+    }
+
+    const entries = getEntries(compilation, this.config);
+    const fileManifest = generateManifest(entries);
+    const fileManifestAsset = formatAsWebpackAsset(fileManifest);
+    const fileManifestHash = getAssetHash(fileManifestAsset);
+    const manifestFilename = `precache-manifest.${fileManifestHash}.js`;
+    compilation.assets[manifestFilename] = fileManifestAsset;
+
+    this.config.importScripts = (this.config.importScripts || [])
```
gauntface: Could the initialisation be moved as well.

packages/workbox-webpack-plugin/src/lib/generate-or-copy-sw.js
```diff
@@ -31,8 +31,7 @@ function sanitizeConfig(config) {
   const propertiesToRemove = [
     'chunks',
     'excludeChunks',
-    'filename',
-    'manifestFilename',
```
gauntface: Are these completely removed as options or is it just moved logic wise to somewhere else?

jeffposnick: They're removed (and hardcoded for the time being; some of this will be revised when we support `GenerateSW` vs. `InjectManifest`).

jeffposnick:
Thanks; I've written up the state of the `webpack` stuff at https://github.com/GoogleChrome/workbox/wiki/webpack%2C-assets%2C-and-chunks-%28oh-my%21%29 which was a good exercise anyway.

PR #1143, opened 2017-12-21
Explicit GenerateSW or InjectManifest webpack modes

### LINKED ISSUE #1136: Make swSrc webpack plugin mode toggling explicit?
**Library Affected**:
*workbox-webpack-plugin*

As of v3-alpha.3, the `workbox-webpack-plugin` supports two distinct modes, which roughly translate to what we refer to as `generateSW` and `injectManifest` elsewhere in Workbox.

- If `swSrc` *is not* configured, then the plugin is in `generateSW` mode, where the entire service worker file is created "from scratch" each time a build occurs.
- If `swSrc` *is* configured, then we're in a rough equivalent of `injectManifest` mode, where a source service worker file is mostly left as-is, with the only changes being those necessary to import the Workbox runtime and fresh precache manifest. (This isn't 100% the same as `injectManifest` elsewhere, since it's a call to `importScripts()` that's injected, rather than values in the `precacheAndRoute()` array. But it's functionally equivalent.)

For `workbox-build` and `workbox-cli`, we decided that developers should explicitly choose which mode they wanted, instead of inferring the mode based on the presence/absence of a config parameter. This makes it easier to reason about the different ways of using the plugin, without having to read through each parameter description and realize that something is fundamentally different when one of those parameters is set.

If we were to make the same decision for `workbox-webpack-plugin`, we could do that by removing the current default export of the `WorkboxWebpackPlugin` class from our source code and instead use two named exports of new classes, `GenerateSW` and `InjectManifest`.

So the current source code

```js
// Inside workbox-webpack-plugin/src/index.js:
module.exports = WorkboxWebpackPlugin;
```

would change to

```js
// Inside workbox-webpack-plugin/src/index.js:
module.exports = {
  GenerateSW,
  InjectManifest,
};
```

and usage would be:

```js
// Inside webpack.config.js:
const WorkboxPlugin = require('workbox-webpack-plugin');
module.exports = {
  plugins: [
    new WorkboxPlugin.GenerateSW({})
  ],
};
```

CC: @gauntface @goldhand for thoughts.

jeffposnick:
R: @goldhand @prateekbh @gauntface

Fixes #1136

There's a lot of refactoring in this PR to move some of the internal methods in `workbox-webpack-plugin` around, and share logic between the two plugins.

As part of this PR it also made sense for me to change the global default value for `globPatterns` (defined in `workbox-build`) from `['**/*.{js,css,html}']` to `[]`. We had introduced a number of overrides to that global default previously, and would have to introduce additional ones here, and it got to the point that defaulting everything to `[]` seemed cleaner. (Developers going through the `workbox wizard` onboarding flow won't be affected by this change, since they're explicitly asked which file extensions to precache.)

Assuming this PR is merged, I will put together a follow-up PR to reflect the changes that are needed to the relevant sections of the docs:

https://github.com/google/WebFundamentals/tree/master/src/content/en/tools/workbox/guides/_shared
https://github.com/google/WebFundamentals/blob/master/src/content/en/tools/workbox/modules/workbox-webpack-plugin.md


packages/workbox-build/src/entry-points/options/defaults.js
```diff
@@ -17,10 +17,8 @@
 module.exports = {
   globFollow: true,
   globIgnores: ['**/node_modules/**/*'],
-  globPatterns: ['**/*.{js,css,html}'],
+  globPatterns: [],
```
gauntface: could we keep this default and give the webpack plugin a different default?

I'm guesisng this makes the webpack plugin require no config and requires config in webpack build.

packages/workbox-webpack-plugin/src/lib/read-file-wrapper.js
```diff
@@ -15,35 +15,16 @@
 */
 
 /**
- * Use the `compiler.inputFileSystem._readFile` method instead of `fs.readFile`,
- * `readFile` is configured to use `compiler.inputFileSystem._readFile` during
- * the run phase of the webpack compilation lifecycle by passing the function
- * to the `setReadFile` function.
+ * A wrapper that calls readFileFn and returns a promise for the contents of
```
gauntface: Could you explain what readFileFn is it - I'm assuming it's a webpack provided function?

packages/workbox-build/src/entry-points/options/defaults.js
jeffposnick: There's always going to be config needed in the `workbox-build` (I think that's what you're asking about), because we don't know what the `globDirectory` should be.

So yes, this change makes it so that you need to explicitly configure `globPatterns` as well as `globDirectory` when using `workbox-build` directly.

It sounds like you're not in favor of that, so I'll undo that part of this PR and continue overriding `globPatterns` to `[]` when needed, instead of the other way around.

packages/workbox-webpack-plugin/src/lib/read-file-wrapper.js
jeffposnick: Done.

jeffposnick:
I've filed https://github.com/google/WebFundamentals/issues/5567 as a reminder to update the module docs to reflect this new interface and config options.

PR #1148, opened 2018-01-02
Webpack import workbox from local

### LINKED ISSUE #933: Make it easier to use a custom Webpack SW bundle with the plugin

jeffposnick:

With this PR, I think we have enough to fix #933. (There are still docs updates that are needed to reflect all the new options, tracked in https://github.com/google/WebFundamentals/issues/5567).

This adds support for `importWorkboxFrom: 'local'` in the `GenerateSW` mode of the webpack plugin.

(As with the underlying `workbox-build` library, `'local'` mode isn't supported in the inject manifest use case.)

### DIFF
--- a/packages/workbox-webpack-plugin/src/lib/get-workbox-sw-imports.js
+++ b/packages/workbox-webpack-plugin/src/lib/get-workbox-sw-imports.js
@@ -14,26 +14,39 @@
   limitations under the License.
 */
 
-const {getModuleUrl} = require('workbox-build');
+const {copyWorkboxLibraries, getModuleUrl} = require('workbox-build');
 
 /**
  * @param {Object} compilation The webpack compilation.
  * @param {Object} config The options passed to the plugin constructor.
- * config.excludeChunks may be modified by this function if
- * config.importWorkboxFrom is set to a chunk name.
+ * - config.excludeChunks may be modified by this function if
+ *   config.importWorkboxFrom is set to a chunk name.
+ * - config.modulePathPrefix may be modified by this function if
+ *   config.importWorkboxFrom is set to 'local'.
  * @return {Array<String>|null} A list of URLs to use to import the Workbox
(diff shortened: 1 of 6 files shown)


packages/workbox-webpack-plugin/src/lib/get-workbox-sw-imports.js
```diff
@@ -14,26 +14,39 @@
...
-const {getModuleUrl} = require('workbox-build');
+const {copyWorkboxLibraries, getModuleUrl} = require('workbox-build');
 
 /**
  * @param {Object} compilation The webpack compilation.
  * @param {Object} config The options passed to the plugin constructor.
- * config.excludeChunks may be modified by this function if
- * config.importWorkboxFrom is set to a chunk name.
+ * - config.excludeChunks may be modified by this function if
+ *   config.importWorkboxFrom is set to a chunk name.
```
gauntface: I don't know if it's just me, but I don't fully understand the 'may be modified by this function' part. What function and why does excludeChunks have to have a comment regarding importWorkboxFrom is a chunk, I would have assumed that would be expected.

PR #1149, opened 2018-01-03
Add in asset filename filtering via {test, include, exclude}

### LINKED ISSUE #935: Flexible filtering of build output via the Webpack plugin
Adapted from @anilanar's comment at https://github.com/GoogleChrome/workbox/pull/808#discussion_r144999358:

> I guess source-maps can be filtered out by default. However I'm not sure how that would affect development experience when testing offline caching. Need to test it.
> 
> This problem is similar to how loaders handle file inclusion/exclusion using three properties: `{test, include, exclude}` and the code for that is in https://github.com/webpack/webpack/blob/master/lib/ModuleFilenameHelpers.js, look for `matchObject`. That's what `uglifyjs-webpack-plugin` is also using in  https://github.com/webpack-contrib/uglifyjs-webpack-plugin/blob/master/src/index.js.
> 
> BTW the same can be used to filter chunks, that would allow a unified API for all kind of filtering and webpack users know how to use `{test, include, exclude}`.

jeffposnick:
CC: @anilanar

Fixes #935

As mentioned in the comments/the original issue, this takes advantage of a pattern native to webpack and used in other plugins (like UglifyJS) for filtering the precache manifest of webpack assets based on an object with `test`, `include`, and `exclude` properties.

It defaults to using RegExps for `exclude` that will match `.map` files and `manifest*.js(on)` files. While webpack gives developers flexibility to name their sourcemaps and asset manifest files whatever they'd like, those seem like the most common naming schemes (and they can be overridden).

The drawback of this PR is that it makes filtering more complicated: prior to this PR we implemented filtering based on chunk name white/blacklists, and now this introduces filtering based on asset name white/blacklists. (The chunk filtering is run first, then asset filtering.) There seems to be developer demand for supporting both, but this is another thing to try to explain clearly in the docs update (https://github.com/google/WebFundamentals/issues/5567).

### DIFF
--- a/packages/workbox-webpack-plugin/package.json
+++ b/packages/workbox-webpack-plugin/package.json
@@ -31,6 +31,9 @@
     "json-stable-stringify": "^1.0.1",
     "workbox-build": "^3.0.0-alpha.3"
   },
+  "peerDependencies": {
+    "webpack": "^2.0.0 || ^3.0.0"
+  },
   "author": "Will Farley <a.will.farley@gmail.com> (http://www.willfarley.org/)",
   "license": "Apache-2.0",
   "repository": "googlechrome/workbox",
--- a/packages/workbox-webpack-plugin/src/lib/get-manifest-entries-from-compilation.js
+++ b/packages/workbox-webpack-plugin/src/lib/get-manifest-entries-from-compilation.js
@@ -14,6 +14,8 @@
   limitations under the License.
 */
 
+const ModuleFilenameHelpers = require('webpack/lib/ModuleFilenameHelpers');
+
 const getAssetHash = require('./get-asset-hash');
 const resolveWebpackUrl = require('./resolve-webpack-url');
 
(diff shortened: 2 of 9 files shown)


packages/workbox-webpack-plugin/package.json
```diff
@@ -31,6 +31,9 @@
     "json-stable-stringify": "^1.0.1",
     "workbox-build": "^3.0.0-alpha.3"
   },
+  "peerDependencies": {
+    "webpack": "^2.0.0 || ^3.0.0"
```
gauntface: This seems unusual to me, is it stipulating that workbox only works with Webpack 2 and 3 (i.e. not 4)?

packages/workbox-webpack-plugin/src/lib/get-manifest-entries-from-compilation.js
```diff
@@ -177,6 +179,13 @@ function getManifestEntriesFromCompilation(compilation, config) {
 
   const manifestEntries = [];
   for (const [file, metadata] of Object.entries(filteredAssetMetadata)) {
+    // Filter based on test/include/exclude options set in the webpack config.
```
gauntface: Can you just add a comment on the end of this sentence along the lines "This matches behaviour of uglifyJS". Copying and pasting links is O.T.T. to get the gist of the point your are getting across.

gauntface: It would also be worth calling out that we can take advantage of webpacks built in filtering (which for me is a better reason for this approach over something uglify does).

packages/workbox-webpack-plugin/package.json
jeffposnick: I am following UglifyJS's lead here in limited the acceptable peer dependency: https://github.com/webpack-contrib/uglifyjs-webpack-plugin/blob/025262284c86196d0a4c4fe71e186132579629ec/package.json#L75-L77

(My understanding is that webpack 4 changes the plugin architecture anyway, so I think we should track that separately: https://github.com/GoogleChrome/workbox/issues/1151)

Issue #903, opened 2017-10-16
Add docs on how to handle pretty URLS.


gauntface:
Taken from #640 

> So I went offline, and things weren't working as expected. I eventually realized that the links throughout my page all use "pretty URLs", whereas WB cached the full file names, including the .html extension. Seems like a pretty common scenario, maybe WB should have a flag that handles this?
>
> At this point, I realized that I probably need to my own service worker with a fetch listener that handles the pretty URLs. It was surprisingly hard to find out how to do this on the docs site. This seems like a core user-journey, I think it should have an entry on the "How To" page. I think I need to use the injectManifest() method, but my first attempt hasn't worked yet.

gauntface:
Thinking on this a little more, this is messy for precaching.

Let's say you have a precache of the following:

- /index.html
- /about.html
- /something/nested.html

At the moment (Out of the box) we will do the following:

- Match for the URLs out right (i.e. '/index.html', '/about.html', '/something/nested.html).
- If that fails, we match minus search params
- If that fails and the URL **DOES** end in '/', we add `index.html` and see if that matches.

This means we'll handle '/'. However, we won't handle the pretty URL case of '/about' or '/something/nested'.

@jeffposnick would you be happy to add this into precaching's cache matching? This isn't an issue on template matching or runtime routes.

jeffposnick:
I'm okay with that additional "magic" being added in, and perhaps on by default, with the option to disable it via `prettyUrls: false`.

FWIW, the other option is to create a recipe that they used in their build process along the lines of:

```js
buildConfig.templatedUrls = glob.sync('**/*.html').map((htmlFile) => {
  const url = path.basename(htmlFile, '.html');
  return {[url]: htmlFile};
});
```

That's pretty gnarly, and wouldn't work well for developers who are just using the CLI or webpack plugin. So we can support this by default instead.

raviqqe:
Is there any workaround so far? I really want this feature so that my websites work offline as it is without rewriting `<a/>` URLs.

jeffposnick:
A workaround is to use a config like the one in https://github.com/GoogleChrome/workbox/issues/903#issuecomment-348611755, where you explicitly set `templatedUrls` to a mapping of "pretty" URL => the underlying `.html` file.

Have you had any luck with that?

raviqqe:
I forgot to mention that I use the CLI version.

jeffposnick:
You could actually do this with the CLI version, too 😄 

[...]

PR #1154, opened 2018-01-04
Url manipulation

gauntface:
R: @jeffposnick 

This is a step towards closing #903 (although will add some docs once this PR is approved before closing it).

- Adds `cleanUrls` -> Name is to match firebase config, happy to change if something like `prettyUrls` is better
- Adds `urlManipulation` -> This is a function that allows developers to add any additional permutations of a URL that could be precached. I found this will be used, but serves as a helpful escape hatch for any further alterations developers need.


packages/workbox-precaching/_default.mjs
```diff
@@ -76,44 +76,53 @@ const _removeIgnoreUrlParams = (origUrlObject, ignoreUrlParametersMatching) => {
...
-    // It's a perfect match
-    if (process.env.NODE_ENV !== 'production') {
-      logger.debug(`Precaching found an exact URL match for ` +
-        getFriendlyURL(urlObject.toString()));
-    }
-    return urlObject.href;
-  }
-
-  let strippedUrl = _removeIgnoreUrlParams(
+  const noSearchUrl = _removeIgnoreUrlParams(
```
jeffposnick: `noSearchUrl` sounds like it's completely dropping the `search` portion, which isn't the case. We're just removing the ignored search parameters. Could you name it `urlWithoutIgnoredParams` or something similar?

PR #1155, opened 2018-01-04
Throwing error instead of re registering sync on failure

### LINKED ISSUE #1145: Infinite service calls in offline mode when using BG sync

**Library Affected**:
*workbox-backgroundSync

**Browser & Platform**:
*Google Chrome v51.0.1

**Infinite service calls when in offline mode**:
```javascript
importScripts('https://storage.googleapis.com/workbox-cdn/releases/3.0.0-alpha.3/workbox-sw.js');

if (workbox) {
  console.log(`Yay! Workbox is loaded 🎉`);
} else {
  console.log(`Boo! Workbox didn't load 😬`);
}

const queue = new workbox.backgroundSync.Queue('myQueueName');

const request = new Request('https://jsonplaceholder.typicode.com/posts', {
    method: 'GET',
});

setTimeout(() => {
    fetch(request)
    .then(res => res.json(data => console.log(data)))
    .catch(e => queue.addRequest(request))
  }, 5000);
```

screen shot 2017-12-22 at 5 28 43 pm

prateekbh:
R:  @philipwalton 
CC: @jeffposnick @addyosmani @gauntface

Fixes #1145

Currently, on any failures in queue replay it results in re-registering of sync queue tag.
This causes infinite loop in case of `offline` from devtools + IIRC will not re-play the failed requests as this will re-register the same tag and also resolve the promise inside `event.waitUntil` of `sync` event. Hence telling the browser that we've successfully done everything related to this tag.

Instead to indicate any failure it is require that the promise passed to `event.waitUntil` should be rejected. ref: https://github.com/WICG/BackgroundSync/blob/master/explainer.md#the-api

> The promise passed to waitUntil is a signal to the user agent that the sync event is ongoing and that it should keep the service worker alive if possible. Rejection of the event signals to the user agent that the sync failed. Upon rejection the user agent should reschedule (likely with a user agent determined backoff).

### DIFF
--- a/packages/workbox-background-sync/Queue.mjs
+++ b/packages/workbox-background-sync/Queue.mjs
@@ -104,8 +104,6 @@ class Queue {
    * `queueDidReplay` callback is invoked (which implies the queue is
    * now empty). If any of the requests fail, a new sync registration is
    * created to retry again later.
-   *
-   * @return {Promise}
    */
   async replayRequests() {
     const now = Date.now();
@@ -137,13 +135,14 @@ class Queue {
     await this._runCallback('queueDidReplay', replayedRequests);
 
     // If any requests failed, put the failed requests back in the queue
-    // and reject promise.
+    // and rethrow the failed requests count.
     if (failedRequests.length) {
       await Promise.all(failedRequests.map((storableRequest) => {
         return this._queueStore.addEntry(storableRequest);
       }));
 
-      return Promise.reject(failedRequests);
+      throw new WorkboxError('queue-replay-failed',
+        {name: this._name, count: failedRequests.length});
     }
   }
 
--- a/packages/workbox-background-sync/Queue.mjs
+++ b/packages/workbox-background-sync/Queue.mjs
@@ -104,6 +104,8 @@ class Queue {
    * `queueDidReplay` callback is invoked (which implies the queue is
    * now empty). If any of the requests fail, a new sync registration is
    * created to retry again later.
+   *
+   * @return {Promise}
    */
   async replayRequests() {
     const now = Date.now();
@@ -132,17 +134,17 @@ class Queue {
       replayedRequests.push(replay);
     }
 
+    await this._runCallback('queueDidReplay', replayedRequests);
+
     // If any requests failed, put the failed requests back in the queue
-    // and register for another sync.
+    // and reject promise.
     if (failedRequests.length) {
       await Promise.all(failedRequests.map((storableRequest) => {
         return this._queueStore.addEntry(storableRequest);
       }));
 
-      await this._registerSync();
+      return Promise.reject(failedRequests);
     }
-
-    await this._runCallback('queueDidReplay', replayedRequests);
   }
 
   /**
--- a/test/workbox-background-sync/node/lib/test-Queue.mjs
+++ b/test/workbox-background-sync/node/lib/test-Queue.mjs
@@ -293,37 +293,43 @@ describe(`[workbox-background-sync] Queue`, function() {
       await queue.addRequest(new Request('/three'));
       await queue.addRequest(new Request('/four'));
       await queue.addRequest(new Request('/five'));
-      await queue.replayRequests(); // The 2nd and 4th requests should fail.
 
+      try {
+        await queue.replayRequests(); // The 2nd and 4th requests should fail.
+      } catch (error) {
+        const entries = await getObjectStoreEntries();
+        expect(entries.length).to.equal(2);
+        expect(entries[0].storableRequest.url).to.equal('/two');
+        expect(entries[1].storableRequest.url).to.equal('/four');
+        return;
+      }
 
-      const entries = await getObjectStoreEntries();
-      expect(entries.length).to.equal(2);
-      expect(entries[0].storableRequest.url).to.equal('/two');
-      expect(entries[1].storableRequest.url).to.equal('/four');
+      return Promise.reject('should have exit from catch');
     });
 
-    it(`should re-register for a sync event if re-fetching fails`,
+    it(`should reject replayRequests promise if re-fetching fails`,
         async function() {
-      sandbox.stub(self.registration, 'sync').value({
-        register: sinon.stub().resolves(),
-      });
       sandbox.stub(self, 'fetch')
           .onCall(1).rejects(new Error())
           .callThrough();
 
+      const failureURL = '/two';
       const queue = new Queue('foo');
 
       // Add requests for both queues to ensure only the requests from
       // the matching queue are replayed.
       await queue.addRequest(new Request('/one'));
-      await queue.addRequest(new Request('/two'));
+      await queue.addRequest(new Request(failureURL));
 
-      self.registration.sync.register.reset();
-      await queue.replayRequests(); // The second request should fail.
+      try {
+        await queue.replayRequests(); // The second request should fail.
+      } catch (error) {
+        expect(error.length).to.be.equal(1);
+        expect(error[0].url).to.be.equal(failureURL);
+        return;
+      }
 
-      expect(self.registration.sync.register.calledOnce).to.be.true;
-      expect(self.registration.sync.register.calledWith(
-          'workbox-background-sync:foo')).to.be.true;
+      return Promise.reject('should have exit from catch');
     });
 
     it(`should invoke all replay callbacks`, async function() {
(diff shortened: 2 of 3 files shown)


packages/workbox-background-sync/Queue.mjs
```diff
@@ -132,17 +134,17 @@ class Queue {
...
     // If any requests failed, put the failed requests back in the queue
-    // and register for another sync.
+    // and reject promise.
     if (failedRequests.length) {
       await Promise.all(failedRequests.map((storableRequest) => {
         return this._queueStore.addEntry(storableRequest);
       }));
 
-      await this._registerSync();
+      return Promise.reject(failedRequests);
```
philipwalton: Nit: can we throw here instead of returning a rejected promise? I think it's better to use async function patterns when inside an async function (rather that promise patterns).

test/workbox-background-sync/node/lib/test-Queue.mjs
```diff
@@ -293,37 +293,43 @@ describe(`[workbox-background-sync] Queue`, function() {
...
+        await queue.replayRequests(); // The 2nd and 4th requests should fail.
+      } catch (error) {
+        const entries = await getObjectStoreEntries();
+        expect(entries.length).to.equal(2);
+        expect(entries[0].storableRequest.url).to.equal('/two');
+        expect(entries[1].storableRequest.url).to.equal('/four');
+        return;
+      }
+
+      return Promise.reject('should have exit from catch');
```
philipwalton: Same as above, can we throw here instead of returning a rejected promise?

test/workbox-background-sync/node/lib/test-Queue.mjs
```diff
@@ -293,37 +293,43 @@ describe(`[workbox-background-sync] Queue`, function() {
...
+      } catch (error) {
+        expect(error.length).to.be.equal(1);
+        expect(error[0].url).to.be.equal(failureURL);
+        return;
+      }
 
-      expect(self.registration.sync.register.calledOnce).to.be.true;
-      expect(self.registration.sync.register.calledWith(
-          'workbox-background-sync:foo')).to.be.true;
+      return Promise.reject('should have exit from catch');
```
philipwalton: Same as above, can we throw here instead of returning a rejected promise?

packages/workbox-background-sync/Queue.mjs
```diff
@@ -132,17 +134,17 @@ class Queue {
       replayedRequests.push(replay);
     }
 
+    await this._runCallback('queueDidReplay', replayedRequests);
+
     // If any requests failed, put the failed requests back in the queue
-    // and register for another sync.
+    // and reject promise.
```
philipwalton: Maybe reword this to: "and rethrow the failed requests"?

Note: this comment I'm less concerned about changing...

packages/workbox-background-sync/Queue.mjs
prateekbh: so `throw new Error(failedRequests);`?

prateekbh:
@philipwalton correct me if I am wrong, but `throw new Error...` can only take string as param. So tops we can pass back is how many requests failed. and not the actual requests

packages/workbox-background-sync/Queue.mjs
philipwalton: I was thinking just:

```js
throw failedRequests;
```

In JS you can throw any value.

cc: @gauntface and @jeffposnick to see if they have thoughts on this.

In general I think we should stick to the pattern that async function use `await`, `try`, `catch`, and `throw`, and regular functions use `resolve` and `reject`, though I guess this may be a bit of a special case...

philipwalton:
Plain `Error` objects can only take a string, but AFAIK in JS you can throw any value. See https://github.com/GoogleChrome/workbox/pull/1155#discussion_r159789882 for my comments.

packages/workbox-background-sync/Queue.mjs
jeffposnick: Does anything end up consuming the rejected promise that's returned by this function? Or is the fact that it's a rejected promise all that matters, with the rejection value ignored?

If the value is ignored, then I'd rather see

```
throw new Error('<some meaningful message>');
```

so that a developer reading the source has a bit more context. And if at some point in the future logging is added in, it will end up with a meaningful log message.

If the value isn't ignored, and it needs to be that array, then using `Promise.reject()` seems fine to me, but I'd ask that you add in a comment explaining what's going on.

prateekbh: yeah nothing really consumes it.

+ i was thinking something like`throw new Error('X number of requests failed');`

gauntface: Please use a WorkboxError.

gauntface:
As Jeff pointed out - we should be using useful information to explain what is happening if a developer is viewing this.

The scenario you described sounds like an expected and valid scenario (i.e. the request failed to work - no worries, we'll still have the sync set up). For this - use a WorkboxError with a valid string to explain what has happened.

This means the developer will know the error is from Workbox (because it's a `WorkboxError`) and the message will highlight this is expected and not to worry.

packages/workbox-background-sync/Queue.mjs
prateekbh: Done

packages/workbox-background-sync/Queue.mjs
prateekbh: Done

test/workbox-background-sync/node/lib/test-Queue.mjs
prateekbh: Done

test/workbox-background-sync/node/lib/test-Queue.mjs
prateekbh: Done

test/workbox-background-sync/node/lib/test-Queue.mjs
```diff
@@ -293,37 +294,42 @@ describe(`[workbox-background-sync] Queue`, function() {
...
+        await queue.replayRequests(); // The 2nd and 4th requests should fail.
+      } catch (error) {
+        const entries = await getObjectStoreEntries();
+        expect(entries.length).to.equal(2);
+        expect(entries[0].storableRequest.url).to.equal('/two');
+        expect(entries[1].storableRequest.url).to.equal('/four');
+        return;
+      }
+
+      throw new Error('should have exit from catch');
```
gauntface: Could you use the helper: https://github.com/GoogleChrome/workbox/blob/v3/infra/testing/expectError.js

This will check the error is a WorkboxError with an expected code and it returns a promise, so you can await for the response and check the object store entries after.

test/workbox-background-sync/node/lib/test-Queue.mjs
```diff
@@ -293,37 +294,42 @@ describe(`[workbox-background-sync] Queue`, function() {
...
       await queue.addRequest(new Request('/one'));
-      await queue.addRequest(new Request('/two'));
+      await queue.addRequest(new Request(failureURL));
 
-      self.registration.sync.register.reset();
-      await queue.replayRequests(); // The second request should fail.
+      try {
+        await queue.replayRequests(); // The second request should fail.
+      } catch (error) {
+        expect(error).to.be.instanceof(WorkboxError);
```
gauntface: Please use expectError helper here as well.

test/workbox-background-sync/node/lib/test-Queue.mjs
```diff
@@ -293,37 +294,42 @@ describe(`[workbox-background-sync] Queue`, function() {
...
+        return;
+      }
 
-      expect(self.registration.sync.register.calledOnce).to.be.true;
-      expect(self.registration.sync.register.calledWith(
-          'workbox-background-sync:foo')).to.be.true;
+      throw new Error('should have exit from catch');
```
gauntface: This wont be needed after switching to expectError().

test/workbox-background-sync/node/lib/test-Queue.mjs
```diff
@@ -376,7 +382,13 @@ describe(`[workbox-background-sync] Queue`, function() {
 
       await queue.addRequest(new Request('/three'));
       await queue.addRequest(new Request('/four'));
-      await queue.replayRequests();
+      try {
+        await queue.replayRequests();
```
gauntface: Please use expectError here and make sure the error is what you expect.

test/workbox-background-sync/node/lib/test-Queue.mjs
prateekbh: Done

test/workbox-background-sync/node/lib/test-Queue.mjs
prateekbh: Done

test/workbox-background-sync/node/lib/test-Queue.mjs
prateekbh: Done

test/workbox-background-sync/node/lib/test-Queue.mjs
prateekbh: Done

PR #1160, opened 2018-01-05
Revamp the runtime caching option conversion.

### LINKED ISSUE #1096: Rethink valid runtimeCaching.options
**Library Affected**:
*workbox-build*

Working on #1095 made me realize that there's some cleanup to be done related to what we accept for `runtimeCaching.options` in v3. In v2, we basically supported the same set of options that [`sw-precache` supported](https://github.com/GoogleChromeLabs/sw-precache#runtimecaching-arrayobject) (plus a `plugins` property that could be used to provide additional lifecycle callback handlers).

As of what's checked in to the v3 branch today, our `joi` parameter validation looks like

https://github.com/GoogleChrome/workbox/blob/7a74459c965d03aa98a57cb7f5caf04b5bc5b85e/packages/workbox-build/src/entry-points/options/common-generate-schema.js#L40-L51

This is a departure from what we supported in v2/`sw-precache`, but there wasn't previously a `breaking-change` issue tracking that, so if nothing else, this serves to keep track of that.

But I realize that we could tweak things further in v3, by
- supporting runtimeCaching.options.broadcastUpdate, which would configure a `workbox.broadcastUpdate.Plugin` instance.
- renaming `runtimeCaching.options.cacheExpiration` to `expiration`, to reflect the fact that the module is exposed as `workbox.expiration`.
- (Re-adding support for runtimeCaching.options.networkTimeoutSeconds, which I inadvertently removed from the `joi` config.)

So, I'm proposing that in v3 we'd support a full set of configuration that looked like:

```js
runtimeCaching: [{
  urlPattern: /api/,
  handler: 'networkFirst',
  options: {
    networkTimeoutSeconds: 10,
    cacheName: 'my-api-cache',
    expiration: {
      maxEntries: 5,
      maxAgeSeconds: 60,
    },
    cacheableResponse: {
      statuses: [0, 200],
      headers: {'x-test': 'true'},
    },
    broadcastUpdate: {
      channelName: 'my-update-channel',
    },
    plugins: [
      {cacheDidUpdate: () => /* custom plugin code */}
    ],
  },
}],
```

Any concerns there?

jeffposnick:

Fixes #1096, which also contains an explanation of what's changing and why.

### DIFF
--- a/packages/workbox-build/src/entry-points/options/common-generate-schema.js
+++ b/packages/workbox-build/src/entry-points/options/common-generate-schema.js
@@ -38,16 +38,25 @@ module.exports = baseSchema.keys({
       'staleWhileRevalidate'
     )],
     options: joi.object().keys({
-      cacheName: joi.string(),
-      plugins: joi.array().items(joi.object()),
-      cacheExpiration: joi.object().keys({
-        maxEntries: joi.number().min(1),
-        maxAgeSeconds: joi.number().min(1),
-      }).or('maxEntries', 'maxAgeSeconds'),
+      backgroundSync: joi.object().keys({
+        name: joi.string().required(),
+        options: joi.object(),
+      }),
+      broadcastCacheUpdate: joi.object().keys({
+        channelName: joi.string().required(),
+        options: joi.object(),
+      }),
       cacheableResponse: joi.object().keys({
         statuses: joi.array().items(joi.number().min(0).max(599)),
         headers: joi.object(),
       }).or('statuses', 'headers'),
+      cacheName: joi.string(),
+      expiration: joi.object().keys({
+        maxEntries: joi.number().min(1),
+        maxAgeSeconds: joi.number().min(1),
+      }).or('maxEntries', 'maxAgeSeconds'),
+      networkTimeoutSeconds: joi.number().min(1),
+      plugins: joi.array().items(joi.object()),
     }),
   }).requiredKeys('urlPattern', 'handler')),
   skipWaiting: joi.boolean().default(defaults.skipWaiting),
--- a/infra/testing/validator/service-worker-runtime.js
+++ b/infra/testing/validator/service-worker-runtime.js
@@ -33,6 +33,7 @@ function setupSpiesAndContext() {
     // To make testing easier, return the name of the strategy.
     strategies: {
       cacheFirst: sinon.stub().returns('cacheFirst'),
+      networkFirst: sinon.stub().returns('networkFirst'),
     },
   };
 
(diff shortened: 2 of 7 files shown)


packages/workbox-build/src/entry-points/options/common-generate-schema.js
```diff
@@ -38,16 +38,25 @@ module.exports = baseSchema.keys({
...
-        maxEntries: joi.number().min(1),
-        maxAgeSeconds: joi.number().min(1),
-      }).or('maxEntries', 'maxAgeSeconds'),
+      backgroundSync: joi.object().keys({
+        name: joi.string().required(),
+        options: joi.object(),
+      }),
+      broadcastCacheUpdate: joi.object().keys({
+        channelName: joi.string().required(),
+        options: joi.object(),
```
gauntface: I believe the broadcast cache update object needs a channel name.

packages/workbox-build/src/entry-points/options/common-generate-schema.js
```diff
@@ -38,16 +38,25 @@ module.exports = baseSchema.keys({
...
         statuses: joi.array().items(joi.number().min(0).max(599)),
         headers: joi.object(),
       }).or('statuses', 'headers'),
+      cacheName: joi.string(),
+      expiration: joi.object().keys({
+        maxEntries: joi.number().min(1),
+        maxAgeSeconds: joi.number().min(1),
+      }).or('maxEntries', 'maxAgeSeconds'),
+      networkTimeoutSeconds: joi.number().min(1),
+      plugins: joi.array().items(joi.object()),
```
gauntface: How does this work? I.e. how does code in the node space get into the service worker? My main consideration here is that if this works, why support one way in the build process and a different way in the service worker?

packages/workbox-build/src/entry-points/options/common-generate-schema.js
jeffposnick: Yup—that requirement is enforced one line up, on line 45.

packages/workbox-build/src/entry-points/options/common-generate-schema.js
jeffposnick: The code that converts whatever is configured in the `node` build tools into a string representation that could be inserted into a service worker file is

https://github.com/GoogleChrome/workbox/blob/v3/packages/workbox-build/src/lib/runtime-caching-converter.js

and the string is used in the template at

https://github.com/GoogleChrome/workbox/blob/dafc65ba54f4f320e89df13d3e8a3ee2ed710662/packages/workbox-build/src/templates/sw-template.js#L62

(None of that has changed recently.)

The syntax used to generate the service worker code is not exactly the same as the syntax used when executing the service worker runtime. Service worker execution implies that we're inside of the `ServiceWorkerGlobalScope` with the various definitions in the `workbox` namespace available. When the build configuration is used, we're inside of the `node` runtime environment, and we maintain a mapping of the keys in the config that correspond to the resulting runtime code.

I think in this PR the mapping between configuration supplied to `node` and resulting Workbox v3 service worker runtime code gets much closer to being 1:1, as we removed some differences that were mainly around to match `sw-precache`'s config.

jeffposnick:
Based on the approval, I'm going to merge. Matt, if you've got any larger concerns about the way the `runtimeCaching` config is translated into the final service worker runtime code, I'm happy to hash those out in a separate issue.

PR #1166, opened 2018-01-09
Update gulp patterns

phated:

Hey all! I just wanted to say that the gulpfile & tasks were super informative for me to look through in this project.

As with all Google's usage of gulp, we love to have them be best-in-class and a place to point people for learning.  So I wanted to update this to use our recommended patterns with gulp 4 -  namely, functions instead of strings for composition, `.displayName` property (I'd love to update with `.flags` and `.description`), and exported tasks instead of using `gulp.task`.

There were a bunch of methods registered with `gulp.task` because of string referencing so I wasn't sure if they should even be exposed (composing functions allows for "private" tasks that can still be used within `series`/`parallel`).

I also added some inline comments/questions labeled with `// GULP:` - I'd appreciate any feedback.

Cheers,
-Blaine

### DIFF
--- a/gulp-tasks/analyze-properties.js
+++ b/gulp-tasks/analyze-properties.js
@@ -1,14 +1,24 @@
 const gulp = require('gulp');
 
+const build = require('./build');
+
 const AnalyseBuildForProperties = require('./utils/analyse-properties');
 
-gulp.task('analyze-properties:run', async () => {
+const runAnalyzeProperties = async () => {
   const analysisTool = new AnalyseBuildForProperties();
   const results = await analysisTool.run();
   results.forEach((entry) => {
     analysisTool.printDetails(entry);
   });
-});
+};
+runAnalyzeProperties.displayName = 'analyze-properties:run';
+// GULP: Is this exposed to the CLI?
+gulp.task(runAnalyzeProperties);
 
-gulp.task('analyze-properties',
-  gulp.series(['build', 'analyze-properties:run']));
+const analyzeProperties = gulp.series(
+  build,
+  runAnalyzeProperties,
+);
+analyzeProperties.displayName = 'analyze-properties';
+
+module.exports = analyzeProperties;
--- a/gulp-tasks/analyze-properties.js
+++ b/gulp-tasks/analyze-properties.js
@@ -12,8 +12,6 @@ const runAnalyzeProperties = async () => {
   });
 };
 runAnalyzeProperties.displayName = 'analyze-properties:run';
-// GULP: Is this exposed to the CLI?
-gulp.task(runAnalyzeProperties);
 
 const analyzeProperties = gulp.series(
   build,
--- a/gulp-tasks/build-browser-packages.js
+++ b/gulp-tasks/build-browser-packages.js
@@ -14,8 +14,6 @@ const browserBundle = gulp.series(
   ))
 );
 browserBundle.displayName = 'build-browser-packages:browser-bundle';
-// GULP: Is this exposed to the CLI?
-gulp.task(browserBundle);
 
 const version = gulp.series(
   Object.keys(constants.BUILD_TYPES).map((buildKey) => packageRunnner(
@@ -26,8 +24,6 @@ const version = gulp.series(
   ))
 );
 version.displayName = 'build-browser-packages:version-module';
-// GULP: Is this exposed to the CLI?
-gulp.task(version);
 
 const buildPackages = gulp.series(
   version,
--- a/gulp-tasks/build-browser-packages.js
+++ b/gulp-tasks/build-browser-packages.js
@@ -5,25 +5,34 @@ const versionModule = require('./utils/version-module');
 const constants = require('./utils/constants');
 const packageRunnner = require('./utils/package-runner');
 
-gulp.task('build-browser-packages:browser-bundle', gulp.series(
+const browserBundle = gulp.series(
   Object.keys(constants.BUILD_TYPES).map((buildKey) => packageRunnner(
     'build-browser-packages:browser-bundle',
     'browser',
     buildBrowserBundle,
     constants.BUILD_TYPES[buildKey],
   ))
-));
+);
+browserBundle.displayName = 'build-browser-packages:browser-bundle';
+// GULP: Is this exposed to the CLI?
+gulp.task(browserBundle);
 
-gulp.task('build-browser-packages:version-module', gulp.series(
+const version = gulp.series(
   Object.keys(constants.BUILD_TYPES).map((buildKey) => packageRunnner(
     'build-browser-packages:version-module',
     'browser',
     versionModule,
     constants.BUILD_TYPES[buildKey],
   ))
-));
+);
+version.displayName = 'build-browser-packages:version-module';
+// GULP: Is this exposed to the CLI?
+gulp.task(version);
 
-gulp.task('build-browser-packages', gulp.series(
-  'build-browser-packages:version-module',
-  'build-browser-packages:browser-bundle',
-));
+const buildPackages = gulp.series(
+  version,
+  browserBundle,
+);
+buildPackages.displayName = 'build-browser-packages';
+
+module.exports = buildPackages;
--- a/gulp-tasks/utils/package-runner.js
+++ b/gulp-tasks/utils/package-runner.js
@@ -113,6 +113,7 @@ module.exports = (displayName, packageType, func, ...args) => {
   // We need to return a single no-op rather than an empty array, or else gulp
   // will throw 'One or more tasks should be combined using series or parallel'.
   if (packagePaths.length === 0) {
+    // TODO: should gulp handle this differently?
     const noOp = (done) => done();
     noOp.displayName = displayName;
     return [noOp];
--- a/gulp-tasks/watch.js
+++ b/gulp-tasks/watch.js
@@ -1,9 +1,11 @@
 const gulp = require('gulp');
 
+const build = require('./build');
+
 const logHelper = require('../infra/utils/log-helper');
 
-gulp.task('watch', gulp.series(
-  'build',
+const watch = gulp.series(
+  build,
   () => {
     gulp.watch(
       [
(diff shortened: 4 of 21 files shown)


phated:
Not too sure why JSDoc exploded on travis

gulp-tasks/analyze-properties.js
```diff
@@ -1,14 +1,24 @@
...
+const runAnalyzeProperties = async () => {
   const analysisTool = new AnalyseBuildForProperties();
   const results = await analysisTool.run();
   results.forEach((entry) => {
     analysisTool.printDetails(entry);
   });
-});
+};
+runAnalyzeProperties.displayName = 'analyze-properties:run';
+// GULP: Is this exposed to the CLI?
```
gauntface: Yes it is (this is a somewhat weird task I concocted to try and spot missed minification opportunities).

gauntface:
I'm really on the fence about this.

I love the fact that we can clean up our gulp --tasks output with this, but I'm not a fan of pulling everything out into a function and then adding a `displayName` vs just `gulp.task(name, function)`.

Is there any non-obvious patterns of one API over the other?

gauntface:
Sidenote: I kicked travis.

phated:
> I'm not a fan of pulling everything out into a function and then adding a displayName vs just gulp.task(name, function)

I wanted to open this PR as a discussion point for private vs public tasks - it seems like a lot of these tasks are just private tasks that are being registered into the undertaker registry just for lookup pretty soon after.  I would love to remove all of the `gulp.task` calls, as that's not a suggested pattern going forward.  The `displayName` stuff is a little harder since the project relies heavily on `-` and `:` in task names so we can't use named functions, which is the recommended pattern.  I'd be happy to help with a refactor for that too, if you have any suggestions.

undertaker was designed with the paradigm that you are just composing functions.  The `gulp.task(string, fn)` API was added to smooth the upgrade path but the `gulp.task` API should only be used for tasks that are meant to be run at the command line.  When using strings, order is much too important and we should just be using plain JS.  There's quite a bit of overhead in using task registration on every method.

The `gulp --tasks` command is very important in knowing what is available to the project (which is why we added `.description` and `.flags` properties for documentation). For example - when I first started looking at this project, I couldn't figure out what the top-level commands were, as I don't see them documented, nor are they aliased as npm-scripts.

Ideally, `gulp --tasks --depth 1` should show only the tasks I should run at the command line with descriptions and flags.  As opposed to: (which doesn't all fit on my screen)
screen shot 2018-01-09 at 3 07 54 pm

Hopefully this explains my thoughts on the patterns - let me know if I can detail anything further.

gauntface:
Awesome sauce.

I totally agree that soooo many of these tasks can and should be private and if the function names *can* we used then I'm always happy to consider a different naming pattern if it is a recommended pattern. I just don't like the `displayName` step.

As stupid as this may sound - the top level tasks are actually defined by the file names in `gulp-tasks/*.js`.

Is it cool to use `gulp.task(<string>, ...)` for top level tasks we expect people to use and then just move everything else to functions?

phated:
I've pushed another commit that removes all `gulp.task` calls.

Here's the resulting `gulp --tasks --depth 0` output:
screen shot 2018-01-09 at 3 51 20 pm

phated:
I had to make a change to the package.json in `workbox-build` because one of the tasks became private.  Let me know if that's wrong.

packages/workbox-build/package.json
```diff
@@ -39,7 +39,7 @@
   },
   "main": "build/index.js",
   "scripts": {
-    "build": "gulp build:update-cdn-details && gulp build-packages --package workbox-build",
+    "build": "gulp build-packages --package workbox-build",
```
gauntface: This change is likely to be a problem. There is a subtle dance around CDN updates that is sadly fragile and I'm not sure this will have the same output.

gulpfile.js
```diff
@@ -36,7 +36,9 @@ global.cliOptions = options;
 
 // Forward referencing means the order of gulp-task
 // requires is important
-const gulpTaskFiles = [
+const exportedTasks = [
```
gauntface: This could presumably just glob for files and require them all.

gulp-tasks/watch.js
```diff
@@ -1,21 +1,27 @@
...
     )
+    // GULP: why is this here?
     .on('error', () => {})
     .on('change', function(path, stats) {
       logHelper.log(`gulp.watch() is running due to a change in '${path}'`);
     });
   }
-));
+);
+watch.displayName = 'watch';
```
gauntface: Is this needed?

gulp-tasks/watch.js
```diff
@@ -1,21 +1,27 @@
...
     gulp.watch(
       [
         `packages/${global.packageOrStar}/**/*.mjs`,
         '!packages/**/_version.mjs',
         '!packages/**/build/**/*',
         '!packages/**/node_modules/**/*',
-      ], gulp.series('build'),
+      ], build,
     )
+    // GULP: why is this here?
```
gauntface: I think the idea is that during a watch, the build can fail, just watch for the next code change to fix it.

gulp-tasks/utils/package-runner.js
```diff
@@ -113,6 +113,7 @@ module.exports = (displayName, packageType, func, ...args) => {
   // We need to return a single no-op rather than an empty array, or else gulp
   // will throw 'One or more tasks should be combined using series or parallel'.
   if (packagePaths.length === 0) {
+    // TODO: should gulp handle this differently?
```
gauntface: There are scenarios where the flags result in us running gulp tasks that ultimately do nothing.

It's arguable what should happen here. The following snippet looks like developer error, but it's essentially what is happening here.

```javascript
gulp.series([])

// or

gulp.parallel([])
```

The other side effect is that we have to return a promise or use the callback since a synchronous function didn't work with gulp when I tried it.

phated:
@gauntface Sorry I haven't gotten back on these things - just got a new computer that I'm setting up.  I'll respond inline.

packages/workbox-build/package.json
phated: I'll split this task into it's own file and expose it to the CLI and put this back.

gulpfile.js
phated: Yeah, with the move to functions instead of strings, the order tasks are registered wouldn't matter.  I really like `export default as "build" from './gulp-tasks/build'` though - might be easier to convert later if these were `exports.build = require('./gulp-tasks/build')`. This would require the top-level tasks to avoid `-` or `:` in their names.

gulp-tasks/watch.js
phated: Nope, this would be better served as a named function.  I started with that but switched it to an arrow function with `displayName` to keep things consistent.

gulp-tasks/watch.js
phated: Does the latest gulp stop the watcher if an error happens?  I didn't think it did but I might have to take another look.

gulp-tasks/utils/package-runner.js
phated: Yeah, that's tough. We want things to be usable without much ceremony (remember run-sequence, gulp-watch and gulp-plumber?) but we also don't want to introduce footguns like `gulp.series([])`.  I'm wondering if we can revisit the usage of this and see if there is a compromise between the 2 ends of the spectrum.

> The other side effect is that we have to return a promise or use the callback since a synchronous function didn't work with gulp when I tried it.

We did away with sync tasks because they generally were another footgun that people didn't realize.  All the benchmarks where people didn't return streams (and had sub-millisecond build times!) or had a runSequence but their clean task didn't finish before their build started were far too common.

phated: Maybe all tasks could just be marked as `async`?

gulp-tasks/watch.js
gauntface: This could easily have been copy and pasted and assumed to be correct.

If the latest watcher doesn't stop that would make me soooo happy.

packages/workbox-build/package.json
phated: I'm fine with whatever. I don't have a job right now so I'm just looking through large uses of gulp 4 to learn people's patterns and make suggestions. Let me know what you'd like to do.

gulp-tasks/analyze-properties.js
```diff
@@ -1,14 +1,22 @@
...
-gulp.task('analyze-properties:run', async () => {
+const runAnalyzeProperties = async () => {
   const analysisTool = new AnalyseBuildForProperties();
   const results = await analysisTool.run();
   results.forEach((entry) => {
     analysisTool.printDetails(entry);
   });
-});
+};
+runAnalyzeProperties.displayName = 'analyze-properties:run';
```
gauntface: Can drop this, it can be private.

gulp-tasks/analyze-properties.js
```diff
@@ -1,14 +1,22 @@
...
+
+const analyzeProperties = gulp.series(
+  build,
+  runAnalyzeProperties,
+);
+analyzeProperties.displayName = 'analyze-properties';
```
gauntface: Drop this - I think it's ok to be called analyzeProperties.

gulp-tasks/build-browser-packages.js
```diff
@@ -5,25 +5,30 @@ const versionModule = require('./utils/version-module');
...
+const browserBundle = gulp.series(
   Object.keys(constants.BUILD_TYPES).map((buildKey) => packageRunnner(
     'build-browser-packages:browser-bundle',
     'browser',
     buildBrowserBundle,
     constants.BUILD_TYPES[buildKey],
   ))
-));
+);
+browserBundle.displayName = 'build-browser-packages:browser-bundle';
```
gauntface: This can probably be private as well.

gulp-tasks/build-browser-packages.js
```diff
@@ -5,25 +5,30 @@ const versionModule = require('./utils/version-module');
...
+const version = gulp.series(
   Object.keys(constants.BUILD_TYPES).map((buildKey) => packageRunnner(
     'build-browser-packages:version-module',
     'browser',
     versionModule,
     constants.BUILD_TYPES[buildKey],
   ))
-));
+);
+version.displayName = 'build-browser-packages:version-module';
```
gauntface: Can be private.

gauntface:
@phated I'm really sorry but would you mind if I re-did this PR in smaller chunks?

I do want this to land so that we can be more inline with best practice of Gulp 4 but I'm worried that doing this all at once might result in uncaught problems at deploy time.

phated:
> I'm really sorry but would you mind if I re-did this PR in smaller chunks?

Yeah, go ahead and ping me on stuff.  I just wanted to get this sent to start the discussion

phated:
@gauntface do you want me to close this?

gauntface:
@phated if you could and I'll cc you on some bugs reallll soon,

phated:
@gauntface perfect. closing now.

PR #1171, opened 2018-01-12
Support --injectManifest in the workbox-cli wizard.

### LINKED ISSUE #1052: Workbox CLI has an Awkward Start
Using the Workbox CLI there are two big wrinkles:

1. The workbox config **must** be defined in the workbox commands even thought it's using the default file in the default location
1. You can't inject a manifest with the provided config. It would be good to ask this during the configuration rather than leave the developer go in and manually edit the config.json

jeffposnick:
R: @addyosmani @gauntface
CC: @DavidScales @nasearle

Fixes #1052

```sh
workbox wizard --swSrc='/path/to/sw.js'
```

will work now.

### DIFF
--- a/packages/workbox-cli/src/lib/help-text.js
+++ b/packages/workbox-cli/src/lib/help-text.js
@@ -18,9 +18,13 @@ module.exports = `Usage:
 $ workbox <command> [options]
 
 Commands:
-  wizard
+  wizard [--swSrc=<path/to/service-worker.js>]
     Runs the configuration wizard, which will generate a
     config file based on answers to questions.
+    If --swSrc is omitted, the configuration will be tailored
+    to the generateSW use case.
+    If --swSrc is provided, the configuration will be tailored
+    to the injectManifest use case.
 
   generateSW [<path/to/config.js>]
     Creates a new service worker file based on the options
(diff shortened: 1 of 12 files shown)


packages/workbox-cli/src/lib/help-text.js
```diff
@@ -18,9 +18,13 @@ module.exports = `Usage:
 $ workbox <command> [options]
 
 Commands:
-  wizard
+  wizard [--swSrc=<path/to/service-worker.js>]
```
addyosmani: A lot of tools opt for `<name> init` or `<name> create` to designate the action you're about to perform. Is there a reason we felt `wizard` made more sense here?

addyosmani:
@gauntface was your suggestion that `--inject-manifest` trigger a prompt? I like the general idea here. Not a huge fan of `--swSrc=` (perhaps we could simplify that further or opt for the prompt suggestion?)

jeffposnick:
Oh, I thought https://github.com/GoogleChrome/workbox/issues/1052#issuecomment-356352725 settled it on `--swSrc`.

But `--injectManifest` it is, then. And then I'll add in an additional question to ask for the `swSrc` value.

jeffposnick:
PTAL.

nasearle:
Is there any way I can test this change before it gets published?

PR #1173, opened 2018-01-12
Properly deal with webpack's publicPath + importWorkboxFrom: 'local'

### LINKED ISSUE #1172: Issue with importWorkboxFrom: 'local' in webpack plugin
**Library Affected**:
*workbox-webpack-plugin*

When trying out `GenerateSW({importWorkboxFrom: 'local'})` I ended up with the following:

```js
importScripts(
  "precache-manifest.d7157b2a92a7f026845b1dc0da6f7da1.js",
  "/"
);

workbox.setConfig({modulePathPrefix: "workbox-v3.0.0-alpha.5"});
```

That `"/"` should have been `"workbox-v3.0.0-alpha.5/workbox-sw.js"`.

jeffposnick:
R: @addyosmani @gauntface

Fixes #1172 

This was due to a missing `()` around an `||` operator.

### DIFF
--- a/test/workbox-webpack-plugin/node/generate-sw.js
+++ b/test/workbox-webpack-plugin/node/generate-sw.js
@@ -18,6 +18,46 @@ describe(`[workbox-webpack-plugin] GenerateSW (End to End)`, function() {
   const WORKBOX_DIRECTORY_PREFIX = 'workbox-';
   const WORKBOX_SW_FILE_NAME = getModuleUrl('workbox-sw');
   const SRC_DIR = path.join(__dirname, '..', 'static', 'example-project-1');
+  const ALL_WORKBOX_FILES = [
+    'workbox-background-sync.dev.js',
+    'workbox-background-sync.dev.js.map',
+    'workbox-background-sync.prod.js',
+    'workbox-background-sync.prod.js.map',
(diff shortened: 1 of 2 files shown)


test/workbox-webpack-plugin/node/generate-sw.js
```diff
@@ -274,6 +274,120 @@ describe(`[workbox-webpack-plugin] GenerateSW (End to End)`, function() {
...
+        const swFile = path.join(outputDir, 'service-worker.js');
+        try {
+          // Validate the copied library files.
+          const libraryFiles = glob.sync(`${WORKBOX_DIRECTORY_PREFIX}*/*.js*`,
+            {cwd: outputDir});
+
+          const modulePathPrefix = path.dirname(libraryFiles[0]);
+
+          const basenames = libraryFiles.map((file) => path.basename(file));
+          expect(basenames).to.eql([
```
addyosmani: My only passing comment is that it appears lines 197 and 311 execute `expect(basenames).to.eql()` against the same list of dev, prod and map files - should this list of basenames be turned into a variable if they're the same between both tests? If I missed something and they're not exactly 1:1, feel free to ignore.

jeffposnick: Good point. Changing!

PR #1177, opened 2018-01-16
cdnUtils fix + tests.

### LINKED ISSUE #1174: workbox-build getModuleUrl should throw for no input
Running:

```javascript
workboxBuild.getModuleUrl()
```

Will return the URL: `https://storage.googleapis.com/workbox-cdn/releases/3.0.0-alpha.5/undefined.js`

It should throw when no module name is passed in.

jeffposnick:
R: @addyosmani @gauntface

Fixes #1174

Added some missing tests for that module as well.

### DIFF
--- a/packages/workbox-build/src/lib/cdn-utils.js
+++ b/packages/workbox-build/src/lib/cdn-utils.js
@@ -13,7 +13,11 @@
   See the License for the specific language governing permissions and
   limitations under the License.
 */
+
+const assert = require('assert');
+
 const cdn = require('../cdn-details.json');
+const errors = require('./errors');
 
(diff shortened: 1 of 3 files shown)


jeffposnick:
I'll update https://github.com/GoogleChrome/workbox/pull/1177/files#diff-42e001337515b82bde40434a491fc0c3R28 to use `workbox-routing` instead, though I think the same codepath will be covered.

gauntface:
@jeffposnick it's more that workbox-sw is a special case compared to all the other modules, so checking for workbox-sw.js and workbox-routing.dev.js and workbox-routing.prod.js seem like all the possible scenarios.

PR #1179, opened 2018-01-16
Use the native archiver.finalize() promise

### LINKED ISSUE #1165: CDN publishing fails when part of the larger gulp publish chain
This is the error thrown by `gulp publish-github` when we do a release via `gulp publish` and all steps are chained.

```
[10:27:06] 'publish-github:generate-from-tags' errored after 1.05 min
[10:27:06] TypeError: Cannot read property 'objectMode' of undefined
    at Readable.Writable.write (/usr/local/google/home/mattgaunt/Projects/Code/workbox/node_modules/readable-stream/lib/_stream_writable.js:305:45)
    at ReadStream.ondata (_stream_readable.js:628:20)
    at emitOne (events.js:115:13)
    at ReadStream.emit (events.js:210:7)
    at addChunk (_stream_readable.js:252:12)
    at readableAddChunk (_stream_readable.js:239:11)
    at ReadStream.Readable.push (_stream_readable.js:197:10)
    at onread (fs.js:2003:12)
    at FSReqWrap.wrapper [as oncomplete] (fs.js:664:17)
```

jeffposnick:
R: @addyosmani @gauntface

This *might* deal with #1165, though we should probably leave that bug open until we can confirm during the next release.

Regardless of whether it fixes it or not, it seems like a cleaner approach to creating our own Promise wrapper.

### DIFF
--- a/gulp-tasks/utils/publish-helpers.js
+++ b/gulp-tasks/utils/publish-helpers.js
@@ -140,17 +140,12 @@ const createArchive = async (tagName, fileExtension, format, options) => {
   const groupedBuildFiles =
     path.join(getBuildPath(tagName), GROUPED_BUILD_FILES);
 
-  await new Promise((resolve, reject) => {
-    const writeStream = fs.createWriteStream(outputFilePath);
-    writeStream.on('close', () => resolve(outputFilePath));
-
-    const archive = archiver(format, options);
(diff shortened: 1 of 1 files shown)


gauntface:
I reckon this would fix the issue we've been having. Would explain the initial failure and success after.

PR #1183, opened 2018-01-17
Minify self.babelHelpers.asyncToGenerator()

### LINKED ISSUE #1061: Figure out how to mangle the babelHelpers.asyncToGenerator variable name
**Library Affected**:
*workbox-core*

See https://github.com/GoogleChrome/workbox/pull/1048#issuecomment-344658853 for context. Shortening this shared variable name should shave off a few bytes across many of the bundles.

jeffposnick:

Fixes #1061

This might be *really* hacky, so feedback/pushback is welcome. In particular, I'm not 100% sure that I'm adding in the callbacks to load/save the name cache in the best spots, to avoid any edge cases where a different entry point to the build process leads to that code not being executed.

(I'm also filing this PR to get a rundown of what happens to the bundle sizes as logged by the PR Bot. Local testing indicates that we drop like ~300 bytes from `workbox-core.prod.js`, at least.)

### DIFF
--- a/gulp-tasks/utils/rollup-helper.js
+++ b/gulp-tasks/utils/rollup-helper.js
@@ -11,6 +11,10 @@ module.exports = {
   // plugin set up and used to ensure as consist set of tests
   // as possible.
   getDefaultPlugins: (buildType) => {
+    if (!global.uglifyNameCache) {
+      throw Error('global.uglifyNameCache needs to be initialized.');
+    }
+
     const plugins = [];
 
     const babelConfig = {
@@ -47,16 +51,18 @@ module.exports = {
             // private variable and names that start with a single underscore
             // followed by a letter. This restriction to avoid mangling
             // unintentional fields in our or other libraries code.
-            regex: /^_[A-Za-z]/,
+            regex: /(?:^_[A-Za-z])|(?:babelHelpers)|(?:asyncToGenerator)/,
             // If you are getting an error due to a property mangle
             // set this flag to true and the property will be changed
             // from '_foo' to '$_foo$' to help diagnose the problem.
             debug: false,
           },
         },
+        nameCache: global.uglifyNameCache,
       };
       plugins.push(
-        uglifyPlugin(uglifyOptions, minify),
+        replace({babelHelpers: 'self.babelHelpers'}),
+        uglifyPlugin(uglifyOptions, minify)
       );
     }
 
--- a/gulp-tasks/utils/rollup-helper.js
+++ b/gulp-tasks/utils/rollup-helper.js
@@ -5,15 +5,14 @@ const uglifyPlugin = require('rollup-plugin-uglify');
 
 const constants = require('./constants');
 const getVersionsCDNUrl = require('./versioned-cdn-url');
+const uglifyNameCacheHelper = require('./uglify-name-cache-helper');
 
 module.exports = {
   // Every use of rollup should have minification and the replace
   // plugin set up and used to ensure as consist set of tests
   // as possible.
   getDefaultPlugins: (buildType) => {
-    if (!global.uglifyNameCache) {
-      throw Error('global.uglifyNameCache needs to be initialized.');
-    }
+    const uglifyNameCache = uglifyNameCacheHelper.load();
 
     const plugins = [];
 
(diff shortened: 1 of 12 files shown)


gauntface:
The size drops look promising, but I'm pretty confused with what you are doing with the name cache and why it's required - could you explain a little more what is happening in these changes?

jeffposnick:
Writing the cache out to disk is required since when running `gulp build`, each package is built in its own process, kicked off by `lerna boostrap`. You can't share the build cache in memory, and need to serialize it to disk. (I'll add some comments to that effect.)

I'm going to spend a bit more time trying to understand the various entry points, other than `gulp build`, that someone might go through when producing the final production bundles, and ensure that the cache is cleaned up appropriately in between builds.

But at least the byte savings looks decent, so this is worth pursuing further.

gauntface:
@jeffposnick why are you writing the cache out to disk? Sorry I think is just a set of steps you've taken that I have no idea how or why that ended in this - not that it's a problem, just trying to get to the core reasoning for all of this.

gulp-tasks/utils/uglify-name-cache-helper.js
```diff
@@ -0,0 +1,18 @@
+const fs = require('fs-extra');
+
+const constants = require('./constants');
+
+module.exports = {
```
gauntface: Adding a comment here along the lines of:

To enable uglify to minify the async and await generators and the babel helpers, we need to share the cache names between the individual module builds.

This file will stashes the cache names locally and saves the cache names after each module build.

gulp-tasks/utils/rollup-helper.js
```diff
@@ -47,16 +51,18 @@ module.exports = {
...
             // unintentional fields in our or other libraries code.
-            regex: /^_[A-Za-z]/,
+            regex: /(?:^_[A-Za-z])|(?:babelHelpers)|(?:asyncToGenerator)/,
             // If you are getting an error due to a property mangle
             // set this flag to true and the property will be changed
             // from '_foo' to '$_foo$' to help diagnose the problem.
             debug: false,
           },
         },
+        nameCache: global.uglifyNameCache,
```
gauntface: Nit: does this need to be on the global scope? Could we require the ulgify helper and reference an object in the module?

jeffposnick:
- I moved the in-memory cache out of `global` and into an instance variable.
- There's a task that will clean up the JSON cache file at the start of `gulp build`.
- There are additional comments.

The one thing that I'm still worried about is what would happen if we cut a release in which only half of the modules end up getting tagged with a new version, and Uglify uses a different symbol for `asyncToGenerator()` for the new modules vs. what it used for the previous ones.

@gauntface, in v3, are there scenarios in which we'd end up cutting releases that only affect a few modules? I know lerna supports it, but at the same time, our CDN strategy implies that we'd always have every module tagged with the same version.

gauntface:
@jeffposnick The current build process is:

- Run lerna to bootstrap everything and bump version numbers
- Download the tagged source code from github
- Build the tagged source code (i.e. npm install, npm run build etc)
- Copy the builds files for all modules for this taggs version
- Upload them to CDN
- Update them to Github
- Publish demo updates using new version

This means that while the NPM files will be published out of sync with each other, this is completely seperate from the CDN and Github published builds.
 
The CDN and Github bundles will be a snapshot of all the modules built at that time.

The problem occurs with NPM builds when developers are using the built version of the files.

Main options are:
1. ensure babel helpers are consistently named in the build process
1. We publish new builds for packages (kind of breaking versioning for developers using src/).

gauntface:
Stupid Q: Why not just reduce the naming of babelHelpers to 'b' and the name of asyncToGenerator to 'a' ? 

This is the majority of the weight (I assume). If we can drop the `function` names as well it should improve size as well.

jeffposnick:
Hmm... so if we were going to brute-force things, I think I would prefer something similar to approach 1. in which we just used `rollup-replace-plugin` to rename all instances of `self.babelHelpers.asyncToGenerator` and `babelHelpers.asyncToGenerator` to `self.a`.

This could be done before UglifyJS runs, and I think it's safe to assume that UglifyJS would be smart enough to never overwrite `self.a` with its own renaming.

That's still a hack, but maybe it's a preferable hack, in that it doesn't require any of the UglifyJS name cache coordination that this current PR introduces?

gauntface:
@jeffposnick if we can get away with renaming like that, I'm 100% on board with that approach.

jeffposnick:
Closing in favor of #1231

PR #1213, opened 2018-01-19
Adding a bg sync demo

### LINKED ISSUE #1198: Add Demo for workbox-background-sync

gauntface:
R: @jeffposnick @philipwalton 

Adds a basic bg sync demo and a log whenever when the sync event first.

Fixes #1198.

### DIFF
--- a/demos/functions/views/demo/workbox-background-sync.hbs
+++ b/demos/functions/views/demo/workbox-background-sync.hbs
@@ -1,0 +1,24 @@
+<section>
+  <div class="container">
+    <div class="row">
+      <div class="col-lg-8 mx-auto">
+        <ol>
+          <li>Open DevTools</li>
+          <li>Go to the Console</li>
+          <li>Click this button:<br /><button class="register-event btn btn-outline-primary">Fetch Request for BG Sync</button></li>
+        </ol>
+      </div>
+    </div>
+  </div>
+</section>
+
+<script>
+  const registerEventBtn = document.querySelector('.register-event');
+
+  navigator.serviceWorker.register('/demo/workbox-background-sync/sw.js')
+  .then(() => {
+    registerEventBtn.addEventListener('click', () => {
+      fetch(`./example.txt`);
+    });
+  });
+</script>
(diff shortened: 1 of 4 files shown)


demos/functions/views/demo/workbox-background-sync.hbs
```diff
@@ -0,0 +1,24 @@
+<section>
+  <div class="container">
+    <div class="row">
+      <div class="col-lg-8 mx-auto">
+        <ol>
+          <li>Open DevTools</li>
+          <li>Go to the Console</li>
+          <li>Click this button:<br /><button class="register-event btn btn-outline-primary">Fetch Request for BG Sync</button></li>
```
prateekbh: we need to make network offline before this step.

demos/functions/views/demo/workbox-background-sync/sw.hbs
```diff
@@ -0,0 +1,14 @@
+{{{WORKBOX_SW_IMPORT}}}
+
+const queue = new workbox.backgroundSync.Queue('myQueueName');
+
+self.addEventListener('fetch', (event) => {
+  const pathname = new URL(event.request.url).pathname;
+  if (pathname === '/demo/workbox-background-sync/example.txt') {
```
prateekbh: I feel we should show how to add this to queue when fetch fails.

demo might make more sense that ways

demos/functions/views/demo/workbox-background-sync.hbs
gauntface: does that work with devtools?

Secondly it still shows that the request being replayed with the log message (although I will say I'm sad there isn't more info - like a summary of requests that succeeded, failed etc).

demos/functions/views/demo/workbox-background-sync/sw.hbs
gauntface: Do you fancy taking on this issue and doing this? I agree it makes more sense (although I must confess the current flow actually works as it is, so why wait until an error when you can just queue it?)

demos/functions/views/demo/workbox-background-sync.hbs
prateekbh: Doesn't it plays immediately, without it?

demos/functions/views/demo/workbox-background-sync/sw.hbs
prateekbh: Sure thing, i'll get it done. But the reason i'd like it this way is because

a.) Its more close to the way it would be used at times

b.) It shows how it would actually play after network coming back

gauntface:
@prateekbh is doing this.

PR #1217, opened 2018-01-22
BroadcastCacheUpdate.Plugin integration test

### LINKED ISSUE #1191: Add Integration Test to workbox-broadcast-cache-update

jeffposnick:
R: @addyosmani @gauntface

Fixes #1191 

In the course of writing this test I found two problems with the implementation, which this PR also fixes:

- The v3 `cacheDidUpdate` lifecycle callback no longer sets the `url` property on the parameter. It still sets the `request` property, and you can get the `url` from that, so I just switched over to getting it from there.
- The move from named properties for parameters to positional parameters led to an issue where `notifyIfUpdated()` was being called with the `url` and `cacheName` parameters transposed.

(I guess that's why we write integration tests!)

### DIFF
--- a/infra/testing/test-server.js
+++ b/infra/testing/test-server.js
@@ -40,6 +40,13 @@ app.get(/\/__WORKBOX\/buildFile\/(workbox-[A-z|-]*)(\.(?:dev|prod)\.(.*))*/, (re
   res.sendFile(path.join(libraryPath, libraryFileName));
 });
 
+let eTagCounter = 0;
+app.get('/test/uniqueETag', (req, res) => {
+  res.header('Cache-Control', 'no-cache');
+  res.header('ETag', eTagCounter++);
+  res.send('ignored');
+});
+
 let server = null;
 let requestCounts = {};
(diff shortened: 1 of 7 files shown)


infra/testing/test-server.js
```diff
@@ -40,6 +40,13 @@ app.get(/\/__WORKBOX\/buildFile\/(workbox-[A-z|-]*)(\.(?:dev|prod)\.(.*))*/, (re
   res.sendFile(path.join(libraryPath, libraryFileName));
 });
 
+let eTagCounter = 0;
```
jeffposnick: I noticed that all the "helper" routes that used to exist in v2 have been removed from the test server.

@gauntface, are you cool with this being added to the test server? It greatly simplifies the test logic. If you'd rather see the route registered in a different way, I'm happy to refactor.

gauntface: It's fine with me - I just removed them as I was unsure as to what would be good and what wouldn't needed so please add what you want.

PR #1221, opened 2018-01-23
rangeRequests.Plugin integration test

### LINKED ISSUE #1194: Add Integration Test to workbox-range-requests

jeffposnick:
R: @addyosmani @gauntface

Fixes #1194


jeffposnick:
(I tweaked the test a bit to use a synthetic cached response, and a route that matches the URL, instead of a file that exists on the server. This ensures that we don't get a false positive due to the server handling the `Range:` request without service worker intervention.)

test/workbox-range-requests/static/sw.js
```diff
@@ -0,0 +1,17 @@
+importScripts('/__WORKBOX/buildFile/workbox-core');
+importScripts('/__WORKBOX/buildFile/workbox-range-requests');
+importScripts('/__WORKBOX/buildFile/workbox-routing');
+importScripts('/__WORKBOX/buildFile/workbox-strategies');
+
+workbox.routing.registerRoute(
+  new RegExp('dummy\\.txt$'),
+  workbox.strategies.cacheFirst({
```
gauntface: Can you change this to cacheOnly?

test/workbox-range-requests/integration/range-requests-plugin.js
```diff
@@ -0,0 +1,76 @@
...
+
+    if (error) {
+      throw error;
+    }
+  };
+
+  it(`should return a partial response that satisfies the request's Range: header, and an error response when it can't be satisfied`, async function() {
+    const testPageUrl = `${testServerAddress}/test/workbox-range-requests/static/`;
+    const swUrl = `${testPageUrl}sw.js`;
+    const dummyUrl = `${testPageUrl}dummy.txt`;
```
gauntface: Stupid thing, but could you change `dummy.txt` to `this-file-doesnt-exist.txt` just to make it painfully clear that it HAS to come from the cache and via the SW.

PR #1231, opened 2018-01-25
babelHelpers.asyncToGenerator minification, take two.

### LINKED ISSUE #1061: Figure out how to mangle the babelHelpers.asyncToGenerator variable name

jeffposnick:

Fixes #1061 

This is an alternative to #1183 that might be less finicky.

Instead of relying on UglifyJS to rename things for us, this brute-forces exposing the `babelHelpers.asyncToGenerator` helper as `self.Z` (an arbitrary 1 character property name) and then imposes a string replacement in the rollup bundling process to convert `babelHelpers.asyncToGenerator` to `self.Z`.

### DIFF
--- a/gulp-tasks/utils/external-helpers.js
+++ b/gulp-tasks/utils/external-helpers.js
@@ -2,35 +2,36 @@
 // This is extracted from the Babel runtime (original source: https://github.com/babel/babel/blob/9e0f5235b1ca5167c368a576ad7c5af62d20b0e3/packages/babel-helpers/src/helpers.js#L240).
 // As part of the Rollup bundling process, it's injected once into workbox-core
 // and reused throughout all of the other modules, avoiding code duplication.
+// It's exposed as self.Z as a brute-force minification technique, and rollup
+// replaces references to babelHelpers.asyncToGenerator with self.Z as part of
+// the bundling process.
 // See https://github.com/GoogleChrome/workbox/pull/1048#issuecomment-344698046
 // for further background.
-self.babelHelpers = {
-  asyncToGenerator: function(fn) {
-    return function() {
-      var gen = fn.apply(this, arguments);
-      return new Promise(function(resolve, reject) {
-        function step(key, arg) {
-          try {
-            var info = gen[key](arg);
-            var value = info.value;
-          } catch (error) {
-            reject(error);
-            return;
-          }
-
-          if (info.done) {
-            resolve(value);
-          } else {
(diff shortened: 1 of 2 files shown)


gulp-tasks/utils/external-helpers.js
```diff
@@ -2,35 +2,36 @@
...
-      return new Promise(function(resolve, reject) {
-        function step(key, arg) {
-          try {
-            var info = gen[key](arg);
-            var value = info.value;
-          } catch (error) {
-            reject(error);
-            return;
-          }
+self.Z = function(fn) {
```
gauntface: Two things:

1.) Can we drop `function` for arrows (should be smaller).
2.) Should this be workbox.Z (or something to that affect. Could be workbox._). Main thinking is that self.Z is still user space.

Comparing:
```
self.Z
self._wb
self.workbox._
self.babelHelpers.asyncToGenerator
```

gulp-tasks/utils/rollup-helper.js
```diff
@@ -32,6 +32,16 @@ module.exports = {
     };
     plugins.push(babel(babelConfig));
 
+    // This is a brute-force approach to minifying the babel-helpers code.
+    // self.Z is defined in our external-helper.js file and is used as the
+    // exposed interface for alias for babelHelpers.asyncToGenerator.
+    // See https://github.com/GoogleChrome/workbox/issues/1061
+    plugins.push(
+      replace({
+        'babelHelpers.asyncToGenerator': 'self.Z',
```
gauntface: do we need to include the `self.` with this or is this just a safety thing?

gauntface:
I think this is the way we should go. I am a little concerned with naming of just `self.Z` Could we do `workbox._` or something along those lines?

gauntface:
@jeffposnick thoughts on the comment above?

gauntface:
Gonna close this - please re-open if you want to discuss it further.

jeffposnick:
Oops! Sorry for letting this drop.

It boils down to not understanding how the combination of minification and rollup code would deal with various combinations of suggested solutions, as well as where/when we'd have to initialize the `workbox` global object to use as a namespace (since there's already code that does that automatically).

It was hard prioritizing those explorations, and given that the current file size doesn't seem objectionable, I'm fine with keeping this closed.

PR #1245, opened 2018-01-26
Adding range request demo and making function - non-static

### LINKED ISSUE #1200: Add Demo for workbox-range-request

gauntface:
R: @jeffposnick @addyosmani 

Fixes #1200

I added the demo for range requests but made one change to the range request plugin.

At the moment we have a number of plugins exposed but they all require instantiating since they have their own constructor options. Range raquests has a single static method meaning it **doesn't** need instantiating like the others. This felt pretty weird so I've undone it in this PR.

Hopefully that's ok in the name of consistency.

### DIFF
--- a/demos/functions/views/demo/workbox-range-requests.hbs
+++ b/demos/functions/views/demo/workbox-range-requests.hbs
@@ -1,0 +1,36 @@
+<section>
+  <div class="container">
+    <div class="row">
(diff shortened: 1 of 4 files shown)


gauntface:
Tested manually - going to merge - need to investigate the timeouts that have started happening (Issue here: https://github.com/GoogleChrome/workbox/issues/1251)

PR #1257, opened 2018-01-30
Validate plugin input / output

### LINKED ISSUE #1206: Add assertion checks to plugin return types.
I don't think it's particularly strong in terms of checks and resulted in some issues already.

gauntface:

Fixes #1206

This adds some additional checks to plugin return types and inputs.

### DIFF
--- a/packages/workbox-core/_private/cacheWrapper.mjs
+++ b/packages/workbox-core/_private/cacheWrapper.mjs
@@ -107,7 +107,7 @@ const matchWrapper = async (cacheName, request, matchOptions, plugins = []) => {
         });
       if (process.env.NODE_ENV !== 'production') {
         if (cachedResponse) {
-          assert.isInstance(cachedResponse, 'Response', {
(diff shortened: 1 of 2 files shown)


packages/workbox-core/_private/cacheWrapper.mjs
```diff
@@ -38,9 +38,11 @@ const putWrapper = async (cacheName, request, response, plugins = []) => {
   let responseToCache = await _isResponseSafeToCache(
     request, response, plugins);
 
-  // TODO If response is not safe to cache - print info to log.
-
   if (!responseToCache) {
+    if (process.env.NODE_ENV !== 'production') {
+      logger.debug(`Response '${getFriendlyURL(request.url)}' is not safe ` +
```
jeffposnick: Can you change this message to something like "...does not meet the criteria to be cached?" instead of talking about safety? I could see a developer being freaked out if they think that something about their web traffic is unsafe.

Issue #1261, opened 2018-01-31
InjectManifest in Webpack doesn't support publicPath for `workbox-sw` importWorkboxFrom as a chunk


raejin:
**Library Affected**:
workbox-webpack-plugin@3.0.0-alpha.6

**Browser & Platform**:
all browsers

**Issue or Feature Request Description**:

### Update 1/31/18
Now that I actually tried including the bundle, I don't believe this would have worked since the service worker script will not be aware of Webpack bundling. Maybe InjectManifest doesn't support importing `workbox-sw` from a Webpack chunk? I guess the question is: how do I import a local copy of `workbox-sw` from a Webpack bundle without doing copypasta? As according to [the doc](https://developers.google.com/web/tools/workbox/modules/workbox-webpack-plugin#injectmanifest_plugin), this is possible.

---
When using InjectManifest to configure Workbox and to include `workbox-sw` as a Webpack chunk, the generated service worker script will import the `workbox-sw` bundle as a relative path.

The Workbox Webpack plugin configuration looks like the following:

```js
module.exports = {
  /* Webpack specific configuration */
  importWorkboxFrom: 'workbox-sw.chunk.name',
  swSrc: require.resolve('./sw-template'),

  chunks: chunksToBeCached,

  swDest: 'sw.js',

  /* Workbox specific configuration */
  globDirectory: path.join(webpackConstants.rootPath),
};
```

Given the `compilation.options.output.publicPath` is `//localhost.subdomain.com/static/`
The above setting will generate the following service worker script:

```js
importScripts("//localhost.subdomain.com/static/precache-manifest.8bd42eef2c7ecd22.js", "workbox-sw.chunk.name-c7ecd22ced9716dbf6b2c15.js");

// other stuff
```

Which I would like the imported `workbox-sw` chunk path to also prefix with `compilation.options.output.publicPath`.

I quickly tried patching this by prefixing `compilation.options.output.publicPath` to the `chunk.files` array. https://github.com/GoogleChrome/workbox/blob/a100a513e0e259a35d93bc2c852a1eed02436501/packages/workbox-webpack-plugin/src/lib/get-workbox-sw-imports.js#L60-L66

My attempt fix: https://github.com/GoogleChrome/workbox/commit/40132067a707c1ce7a1a8da05b10eedefbc1c297

gauntface:
@jeffposnick this was something I struggled with.

[...]

PR #1265, opened 2018-01-31
Make workbox sw chunk prefix with public path

raejin:

Fixes https://github.com/GoogleChrome/workbox/issues/1261

This fix is intended for the following two configurations:
- Webpack plugin InjectManifest is used
- Webpack compiler is configured with `output.publicPath`
- `importWorkboxFrom` plugin option is set to a string, which points to a valid Webpack chunk name.

The fix is to ensure the generated `importScripts` statement will prefix the Webpack workbox chunk with `output.publicPath`


raejin:
I signed it!

PR #1268, opened 2018-02-01
Convert bg sync maxRetentionTime to minutes

### LINKED ISSUE #1254: maxRetentionTime in bg sync
I just noticed that `maxRententionTime` is in ms, which I'm not sure makes sense given that sync events generally takes minutes to trigger if they aren't instant, so should we change this to minutes instead of ms?

philipwalton:
R: @jeffposnick @prateekbh  @gauntface

Fixes #1254

Also fixes some other uncaught testing issues and moves the `test-Queue.mjs` file into the correct folder.

### DIFF
--- a/test/workbox-background-sync/node/test-Queue.mjs
+++ b/test/workbox-background-sync/node/test-Queue.mjs
@@ -16,13 +16,13 @@
 import {expect} from 'chai';
 import {reset as iDBReset} from 'shelving-mock-indexeddb';
 import sinon from 'sinon';
-import expectError from '../../../../infra/testing/expectError';
-import {Queue} from '../../../../packages/workbox-background-sync/Queue.mjs';
+import expectError from '../../../infra/testing/expectError';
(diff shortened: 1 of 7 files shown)


test/workbox-background-sync/node/test-Queue.mjs
```diff
@@ -258,7 +260,7 @@ describe(`[workbox-background-sync] Queue`, function() {
       });
 
       const queue = new Queue('foo', {
-        maxRetentionTime: 1000,
+        maxRetentionTime: 1000 / MINUTES,
```
gauntface: Nit: I'm not sure what this is doing - 1000 / 60000 === 0.016666667 minutes? Could you just make this 1 minute and tick the clock by one minute and 1 ms?

philipwalton: Ahh, yeah, for some reason I was thinking that we'd actually have to wait that long in the test.

philipwalton: Done.

PR #1275, opened 2018-02-02
Support for the new webpack v4 plugin architecture.

### LINKED ISSUE #1151: Support for webpack 4 plugin system
**Library Affected**:
*workbox-webpack-plugin*

As mentioned in https://medium.com/webpack/the-new-plugin-system-week-22-23-c24e3b22e95 and probably elsewhere, the plugin system is changing in webpack 4.

We need to evaluate those changes and release a compatible version of `workbox-webpack-plugin` in the future.

(If there's anyone in the community familiar with migrating to webpack 4's architecture, we'd love some help!)

jeffposnick:
R: @addyosmani @gauntface
CC: @goldhand @TheLarkInn

Fixes #1151 

I'd like to get this out there in the next release to at least ensure that folks who are early adopters of webpack v4 get a chance to try it out in real-world scenarios. It's a small change that should be backwards compatible with users still on webpack v3.

This has been a bit hard to test manually, since most of our webpack integration tests rely on other plugins that aren't yet webpack v4 compatible. I did some manual tests of basic compilations using the latest webpack v4 beta, and it seemed to work, though.

I'm also not sure how we could update our tests to work with both webpack v3 and v4, along with the corresponding versions of plugins that are compatible with each. The strategy moving forward might be just to keep our tests against webpack v3 for now, and migrate to testing against webpack v4 once that is fully released and gets some traction.

### DIFF
--- a/packages/workbox-webpack-plugin/package.json
+++ b/packages/workbox-webpack-plugin/package.json
@@ -32,7 +32,7 @@
     "workbox-build": "^3.0.0-alpha.6"
   },
   "peerDependencies": {
-    "webpack": "^2.0.0 || ^3.0.0"
+    "webpack": "^2.0.0 || ^3.0.0 || ^4.0.0"
   },
   "author": "Will Farley <a.will.farley@gmail.com> (http://www.willfarley.org/)",
   "license": "Apache-2.0",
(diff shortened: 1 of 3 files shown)


gauntface:
@jeff could you have tests that have a package.json file, run npm install and then run our plugins against a webpack config? Just as a smoke test would be helpful.

jeffposnick:
Coming up with a reasonable way to test across webpack v3 and v4 is tracked in #1281.

Given that it would be nice to make this available to folks currently trying out webpack v4 sooner rather than later, I'm going to merge and we can follow-up there.

TheLarkInn:
Thanks for all the support here! 💪

Issue #1267, opened 2018-02-01
`importScripts` include duplicated entries in Webpack development


raejin:
**Library Affected**:
workbox-webpack-plugin@3.0.0-alpha.6

**Browser & Platform**:
all browsers

**Issue or Feature Request Description**:
We are using Webpack dev server to serve JavaScript from a separate subdomain, as per new changes come in the dev server will re-compile and update the asset. I notice the `importScripts` statement started to include multiple precache manifest script URLs like the following:

```js
importScripts("//webpack.localhost.com/static/workbox-sw.js", "//webpack.localhost.com/static/precache-manifest.ac66c5c0aebbbfd12eeeb04b549b27f6.js", "//webpack.localhost.com/static/precache-manifest.e9a5c1bdd2a12c4a2391588f6670ea25.js", "//webpack.localhost.com/static/precache-manifest.ac66c5c0aebbbfd12eeeb04b549b27f6.js", "//webpack.localhost.com/static/precache-manifest.edc98274e5c08f4c770765983151ad0e.js", "//webpack.localhost.com/static/precache-manifest.09533f69635032c7cded7a71bba623d4.js");
```

By digging into the source, I noticed that we are pushing a new precache-manifest path to the same array on every Webpack emit:
https://github.com/GoogleChrome/workbox/blob/f2adc5664c7827a0a926f0896f808d8849583c5c/packages/workbox-webpack-plugin/src/inject-manifest.js#L101-L102

I managed to fix the problem by constructing a new array every time, as to make sure we don't mutate the plugin config:
https://github.com/raejin/workbox/commit/44b2d9cbd9ce1673818a095a555f970b334f8c80#diff-4b4d07cea6eee21561a9e810df4793bcL101

If this sounds reasonable, I'm happy to submit a PR with tests. Thanks!

gauntface:
That PR looks great to me although I'm not 100% what this is for: `Array.prototype.push.apply(importScriptsArray, workboxSWImports);`

jeffposnick:
Thanks for pointing this out, @raejin!

The repeated compilations use case isn't something that's been on our radar, so I wonder whether there are other areas that we would to account for in regards to resetting back to initial state.

Do you have a good way to reproduce this without fully adopting your webpack dev server setup so that we can confirm whether anything else might be amiss?

raejin:
[...]

PR #1276, opened 2018-02-02
Prevent multiple precache manifest in webpack development

raejin:

Fixes https://github.com/GoogleChrome/workbox/issues/1267

When in Webpack development mode where, regardless of using Webpack dev server, if Webpack watches the source change and re-compiles accordingly, the bug manifests itself as:

- precache-mainfest.js & workbox-sw path will be included multiple times, within `importScripts` from the generated service worker script.
- It appears in both `InjectManifest` and `GenerateSW`

### How to reproduce the bug
By `git checkout` the fixes on the latest `v3`, and run it against the tests.


raejin:
screen shot 2018-02-02 at 14 39 49

Could this be due to a flakey integration test?

PR #1280, opened 2018-02-05
adding bg-sync tag name check

prateekbh:

Adds a check for tag name before replaying the requests in background sync queue.

### DIFF
--- a/packages/workbox-background-sync/Queue.mjs
+++ b/packages/workbox-background-sync/Queue.mjs
@@ -94,8 +94,10 @@ class Queue {
     await this._runCallback('requestWillEnqueue', storableRequest);
     await this._queueStore.addEntry(storableRequest);
     await this._registerSync();
-    logger.log(`Request for '${request.url}' has been added to
-        background sync queue '${this._name}'.`);
+    if (process.env.NODE_ENV !== 'production') {
+      logger.log(`Request for '${request.url}' has been added to
+          background sync queue '${this._name}'.`);
+    }
   }
 
   /**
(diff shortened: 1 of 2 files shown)


packages/workbox-background-sync/Queue.mjs
```diff
@@ -94,6 +94,8 @@ class Queue {
     await this._runCallback('requestWillEnqueue', storableRequest);
     await this._queueStore.addEntry(storableRequest);
     await this._registerSync();
+    logger.log(`Request for '${request.url}' has been added to
```
gauntface: please wrap this in an if statement for `if (process.env.NODE_ENV !== 'production'){ ...log.... }` This allows rollup to remove the code in prod builds.

packages/workbox-background-sync/Queue.mjs
```diff
@@ -168,7 +170,9 @@ class Queue {
   _addSyncListener() {
     if ('sync' in registration) {
       self.addEventListener('sync', (event) => {
-        event.waitUntil(this.replayRequests());
+        if (event.tag === `${TAG_PREFIX}:${this._name}`) {
+          event.waitUntil(this.replayRequests());
```
gauntface: Can you add a log here to signal the sync event was received and replaying is being started.

packages/workbox-background-sync/Queue.mjs
prateekbh: Done

packages/workbox-background-sync/Queue.mjs
prateekbh: Done

PR #1284, opened 2018-02-07
Refactor intergration tests

### LINKED ISSUE #1251: Timeouts Occuring in Integration Tests
```
  1) [workbox-core] Load core in the browser
       "before each" hook for "should load workbox-core in a service worker.":
     Error: Timeout of 180000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

I've been putting off any work in trying to speed up our integration tests, so this may be a good time to do some house keeping and sure up our integration tests.

jeffposnick:
Fixes #1251 

Don't merge for now; I want to try running this on Travis and tweak some things as needed.

### DIFF
--- a/gulp-tasks/test-integration.js
+++ b/gulp-tasks/test-integration.js
@@ -33,8 +33,8 @@ const runFile = (filePath) => {
   });
 };
 
-const runIntegrationTestSuite =
-async (testPath, nodeEnv, seleniumBrowser) => {
+const runIntegrationTestSuite = async (testPath, nodeEnv, seleniumBrowser,
+                                       webdriver) => {
   logHelper.log(oneLine`
     Running Integration test on ${logHelper.highlight(testPath)}
     with NODE_ENV '${nodeEnv}'
@@ -51,13 +51,14 @@ async (testPath, nodeEnv, seleniumBrowser) => {
 
   try {
     global.__workbox = {
+      webdriver,
       seleniumBrowser,
       server: testServer,
     };
 
     const testFiles = glob.sync(
       path.posix.join(__dirname, '..', testPath, '*.js'));
-    for (let testFile of testFiles) {
+    for (const testFile of testFiles) {
       testServer.reset();
       await runFile(testFile);
     }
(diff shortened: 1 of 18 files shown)


jeffposnick:
Okay, this should be ready for review, @gauntface.

The major structural change is to reorder the top-level test runner so that the entire suite of tests across all packages, for a given browser + `NODE_ENV`, can be run under a single `webdriver` instance. This should speed things up because we're no longer killing/relaunching the browser instances for each individual test file.

In order to accommodate that change, I had to refactor a number of the individual test files. This was mainly a cleanup, and should result in more consistency and code reuse, but there were a few individual tests that required either some tweaking to the logic, or a new `before()` setup step, to clear out state, because they were previously written with the assumption that there would be a fresh browser instance every time they started.

gulp-tasks/test-integration.js
```diff
@@ -92,13 +103,13 @@ gulp.task('test-integration', async () => {
   }
 
   logHelper.log(`Downloading browsers......`);
-  await seleniumAssistant.downloadLocalBrowser('chrome', 'stable', 24);
-  await seleniumAssistant.downloadLocalBrowser('chrome', 'beta', 24);
-  await seleniumAssistant.downloadLocalBrowser('firefox', 'stable', 24);
-  await seleniumAssistant.downloadLocalBrowser('firefox', 'beta', 24);
+  const expiration = 24;
+  await seleniumAssistant.downloadLocalBrowser('chrome', 'stable', expiration);
```
gauntface: Why do one at a time? Maybe add a comment.

jeffposnick: It was previously one at a time. I tried doing them all at once in an interim commit as part of this PR, and it led to the following Travis CI run: https://travis-ci.org/GoogleChrome/workbox/builds/338757961#L3887

That shows that the Firefox tests never ran and Travis eventually timed out, so I just went back to the original sequential download.

I'll open a `selenium-assistant` issue to track whether there's any technical reason why simultaneous downloads might not work.

jeffposnick: See https://github.com/GoogleChromeLabs/selenium-assistant/issues/110

PR #1290, opened 2018-02-09
Support importWorkboxFrom: 'local' in the InjectManifest webpack plugin

### LINKED ISSUE #1260: GenerateSW vs InjectManifest - importWorkboxFrom
I'm little confused by the differences in the GenerateSW and InjectManifest allowed values for `importWorkboxFrom`.

Why does inject manifest not allow 'local'?

jeffposnick:
R: @gauntface
CC: @raejin 

Fixes #1260 

This adds `importWorkboxFrom: 'local'` support to the `InjectManifest` webpack plugin. (It's still not supported in `injectManifest` mode of the `workbox-build` or `workbox-cli`.)

Assuming this PR is merged, I'll need to update https://github.com/google/WebFundamentals to reflect the new config.


test/workbox-webpack-plugin/node/generate-sw.js
```diff
@@ -80,10 +80,13 @@ describe(`[workbox-webpack-plugin] GenerateSW (End to End)`, function() {
       const compiler = webpack(config);
       compiler.run((webpackError) => {
         if (webpackError) {
-          expect(webpackError.message.includes('importWorkboxFrom'));
-          done();
+          if (webpackError.message.includes('importWorkboxFrom')) {
```
jeffposnick: FYI, in the course of writing this PR, I discovered that this test pattern wasn't actually working as it was initially implemented: it should have been `expect(webpackError.message.includes('importWorkboxFrom')).to.be.true`, although that would then cause an unhandled promise rejection rather than a clean error to report back to `mocha`.

I took the opportunity to clean up this test along with the corresponding test in the `inject-manifest.js` suite so that it properly detects the failure and reports it using the `done()` callback.

packages/workbox-webpack-plugin/src/inject-manifest.js
```diff
@@ -113,8 +116,13 @@ class InjectManifest {
       .map(JSON.stringify)
       .join(', ');
 
-    const postInjectionSWString = `importScripts(${importScriptsString});
+    const setConfigString = modulePathPrefix
+      ? `workbox.setConfig({modulePathPrefix: ` +
+        `${JSON.stringify(modulePathPrefix)}});`
```
raejin: Do you think `JSON.stringify` is necessary?

test/workbox-webpack-plugin/node/inject-manifest.js
```diff
@@ -240,12 +256,87 @@ describe(`[workbox-webpack-plugin] InjectManifest (End to End)`, function() {
...
+          expect(context.self.__precacheManifest).to.eql(expectedEntries);
+
+          done();
+        } catch (error) {
+          done(error);
+        }
+      });
+    });
+
+    it(`should support setting importWorkboxFrom to 'local'`, function(done) {
```
raejin: It'd be cool to include a test where Webpack `publicPath` is specified

packages/workbox-webpack-plugin/src/inject-manifest.js
jeffposnick: It's safer than wrapping the value in either `"..."` or `'...'`, just on the off chance modulePathPrefix contains any quote characters. (Not that it should, but still.)

`JSON.stringify()` is guaranteed to return a properly escaped string.

test/workbox-webpack-plugin/node/inject-manifest.js
jeffposnick: Done.

raejin:
@gauntface @jeffposnick 

After testing out `v3.0.0-beta.1`, I think this is not compatible with static asset that gets served from a different domain. By tracing the source a bit, we are copying the file on the disk instead of copying the file to the Webpack compilation. 

https://github.com/raejin/workbox/blob/17c38de990585d7076b03ecca55d33b3166e53e9/packages/workbox-build/src/lib/copy-workbox-libraries.js#L78-L79

What do you guys think of a different implementation on copying on disk versus Webpack compilation?

jeffposnick:
Rather than discussing this on a merged PR, I created https://github.com/GoogleChrome/workbox/issues/1338

PR #1298, opened 2018-02-10
Safari tech preview

gauntface:

This makes the tests work on safari tech preview. I won't lie there could easily be stuff that isn't ideal in here but I've tried to seperate Safari from everything else. This will probably need local testing. I need to figure out a way to validate version in selenium-assistant if we need Safari 49.


jeffposnick:
Thanks for fighting through the mess.

Issue #1188, opened 2018-01-18
Handle cache-misses by falling back to network (by default?)


jeffposnick:
**Library Affected**:
`workbox-strategies`, `workbox-precaching`, `workbox-routing`

There are three places where cached responses would normally be used, but, in the case of a cache miss, falling back to the network might be preferable to a hard failure.

- [x] One is in the [`NetworkFirst` strategy](https://github.com/GoogleChrome/workbox/blob/v3/packages/workbox-strategies/NetworkFirst.mjs#L115) when `networkFirstTimeout` is set. You could imagine that strategy being applied with an empty cache, and the timeout being reached, but there not being any cached response to end up using. In the current codebase, we'd end up resolving with `null` as soon as the timeout hits rather than waiting for an eventual network response.

- [x] Another is in [`workbox.precaching`'s `default` export](https://github.com/GoogleChrome/workbox/blob/v3/packages/workbox-precaching/_default.mjs#L213), where a user who somehow ends up with a missing precached entry, that will result in the request failing immediately, instead of falling back to the network. (FWIW, `sw-precache` [automatically fell back](https://github.com/GoogleChromeLabs/sw-precache/blob/master/service-worker.tmpl#L171) to the network in this scenario.)

- [x] And finally, [`workbox.routing`'s `default` export](https://github.com/GoogleChrome/workbox/blob/v3/packages/workbox-routing/_default.mjs#L164) has a `registerNavigationRoute()` method that only does a `caches.match()` without a fallback.

In v2, you could accomplish something similar `workboxSW.router.setCatchHandler()`, but in v3, `workbox.precaching` doesn't end up using `Router` under the hood.

What are folks' thoughts on adding in network fallbacks by default in v3? I don't know if we'd need to allow developers to opt-out of the automatic fallbacks, but if we did, it might get a little messy figuring out how to do that, since it would need to be a setting that both `workbox.precaching` and `workbox.routing` could read. I guess it could be in `workbox.core`?

gauntface:
1. We could definitely be smarter here.
1. I'm lost as to when this could be the case unless we aren't throwing an error (resulting in an aborted sw install). But still I'm ok with adding it in as a last resort.
1. Again agree we should change this behavior, although we should probably throw a warning in dev when caches.match is a miss and we fall back to the network - just to encourage precaching.

DavidScales:
[...]

PR #1300, opened 2018-02-10
Fallback to network on navigationRoute cache miss

jeffposnick:
R: @addyosmani @gauntface

This is 1/3 of #1188

It adds network fallback behavior to the `workbox-routing`'s navigation route in the default export.

### DIFF
--- a/packages/workbox-routing/_default.mjs
+++ b/packages/workbox-routing/_default.mjs
@@ -161,14 +161,27 @@ class DefaultRouter extends Router {
     }
 
     const cacheName = cacheNames.getPrecacheName(options.cacheName);
-    const handler = () => caches.match(cachedAssetUrl, {cacheName});
+    const handler = () => caches.match(cachedAssetUrl, {cacheName})
+      .then((response) => {
+        if (response) {
+          return response;
+        }
+        throw new Error(`The cache ${cacheName} did not have an entry for ` +
+          `${cachedAssetUrl}.`);
+      }).catch((error) => {
+        if (process.env.NODE_ENV !== 'production') {
+          logger.debug(`Unable to respond to navigation request with cached ` +
+            `response: ${error}. Falling back to network.`);
+        }
+        return fetch(cachedAssetUrl);
+      });
+
     const route = new NavigationRoute(handler, {
       whitelist: options.whitelist,
       blacklist: options.blacklist,
     });
-    super.registerRoute(
-      route
-    );
+    super.registerRoute(route);
+
     return route;
   }
 }
--- a/test/workbox-routing/node/test-default.mjs
+++ b/test/workbox-routing/node/test-default.mjs
@@ -316,6 +316,27 @@ describe(`[workbox-routing] Default Router`, function() {
       }));
       expect(response).to.equal(injectedResponse);
     });
+
+    it(`should fetch() when there's a cache miss for the registered URL`, async function() {
+      const fakeResponse = new Response('fake');
+      const fetchStub = sandbox.stub(self, 'fetch').returns(fakeResponse);
+
+      const cacheName = 'does-not-exist';
+      const shellUrl = '/does-not-exist.html';
+
+      const route = defaultRouter.registerNavigationRoute(shellUrl, {
+        cacheName,
+      });
+      const response = await route.handler.handle(new FetchEvent('fetch', {
+        request: new Request('/random/navigation.html', {
+          mode: 'navigate',
+        }),
+      }));
+
+      expect(fetchStub.calledOnce).to.be.true;
+      expect(fetchStub.firstCall.args[0]).to.eql(shellUrl);
+      expect(response).to.eql(fakeResponse);
+    });
   });
 
   describe(`Fetch Events`, function() {


packages/workbox-routing/_default.mjs
```diff
@@ -161,14 +161,27 @@ class DefaultRouter extends Router {
...
+      .then((response) => {
+        if (response) {
+          return response;
+        }
+        throw new Error(`The cache ${cacheName} did not have an entry for ` +
+          `${cachedAssetUrl}.`);
+      }).catch((error) => {
+        if (process.env.NODE_ENV !== 'production') {
+          logger.debug(`Unable to respond to navigation request with cached ` +
+            `response: ${error}. Falling back to network.`);
```
gauntface: This should be `error.message` or you should log it as a separate log call (possibly in a group).

packages/workbox-routing/_default.mjs
```diff
@@ -161,14 +161,27 @@ class DefaultRouter extends Router {
...
+          return response;
+        }
+        throw new Error(`The cache ${cacheName} did not have an entry for ` +
+          `${cachedAssetUrl}.`);
+      }).catch((error) => {
+        if (process.env.NODE_ENV !== 'production') {
+          logger.debug(`Unable to respond to navigation request with cached ` +
+            `response: ${error.message}. Falling back to network.`);
+        }
+        return fetch(cachedAssetUrl);
```
gauntface: If I set a `setCatchHandler` what happens? Won't this now block it?

Why don't we just have a default catch handler that is just the network.

jeffposnick:
> If I set a `setCatchHandler` what happens? Won't this change block it from working as intended?

Prior to this PR, any `catchHandler` you registered will be triggered when there's a cache miss and either a network failure or no response from the network for `networkTimeoutSeconds`.

With this PR, any `catchHandler` you registered will be triggered when there's a cache miss and a network failure, independent of `networkTimeoutSeconds`.

So that's a change in behavior for developers who are already using `catchHandlers`.

The argument in favor of this PR is that if we're in a world where "unexpected" cache misses are more likely, developers who would not think to use a `catchHandler` could benefit from Workbox doing more to provide a response by default.

> Why don't we just have a default catch handler that is just the network given the lengths we are going to basically try and fallback to the network.

`catchHandlers` are tied to a `Router` instance. Making the change to the underlying strategy, like we do in this PR, let's us default to behaviors for certain strategies that isn't tied to using `Router`.

packages/workbox-routing/_default.mjs
gauntface: Why isn't this the default setCatchHandler? It's in the router module, not the strategy.

gauntface:
@jeffposnick I'm confused by your comment:

> Making the change to the underlying strategy, like we do in this PR, let's us default to behaviors for certain strategies that isn't tied to using Router.

This PR is a single change to the Router module.

jeffposnick:
Apologies about that confusion. Serves me right for putting out different PRs in flight at the same time.

gauntface:
@jeffposnick is this still relevant?

jeffposnick:
It's still relevant, but it's not something we need to get implemented in time for v3.

I'd like to revisit some of the "cache resiliency" issues for a minor release, post v3.

jeffposnick:
CC: @philipwalton 

So I wanted to revisit this, particularly in light of https://github.com/GoogleChrome/workbox/issues/1441, in which I believe @coryrylan is bumping up against this edge case. (As far as I can tell this scenario the most likely culprit.)

To clarify some of the crossed wires in response to the initial feedback:

This is a focused changes that detects one of two possible failure modes when the handler associated with `workbox.routing.registerNavigationRoute()` is executing, and it attempts to read the "fallback ULR" from the cache:
- Either there's an exception that occurs during the initial `caches.match()`
- Or there's a cache miss
In either case, this PR will cause the handler to attempt to obtain the fallback URL via the network using `fetch()`, rather than failing completely (and resulting in a broken navigation).

If that `fetch()` fails, then any default `catchHandler` that's registered with `workbox.routing` will still get a chance to deal with *that* failure. So this change neither requires or prevents a user from registering a global `catchHandler`. 

Asking a user to manually deal with this situation via their own global `catchHandler` is problematic, because:
- The fallback URL and the request's URL aren't the same, and the `catchHandler` would only have access to the request's URL. So you wouldn't have the context to know that it's the fallback URL which should be `fetch()`ed inside of the global `catchHandler`.
- Adding in a default, global `catchHandler` that always did a `fetch()` can lead to some pathological scenarios. E.g., a network-first strategy times out after X seconds (and there's nothing in the cache), leading to an exception. If that exception triggers another `fetch()` automatically, then we're going to end up waiting another X seconds for the second failure before we finally give up. This PR just adds in a single `fetch()` that would execute if there's an issue with the cached response.

I took another pass through the code and added some comments to clarify these points, but I think it's a safe bet to move ahead with. (And it matches [what we were doing](https://github.com/GoogleChromeLabs/sw-precache/blob/master/service-worker.tmpl#L163-L177) in `sw-precache` previously.)

jeffposnick:
(Also, this PR uses `v3` as its base, and it's going to be a mess to rebase to the current `master`. Assuming there's a 👍, I'm going to close this out and create a fresh PR against `master`.)

philipwalton:
I'm good with this change, and I think falling back to network anytime we're dealing with cache (with the possible exception of cache only) is probably safest.

PR #1302, opened 2018-02-11
Fallback to fetch() if the precached response is missing.

jeffposnick:
R: @addyosmani @gauntface

This is the last 1/3 of #1188 

If the browser doesn't have a cache match for a URL that should have been precached, this will cause it to go to the network instead of returning an error.

### DIFF
--- a/packages/workbox-precaching/_default.mjs
+++ b/packages/workbox-precaching/_default.mjs
@@ -211,9 +211,24 @@ moduleExports.addRoute = (options) => {
     }
 
     let responsePromise = caches.open(cacheName)
-    .then((cache) => {
-      return cache.match(precachedUrl);
-    });
+      .then((cache) => {
+        return cache.match(precachedUrl);
+      }).then((cachedResponse) => {
+        if (cachedResponse) {
+          return cachedResponse;
+        }
+
+        // Fall back to the network if we don't have a cached response (perhaps
+        // due to manual cache cleanup).
+        if (process.env.NODE_ENV !== 'production') {
+          logger.debug(`The precached response for ` +
+            `${getFriendlyURL(precachedUrl)} in ${cacheName} was not found. ` +
+            `Falling back to the network instead.`);
+        }
+
+        return fetch(precachedUrl);
+      });
+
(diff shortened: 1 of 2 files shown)


packages/workbox-precaching/_default.mjs
```diff
@@ -211,9 +211,24 @@ moduleExports.addRoute = (options) => {
...
+
+        // Fall back to the network if we don't have a cached response (perhaps
+        // due to manual cache cleanup).
+        if (process.env.NODE_ENV !== 'production') {
+          logger.debug(`The precached response for ` +
+            `${getFriendlyURL(precachedUrl)} in ${cacheName} was not found. ` +
+            `Falling back to the network instead.`);
+        }
+
+        return fetch(precachedUrl);
```
gauntface: Should we cache the file as well?

jeffposnick: Good question. The downside of caching is that we could end up in a state where we have a response cached that doesn't match the revision information inlined in the manifest/stored in IDB.

I don't know if that's preferable to having nothing cached at all.

One advantage to just logging rather than also attempting to "repair" the cache is that a developer who sees this message would know that their cache is in an inconsistent state. If we attempted to repair the cache but end up caching a different revision of a given URL, that won't end up logging anything the next time the same resource is requested, and the inconsistency would be hidden.

When implementing something similar in `sw-precache`, I just went with the network + logging route:

https://github.com/GoogleChromeLabs/sw-precache/blob/2ee24a96bb4812e83f47f6859361ba5e1bd535f5/service-worker.tmpl#L171-L176

th3hunt:
Oh thank you!

After reading this about Safari, 
> To keep only the stored information that is useful to the user, WebKit will remove unused service worker registrations after a period of a few weeks. Caches that do not get opened after a few weeks will also be removed. Web Applications must be resilient to any individual cache, cache entry or service worker being removed.
https://webkit.org/blog/8090/workers-at-your-service/

I was looking for a workaround using some of the less popular workbox 2.x APIs.

Seeing this in v3 is a relief :)

PR #1307, opened 2018-02-13
Adds feature detection for BroadcastChannel

### LINKED ISSUE #1304: Broadcast Cache Update on v3 in Safari
Like V2 - we should ensure v3 doesn't break in safari if used.

jeffposnick:
R: @addyosmani @gauntface

Fixes #1304

### DIFF
--- a/packages/workbox-broadcast-cache-update/BroadcastCacheUpdate.mjs
+++ b/packages/workbox-broadcast-cache-update/BroadcastCacheUpdate.mjs
@@ -61,8 +61,9 @@ class BroadcastCacheUpdate {
   }
 
   /**
-   * @return {BroadcastChannel} The BroadcastChannel instance used for
-   * broadcasting updates.
+   * @return {BroadcastChannel|undefined} The BroadcastChannel instance used for
+   * broadcasting updates, or undefined if the browser doesn't support the
+   * Broadcast Channel API.
    *
    * @private
    */
--- a/packages/workbox-broadcast-cache-update/BroadcastCacheUpdate.mjs
+++ b/packages/workbox-broadcast-cache-update/BroadcastCacheUpdate.mjs
@@ -68,7 +68,7 @@ class BroadcastCacheUpdate {
    * @private
    */
   _getChannel() {
-    if (!this._channel) {
+    if (('BroadcastChannel' in self) && !this._channel) {
       this._channel = new BroadcastChannel(this._channelName);
     }
     return this._channel;
--- a/packages/workbox-broadcast-cache-update/broadcastUpdate.mjs
+++ b/packages/workbox-broadcast-cache-update/broadcastUpdate.mjs
@@ -14,6 +14,7 @@
 */
 
 import {assert} from 'workbox-core/_private/assert.mjs';
+import {logger} from 'workbox-core/_private/logger.mjs';
 import messageTypes from './messageTypes.mjs';
 import './_version.mjs';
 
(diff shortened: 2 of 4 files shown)


packages/workbox-broadcast-cache-update/broadcastUpdate.mjs
```diff
@@ -55,6 +56,18 @@ import './_version.mjs';
  * @memberof workbox.broadcastUpdate
  */
 const broadcastUpdate = (channel, cacheName, url, source) => {
+  // There are browsers which support service workers but don't support the
+  // Broadcast Channel API.
+  // See https://github.com/GoogleChrome/workbox/issues/1304
+  if (!('BroadcastChannel' in self)) {
```
gauntface: You need to check that the channel exists along with this - since you are setting the channel to undefined if it's not supported.

packages/workbox-broadcast-cache-update/BroadcastCacheUpdate.mjs
```diff
@@ -61,8 +61,9 @@ class BroadcastCacheUpdate {
   }
 
   /**
-   * @return {BroadcastChannel} The BroadcastChannel instance used for
-   * broadcasting updates.
+   * @return {BroadcastChannel|undefined} The BroadcastChannel instance used for
+   * broadcasting updates, or undefined if the browser doesn't support the
```
gauntface: This doesn't actually make sense - looking at the getChannel code - it's always creating a broadcast channel - so it never returns undefined.

packages/workbox-broadcast-cache-update/broadcastUpdate.mjs
gauntface: This should be checking if the channel exists as well

packages/workbox-broadcast-cache-update/BroadcastCacheUpdate.mjs
jeffposnick: Uch, I was locally running some test scenarios and I undid the line of code that was the whole point of this PR, and then submitted it.

jeffposnick:
Sorry, I accidentally committed a version of the code that was missing the actual check. PTAL.

packages/workbox-broadcast-cache-update/broadcastUpdate.mjs
gauntface: This still holds true for the PR I think - we should check if the channel exists and return if not.

packages/workbox-broadcast-cache-update/broadcastUpdate.mjs
```diff
@@ -55,6 +56,17 @@ import './_version.mjs';
  * @memberof workbox.broadcastUpdate
  */
 const broadcastUpdate = (channel, cacheName, url, source) => {
+  // There are browsers which support service workers but don't support the
+  // Broadcast Channel API.
+  // See https://github.com/GoogleChrome/workbox/issues/1304
+  if (!(('BroadcastChannel' in self) && channel)) {
```
gauntface: This just feels error prone - why don't we just check for no channel?

The current AND'ing of support with channel just feels OTT. We know our code won't have channel when the `BroadcastChannel` API is unavailable.

henrylearn2rock:
It would be great to include a note in the doc that 
Workbox Broadcast Cache Update does not work in Safari and mobile Safari https://developers.google.com/web/tools/workbox/modules/workbox-broadcast-cache-update

https://caniuse.com/#feat=broadcastchannel

philipwalton:
Since v4 this package has supported a fallback to `postMessage()` in Safari: https://github.com/GoogleChrome/workbox/releases/tag/v4.0.0

In v5 we'll likely switch to using `postMessage()` everywhere (i.e. not use `BroadcastChannel` at all) since it has some other benefits as well.

But yeah, it's probably worth mentioning in the docs that this **does** work in Safari.

Issue #1287, opened 2018-02-08
Fail the Webpack build if a chunk doesn't exist


raejin:
**Library Affected**:
workbox-webpack-plugin@^3.0.0-beta.0

**Browser & Platform**:
all browsers

**Issue or Feature Request Description**:
It would be problematic if we cannot track if a specified precaching Webpack chunk doesn't actually exist in the compilation. According to the source, `chunks` option is used as a whitelist to ensure the specified chunks to be considered from the compilation. 

Perhaps we can consider pushing an error to the compilation when a chunk is not found? This would make the precache-manifest more robust.

https://github.com/GoogleChrome/workbox/blob/v3/packages/workbox-webpack-plugin/src/lib/get-manifest-entries-from-compilation.js#L166-L172

gauntface:
@jeffposnick does this make sense? I'm 50/50. It's a whitelist and blacklist (i.e. allow these things to be included or not) but I'm not sure that it's a requirement that these things be included (at least, that's not how I think of whitelist and blacklist).

That being said, I don't know what a "chunk" is and I don't know how a chunk is formed and what the end result looks like in the config, so maybe this does make sense.

jeffposnick:
In v5, I'll add info about any mismatch between what's provided in `chunks` and what's found in the compilation to the webpack warnings. I don't think it's worth treating it as a more serious error, though.

jeffposnick:
This should be addressed by the current Workbox v5.0.0 alpha.

PR #1313, opened 2018-02-14
Fail Webpack build when `chunks` and `excludeChunks` are missing from compilation

raejin:
Explicit error messages thrown if any of the specified `chunks` or
`excludeChunks` is missing from the compilation. This is ensuring that
the workbox configuration is robust against any changes to the
compilation, resulting in confidence in precaching strategy.

Fixes https://github.com/GoogleChrome/workbox/issues/1287

This PR adds robustness to the Workbox Webpack configuration as we can be confident of the specified `chunks` and `excludeChunks` will be captured in the precache-manifest. If that is not the case, we would add to the compilation errors detailing what was missing.

### DIFF
--- a/packages/workbox-webpack-plugin/src/lib/get-manifest-entries-from-compilation.js
+++ b/packages/workbox-webpack-plugin/src/lib/get-manifest-entries-from-compilation.js
@@ -55,6 +55,7 @@ function getEntry(knownHashes, url, revision) {
  *  but filter for [/\.map$/, /asset-manifest\.json$/] by default:
  *    https://github.com/GoogleChrome/workbox/pull/808#discussion_r140565156
  *
+ * @param {Object} compilation webpack compilation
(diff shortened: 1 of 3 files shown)


raejin:
Hmm, I think we should also make this a configurable behavior.

jeffposnick:
It looks like we're going to head in the direction of allowing build methods to bubble up warnings to the wrapper that called them (either `workbox-build` or `workbox-webpack-plugin`). See https://github.com/GoogleChrome/workbox/issues/1315#issuecomment-366132669

My inclination is to try to lay the foundation for general warning reporting in a new PR, and then turn the logic that's currently in this PR into something that used that new mechanism. Does that sound reasonable? If so, I'd like to add the `chilling` label for now, and we can revisit in the near future.

raejin:
@jeffposnick 

That sounds great, thanks for context around future iteration!

PR #1317, opened 2018-02-16
Log warning for bad cache control

### LINKED ISSUE #620: Add Log to Highlight When a Caching Header if used for a service-worker js file
> This is, in my opinion, the issue most likely to cause developer frustration and confusion.
>
> It's very easy to unknowingly cache your service-worker.js file for even a short period of time (e.g. by deploying to GitHub pages, which caches everything for 10 minutes). This cache reuse has an upper bound of 24 hours, so it won't stretch on indefinitely, but during the window when the cache is in effect, the potential for developer (and operations) pain is very high.

We could request the service worker, check for cache headers and log a message to highlight when it's not zero. This could be part of a larger logging / welcome message.

gauntface:

Fixes #620

Adds a basic test for cache-control header.

### DIFF
--- a/packages/workbox-core/_default.mjs
+++ b/packages/workbox-core/_default.mjs
@@ -19,6 +19,7 @@ import {WorkboxError} from './_private/WorkboxError.mjs';
 import {cacheNames} from './_private/cacheNames.mjs';
 import {logger} from './_private/logger.mjs';
 import {assert} from './_private/assert.mjs';
+import {checkSWFileCacheHeaders} from './_private/checkSWFileCacheHeaders.mjs';
 import {setLoggerLevel, getLoggerLevel} from './_private/logger.mjs';
 import './_version.mjs';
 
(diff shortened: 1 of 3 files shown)


gauntface:
Chillin till I can checkout the size growth (This should only be a dev feature).

gauntface:
@jeffposnick a review would be appreciated - I figured out how to make rollup *not* include this stuff.

packages/workbox-core/_default.mjs
```diff
@@ -61,6 +62,8 @@ class WorkboxCore {
         `${padding}https://github.com/GoogleChrome/workbox/issues/new`
       );
       logger.groupEnd();
+
+      checkSWFileCacheHeaders();
```
jeffposnick: Small nit: could you wrap this in

```js
if (typeof checkSWFileCacheHeaders === 'function') {
  checkSWFileCacheHeaders();
}
```

just to double-check that we're not somehow in a scenario in which it's set to `null`? I can't really imagine why that would happen with our normal builds, since this is in a `NODE_ENV !== 'production'` block , but I wonder if some custom build scenarios might end up leading to a mismatch.

gauntface:
@jeffposnick it shouldn't ever happen (It's following the same approach as assert class - but I added in the check regardless)

PR #1318, opened 2018-02-16
Bumping depecencies

### LINKED ISSUE #924: Service Worker Mock Update
Hey folks. Just pushed an update to service-worker-mock (1.6.0) that should help decrease the custom mocks you have in codebase. Let me know if there are any upgrade issues!

## Changes

- Everything uses URL internally (self.location, request.url, etc)
- Updated FetchEvent mock to match spec
- Fixed Response.type for opaque responses
- Event was renamed to ExtendableEvent to match spec
- Header is now a proper Map
- Request extends Body and has all the corresponding methods (json, text, blob, etc)

gauntface:

Fixes #924 

Also brings us up to latest @std/esm and it seems to work with Proxyquire which is awesomesauce.

The naming of `const exports =` was to fix some some failures with `@std/esm` being loaded in tests - it seems to hit a name clash issue, but it's cleaner code after the change.


test/workbox-cli/node/app.js
```diff
@@ -71,6 +71,8 @@ describe(`[workbox-cli] app.js`, function() {
           expect(
             loggerErrorStub.alwaysCalledWithExactly(errors['invalid-common-js-module'])
           ).to.be.true;
+          /* eslint-disable */
+          console.log(error.message);
```
jeffposnick: Did you intend to keep this?

gauntface: No but I'm trying to figure out the AppVeyor issue (I stopped yesterday after realising this was going to be painful)

PR #1319, opened 2018-02-16
Analytics fixes

### LINKED ISSUE #1309: v3: workbox-analytics throws error on sync
**Library Affected**:
*v3 beta workbox-google-analytics*

**Browser & Platform**:
*Google Chrome v64*

Hi again guys!

I'm trying trying to use the v3 beta workbox-google-analytics library for offline analytics. 

While offline, events are being stored in IndexedDB, but when returning online, I get this error:

```
Uncaught (in promise) TypeError: Failed to construct 'Request': The provided value is not of type '(sequence<sequence<ByteString>> or record<ByteString, ByteString>)'
    at n.toRequest (workbox-background-sync.prod.js:1)
    at workbox-background-sync.prod.js:1
    at Generator.next (<anonymous>)
    at n (workbox-core.prod.js:1)
    at workbox-core.prod.js:1
    at <anonymous>
toRequest | @ | workbox-background-sync.prod.js:1
-- | -- | --
  | (anonymous) | @ | workbox-background-sync.prod.js:1
  | n | @ | workbox-core.prod.js:1
  | (anonymous) | @ | workbox-core.prod.js:1
```

And from the network tab it doesn't appear that the stored requests are being sent. 

Note: The IndexedDB entry is removed nonetheless. Also, if I queue up multiple events in IndexedDB, it looks like only one is removed. 

I have a [glitch demo](https://alike-pyjama.glitch.me/) with [code](https://glitch.com/edit/#!/alike-pyjama?path=sw.js:4:57). Any ideas?

philipwalton:

Fixes #1309 

This PR fixes the error in #1309, which was resulting from `workbox-google-analytics` setting the replayed request headers as the string ` '[["Content-Type", "text/plain"]]'` instead of the object.

This likely means none of the replays were working, something that ideally should have been caught by an integration test. I'll put together another PR to make sure an integration test exists that ensures the entire process works. (Note, the node tests didn't catch this because we use header mock that didn't error here when they should have. Not sure if anything should be done about that though.)

In the process of manually testing it, I noticed a few other issues that I fixed as well.

- Failed `workbox-background-sync` requests that were modified were storing the modified versions, I updated it to store the original.
- `workbox-google-analytics` requests that already included a `qt` param were having that param overridden when the correct behavior should be to update it instead.

/cc: @DavidScales

### DIFF
--- a/packages/workbox-background-sync/Queue.mjs
+++ b/packages/workbox-background-sync/Queue.mjs
@@ -124,6 +124,10 @@ class Queue {
 
     let storableRequest;
     while (storableRequest = await this._queueStore.getAndRemoveOldestEntry()) {
+      // Make a copy so the unmodified request can be stored]
+      // in the event of a replay failure.
+      const storableRequestClone = storableRequest.clone();
+
       // Ignore requests older than maxRetentionTime.
       const maxRetentionTimeInMs = this._maxRetentionTime * 60 * 1000;
       if (now - storableRequest.timestamp > maxRetentionTimeInMs) {
@@ -139,7 +143,7 @@ class Queue {
         replay.response = await fetch(replay.request.clone());
       } catch (err) {
         replay.error = err;
-        failedRequests.push(storableRequest);
+        failedRequests.push(storableRequestClone);
       }
 
       replayedRequests.push(replay);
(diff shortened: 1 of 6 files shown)


packages/workbox-background-sync/Queue.mjs
```diff
@@ -124,6 +124,10 @@ class Queue {
 
     let storableRequest;
     while (storableRequest = await this._queueStore.getAndRemoveOldestEntry()) {
+      // Make a copy so the unmodified request can be stored]
```
gauntface: Can you remove the `]` from `stored]`

DavidScales:
@philipwalton @gauntface working for me now 👍 Thanks guys

PR #1321, opened 2018-02-18
Add an integration test for the entire offline analytics replay process

philipwalton:

This PR updates the `workbox-google-analytics` example, so it's easier to test the sending of hits and replaying of failed hits both manually and via webdriver.

The failed hits are simulated by overriding `self.fetch` in the service worker to throw `Response.error()` if the page's `simulate-offline` checkbox is checked. And the `sync` event is faked via `self.dispatchEvent(new SyncEvent(...))`. Unfortunately, this technique doesn't work in FF, so I'm skipping that in non-sync-supporting browsers at the moment. If anyone has any ideas how to fake this outside of sync events, let me know, but I think this is good enough for now (i.e. better than nothing).


gauntface:
@philipwalton does it make sense to make this an extra test for google analytics?

At the moment, it seems better to have a super gentle test for all browsers and then a second more specific test for Chrome.

But I might be missing some context.

philipwalton:
> @philipwalton does it make sense to make this an extra test for google analytics?

I'm not sure what you mean by an "extra test".

> At the moment, it seems better to have a super gentle test for all browsers and then a second more specific test for Chrome.

Is there already a mechanism set up to do this?

In this case since the behavior is subtly different for background-sync and non-background-sync supporting browser, not having a cross-browser test makes me a bit nervous. That being said, I'm not sure the best way to test this behavior.

The non-background-sync behavior runs code on service worker startup, and I don't know of a good way to force that to happen.

gauntface:
@philipwalton the current test that you've edited just makes sure that a google analytics request can be made. It doesn't test the bg sync behaviour, but it at least does a gut check on on bg-sync browsers.

By extra test - I meant - keep the test that is in the v3 branch and then have this as a second test that only runs on bg sync supporting browsers.

philipwalton:
> @philipwalton the current test that you've edited just makes sure that a google analytics request can be made. It doesn't test the bg sync behaviour, but it at least does a gut check on on bg-sync browsers.

It does test the bg-sync behavior. The test I wrote simulates offline by overriding `fetch()` in the service worker to trigger the bg-sync logic. It then dispatches a sync event and asserts `fetch()` was re-called with the modified GA requests.

If any of the bg-sync logic weren't working, this test would fail.

> By extra test - I meant - keep the test that is in the v3 branch and then have this as a second test that only runs on bg sync supporting browsers.

I'm confused by what you mean because that's exactly what I'm doing in this PR. I didn't remove the original test; it's still there. And for my added test I'm skipping it in non-bg-sync-supporting browsers.

PR #1323, opened 2018-02-20
Bg sync logs

### LINKED ISSUE #1226: Background sync is a black box - Add logging
There should be logging to make it clear what is happening so figuring out what events are occuring, when and the result of the events is impossible :(

prateekbh:

Fixes #1226

Added more logs to background-sync


prateekbh:
@gauntface done!

PR #1325, opened 2018-02-20
Adds non-fatal warnings to the build tools

### LINKED ISSUE #1176: Geting error: One of the glob patterns doesn't match any files....
Tried many thing but it still saying "One of the glob patterns doesn't match any files". Please remove or fix the following:

```
{
"globDirectory":"E:\\webpack_practice\\mn6\\dist",
"globPattern": "*.{html, js,css}",
"globIgnores":[
       "node_modules/**/*",
       "workbox-sw.prod.v2.1.2.js",
       "sw.js"
  ]
}
```

**My webpack config file as  #below...**

```js
const path = require("path");
const src_path = path.resolve("src/");
const dist_path = path.resolve("dist/");
const extract_text = require("extract-text-webpack-plugin"); 
const html_plugin = require("html-webpack-plugin");
const commons_chunk = require("webpack/lib/optimize/CommonsChunkPlugin");
const CopyWebpackPlugin = require("copy-webpack-plugin");
const ImageminPlugin = require("imagemin-webpack-plugin").default;
const HardSourceWebpackPlugin = require("hard-source-webpack-plugin");
const Uglify = require("webpack/lib/optimize/UglifyJsPlugin");
const BundleBuddyWebpackPlugin = require("bundle-buddy-webpack-plugin");
const BundleAnalyzerPlugin = require("webpack-bundle-analyzer").BundleAnalyzerPlugin;
const CleanWebpackPlugin = require("clean-webpack-plugin");
const WebpackPwaManifest = require("webpack-pwa-manifest");
const FaviconsWebpackPlugin = require("favicons-webpack-plugin");
const Workbox = require("workbox-webpack-plugin");
const settings={
	plugins:[]
};
const entries={
 	index:path.resolve(src_path,"js/index.js")
}
	settings.entry = entries;
	settings.output = {
		filename:"js/[name].[chunkhash:8].js",
		path:path.resolve(dist_path),
		publicPath:process.env.NODE_ENV == "Client "?"/":""
	};
	settings.devtool= "source-map";
	settings.module={
		rules:[
			{
				test:/.(css|sass)$/,
				use:extract_text.extract({
					fallback:"style-loader",
					use:["css-loader", "postcss-loader", "sass-loader"],
					publicPath:"../"
				})
			},
			{
				test:/.jsx?$/,
				use:"babel-loader",
				exclude:/node-modules/
			},
			{
				test:/.(jpe?g|png|svg|gif)$/,
				use:"url-loader?limit=10000&name=images/[name].[ext]"
			},
			{
				test:/.(ttf|otf|eot|woff|woff2)$/,
				use:"file-loader?name=fonts/[name].[ext]"
			},
			{
				test:/.html$/,
				use:"html-loader?interpolate=require!"
			}
		]
	};
	settings.plugins.push(
		new WebpackPwaManifest({
			name: "My Progressive Web App",
			short_name: "MyPWA",
			description: "My awesome Progressive Web App!",
			background_color: "#cccccc",
			display: "standalone",
			start_url:".",
	    	icons: [
	      		{
	        		src: path.resolve(src_path, "icon.png"),
					sizes: [16, 32, 57, 60,72, 76, 96, 114, 120, 128, 144, 152,180, 192, 256, 384, 512]
	      		}
	    	]
	  		}),
			new FaviconsWebpackPlugin({
				logo: path.resolve(src_path, "icon.png"),
				persistentCache: true,
				inject: true,
				background: "#cccccc",
				title: "My Progressive Web App",
				icons: {
      				android: true,
      				appleIcon: true,
      				appleStartup: true,
      				coast: false,
      				favicons: true,
      				firefox: true,
      				opengraph: false,
      				twitter: true,
      				yandex: false,
      				windows: true
    			}
			}),
 			new CleanWebpackPlugin([dist_path]),
			new extract_text({
 				allChunks: true,
				filename:"css/[name].[contenthash:8].css"
			}),
			function(){
				if(process.env.NODE_ENV == "Production "){
	 				return new Uglify({sourceMap:true})
	 			}else{
	 				return ()=>{};
	 			}
	 		},
	 		function(){
				if(process.env.NODE_ENV == "Production "){
					return new ImageminPlugin({ test: /.(jpe?g|png|gif|svg)$/i })
				}else{
					return ()=>{};
				}
			}(),
			new commons_chunk({async:"commons", children:true, minChunks:4}),
			function(){
			     Object.keys(entries).forEach(function(a){	
			       var flnm = a+".html";
			       var dist_flnm = path.resolve(dist_path, flnm);
			       var src_flnm = path.resolve(src_path, flnm)
				      settings.plugins.push(new html_plugin({
				     	  favicon:false,
 						  hash:true,
					      filename:dist_flnm,
					      template:src_flnm,
					      inject:true,
					      chunks:["commons", a],
					      optimize:{
					       	collapseWhiteSpace:false
					      }
					 }))
			    });
			    return ()=>{};
			}(),
			new HardSourceWebpackPlugin(),
	 		function(){
				if(process.env.NODE_ENV == "Check "){
					return new BundleBuddyWebpackPlugin({sam: true})
				}else{
					return ()=>{};
				}
			}(),
			function(){
				if(process.env.NODE_ENV == "Check "){
					return new BundleAnalyzerPlugin({analyzerPort: 1025})
				}else{
					return ()=>{};
				}
			}(),
			new Workbox({
			    globDirectory: dist_path,
			    globPatterns: ['**/*.{html,js,css}'],
			    swDest: path.join(dist_path, 'sw.js'),
			    clientsClaim: true,
			    skipWaiting: true,
		  	})
	);
	settings.devServer={
		historyApiFallback: true,
		port:9000,
		compress:true,
		open:true,
		https:false,
		open:true
	};
module.exports = settings;
```

I have removed many plugins like hard-source-webpack-plugin, clean-webpack-plugin, webpack-pwa-manifest, favicons-webpack-plugin and tried but issue remain same
Please help me out

### LINKED ISSUE #1315: Warn/inform the user when files will be dropped due to `maximumFileSizeToCacheInBytes`
This is a UX related issue. 

During development, it's often the case that files are not fully minified in order to speed up the workflow.

For example, running the `r.js` build optimizer tool on a large ecommerce Magento store can take **minutes** if `uglifyjs` is enabled.

To combat this, during development we will only *combine* assets into a single or multiple files, but will wait until the production deploy to apply the final step of optimization (which reduces the file size dramatically)

And so onto my issue, because of the following code, `workbox-build` will silently drop any files that exceed the `maximumFileSizeToCacheInBytes` - this is an extreme, unexpected occurrence.

https://github.com/GoogleChrome/workbox/blob/0d5e650147e916f8e21b43ed2e888b41391fe62a/packages/workbox-build/src/lib/filter-files.js#L86-L88

Whilst the user can configure this number, I believe we should incorporate some type of logging/warning system to indicate when & why this is happening.

**Library Affected**:
*workbox-build*

**Browser & Platform**:
Not relevant

Warn/inform the user when files will be dropped due to `maximumFileSizeToCacheInBytes`

jeffposnick:
R: @addyosmani @gauntface

Fixes #1315, fixes #1176 

This adds in support for non-fatal warnings in `workbox-build`, as well as reporting those errors back to the developers via `workbox-cli` an `workbox-webpack-plugin`.

Two classes of problems are treated as warnings now:
- A file that exceeded `maximumFileSizeToCacheInBytes`
- Something going wrong when processing a single glob pattern, like that pattern not matching anything.


packages/workbox-build/src/lib/filter-files.js
```diff
@@ -33,33 +33,38 @@ const noRevisionForUrlsMatchingTransform =
...
  *   });
+ *   return {manifest, warnings: []};
  * };
  *
  * @callback ManifestTransform
  * @param {Array<ManifestEntry>} manifestEntries The full array of entries,
  * prior to the current transformation.
- * @return {Array<ManifestEntry>} The array of entries with the transformation
- * applied.
+ * @return {{manifest: Array<ManifestEntry>, warnings: Array<String>}}
```
gauntface: Make the warnings optional to return.

packages/workbox-build/src/lib/modify-url-prefix-transform.js
```diff
@@ -60,15 +60,19 @@ module.exports = (modifyUrlPrefix) => {
...
+      });
 
-    entry.url = entry.url.replace(modifyRegex, (match) => {
-      return modifyUrlPrefix[match];
+      return entry;
     });
 
-    return entry;
-  });
+    return {manifest, warnings: []};
```
gauntface: Exclude the warnings here.

packages/workbox-build/src/lib/no-revision-for-urls-matching-transform.js
```diff
@@ -21,14 +21,19 @@ module.exports = (regexp) => {
...
+      }
+
+      if (entry.url.match(regexp)) {
+        delete entry.revision;
+      }
+
+      return entry;
+    });
+
+    return {manifest, warnings: []};
```
gauntface: Exclude the warnings here.

jeffposnick:
I'm happy with returning the `warnings` in the response for the `gulp` use case rather than trying to log anything meaningful ourselves.

E.g., we handle errors from modules in our own first-party `gulp` tasks by wrapping them in our own logging helper anyway. That's not uncommon, and I find it distracting when there are two copies of the same info logged.

shakyShane:
@jeffposnick thanks so much!

PR #1327, opened 2018-02-21
Add a new workboxSWImport option to the template.

### LINKED ISSUE #1324: importScripts using Workbox + `workbox-build` can lead to broken SW
@beatrizdemiguelperez pointed this out on the Slack channel.

If you add an `importScript` that uses the `workbox` API and `workbox-build` calls `workbox.setConfig()`, the service worker throws with `workbox-sw.js:1 Uncaught Error: Config must be set before accessing workbox.* modules ` to account for the fact that this is probably not intended.

@jeffposnick the easiest solution is to call setConfig *before* importing additional scripts.

jeffposnick:
R: @addyosmani @gauntface
CC: @beatrizdemiguelperez 

Fixes #1324 

Should hopefully be a clean fix to the issue.

### DIFF
--- a/packages/workbox-webpack-plugin/src/generate-sw.js
+++ b/packages/workbox-webpack-plugin/src/generate-sw.js
@@ -77,7 +77,10 @@ class GenerateSW {
       (compilation.options.output.publicPath || '') + manifestFilename);
 
     // workboxSWImports might be null if importWorkboxFrom is 'disabled'.
+    let workboxSWImport;
     if (workboxSWImports) {
+      // Get the Workbox SW import from the first element in the array.
+      workboxSWImport = workboxSWImports.shift();
       importScriptsArray.push(...workboxSWImports);
     }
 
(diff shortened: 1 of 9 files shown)


packages/workbox-webpack-plugin/src/generate-sw.js
```diff
@@ -77,7 +77,10 @@ class GenerateSW {
       (compilation.options.output.publicPath || '') + manifestFilename);
 
     // workboxSWImports might be null if importWorkboxFrom is 'disabled'.
+    let workboxSWImport;
     if (workboxSWImports) {
+      // Get the Workbox SW import from the first element in the array.
+      workboxSWImport = workboxSWImports.shift();
```
gauntface: I'm not sure why you are shifting here. To me it feels safer to have all `workboxSWImports` together, then the possible config and then the developers import scripts.

jeffposnick:
The only weird case is when someone uses `importWorkboxFrom: <chunkname>` and that chunk contains multiple JavaScript assets. I've modified the PR to group everything together at the front of the `importScripts` for that specific use case. (We don't end up calling `workbox.setConfig()` in that scenario.)

PR #1335, opened 2018-02-28
Fix for runtimeCaching plugin config conversion

### LINKED ISSUE #1334: workbox-build: broadcastUpdate channelName
Again `broadcastUpdate` with workbox-build@3.0.0-beta.1: 

workbox-build wants this:
```js
"options": {
  "broadcastUpdate": {
    "channelName": "my-update-channel"
  }
}
```
(resulting in `new workbox.broadcastUpdate.Plugin({"channelName":"my-update-channel"})`)

but it should be:
```js
"options": {
  "broadcastUpdate": "my-update-channel"
}
```
(for `new workbox.broadcastUpdate.Plugin("my-update-channel")`, which actually works)

Without channelName, it doesn't get accepted: 

> Service worker generation failed: ValidationError: child "runtimeCaching" fails because ["runtimeCaching" at position 1 fails because [child "options" fails because [child "broadcastUpdate" fails because ["broadcastUpdate" must be an object]]]]

There is no `{channelName: ...}` in the test either:
https://github.com/GoogleChrome/workbox/blob/86c7ef7a8b6fe768c7fc52ba9e18aed8dee338b1/test/workbox-broadcast-cache-update/static/sw.js#L12-L14

jeffposnick:
R: @addyosmani @gauntface

Fixes #1334

With additional tests as well.

### DIFF
--- a/packages/workbox-build/src/lib/runtime-caching-converter.js
+++ b/packages/workbox-build/src/lib/runtime-caching-converter.js
@@ -60,7 +60,40 @@ function getOptionsString(options = {}) {
       throw new Error(`${errors['bad-runtime-caching-config']} ${pluginName}`);
     }
 
-    plugins.push(`new ${pluginString}(${JSON.stringify(pluginConfig)})`);
+    let pluginCode;
+    switch (pluginName) {
+      // Special case logic for plugins that have a required parameter, and then
+      // an additional optional config parameter.
+      case 'backgroundSync': {
+        const name = pluginConfig.name;
+        pluginCode = `new ${pluginString}(${JSON.stringify(name)}`;
+        if ('options' in pluginConfig) {
+          pluginCode += `, ${JSON.stringify(pluginConfig.options)}`;
+        }
+        pluginCode += `)`;
+
+        break;
+      }
+
+      case 'broadcastUpdate': {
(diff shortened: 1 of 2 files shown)


packages/workbox-build/src/lib/runtime-caching-converter.js
```diff
@@ -60,7 +60,40 @@ function getOptionsString(options = {}) {
       throw new Error(`${errors['bad-runtime-caching-config']} ${pluginName}`);
     }
 
-    plugins.push(`new ${pluginString}(${JSON.stringify(pluginConfig)})`);
+    let pluginCode;
+    switch (pluginName) {
+      // Special case logic for plugins that have a required parameter, and then
+      // an additional optional config parameter.
+      case 'backgroundSync': {
```
jeffposnick: I was thinking of combining this and the next `case`, which would prevent some code duplication. One reads the required value from the `name` parameter, and the other reads it from `channelName`, though, so there would be some extra logic that would need to be added.

I ended up keeping them separate, but if folks would prefer them combined, I could do that.

gauntface: I think seperate makes sense - makes the tests super easy to read.

Issue #1131, opened 2017-12-14
Docs on how to set credentials


gauntface:
This Q: https://stackoverflow.com/questions/47808749/how-to-add-fetch-options-to-workbox

How do we do this for v3?

See: https://github.com/GoogleChrome/workbox/issues/179#issuecomment-351639894 for info.

jeffposnick:
I just took a look at the codebase, and it's not currently possible in v3 for runtime caching strategies.

PR #1340, opened 2018-03-02
Adds support for changing the options used when calling fetch()

jeffposnick:
R: @addyosmani @gauntface

This is related to #1131.

I don't know if we want to mark it as fixed once this is merged—that issue tracks docs, and while this PR adds the code that's required, along with JSDocs, I guess we could still put together a standalone guide on d.g.c.


packages/workbox-strategies/NetworkFirst.mjs
```diff
@@ -181,6 +186,7 @@ class NetworkFirst {
     try {
       response = await fetchWrapper.fetch(
         event.request,
+        this._fetchOptions,
```
jeffposnick: There's a bug in the current codebase—the plugins array was being passed as the second parameter in this invocation, instead of the third parameter.

(The joy of positional arguments...)

PR #1346, opened 2018-03-05
Using cache name as idb name for precaching

### LINKED ISSUE #1344: Hosting multiple service workers on the same origin deletes the cache from other subdirectories
**Library Affected**:
workbox-precaching

**Browser & Platform**:
all browsers

I have multiple web apps hosted on the same domain that each have a service worker set up to allow the use of the app offline. For example, https://benlubar.github.io/cmv2mp4/ and https://benlubar.github.io/df-ai/plan-editor/.

Setting a different cache name prefix on each only affects the name of the cache, which already contains the URL of the directory where the service worker is hosted. The IndexedDB is always named `workbox-precaching` and contains no information about which subdirectory the service worker is hosted in.

gauntface:
Fixes #1344

### DIFF
--- a/packages/workbox-precaching/models/PrecachedDetailsModel.mjs
+++ b/packages/workbox-precaching/models/PrecachedDetailsModel.mjs
@@ -31,10 +31,13 @@ class PrecachedDetailsModel {
   /**
    * Construct a new model for a specific cache.
    *
+   * @param {string} dbName
    * @private
    */
-  constructor() {
-    this._db = new DBWrapper(`workbox-precaching`, 2, {
+  constructor(dbName) {
+    // This ensures the db name contains only letters, numbers, '-', '_' and '$'
+    const filteredDBName = dbName.replace(/[^\w-]/g, '_');
+    this._db = new DBWrapper(filteredDBName, 2, {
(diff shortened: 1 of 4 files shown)


packages/workbox-precaching/models/PrecachedDetailsModel.mjs
```diff
@@ -31,10 +31,13 @@ class PrecachedDetailsModel {
   /**
    * Construct a new model for a specific cache.
    *
+   * @param {string} dbName
    * @private
    */
-  constructor() {
-    this._db = new DBWrapper(`workbox-precaching`, 2, {
+  constructor(dbName) {
```
jeffposnick: Just wanted to double-check: it looks like the existing codebase [is already passing in](https://github.com/GoogleChrome/workbox/blob/f5792fd9c130fbd7916c7b3763752cf795d9b8a5/packages/workbox-precaching/controllers/PrecacheController.mjs#L44) a string parameter when calling this constructor, even though it was previously being ignored. So now the change is that you're using that string value inside of this constructor.

gauntface: Exactly.

Issue #1368, opened 2018-03-15
Occasional responseToCache undefined


djeeg:
**Library Affected**:

https://github.com/GoogleChrome/workbox/blob/master/packages/workbox-core/_private/cacheWrapper.mjs#L174
Getting TypeError: Cannot read property 'ok' of undefined

**Browser & Platform**:
    "workbox-sw": "^3.0.0",
    "workbox-webpack-plugin": "^3.0.0"
Chrome 64

**Issue or Feature Request Description**:

How do i cause this .. possibly by deleting application in chrome/refreshing service worker a couple of times in a row. Sometimes deleting the application and a new tab helps. Sometimes waiting a bit clears it.

Maybe just another  `if (!responseToCache)` check is needed (like a couple of lines above it)

webmaxru:
Same for me.
Website: https://workbox-workshop.firebaseapp.com/
SW: https://workbox-workshop.firebaseapp.com/sw.js

gauntface:
Eek - annoyingly sporadic.

Thanks for the demo @webmaxru. I can raise a PR that checks for a response, but would have really liked a clearer explanation of why the response is undefined while it's being put in the cache and the PR could bury it..

gauntface:
@webmaxru could you try your demo without skipWaiting and clientsClaim and see if you can reproduce the error?

@djeeg were you using skipWaiting and clientsClaim in your service worker?

djeeg:
No I'm not using either.

> is actually an issue with precaching going from temp -> final cache

This is about the only time I the -temp cache under Cache Storage in Chrome

![image](https://user-images.githubusercontent.com/4269005/37498153-8fc79aba-28b4-11e8-9138-e601c73eda43.png)

gauntface:
@djeeg is there anything in the caches for v1 or v1-temp or are they empty?

djeeg:
I have only ever seen temp empty

![image](https://user-images.githubusercontent.com/4269005/37536733-59081678-2943-11e8-874b-0173c4e41d9d.png)

And precache only has 1 or a couple of files in it

![image](https://user-images.githubusercontent.com/4269005/37536772-7815b2be-2943-11e8-8770-f359b3873d44.png)

gauntface:
Done plenty of digging into this with @jeffposnick and we have a theory (albeit difficult to prove).

At the moment the process is that during install we add files to a temporary cache and during activation these requests and responses are copied over to the final cache.

[...]

PR #1369, opened 2018-03-16
Cache put error

gauntface:
Somewhat addresses: #1368

This will throw with a meaningful error if we try to cache a non-existant response, BUT I think the error in #1368 is actually an issue with precaching going from temp -> final cache but I'm not 100%.

### DIFF
--- a/packages/workbox-core/_private/cacheWrapper.mjs
+++ b/packages/workbox-core/_private/cacheWrapper.mjs
@@ -36,6 +36,17 @@ import '../_version.mjs';
  * @memberof module:workbox-core
  */
 const putWrapper = async (cacheName, request, response, plugins = []) => {
+  if (!response) {
+    if (process.env.NODE_ENV !== 'production') {
+      logger.error(`Cannot cache non-existant response for ` +
+        `'${getFriendlyURL(request.url)}'.`);
+    }
+
+    throw new WorkboxError('cache-put-with-no-response', {
+      url: getFriendlyURL(request.url),
+    });
+  }
+
   let responseToCache = await _isResponseSafeToCache(
     request, response, plugins);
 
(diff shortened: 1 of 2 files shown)


packages/workbox-core/_private/cacheWrapper.mjs
```diff
@@ -36,6 +36,16 @@ import '../_version.mjs';
  * @memberof module:workbox-core
  */
 const putWrapper = async (cacheName, request, response, plugins = []) => {
+  if (!response) {
+    if (process.env.NODE_ENV !== 'production') {
+      logger.error(`Request '${getFriendlyURL(request.url)}' had no URL.`,
+        responseToCache);
```
jeffposnick: `responseToCache` is declared later on in the function, so don't use it here.

I think you mean to log `response`, instead. Are you doing that because you want to distinguish between `undefined` and `null`? Do you think that would be a meaningful distinction when debugging?

packages/workbox-core/_private/cacheWrapper.mjs
```diff
@@ -36,6 +36,16 @@ import '../_version.mjs';
  * @memberof module:workbox-core
  */
 const putWrapper = async (cacheName, request, response, plugins = []) => {
+  if (!response) {
+    if (process.env.NODE_ENV !== 'production') {
+      logger.error(`Request '${getFriendlyURL(request.url)}' had no URL.`,
```
jeffposnick: `had no URL` doesn't seem like the right message to log.

How about `"Can't add a non-existent response to the cache for ${getFriendlyURL(request.url)}."`, or something like that?

gauntface:
Apologies for such a terrible PR. Think I was going back and forth on debugging, thinking of cause while deciding how to provide a little more information.

gauntface:
@jeffposnick another look would be appreciated.

PR #1381, opened 2018-03-20
Set webpack's compilation.warnings when glob-related options are used

### LINKED ISSUE #1380: Use webpack's performance.hints for reporting non-fatal errors
**Library Affected**:
*workbox-webpack-plugin*

`workbox-webpack-plugin` v3 introduced a model in which all assets in the webpack compilation are potentially eligible for precaching (with the `include/exclude/chunks/excludeChunks` options available for modifying the defaults).

Because of this new model, it's (usually) not necessary to use the previous `glob`-based configuration options, which matched files in the output filesystem, rather than matching against assets in the webpack compilation. But there are still some scenarios in which they might be necessary, so we didn't remove support, and documented them at https://developers.google.com/web/tools/workbox/modules/workbox-webpack-plugin#cache_additional_non-webpack_assets

If you do use those `glob`-based options in v3, you end up generating an additional list of precached URLs, and options like `manifestTransforms` and `dontCacheBustUrlsMatching` only apply to that additional list, not the list derived from the webpack asset pipeline.

We should explicitly add one of two possible messages to `compilation.warnings` to let folks know about this scenarios. They are not *always* due to mistakes, so there might be some false positives, but if that's the case, it's easy enough to ignore the warning.

- When `globPatterns` or any of the related options is used, warn that it's often not needed, and link to the relevant docs.
- When `manifestTransforms` or any of the related options that modify the `glob`-derived manifest is used, **and** `globPatterns` is **not** used, warn that those options will have no effect (because there is no `glob`-derived manifest) and link to the relevant docs.

jeffposnick:
R: @addyosmani @gauntface

Fixes #1380

Happy to iterate on the warning language.

This PR also includes updates to the existing test suite to ensure that warnings were not previously being set when they were not expected.

### DIFF
--- a/packages/workbox-webpack-plugin/src/lib/warn-about-config.js
+++ b/packages/workbox-webpack-plugin/src/lib/warn-about-config.js
@@ -1,0 +1,54 @@
+/*
+  Copyright 2017 Google Inc.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+      https://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+
+module.exports = (config) => {
+  const moreInfoUrl = 'https://goo.gl/EQ4Rhm';
+
+  const globOptions = [
+    'globDirectory',
+    'globFollow',
+    'globIgnores',
+    'globPatterns',
+    'globStrict',
+  ];
+
+  const usedGlobOptions = globOptions.filter((option) => option in config);
+  if (usedGlobOptions.length > 0) {
+    return `You're using the following Workbox configuration options: ` +
+      `[${usedGlobOptions.join(', ')}] In Workbox v3 and later, these ` +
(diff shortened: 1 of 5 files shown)


packages/workbox-webpack-plugin/src/lib/warn-about-config.js
```diff
@@ -0,0 +1,54 @@
...
+    'globDirectory',
+    'globFollow',
+    'globIgnores',
+    'globPatterns',
+    'globStrict',
+  ];
+
+  const usedGlobOptions = globOptions.filter((option) => option in config);
+  if (usedGlobOptions.length > 0) {
+    return `You're using the following Workbox configuration options: ` +
```
gauntface: Maybe `You're using **one** of the following workbox configuration`

packages/workbox-webpack-plugin/src/lib/warn-about-config.js
```diff
@@ -0,0 +1,54 @@
...
+      `[${usedGlobOptions.join(', ')}] In Workbox v3 and later, these ` +
```
gauntface: Add a fullstip after `]`?

packages/workbox-webpack-plugin/src/lib/warn-about-config.js
jeffposnick: The error message only includes the options that are actually being used, so I think it's accurate. The only nit is whether `options` should be singular when there's one option... so I can change that.

packages/workbox-webpack-plugin/src/lib/warn-about-config.js
jeffposnick: Done.

PR #1388, opened 2018-03-23
Use the correct IDB name in Safari's precaching integration test

### LINKED ISSUE #1383: workbox-precaching integration test times out in Safari TP 52
**Library Affected**:
*workbox-precaching*

**Browser & Platform**:
*[Safari TP 52](https://webkit.org/blog/8165/release-notes-for-safari-technology-preview-52/) on OS X*

https://github.com/GoogleChrome/workbox/blob/master/test/workbox-precaching/integration/precache-and-update.js times out when I run

```sh
gulp test-integration --package='workbox-precaching'
```

(This is not necessarily an issue that was introduced in TP 52, as I don't think I ran the full integration test suite against TP 51.)

There's nothing logged in the JS Console prior to the test timing out:
untitled

jeffposnick:

Fixes #1383, by accounting for the change introduced in https://github.com/GoogleChrome/workbox/pull/1346

### DIFF
--- a/test/workbox-precaching/static/precache-and-update/messenger.js
+++ b/test/workbox-precaching/static/precache-and-update/messenger.js
@@ -1,28 +1,37 @@
+// This needs to be kept in sync with the naming convention used by workbox-precaching.
+const DB_NAME = 'workbox-precache-http___localhost_3004_test_workbox-precaching_static_precache-and-update_';
+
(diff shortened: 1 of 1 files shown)


gauntface:
Thanks for this - it might be worth exploring using saucelabs to test safari tech preview.

PR #1403, opened 2018-03-28
Make webpack path and filename options more customizable

### LINKED ISSUE #1285: InjectManifest Plugin – Set manifest dest
**Library Affected**:
*workbox-webpack-plugin* (v3.0.0-alpha-5)

*Add option to set the `manifestDest` when using InjectManifest Plugin with Webpack*

Since it is important to put the service worker file at the root of the files you want to cache (for scoping reasons), it is possible that the `precache-manifest.js` file could be in a different directory. It seems that the generated import call forces you to have both the service worker and precache manifest in the same directory.

Although this will usually be the case, it would be convenient to have the ability to choose.

jeffposnick:
R: @addyosmani @gauntface
CC: @raejin

Fixes #1285

This adds the following to both the `InjectManifest` and `GenerateSW` webpack plugins:

- An optional `precacheManifestFilename` parameter, defaulting to `'precache-manifest.[manifestHash].js'` (i.e. the current behavior). It can be used to customize the precache manifest name. If the string `'[manifestHash]'` isn't included in your custom name, that's treated as a webpack compilation error.

- An optional `importsDirectory` parameter, defaulting to `''` (i.e. the current behavior). It is treated as a subdirectory of `output.path`. When set, the precache manifest is written to this directory, and if `importWorkboxFrom: 'local'`, the local Workbox libraries are written there as well.

- Automatic adjustments to the URLs used in `importScripts()` to account for both of these potential customizations.

- When `importWorkboxFrom: 'local'` and `output.publicPath` is configured, the `output.publicPath` value will be prepended to the `modulePathPrefix` used to determine where the Workbox libraries are dynamically loaded from. This amounts to a change in the previous behavior, but based on developer expectation, the previous behavior was considered a bug.

There needs to be a corresponding update to the WebFundamentals docs once this is merged.


packages/workbox-webpack-plugin/src/generate-sw.js
```diff
@@ -45,17 +48,7 @@ class GenerateSW {
...
-        // Exclude source maps.
-        /\.map$/,
-        // Exclude anything starting with manifest and ending .js or .json.
-        /^manifest.*\.js(?:on)?$/,
-      ],
-      excludeChunks: [],
-      importScripts: [],
-      importWorkboxFrom: 'cdn',
+    this.config = Object.assign(getDefaultConfig(), {
       swDest: 'service-worker.js',
```
gauntface: Why is `swDest` seperate from the default config?

PR #1408, opened 2018-03-29
Adds makeRequest() method to workbox.strategies

### LINKED ISSUE #1395: Add makeRequest() method to workbox.strategies.* classes
**Library Affected**:
*workbox-strategies*

Extracted from [this larger proposal](https://docs.google.com/document/d/1oZOIsJjZ31DjQksU8Zlq3Tql6f13c-5aWHKFxpZynqo/edit) around the new `workbox.streams.*` modules. There seemed to be general support from @gauntface and @philipwalton.

## Problem
It’s currently difficult to take advantage of the built-in workbox.strategies in a standalone fashion, outside of `workbox.routing.Router`. Methods like `workbox.strategies.staleWhileRevalidate.handle({event})` can’t be used directly, because they expect event to be a real `FetchEvent` instance that contains all of the context needed to implement the strategy.

## Proposal
- Don't modify any existing methods in the `workbox.strategies.*` classes.
- Add a `makeRequest({event, request})` method to each strategy class to make it easier to call each strategy independently of `workbox.routing.Router`.
- `event` is optional, and if set, `event.waitUntil(...)` will be automatically called, waiting on the appropriate promise.
- `request` does not need to correspond to `event.request` (and likely won't, since if it did, you could just call `handle({event})` instead). You can pass in an arbitrary `Request` object, or URL string, for `request`.

This leads to potential usage like:

```js
const httpBinStrategy = workbox.strategies.staleWhileRevalidate({
  cacheName: 'httpbin',
});

workbox.streams.strategy([
  // ... other stream sources ...  
  ({event}) => httpBinStrategy.makeRequest({
    event,
    request: 'https://httpbin.org/bytes/12',
  }),
  // ... other stream sources ...  
]);
```

jeffposnick:
R: @gauntface @philipwalton 

Fixes #1395 

This allows any of the `workbox.strategies.*` classes to be called manually, outside of a `Route` context.

The JSDocs in this PR refer to https://developers.google.com/web/tools/workbox/guides/advanced-recipes#make-requests, which I'll write following this PR.

### DIFF
--- a/packages/workbox-strategies/CacheFirst.mjs
+++ b/packages/workbox-strategies/CacheFirst.mjs
@@ -164,7 +164,7 @@ class CacheFirst {
    * Handles the network and cache part of CacheFirst.
    *
    * @param {Request} request
-   * @param {FetchEvent} event
+   * @param {FetchEvent} [event]
    * @return {Promise<Response>}
    *
    * @private
--- a/packages/workbox-strategies/CacheFirst.mjs
+++ b/packages/workbox-strategies/CacheFirst.mjs
@@ -1,5 +1,5 @@
 /*
- Copyright 2016 Google Inc. All Rights Reserved.
+ Copyright 2018 Google Inc. All Rights Reserved.
  Licensed under the Apache License, Version 2.0 (the "License");
  you may not use this file except in compliance with the License.
  You may obtain a copy of the License at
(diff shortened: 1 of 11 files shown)


gauntface:
Is there any value in keeping the handle() method with this change?

jeffposnick:
Removing `.handle()` entirely would be a breaking change to the interface. While I doubt there are too many folks calling `.handle()` directly, we'd have to make some changes to the `workbox-routing` to go along with removing `.handle()`, and that might be more likely to break existing assumptions.

So I'd rather punt on that until the next round of discussions about breaking interface changes.

Issue #1414, opened 2018-04-03
cacheFirst appears to fetch when the response is already cached


rps:
I've been examining the Network calls for an app using Workbox Service Workers with Webpack to cache static images. After activating the SW and visiting the page, I can confirm that the cache is created and the assets are in the cache. However, when revisiting the page, the network tab shows two calls for the image: an image returned from the SW (expected) and a fetch event from the disk cache originating from Workbox (and suggesting that the SW is making a request, which it shouldn't be).

- What is the second unexpected fetch event and why is it being used in a cacheFirst strategy? I would expect just to see a single event for the image being returned from the SW.
- If I remove the expiration constraints, this issue is resolved. Why is that?

Sample code below. Thanks!

Webpack config:
```
            new GenerateSW({
                skipWaiting: true,
                clientsClaim: true,
                include: [],
                runtimeCaching: [
                    {
                        urlPattern: /^https:\/\/.*\.cloudfront\.net\/.*/,
                        handler: 'cacheFirst',
                        options: {
                            cacheName: 'cloudfront',
                            cacheableResponse: {
                                statuses: [0, 200],
                            },
                            expiration: {
                                maxEntries: 100,
                                maxAgeSeconds: 600,
                            },
                        },
                    },
                ],
            }),
```

Generated:
```
workbox.routing.registerRoute(/^https:\/\/.*\.cloudfront\.net\/.*/, workbox.strategies.cacheFirst({
    cacheName: "cloudfront",
    plugins: [new workbox.cacheableResponse.Plugin({
        "statuses": [0, 200]
    }), new workbox.expiration.Plugin({
        "maxEntries": 100,
        "maxAgeSeconds": 600
    })]
}), 'GET');
```

jeffposnick:
This should have been fixed by #1422, which was deployed as part of the [`3.1.0` release](https://github.com/GoogleChrome/workbox/releases/tag/v3.1.0).

Please reopen if you're still having trouble after upgrading to `3.1.0`!

Issue #1329, opened 2018-02-22
maxAgeSeconds


JasonMuteham:
The following code updates the cache but always hits the network .  However if maxAgeSeconds is removed it works as expected.

```javascript
workbox.routing.registerRoute(
  new RegExp("https://res.cloudinary.com/(.*)"),
  workbox.strategies.cacheFirst({
      cacheName: 'cloud-images',
      plugins: [
        new workbox.expiration.Plugin({
          maxAgeSeconds: 30 * 24 * 60 * 60,
          maxEntries: 100,
        }),
        new workbox.cacheableResponse.Plugin({
          statuses: [0, 200],
        }),
      ],
  }),
);
```

gauntface:
@JasonMuteham is there a site that is deployed anywhere that demonstrates this?

jorgeuriarte:
I'm experiencing this behaviour too.
As soon as I setup *any* value for `maxAgeSeconds`, the strategy (in my case, StaleWhileRevalidate) will stop finding the response in the cache-

"No response found for...  Will wait for the network response" shows in the log.

If I remove the `maxAgeSeconds`, I will get the "Found a cached response..." log.

Is there anything needed in the server-side of the response, any header in a given format, or something that might explain this behaviour?

jorgeuriarte:
```
  _getDateHeaderTimestamp(cachedResponse) {
    const dateHeader = cachedResponse.headers.get('date');
    const parsedDate = new Date(dateHeader);
    const headerTime = parsedDate.getTime();

    // If the Date header was invalid for some reason, parsedDate.getTime()
    // will return NaN.
    if (isNaN(headerTime)) {
      return null;
    }

    return headerTime;
  }
```

this code will return `0` in my case... :(

jorgeuriarte:
Ok, I think I got it. @JasonMuteham I bet your request was CORS? i.e: res.cloudinary.com is not the domain _serving `index.html`_ ?

In my case, it seems it's all due to this restriction:

https://developers.google.com/web/updates/2015/03/introduction-to-fetch#response_types

In few words, `workbox-cache-expiration` needs the `Date` header to be available, but it is *not* due to CORS restrictions.
When there's no 'date' to access, the code shown before will return `0`, and no caching will be done.

In my case, I've included the 'Date' header in the CORS attributes of the server, and it works fine now.
[...]

PR #1422, opened 2018-04-07
Bugfix - Correctly identify when response does not have `date` header set

### LINKED ISSUE #1416: Caching api request calls two request on cached api path
Currently, I am trying to cache api call with service worker. It does cache the api request. However, it does not seems to take advantage of caching. When I check the network tab on devtool, I see request from service worker and workbox both of them and even slower than normal request.

I tried older version with webpack 3, cached api returns almost instantly like 3ms.

**Dev network tab**
app-1.0.js.gz (from ServiceWorker) 3.5s  see orange bar after 
workbox-core.prod.js 72kb, 3.53 

workbox-webpack-plugin config

```
new workboxPlugin.GenerateSW({
  swDest: 'sw.js',
  clientsClaim: true,
  skipWaiting: true,
  runtimeCaching: [
    {
      urlPattern: new RegExp('/'),
      handler: 'staleWhileRevalidate',
    },
    {
      urlPattern: new RegExp('https://api/'),
      handler: 'cacheFirst',
      options: {
        cacheName: 'api',
        expiration: {
          maxEntries: 100,
          maxAgeSeconds: 72 * 60 * 60
        },
        cacheableResponse: { statuses: [0, 200] },
      }
    },
  ]
})

```
Service worker request create some kind of bottle neck ?

matthewjmay:
Fixes bug where all responses were seen as expired if expiry was set: adds a proper check to see if response `date` header is set.

Fixes #1416 , #1414, #1329 

If the response header did not have `date`, then `cachedResponse.headers.get('date')` would return `null`, as described here: https://developer.mozilla.org/en-US/docs/Web/API/Headers/get. Since `new Date(null).getTime()` returns 0, not `NaN`, all requests were seen as expired if an expiry was specified.

I can open an issue as per the guidelines, but will be duplicates of #1416, and #1414

### DIFF
--- a/infra/testing/sw-env-mocks/Headers.js
+++ b/infra/testing/sw-env-mocks/Headers.js
@@ -26,7 +26,7 @@ class Headers {
   }
 
   get(key) {
-    return this.obj[key];
+    return this.has(key) ? this.obj[key] : null;
   }
 
   set(key, value) {
--- a/packages/workbox-cache-expiration/Plugin.mjs
+++ b/packages/workbox-cache-expiration/Plugin.mjs
@@ -169,6 +169,10 @@ class Plugin {
    * @private
    */
   _getDateHeaderTimestamp(cachedResponse) {
+    if (!cachedResponse.headers.has('date')) {
+      return null;
+    }
+
     const dateHeader = cachedResponse.headers.get('date');
     const parsedDate = new Date(dateHeader);
     const headerTime = parsedDate.getTime();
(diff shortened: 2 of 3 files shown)


jeffposnick:
Many thanks, @matthewjmay! That nuance was lost in translation from the v2 branch:

jeffposnick:
My one request prior to merging would be to get some test coverage for this potentiality, to prevent regressions.

Would you be up to adding in a new test case to https://github.com/GoogleChrome/workbox/blob/master/test/workbox-cache-expiration/node/test-Plugin.mjs to cover the different `Date:` header scenarios? If you'd rather not, we can add in the tests on your behalf, but I figured I'd give you first crack at it.

jeffposnick:
Thanks a ton for this, @matthewjmay!

We've generally used the mocks from https://github.com/pinterest/service-workers/tree/master/packages/service-worker-mock, but haven't been great about upstreaming local changes into that project when we find issues, and instead override them locally (as you're doing here).

gauntface:
Thanks for this @matthewjmay

Issue #1393, opened 2018-03-27
[workbox-webpack-plugin#InjectManifest] Relying on private _readFile is unstable


DorianGrey:
**Library Affected**:
workbox-webpack-plugin

**Browser & Platform**:
all browsers & platforms

**Issue or Feature Request Description**:
This is a follow-up of the an issue originally opened here: https://github.com/angular/angular-cli/issues/10064
Overview:
I've recently tried to update one of my projects to `workbox` v3 (see the update state [here](https://github.com/DorianGrey/ng-webpack-template/tree/webpack-4)). It uses the corresponding `webpack` plugin via `InjectManifest`.
However, the update failed due to a type error:
```
TypeError: readFileFn is not a function
```
https://travis-ci.org/DorianGrey/ng-webpack-template/jobs/357274315

Debugging this a bit lead me to this part of the plugin's code:
https://github.com/GoogleChrome/workbox/blob/fa4d8b7dab2b0680c15d6057c12fd9ec9f0932c4/packages/workbox-webpack-plugin/src/inject-manifest.js#L150-L154
It uses a private (at least: intended to be private) method of `compiler.inputFileSystem`, which is unreliable to be present in every case or circumstance.
Even worse, it might be absent or at least non-accessible in case the `inputFileSystem` got decorated before (see the decoration suggestion https://github.com/webpack/enhanced-resolve/issues/98#issuecomment-317516344). In my case, the `AngularCompilerPlugin` did exactly this, causing `compiler.inputFileSystem._readFile` to be no longer existing at the point the plugin tries to use it.
Since even decorated versions of the `inputFileSystem` should fulfill its public API (though they are not enforced to do so), this plugin's code should rely on this part of the API exclusively.

I've tested this a bit and found two ways that rely on the public API:
1. Using `compiler.inputFileSystem.readFile`
```javascript
compiler.hooks.emit.tapPromise( 
   this.constructor.name, 
   (compilation) => this.handleEmit(compilation, 
     compiler.inputFileSystem.readFile.bind(compiler.inputFileSystem)) 
 );
```
2. Using `compilation.inputFileSystem.readFile`
```javascript
compiler.hooks.emit.tapPromise( 
   this.constructor.name, 
   (compilation) => this.handleEmit(compilation, 
     compilation.inputFileSystem.readFile.bind(compilation.inputFileSystem)) 
 );
```
Suppose both end up using the same `inputFileSystem` instance. Manually binding the methods is required, since they are not bound by default.
[...]

PR #1437, opened 2018-04-17
#1393: Only use public API of 'compiler.inputFileSystem'

DorianGrey:
R: @jeffposnick @philipwalton 

Fixes https://github.com/GoogleChrome/workbox/issues/1393.

The issue linked above provides a more detailed explanation about the source of this problem.
Short version: The `InjectManifest` version of `workbox-webpack-plugin` currently relies on an internal function `_readFile` of the filesystem API. This only works under specific circumstances, and is unreliable in a more general case (e.g. in case of a decorated filesystem version).

Changes included in this PR:
- Use the public `readFile` function of `webpack`'s filesystem API. Requires an explicit binding for `this`.
- Omit the encoding in `readFileFn` of `readFileWrapper` - it cannot be provided explicitly in this case.

### DIFF
--- a/packages/workbox-webpack-plugin/src/lib/read-file-wrapper.js
+++ b/packages/workbox-webpack-plugin/src/lib/read-file-wrapper.js
@@ -18,7 +18,7 @@
  * A wrapper that calls readFileFn and returns a promise for the contents of
  * filePath.
  *
- * readFileFn is expected to be set to compiler.inputFileSystem._readFile, to
+ * readFileFn is expected to be set to compiler.inputFileSystem.readFile, to
  * ensure compatibility with webpack dev server's in-memory filesystem.
  *
  * @param {Function} readFileFn The function to use for readFile.
@@ -28,7 +28,7 @@
  */
 function readFileWrapper(readFileFn, filePath) {
   return new Promise((resolve, reject) => {
-    readFileFn(filePath, 'utf8', (error, data) => {
+    readFileFn(filePath, (error, data) => {
       if (error) {
         return reject(error);
       }
--- a/packages/workbox-webpack-plugin/src/inject-manifest.js
+++ b/packages/workbox-webpack-plugin/src/inject-manifest.js
@@ -145,17 +145,18 @@ ${originalSWString}
    * @private
    */
   apply(compiler) {
+    const readFile = compiler.inputFileSystem.readFile
+      .bind(compiler.inputFileSystem);
     if ('hooks' in compiler) {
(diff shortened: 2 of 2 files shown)


packages/workbox-webpack-plugin/src/lib/read-file-wrapper.js
```diff
@@ -28,7 +28,7 @@
  */
 function readFileWrapper(readFileFn, filePath) {
   return new Promise((resolve, reject) => {
-    readFileFn(filePath, 'utf8', (error, data) => {
```
raejin: Would the `utf8` encoding option cause any problem? This seems like the only different in calling the readFile API.

DorianGrey: Yes - it causes a runtime error, since the method from the public API only expects two parameters, the file path and the callback. See the API of the `CachedInputFileSystem` for an example: https://github.com/webpack/enhanced-resolve/blob/master/lib/CachedInputFileSystem.js#L238

philipwalton:
I'm not as familiar with this issue, but this change looks fine to me. @jeffposnick if you're fine with merging, +1 from me.

packages/workbox-webpack-plugin/src/lib/read-file-wrapper.js
jeffposnick: A few data points:

- Two-parameters seems to match how `webpack` calls that function, as per https://github.com/webpack/webpack/blob/eca7bad8de54c39b9cb8b138793362b8a17ac11b/lib/Compiler.js#L349

- `memory-fs` has support for both two-parameter and three-parameter modes: https://github.com/webpack/memory-fs/blob/7ae73ebf34ae2f95266e6c3fef1f8e99f9d2308a/lib/MemoryFileSystem.js#L313

The one usage of `readFileWrapper` is to read in the `swSrc`, and the results of that are then used as part of a ES2015 template string to construct the final `swDest` file.

If the native `fs.readFile()` is called in two-parameter mode, without specifying the encoding, then it returns the contents as a `Buffer` rather than a `string`. *But*, using the `Buffer` within the final ES2015 template string will automatically call `toString()` on in, and the default encoding for that stringification is `utf-8`. So, we should have the same end result for the `fs.readFile()` case as well.

raejin: 👍 Awesome, thank you for the detailed explanation and references to the source code.

PR #1439, opened 2018-04-19
Adds workbox.streams

### LINKED ISSUE #621: Handler for composite Streams-based responses
**Library Affected**:
*workbox-runtime-caching*

The current state-of-the-art for combining multiple asynchronous data sources into a composite `Response` that could be passed to `FetchEvent#respondWith` is detailed at https://jakearchibald.com/2016/streams-ftw/#crossing-the-streams

It's a bit beyond what most developers would be comfortable with implementing on their own, but we might be able to generalize it and package it as a [`Handler`](https://github.com/GoogleChrome/workbox/blob/master/packages/workbox-runtime-caching/src/lib/handler.js) subclass in `workbox-runtime-caching`.

The actual interface would need to be worked out—I could imagine it being constructed with an array of functions, each function returning a Promise for the partial data stream, along with an object used to populate the headers.

It would be nice to recruit @jakearchibald and/or @domenic for at least a review, if not an actual implementation.

jeffposnick:
R: @addyosmani @gauntface

Fixes #621

I'll add some inline comments with anything that I'd imagine could be contentious/questionable to aid in the review.


infra/testing/activate-sw-safari.js
```diff
@@ -3,15 +3,17 @@ module.exports = async (swUrl) => {
   let error = await global.__workbox.webdriver.executeAsyncScript((swUrl, cb) => {
     const onStateChangePromise = (registration, desiredState) => {
       return new Promise((resolve, reject) => {
-        if (registration.installing === null) {
-          throw new Error('Regstration is not installing');
+        if (desiredState === 'activated' && registration.active) {
```
jeffposnick: I made some tweaks to this logic because I've seen some flakiness with the latest Safari TP, where `registration.installing` was `null` and this function gave up.

I think it's less prone to race conditions with these changes, since if we have a `registration.active` and the state we care about is `'activated'`, that should be fine.

infra/testing/sw-env-mocks/Headers.js
```diff
@@ -18,7 +18,7 @@
 // https://fetch.spec.whatwg.org/#headers-class
 class Headers {
   constructor(obj = {}) {
-    this.obj = obj;
+    this.obj = Object.assign({}, obj);
```
jeffposnick: This works around an issue that I encountered in the tests.

https://github.com/pinterest/service-workers/blob/master/packages/service-worker-mock/models/Headers.js looks like it's much more robust than it used to be, and perhaps we should migrate off of our own mock at some point in the future.

packages/workbox-streams/_default.mjs
```diff
@@ -0,0 +1,84 @@
...
+    );
+
+    const headers = createHeaders(headersInit);
+    // Constructing a new Response from a Blob source is well-supported.
+    // So is constructing a new Blob from multiple source Blobs or strings.
+    return new Response(new Blob(parts), {headers});
+  };
+}
+
+export default {
```
jeffposnick: I'm (still) not 100% sure what belongs in the `_default.mjs`'s named exports and what doesn't.

I *think* this leads to a decent public interface, and I can use `workbox.streams.strategy()`, `workbox.streams.isSupported()`, etc. with this setup. But please double-check my thinking.

packages/workbox-streams/_public.mjs
```diff
@@ -0,0 +1,27 @@
...
+  limitations under the License.
+*/
+
+import {concatenate} from './concatenate.mjs';
+import {concatenateToResponse} from './concatenateToResponse.mjs';
+import {isSupported} from './isSupported.mjs';
+
+import './_version.mjs';
+
+export {
```
jeffposnick: Similar to my previous comment about uncertainty re: what should be in the `_default.mjs` export, I'm not 100% sure here, either.

packages/workbox-streams/concatenate.mjs
```diff
@@ -0,0 +1,147 @@
...
+  });
+
+  let fullyStreamedResolve;
+  let fullyStreamedReject;
+  const done = new Promise((resolve, reject) => {
+    fullyStreamedResolve = resolve;
+    fullyStreamedReject = reject;
+  });
+
+  if (process.env.NODE_ENV !== 'production') {
```
jeffposnick: This series of log messages looks like the following in `.dev.`:

screen shot 2018-04-19 at 4 03 10 pm

test/workbox-streams/integration/test.js
```diff
@@ -0,0 +1,41 @@
+const {expect} = require('chai');
```
jeffposnick: Creating meaningful `node`-based mocks for all the primitives needed by `workbox.streams` wasn't feasible, so I'm relying on the integration tests to exercise most of the actual logic.

demos/functions/views/demo/workbox-streams/sw.hbs
```diff
@@ -0,0 +1,44 @@
...
+  cacheName: 'httpbin',
+});
+
+// String together an artificially complex series of stream sources.
+const streamsStrategy = workbox.streams.strategy([
+  () => caches.match(START_CACHE_KEY, {cacheName: CACHE_NAME}),
+  () => `<p>this was executed at ${Date.now()}</p>`,
+  ({event}) => httpBinStrategy.makeRequest({
+    event,
+    request: 'https://httpbin.org/bytes/12',
```
gauntface: don't use a third party.

demos/functions/views/demo/workbox-streams/sw.hbs
```diff
@@ -0,0 +1,44 @@
...
+
+const CACHE_NAME = 'my-cache';
+const START_CACHE_KEY = 'start';
+const END_CACHE_KEY = 'end';
+
+self.addEventListener('install', (event) => {
+  event.waitUntil((async () => {
+    const cache = await caches.open(CACHE_NAME);
+    await Promise.all([
+      cache.put(START_CACHE_KEY, new Response('<html><head></head><body>')),
```
gauntface: Would be preferable to make this a little more "fun".

Perhaps you could have an iframe that displays a HTML page indicating its not streamed and with SW installed it loads a HTML page with a header and footer with a party popper in the middle. This highlights a clear use case of header and footer and party popper highlights its working.

packages/workbox-streams/_default.mjs
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * @param {
+ *   Array<function(workbox.routing.Route~handlerCallback)>} sourceFunctions
+ * Each function should return a {@link workbox.streams.StreamSource} (or a
+ * Promise which resolves to one).
+ * @param {HeadersInit} [headersInit] If there's no `Content-Type` specified,
+ * `'text/html'` will be used by default.
+ * @return {workbox.routing.Route~handlerCallback}
+ *
+ * @memberof workbox.streams
```
gauntface: Does this work if you run gulp docs?

I think you need to define the @module workbox.streams before you assign memberof's

packages/workbox-streams/_default.mjs
```diff
@@ -0,0 +1,85 @@
...
+    // So is constructing a new Blob from multiple source Blobs or strings.
+    return new Response(new Blob(parts), {headers});
+  };
+}
+
+export default {
+  concatenate,
+  concatenateToResponse,
+  isSupported,
+  strategy,
```
gauntface: I'm not sure `strategy` is the best name here. `workbox.streams.join()` `workbox.streams.run()` `workbox.streams.merge()` would explain the intent a little easier.

packages/workbox-streams/concatenate.mjs
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  if (source.getReader) {
+    return source.getReader();
+  }
+
+  // TODO: This should be possible to do by constructing a ReadableStream, but
+  // I can't get it to work. As a hack, construct a new Response, and use the
+  // reader associated with its body.
+  return new Response(source).body.getReader();
```
gauntface: cc @jaffathecake in case he has any ideas

packages/workbox-streams/concatenate.mjs
gauntface: Could you change this to a single group:

> Concatenating 5 sources.

And instead of `...reading...`, `Reached the end of that source` under each group, could you just add success or failure along with the final response. 

While I understand the desire to log the requests as the state changes, it doesn't work with console groups as everything gets out of sync and ends up accidentally nesting and mixing - instead you need to capture the logs and then log everything at once.

packages/workbox-streams/concatenate.mjs
```diff
@@ -0,0 +1,147 @@
...
+              logger.groupCollapsed('Current source is', sourcePromises[i]);
+            }
+          }
+          return reader.read();
+        }).then((result) => {
+          if (result.done) {
+            logStartOfGroup = true;
+            if (process.env.NODE_ENV !== 'production') {
+              logger.log('Reached the end of that source.');
+              logger.groupEnd();
```
gauntface: Dynamic log groups like this will cause issues - we had with some of the previous strategies. Instead you need to stash the logs you want to make and then log everything together.

jeffposnick: Thanks for the heads-up about nesting and groups. I've refactored that.

I don't have access to the new `Response` at this point in the code, only to the `ReadableStream`. That doesn't end up exposing any useful information when it's logged to the console.

I'm going to take advantage of the fact that the `Response` will normally be logged by folks who use our `Router` class, and consider that sufficient.

demos/functions/views/demo/workbox-streams/sw.hbs
jeffposnick: Switched.

packages/workbox-streams/concatenate.mjs
jeffposnick: It now looks like:
screen shot 2018-04-23 at 2 07 06 pm

demos/functions/views/demo/workbox-streams/sw.hbs
jeffposnick: Updated.

packages/workbox-streams/_default.mjs
gauntface: I'm not sure this having the same API means it should have the same name unless it is behaving the same.

packages/workbox-streams/concatenate.mjs
```diff
@@ -0,0 +1,140 @@
...
+    cancel() {
+      if (process.env.NODE_ENV !== 'production') {
+        logger.warn('The ReadableStream was cancelled.');
+      }
+
+      fullyStreamedResolve();
+    },
+  });
+
+  return {done, stream};
```
philipwalton: I remember we had a discussion before about whether `done` was really needed for `waitUntil` or whether `respondWith` was sufficient to keep the SW alive. Did we ever confirm that one way or another?

jeffposnick: I confirmed with Jake that while the relevant specifications are pointing towards this no longer being required in the future, the current implementation in Chrome (and likely other browsers) does require explicitly calling `waitUntil()` to guarantee the SW is kept alive until the stream is drained.

philipwalton:
LGTM

PR #1440, opened 2018-04-20
More context in the welcome log message

jeffposnick:
R: @addyosmani @gauntface

@developit was asking me about when we load `.dev.` vs. `.prod.` bundles, and I realized that we could provide some extra context in our Welcome message to help explain what's going on.

Happy to tweak the actual language if it's too wordy.

### DIFF
--- a/packages/workbox-core/_default.mjs
+++ b/packages/workbox-core/_default.mjs
@@ -51,7 +51,10 @@ class WorkboxCore {
       logger.groupCollapsed('Welcome to Workbox!');
       if (self.location.hostname === 'localhost') {
         logger.unprefixed.log(`The development build with extra logging is ` +
-          `automatically enabled when accessing your web app via localhost.`);
+          `automatically enabled when accessing your web app via localhost.` +
+          `More info is available at\n` +
+          `${padding}https://developers.google.com/web/tools/workbox/` +
+          `guides/troubleshoot-and-debug`);
       }
       logger.unprefixed.log(
         `📖 Read the guides and documentation\n` +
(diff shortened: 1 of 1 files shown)


packages/workbox-core/_default.mjs
```diff
@@ -49,12 +49,16 @@ class WorkboxCore {
     if (process.env.NODE_ENV !== 'production') {
       const padding = '   ';
       logger.groupCollapsed('Welcome to Workbox!');
+      if (self.location.hostname === 'localhost') {
+        logger.unprefixed.log(`The development build with extra logging is ` +
+          `automatically enabled when accessing your web app via localhost.`);
+      }
```
gauntface: I'm fine with the above but it does only cater to one situation where we set dev build on localhost.

We could change this up a little to be more of a statement

> You are using the development build of Workbox. Learn more how to this works here

Where here is either a link to a docs page or the message is a group and we add info on:
- The localhost vs non localhost builds
> The setConfig options to force a build type
> Setting the log level for additional logs

jeffposnick: https://developers.google.com/web/tools/workbox/guides/troubleshoot-and-debug would probably be that doc—I can link to it in this log message, and then we can follow-up on modifying the information there in a PR against WebFundamentals at some point.

packages/workbox-core/_default.mjs
```diff
@@ -49,12 +49,19 @@ class WorkboxCore {
     if (process.env.NODE_ENV !== 'production') {
       const padding = '   ';
       logger.groupCollapsed('Welcome to Workbox!');
+      if (self.location.hostname === 'localhost') {
+        logger.unprefixed.log(`The development build with extra logging is ` +
```
gauntface: This can also happen on localhost because of the config options.

I'm personally leaning towards having a shorter more direct - You are currently using a development build. By default this will switch to prod builds when not on localhost. You can force this with `workbox.setConfig({debug: true|false|)`.

Issue #1450, opened 2018-04-26
node v10 issues with our automated tests


jeffposnick:
Both Travis CI and AppVeyor have node v10 available now, and both are running into a similar issue with our test suite.

https://travis-ci.org/GoogleChrome/workbox/builds/371742444#L506 shows:

```sh
[Workbox Log]: Running Node test on test/workbox-background-sync/node with NODE_ENV 'dev'
> workbox@0.0.0 test /home/travis/build/GoogleChrome/workbox
> nyc --clean false --require @std/esm --require ./infra/testing/sw-env --silent mocha ./infra/testing/auto-stub-logger.mjs "test/workbox-background-sync/node/**/*.{js,mjs}"
/home/travis/.nvm/versions/node/v10.0.0/bin/node[6401]: ../src/node_file.cc:829:void node::fs::Stat(const v8::FunctionCallbackInfo<v8::Value>&): Assertion `(argc) == (3)' failed.
 1: node::Abort() [/home/travis/.nvm/versions/node/v10.0.0/bin/node]
 2: 0x87b6c5 [/home/travis/.nvm/versions/node/v10.0.0/bin/node]
 3: 0x8b2de2 [/home/travis/.nvm/versions/node/v10.0.0/bin/node]
 4: v8::internal::FunctionCallbackArguments::Call(v8::internal::CallHandlerInfo*) [/home/travis/.nvm/versions/node/v10.0.0/bin/node]
 5: 0xad62fa [/home/travis/.nvm/versions/node/v10.0.0/bin/node]
 6: v8::internal::Builtin_HandleApiCall(int, v8::internal::Object**, v8::internal::Isolate*) [/home/travis/.nvm/versions/node/v10.0.0/bin/node]
 7: 0x17db3588427d
Aborted (core dumped)
npm ERR! code ELIFECYCLE
npm ERR! errno 134
npm ERR! workbox@0.0.0 test: `nyc --clean false --require @std/esm --require ./infra/testing/sw-env --silent mocha ./infra/testing/auto-stub-logger.mjs "test/workbox-background-sync/node/**/*.{js,mjs}"`
npm ERR! Exit status 134
```

There's a whole lot going on in that line (`nyc`! `@std/esm`! `mocha`!) and it's not immediately obvious to me what's triggering the error. In the meantime, I'm going to switch over our CI config to use node v9 explicitly.

jeffposnick:
This was fixed by https://github.com/GoogleChrome/workbox/pull/1511

PR #1451, opened 2018-04-26
Explicitly use node v9 in our CI config

jeffposnick:
See https://github.com/GoogleChrome/workbox/issues/1450

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ cache:
     - ~/.selenium-assistant
 
 node_js:
-  - 'stable'
+  - '9'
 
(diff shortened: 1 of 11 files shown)


jeffposnick:
A few additional things have been added to this PR, beyond just locking node to v9 in our CI environment:

- We started getting checksum validation errors due to mismatches in our existing `package-lock.json`, so that needed to be refreshed.
- Updating that pulled in what I guess is a slightly newer version of our JSDoc linting dependency, which then picked up some errors that needed to be fixed or our `pre-push` `git` hook would fail. So I included those in this PR as well.

PR #1459, opened 2018-05-01
RegExp tweak to allow for precacheAndRoute() options

### LINKED ISSUE #1453: Including options with injectManifest
**Library Affected**:
WorkBox 3.1.0

**Browser & Platform**:
All browsers

Hopefully this is a simple yes no answer. I've searched all over the web but haven't been able to find an example of this. Is it possible to include options with the `workbox.precaching.precacheAndRoute([])`?

For example if I wanted to include the `cleanUrls` option, how would I do that? With injectManifest this fails as I guess it isn't exactly what it is looking for.
```js
workbox.precaching.precacheAndRoute([], {/* additional options here */})
````

Is the only option here to use the `getManifest` to generate the array then pass it into `precacheAndRoute()` along with my custom options? And if so does anyone have an example of this setup?

Thanks!

### PR DESCRIPTION
**Pull request by jeffposnick**
R: @philipwalton
CC: @Nooshu to confirm that this behaves as per their expectation.

Fixes #1453 

This tweaks the default `injectionPointRegexp` to support matching calls to `precacheAndRoute` that configure options, while maintaining backwards compatibility.

I.e. both of these can be used now inside the `swSrc`, while in `injectManifest` mode:

```js
workbox.precaching.precacheAndRoute([], {cleanUrls: true});
workbox.precaching.precacheAndRoute([]);
```

### DIFF
--- a/packages/workbox-build/src/entry-points/options/defaults.js
+++ b/packages/workbox-build/src/entry-points/options/defaults.js
@@ -24,5 +24,5 @@ module.exports = {
   navigateFallback: undefined,
   skipWaiting: false,
   importWorkboxFrom: 'cdn',
-  injectionPointRegexp: /(\.precacheAndRoute\()\s*\[\s*\]\s*(\))/,
+  injectionPointRegexp: /(\.precacheAndRoute\()\s*\[\s*\]\s*(\)|,)/,
 };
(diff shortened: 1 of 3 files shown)


Nooshu:
I can confirm that I now get the expected output. See [comment here](https://github.com/GoogleChrome/workbox/issues/1453#issuecomment-385760077)

Issue #1441, opened 2018-04-20
Sometimes user can get into a broken state with service worker


coryrylan:
**Library Affected**:
workbox-build 3.0.1

**Browser & Platform**:
Chrome 65 Windows 10 x64

**Issue or Feature Request Description**:
Not sure if this is the correct place to file this. On our production site using Workbox 3.0.1 with Chrome 65. Sometimes the user gets stuck with the following screen.

![down-marketplace1](https://user-images.githubusercontent.com/2021067/39061679-5b656d9e-4482-11e8-8b95-497747e2c68a.PNG)

Any page within control of the sw is trapped on this view. The only way to get out of this state is to force unregister the sw. 

![down-marketplace2](https://user-images.githubusercontent.com/2021067/39061748-91afb1ca-4482-11e8-8b63-b86a415304f4.PNG)

![down-marketplace3](https://user-images.githubusercontent.com/2021067/39061765-9d454a2c-4482-11e8-92ba-ceac0b115b9f.PNG)

![down-marketplace4](https://user-images.githubusercontent.com/2021067/39061773-a2c56efa-4482-11e8-9d95-91722f6e2b5a.PNG)

Here is the live url but reproducing this is spotty. https://www.estatesales.net/marketplace

This is our workbox-build config currently

```
import { environment } from './../src/environments/environment';

const workboxBuild = require('workbox-build');

const thirtyDays = 2592000;
const oneDay = 86400;
const tenMinutes = 600;
const cacheableResponse = { statuses: [0, 200, 404] };

const v2Matcher = /^\/v2/;
const styleGuideMatcher = /^\/style-guide/;
const marketplaceMatcher = /^(\/account)?\/marketplace/;
const cityPageAndSaleViewMatcher = /^\/[a-z]{2}(-[a-z]{2})?\/[^/]+(\/([\d]{5})(\/([\d])+)?)?$/i;

workboxBuild
  .generateSW(getSwConfig(false))
  .then(() => {
    workboxBuild.generateSW(getSwConfig(true));
  })
  .then(() => {
    console.log('Service Worker generated');
  })
  .catch((error: any) => {
    console.error(error);
  });

function getSwConfig(productionClient: boolean) {
  // https://workboxjs.org/
  return {
    skipWaiting: true,
    swDest: productionClient ? './dist/client--prod/service-worker.js' : './dist/client/service-worker.js',
    navigateFallback: '/app-shell.html',
[...]

PR #1460, opened 2018-05-01
Fallback to network if there's a cache miss for registerNavigationRoute

jeffposnick:
FYI: @philipwalton
CC: @coryrylan

This provides a fallback for the scenario mentioned in #1441, but I'd like to leave that issue open to track the underlying cause.

This is the equivalent of the previously approached #1300, in a fresh PR against the current `master` branch.


philipwalton:
SGTM

PR #1463, opened 2018-05-01
Use an empty object rather than undefined when calling new Headers()

### LINKED ISSUE #1461: new Headers(undefined) doesn't work with Samsung Internet 6.4
**Library Affected**:
*workbox-streams*

I'm testing out a new site using `workbox.streams` with Samsung Internet 6.4, and found the following:

screen shot 2018-05-01 at 2 10 18 pm

That's coming from

```js
const headers = new Headers(headersInit);
```

Manually running

```js
new Headers(undefined);
```

from within the Samsung Internet JS console leads to that same `TypeError: Failed to construct 'Headers': No matching constructor signature.(…)` error, so I guess Chromium 56 didn't support `undefined` as a parameter to the `Headers` constructor.

jeffposnick:
R: @philipwalton

Fixes #1461

Small change needed to accommodate a quirk with the `Headers` constructor in Samsung Internet 6.4 (Chrome 56).

### DIFF
--- a/packages/workbox-streams/utils/createHeaders.mjs
+++ b/packages/workbox-streams/utils/createHeaders.mjs
@@ -29,7 +29,8 @@ import '../_version.mjs';
  * @memberof workbox.streams
  */
 function createHeaders(headersInit) {
-  const headers = new Headers(headersInit);
+  // See https://github.com/GoogleChrome/workbox/issues/1461
+  const headers = new Headers(headersInit || {});
   if (!headers.has('content-type')) {
     headers.set('content-type', 'text/html');
   }
--- a/packages/workbox-streams/utils/createHeaders.mjs
+++ b/packages/workbox-streams/utils/createHeaders.mjs
@@ -28,9 +28,9 @@ import '../_version.mjs';
  *
  * @memberof workbox.streams
  */
-function createHeaders(headersInit) {
+function createHeaders(headersInit = {}) {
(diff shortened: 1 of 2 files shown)


packages/workbox-streams/utils/createHeaders.mjs
```diff
@@ -29,7 +29,8 @@ import '../_version.mjs';
  * @memberof workbox.streams
  */
 function createHeaders(headersInit) {
-  const headers = new Headers(headersInit);
+  // See https://github.com/GoogleChrome/workbox/issues/1461
+  const headers = new Headers(headersInit || {});
```
philipwalton: Optionally, `headersInit` could default to `{}` in the function declaration.

test/workbox-streams/node/utils/test-createHeaders.mjs
```diff
@@ -1,15 +1,26 @@
 import {expect} from 'chai';
+import sinon from 'sinon';
 
 import {createHeaders} from '../../../../packages/workbox-streams/utils/createHeaders.mjs';
 
 describe(`[workbox-streams] utils/createHeaders`, function() {
+  const sandbox = sinon.sandbox.create();
+
+  afterEach(async function() {
```
philipwalton: Does this need to be async?

jeffposnick:
The review feedback's been addressed!

PR #1476, opened 2018-05-08
Runtime check for a functional ReadableStream

### LINKED ISSUE #1473: More nuanced workbox.streams.isSupported() logic
**Library Affected**:
*workbox-streams*

https://github.com/GoogleChrome/workbox/blob/116a873d0957aaa130c814f5fdc5f2a75b27e269/packages/workbox-streams/isSupported.mjs#L29-L31

is our check to see whether the necessary platform features are available in order to use `workbox.streams`.

Unfortunately, this check is `true` in Edge 17, but responding with a `ReadableStream` will fail due to a known issue: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/17375439/

A check like 

```js
try {
  new ReadableStream({start() {}});
  return true;
} catch (error) {
  return false;
}
```

is probably required to actually confirm the functionality works in the meantime.

jeffposnick:
R: @philipwalton

Fixes #1473

Not urgent, but something to get in for our next patch release.

I've got a locally copy of Workbox with this patch deployed for my streams demo, and can confirm that it properly returned `false` for the current buggy Edge 17 build.

### DIFF
--- a/packages/workbox-streams/isSupported.mjs
+++ b/packages/workbox-streams/isSupported.mjs
@@ -15,6 +15,8 @@
 
 import './_version.mjs';
 
+let cachedIsSupported = undefined;
+
 /**
  * This is a utility method that determines whether the current browser supports
  * the features required to create streamed responses. Currently, it checks if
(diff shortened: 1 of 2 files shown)


packages/workbox-streams/isSupported.mjs
```diff
@@ -19,15 +19,20 @@ import './_version.mjs';
...
+ * can be created.
  *
  * @return {boolean} `true`, if the current browser meets the requirements for
  * streaming responses, and `false` otherwise.
  *
  * @memberof workbox.streams
  */
 function isSupported() {
-  return 'ReadableStream' in self;
+  try {
```
philipwalton: It's probably worth adding an explanation or linking to the bug explaining why this is needed.

Also, does this get invoked more than once in a stream strategy context? If so maybe it's worth caching the `isSupported` value?

PR #1488, opened 2018-05-18
Improve the code that swaps 'prod' for 'dev' in copyWorkboxLibraries

### LINKED ISSUE #1478: Webpack plugin replaces "prod" with "dev" in output path and fails.
**Library Affected**:
workbox-webpack-plugin 3.2.0

**Browser & Platform**:
Linux, Node 8.11.1

I'm building on my production server under a path that has `production` in it and for some reason Workbox replaces "prod" with "dev", resulting in "devuction", which obviously doesn't exist, and fails.  The origin seems to be [here](https://github.com/GoogleChrome/workbox/blob/master/packages/workbox-build/src/lib/use-build-type.js#L21).

The error looks like that:
```
 95% emitting GenerateSWError: One or more of the Workbox libraries could not be copied over to the destination directory: Error: ENOENT: no such file or directory, lstat '/var/www/website/devuction/v2.0.0/node_modules/workbox-background-sync/build/workbox-background-sync.prod.js'
    at _callee$ (/var/www/website/production/v2.0.0/node_modules/workbox-build/build/lib/copy-workbox-libraries.js:227:19)
    at tryCatch (/var/www/website/production/v2.0.0/node_modules/regenerator-runtime/runtime.js:62:40)
    at Generator.invoke [as _invoke] (/var/www/website/production/v2.0.0/node_modules/regenerator-runtime/runtime.js:296:22)
    at Generator.prototype.(anonymous function) [as throw] (/var/www/website/production/v2.0.0/node_modules/regenerator-runtime/runtime.js:114:21)
    at step (/var/www/website/production/v2.0.0/node_modules/babel-runtime/helpers/asyncToGenerator.js:17:30)
    at /var/www/website/production/v2.0.0/node_modules/babel-runtime/helpers/asyncToGenerator.js:30:13
    at <anonymous>
```

Renaming the production directory to something else solves it, or, even uglier, creating a `devuction` symlink pointing to `production`. Looking at the code there doesn't seem to be a way to disable that behavior.

Note that this is probably due to the `importWorkboxFrom: 'local'` option I'm using, so I suppose using the CDN would also resolve the issue, but I haven't tested that.

Any suggestions appreciated.

jeffposnick:
R: @philipwalton

Fixes #1478 

This code path is still a bit hacky, and there's a lot of hardcoded assumptions, but at least it should now work as expected when the `source` path contains multiple instances of `DEFAULT_BUILD_TYPE`.

### DIFF
--- a/test/workbox-build/node/lib/use-build-type.js
+++ b/test/workbox-build/node/lib/use-build-type.js
@@ -18,7 +18,7 @@ describe(`[workbox-build] lib/use-build-type.js`, function() {
     expect(result).to.eql('/path/to/production/also.prod.path/workbox.dev.js');
   });
 
-  it(`should not update anything is there is no match for the default build type`, function() {
+  it(`should not update anything if there is no match for the default build type`, function() {
     const result = useBuildType('/does/not/match', 'dev');
     expect(result).to.eql('/does/not/match');
   });
--- a/test/workbox-build/node/lib/use-build-type.js
+++ b/test/workbox-build/node/lib/use-build-type.js
@@ -15,7 +15,7 @@ describe(`[workbox-build] lib/use-build-type.js`, function() {
 
   it(`should only update the last match when buildType is 'dev'`, function() {
     const result = useBuildType('/path/to/production/and.prod.check/workbox.prod.js', 'dev');
-    expect(result).to.eql('/path/to/production/also.prod.path/workbox.dev.js');
+    expect(result).to.eql('/path/to/production/and.prod.check/workbox.dev.js');
   });
 
   it(`should not update anything if there is no match for the default build type`, function() {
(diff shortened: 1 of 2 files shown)


test/workbox-build/node/lib/use-build-type.js
```diff
@@ -0,0 +1,25 @@
...
+    const result = useBuildType('/path/to/workbox.prod.js', 'dev');
+    expect(result).to.eql('/path/to/workbox.dev.js');
+  });
+
+  it(`should only update the last match when buildType is 'dev'`, function() {
+    const result = useBuildType('/path/to/production/and.prod.check/workbox.prod.js', 'dev');
+    expect(result).to.eql('/path/to/production/also.prod.path/workbox.dev.js');
+  });
+
+  it(`should not update anything is there is no match for the default build type`, function() {
```
philipwalton: Change the first "is" to "if".

PR #1492, opened 2018-05-23
Run the webpack test suite against webpack v4

### LINKED ISSUE #1281: webpack v3 + v4 integration tests
**Library Affected**:
*workbox-webpack-plugin*

https://github.com/GoogleChrome/workbox/pull/1275 adds support for webpack v4's plugin architecture. However, our current test suite only runs integration tests against webpack v3.

A module like https://www.npmjs.com/package/npm-install-version might be a sane way to test against multiple versions without having to play games with `package.json`.

Running our full test suite against both v3 and v4 won't be possible until all of the other webpack plugins that we rely on for our integration tests are also upgraded to support v4:

- https://github.com/webpack-contrib/copy-webpack-plugin
- https://github.com/kjbekkelund/generate-asset-webpack-plugin
- https://github.com/jantimon/html-webpack-plugin

jeffposnick:
R: @philipwalton 

Fixes #1281 

Given that there doesn't seem to be a clean way to run our tests under both webpack v3 & v4, and given that webpack v4 has all the momentum behind it, it seems like just switching over to run our test suite against v4 makes the most sense.

This PR does not change our compatibility story with webpack v3—it's just changes to our tests.

### DIFF
--- a/infra/testing/create-webpack-asset-plugin.js
+++ b/infra/testing/create-webpack-asset-plugin.js
@@ -1,0 +1,24 @@
+// A small helper class to generate a "fake" webpack asset for testing purposes.
+// Roughly equivalent to https://www.npmjs.com/package/generate-asset-webpack-plugin,
+// but this uses the webpack v4 syntax.
+
+class CreateWebpackAssetPlugin {
+  constructor(name) {
+    if (typeof name !== 'string') {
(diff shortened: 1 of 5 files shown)


philipwalton:
Makes sense, and SGTM!

PR #1500, opened 2018-05-25
New workbox.Plugin.deleteCacheAndMetadata() method

### LINKED ISSUE #1365: Add helper method to cleanup cache + expiration data in IDB
**Library Affected**:
*workbox-cache-expiration*

Extracting a feature request from https://github.com/GoogleChrome/workbox/issues/1354#issuecomment-373188977, @josephliccini and I write:

> One thing that I'm not sure about is how well the current workbox-cache-expiration code will behave if you do an "out of band" caches.delete() call to remove a specific runtime cache without also deleting the IDB entries corresponding to its expiration, and then later on recreate the same runtime cache. I'd imagine that it could lead to some confusion—@gauntface?
> 
> Maybe the solution there is to expose a workbox.expiration.deleteCache() method that took in a cache name and did both the Cache Storage API deletion as well as the IDB cleanup in one step.
> 
> Yes (regarding your response to 5.) this is basically how we've implemented /logout today, and the next step is to go to IDB and clear those out too. I would love a deleteCache() method that does IDB and Cache deletion for a workbox cache 👍

jeffposnick:
R: @philipwalton
CC: @josephliccini 

Fixes #1365

It took a while to get a hang of what needed to be explicitly `close()`d in order to avoid `blocked` errors, but I think this implementation works.

I envision making use of this as part of a future PR that will start introducing some opt-in automatic cleanup logic to handle, e.g., `QUOTA_EXCEEDED` errors. For the time being, though, this is just a standalone method that won't be called anywhere automatically.

### DIFF
--- a/gulp-tasks/test-integration.js
+++ b/gulp-tasks/test-integration.js
@@ -128,9 +128,7 @@ gulp.task('test-integration', async () => {
           }
           break;
         case 'safari':
-          if (localBrowser.getReleaseName() === 'beta') {
-            await runIntegrationForBrowser(localBrowser);
-          }
+          await runIntegrationForBrowser(localBrowser);
           break;
         default:
           logHelper.warn(oneLine`
--- a/packages/workbox-cache-expiration/CacheExpiration.mjs
+++ b/packages/workbox-cache-expiration/CacheExpiration.mjs
@@ -268,8 +268,6 @@ class CacheExpiration {
   /**
    * Removes the IndexedDB object store used to keep track of cache expiration
    * metadata.
-   *
-   * @return {Promise<void>}
    */
   async delete() {
     // Make sure we don't attempt another rerun if we're called in the middle of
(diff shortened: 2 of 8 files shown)


surajrautela:
I am still not able to use this method . I am using workbox files from https://storage.googleapis.com/workbox-cdn/releases/3.2.0/workbox-sw.js  . On trying to to use this function is throws an error that deleteCacheAndMetadata() is not a method. Can we have a sample code snippet which can help just in case i am doing something wrong

jeffposnick:
Sorry, we haven't yet cut a new release of Workbox that includes the changes in this PR. There are a few other open PRs that it would be good to get merged first, and then we'll cut a new release that includes `deleteCacheAndMetadata()`.

(If you can't wait, you could do a `gulp build` on the `master` branch of a local copy of this repo, and then run `packages/workbox-cli/build/bin.js copyLibraries /path/to/directory` to make a local copy of the latest libraries.)

surajrautela:
Really appreciate such a fast response . By what time are you guys planning to have a new release?

PR #1503, opened 2018-05-28
Add requestDidReplay callback to BackgroundSync Queue (#1502)

### LINKED ISSUE #1502: Feature Request: requestDidReplay callback in background sync queue
**Library Affected**:
*workbox-background-sync*

**Browser & Platform**:
*"all browsers"*

This followup: https://github.com/GoogleChrome/workbox/issues/868#issuecomment-392139983

I need a callback `requestDidReplay` invoked after each replay response to read the effective response from the API before the next request is handled.

_Description of the use case_

Consider a simple CRUD REST API server. The user can create an object and then modify it, regardless he is online or offline, from the webapp.

When he's **online**, 2 REST API calls are done, and the server generates an id on `POST` (first call)

`POST /api/object` (API server generates an `id=1337`)
`PUT /api/object/1337`

When he's **offline**, Service Worker and background-sync module save requests into a Queue, but the id is generated by ServiceWorker instead of API.

`POST /api/object` => The ServiceWorker generates a local temporary negative `id=-1`, and this requests is queued.
`PUT /api/object/-1` => The following update request is also queued with it's temporary id.

When user is back online, those requests can be replayed, but we have to read the API server response from `POST` request to get the database id generated by the server, and replace it into `PUT` request before replaying it.

This can't be done without introducting a new `requestDidReplay` callback. (`requestWillReplay` and a global object to store ids mapping will also be used).

Toilal:
R: @jeffposnick @philipwalton

Fixes #1502

*Adds `requestDidReplay` and `queueWillReplay` callback*

### DIFF
--- a/packages/workbox-background-sync/Queue.mjs
+++ b/packages/workbox-background-sync/Queue.mjs
@@ -123,6 +123,8 @@ class Queue {
     const replayedRequests = [];
     const failedRequests = [];
 
+    await this._runCallback('queueWillReplay');
+
     let storableRequest;
     while (storableRequest = await this._queueStore.getAndRemoveOldestEntry()) {
       // Make a copy so the unmodified request can be stored
@@ -154,6 +156,7 @@ class Queue {
         replay.error = err;
         failedRequests.push(storableRequestClone);
       }
+      await this._runCallback('requestDidReplay', replay);
 
       replayedRequests.push(replay);
     }
--- a/test/workbox-background-sync/node/test-Queue.mjs
+++ b/test/workbox-background-sync/node/test-Queue.mjs
@@ -334,12 +334,16 @@ describe(`[workbox-background-sync] Queue`, function() {
     });
 
     it(`should invoke all replay callbacks`, async function() {
+      const queueWillReplay = sinon.spy();
       const requestWillReplay = sinon.spy();
+      const requestDidReplay = sinon.spy();
       const queueDidReplay = sinon.spy();
 
       const queue = new Queue('foo', {
         callbacks: {
+          queueWillReplay,
           requestWillReplay,
+          requestDidReplay,
           queueDidReplay,
         },
       });
@@ -347,6 +351,8 @@ describe(`[workbox-background-sync] Queue`, function() {
       await queue.addRequest(new Request('/one'));
       await queue.addRequest(new Request('/two'));
       await queue.replayRequests();
+
+      expect(queueWillReplay.calledOnce).to.be.true;
 
       expect(requestWillReplay.calledTwice).to.be.true;
       expect(requestWillReplay.getCall(0).calledWith(sinon.match({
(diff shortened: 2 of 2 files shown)


Toilal:
To implement the use case in #1502, when a requests fail, we have to set a global `let hasFailed= true` to force an error on all following requests to be retried (`throw new Error()`). Then all requests will be requeued in the original order.

It seems like a hack, it's sad that the API is not more opened to implement this use case properly.

philipwalton:
@Toilal I agree, which is why I wanted to make sure your use case was being met before just merging this change. When we designed this API, we didn't have a lot of use cases in mind, so I'm happy to rework it to make it more flexible.

Do you have any ideas?

One option -- which would be the most flexible but also the biggest pain for users to implement -- is to provide raw access to the request queue. This would allow users to manually retry all requests however they way, and if any fail they can then manually add them back to the queue. What do you think?

Toilal:
To cover more real use cases.

- Both callbacks added by this pull requests (`queueWillReplay` and `requestDidReplay`).

- A way to break the current replay loop. A loop `break` could be introduced when `requestDidReplay` or `requestWillReplay` returns `false`. When the loop is interrupted as such, the current request beeing replayed should be restored in the queue at its original position (last). It should still be the first element of the loop when the queue is replayed the next time. In many case, we need to garantee the requests are replayed in order, and don't wan't to go to next one when something goes wrong.

- A way to mutate stored requests (to refresh Authorization header for example, when JWT token is expired).

```js
queue.forEach((request) => doMutate(request))
```

vcastro45:
I allow myself to suggest an idea that goes through my head:

- It will also useful to extend the "case of fail". In fact, currently a request is queued if the `fetch()` call throw an error but sometimes this `fetch()` call is successful but the request has not been handled properly by the server. For example, if the server returns a HTTP Code 401 (when the JWT refresh token is expired), we would like to be able to consider the request as failed and renew the token on the next `queueWillReplay`

Of course, this example is for code 401 but it can be adapted for any code (>= 400 ?)

fomenyesu:
Totally agree with you. @Toilal 
What you said is a normal use case in the real world of background sync.

bhargashah:
Since this work is approved, can this be merged soon?

Toilal:
I have rebased this pull request on current master version (3.6.2), and all checks are now OK.

Please consider merging this one. I'm using it since months at it may solve many use cases.

philipwalton:
Hey all, we're going to address this issue as part of our next major release (since it may affect the public API). We're still exploring whether this PR is sufficient or whether it requires a larger change to handle some of the use cases we've seen.

WadeFanyao:
Hi everyone. I have a simple case and I am not sure the next release will help.  For example,  I have a get request to get data and a post request to send the data to server. The post request needs the data from the get request.  To my understanding, I can implement it with callback or promises when it is online. But I just wondering if the workbox-background-sync(Next release) can queue them together and run behind when the connection is back.

vcastro45:
@WadeFanyao I've done it using the `requestDidReplay()` implementation but it is painful.
In fact, if you want to do it, you have to use temporary datas to link your requests (like temporary negative IDs) and loop through each pending request when a request is replayed to assign the new datas (generated by the server) to the upcoming request using temporary datas before they are sent to the server.

If you didn't understand, no worry, it's because it's awful.

philipwalton:
Hey all, I'm submitted https://github.com/GoogleChrome/workbox/pull/1710, which I believe will address all of the use cases discussed here. Please take a look at that PR and let me know what you think.

philipwalton:
Closing in favor of #1710.

Issue #1312, opened 2018-02-14
Handling of quota exceeded errors in workbox-precaching


jeffposnick:
**Library Affected**:
*workbox-precaching*

This is closely related to https://github.com/GoogleChrome/workbox/issues/1308, but the exact analog is https://github.com/GoogleChromeLabs/sw-precache/issues/345

Let's leave runtime caching aside for this issue, and assume that there's a web app instance that has revision *N* of their assets precached. The web app is updated and deployed, and now there's an attempt to download and cache a few of the new/updated assets that make up revision *N+1*. `workbox-precaching` does this in the `install` handler, and will fail installation if any of the individual `cache.put()` operations fail. (Since all the assets from revision *N* are still cached at this point, it's not too hard to imagine scenarios in which caching the new assets brings you over quota.)

From the point of view of the web page clients, there are events fired on the associated service worker registration showing the service worker progressing from the `installing` state to the `redundant` state. Unless (somehow) quota is freed up at some point, that `installing` -> `redundant` transition will happen indefinitely, leaving the user stuck revision *N* of all the precached assets indefinitely.

We should do something to help developers out here. What about a pattern like:

```js
if ('serviceWorker' in navigator) {
  window.addEventListener('load', function() {
    navigator.serviceWorker.register('/sw.js').then(function(registration) {
      registration.addEventListener('updatefound', function() {
        var newWorker = registration.installing;
        newWorker.addEventListener('statechange', function() {
          if (newWorker.state === 'redundant') {
            if ('storage' in navigator && 'estimate' in navigator.storage) {
              navigator.storage.estimate().then(function(results) {
                var percentUsed = (results.usage / results.quota) * 100;
                // Let's assume that if we're using 95% of our quota, then this failure
                // was due to quota exceeded errors.
                // TODO: Hardcoding a threshold stinks.
                if (percentUsed >= 0.95) {
                  // Get rid of the existing SW so that we're not stuck
                  // with the previously cached content.
                  registration.unregister();

                  // Let's assume that we have some way of doing this without inadvertantly
                  // blowing away storage being used on the origin by something other than
                  // our service worker.
                  // I don't think that the Clear-Site-Data: header helps here, unfortunately.
                  deleteAllCachesAndIBD();
                }
              });
            } else {
              // What about browsers that don't support navigator.storage.estimate()?
              // There's no way of guessing why the service worker is redundant.
            }
          }
        });
      });
    });
  });
}
```

[...]

PR #1505, opened 2018-05-30
Opt-in method for purging (some) caches upon a QuotaExceededError

### LINKED ISSUE #1308: Handling of quota exceeded errors in workbox-strategies
**Library Affected**:
*workbox-strategies*

There's be an uptick of issues related to `DOMException: Quota exceeded.` reported recently, both from Workbox users and from the larger development community. A [few](https://bugs.chromium.org/p/chromium/issues/detail?id=795134) [bugs](https://bugs.chromium.org/p/chromium/issues/detail?id=801024) in Chrome, along with what appears to be a stricter enforcement of per-origin quota limitations (especially when [opaque responses](https://stackoverflow.com/questions/39109789/what-limitations-apply-to-opaque-responses/39109790#39109790) are cached) have all brought this to the forefront.

This is likely to remain an issue if Safari goes ahead with a maximum limit of 50 MiB of cache storage when they ship service workers, as [they've indicated](https://webkit.org/blog/8090/workers-at-your-service/).

We could do more to ensure that, when appropriate, Workbox will catch `DOMException: Quota exceeded.` that occur when writing to caches, and ensure that they lead to failures in passing along a valid response to the client page. 

I think it makes sense to address this at the `workbox-strategies` level, rather than in the underlying `cacheWrapper` from `workbox-core`, because we want the flexibility to treat a `cache.put()` as a fatal exception for some use cases. A failure to write to caches within `workbox-precaching` should still be fatal, and cause the `install` event to fail, for instance.

I'm thinking changing code like

https://github.com/GoogleChrome/workbox/blob/aa95c449b45102e59d5d578738263128eff2b0ec/packages/workbox-strategies/CacheFirst.mjs#L142-L149

to be:

```js
event.waitUntil(
  cacheWrapper.put(
    this._cacheName,
    event.request,
    responseClone,
    this._plugins
  ).catch((error) => {
    // Log the error, but don't reject.
  });
);
```

would help, along with corresponding changes in the other `workbox-strategies`.

What do you think, @gauntface & co.?

jeffposnick:
R: @philipwalton
CC: @raejin @Enalmada @josephliccini

Fixes #1308

There are two changes in this PR:

- There's a check in `workbox-core`'s `cacheWrapper.put()` (which is used "under the hood" throughout Workbox) method to detect when the underlying `cache.put()` fails due to a `QuotaExceededError`. When that happens, one or more previously registered callback functions will automatically be called, and given a chance to free up storage. There is (currently) no attempt made to retry the failed `cache.put()` automatically after the cleanup completes, and the `QuotaExceededError` will be re-thrown and bubble-up to the method that called `cacheWrapper.put()` originally.

- There's a new `purgeOnQuotaError` option added to the `workbox.expiration.Plugin` constructor. It defaults to `false`. When `true`, it will automatically register a quota error callback that will delete the **entire** cache (and IndexedDB) data that the plugin is responsible for. (This is based on #1500.)

Nothing should change by default, but developers who want to opt-in to trying this automatic deletion of caches should make sure that they're using `workbox.expiration.Plugin` for a given strategy, and set `purgeOnQuotaError: true` alongside their `maxEntries` and/or `maxAgeSeconds` configuration when constructing the plugin.

At some point in the future, we might:

- Automatically retry the failed `cache.put()` operation after all the callbacks are complete.
- Default to `purgeOnQuotaError: true` inside of `workbox.expiration.Plugin`.
- Add in some special logic inside `workbox.precaching` to handle quota errors in a different way.

In the meantime, I'd love to start getting some feedback from developers who frequently encounter quota issues with their real-world applications, and figure out how effective opting-in to this new behavior is.

It will have the most significant effect on developers whose runtime caches contribute the largest amounts towards their quota usage (perhaps because they're [caching opaque responses](https://stackoverflow.com/questions/39109789/what-limitations-apply-to-opaque-responses)). If you're precaching a massive amount of data, but not runtime caching much, this PR is not likely to help.

### DIFF
--- a/packages/workbox-cache-expiration/Plugin.mjs
+++ b/packages/workbox-cache-expiration/Plugin.mjs
@@ -15,6 +15,7 @@ import {CacheExpiration} from './CacheExpiration.mjs';
 import {WorkboxError} from 'workbox-core/_private/WorkboxError.mjs';
 import {assert} from 'workbox-core/_private/assert.mjs';
 import {cacheNames} from 'workbox-core/_private/cacheNames.mjs';
+import {registerCallback} from 'workbox-core/_private/quota.mjs';
 
 import './_version.mjs';
 
@@ -43,6 +44,8 @@ class Plugin {
    * Entries used the least will be removed as the maximum is reached.
    * @param {number} [config.maxAgeSeconds] The maximum age of an entry before
    * it's treated as stale and removed.
+   * @param {boolean} [config.purgeOnQuotaError] Whether to opt this cache in to
+   * automatic deletion if the available storage quota has been exceeded.
    */
   constructor(config = {}) {
     if (process.env.NODE_ENV !== 'production') {
@@ -76,6 +79,10 @@ class Plugin {
     this._config = config;
     this._maxAgeSeconds = config.maxAgeSeconds;
     this._cacheExpirations = new Map();
+
+    if (config.purgeOnQuotaError) {
+      registerCallback(() => this.deleteCacheAndMetadata());
+    }
   }
 
   /**
@@ -235,7 +242,7 @@ class Plugin {
    * There is no Workbox-specific method needed for cleanup in that case.
    */
   async deleteCacheAndMetadata() {
-    // Do this one at at a time instance of all at once via `Promise.all()` to
+    // Do this one at at a time instead of all at once via `Promise.all()` to
     // reduce the chance of inconsistency if a promise rejects.
     for (const [cacheName, cacheExpiration] of this._cacheExpirations) {
       await caches.delete(cacheName);
--- a/packages/workbox-cache-expiration/Plugin.mjs
+++ b/packages/workbox-cache-expiration/Plugin.mjs
@@ -242,7 +242,7 @@ class Plugin {
    * There is no Workbox-specific method needed for cleanup in that case.
    */
   async deleteCacheAndMetadata() {
-    // Do this one at at a time instead of all at once via `Promise.all()` to
+    // Do this one at a time instead of all at once via `Promise.all()` to
     // reduce the chance of inconsistency if a promise rejects.
     for (const [cacheName, cacheExpiration] of this._cacheExpirations) {
       await caches.delete(cacheName);
--- a/packages/workbox-core/_private.mjs
+++ b/packages/workbox-core/_private.mjs
@@ -23,7 +23,9 @@ import {cacheWrapper} from './_private/cacheWrapper.mjs';
 import {fetchWrapper} from './_private/fetchWrapper.mjs';
 import {getFriendlyURL} from './_private/getFriendlyURL.mjs';
 import {logger} from './_private/logger.mjs';
-import {registerCallback} from './_private/quota.mjs';
+import {
+  registerCallback as registerQuotaErrorCallback,
+} from './_private/quota.mjs';
 
 import './_version.mjs';
 
@@ -36,5 +38,5 @@ export {
   fetchWrapper,
   getFriendlyURL,
   logger,
-  registerCallback,
+  registerQuotaErrorCallback,
 };
--- a/packages/workbox-core/_private.mjs
+++ b/packages/workbox-core/_private.mjs
@@ -15,23 +15,26 @@
 */
 
 // We either expose defaults or we expose every named export.
+import {DBWrapper} from './_private/DBWrapper.mjs';
+import {WorkboxError} from './_private/WorkboxError.mjs';
 import {assert} from './_private/assert.mjs';
+import {cacheNames} from './_private/cacheNames.mjs';
 import {cacheWrapper} from './_private/cacheWrapper.mjs';
 import {fetchWrapper} from './_private/fetchWrapper.mjs';
-import {DBWrapper} from './_private/DBWrapper.mjs';
-import {logger} from './_private/logger.mjs';
-import {WorkboxError} from './_private/WorkboxError.mjs';
-import {cacheNames} from './_private/cacheNames.mjs';
 import {getFriendlyURL} from './_private/getFriendlyURL.mjs';
+import {logger} from './_private/logger.mjs';
+import {registerCallback} from './_private/quota.mjs';
+
 import './_version.mjs';
 
 export {
-  logger,
+  DBWrapper,
+  WorkboxError,
   assert,
   cacheNames,
   cacheWrapper,
   fetchWrapper,
-  WorkboxError,
-  DBWrapper,
   getFriendlyURL,
+  logger,
+  registerCallback,
 };
--- a/packages/workbox-core/_private/cacheWrapper.mjs
+++ b/packages/workbox-core/_private/cacheWrapper.mjs
@@ -14,12 +14,14 @@
   limitations under the License.
 */
 
-import {logger} from './logger.mjs';
-import {assert} from './assert.mjs';
-import {WorkboxError} from './WorkboxError.mjs';
-import {getFriendlyURL} from '../_private/getFriendlyURL.mjs';
 import pluginEvents from '../models/pluginEvents.mjs';
 import pluginUtils from '../utils/pluginUtils.mjs';
+import {WorkboxError} from './WorkboxError.mjs';
+import {assert} from './assert.mjs';
+import {executeCallbacks} from './quota.mjs';
+import {getFriendlyURL} from './getFriendlyURL.mjs';
+import {logger} from './logger.mjs';
+
 import '../_version.mjs';
 
 /**
@@ -38,7 +40,7 @@ import '../_version.mjs';
 const putWrapper = async (cacheName, request, response, plugins = []) => {
   if (!response) {
     if (process.env.NODE_ENV !== 'production') {
-      logger.error(`Cannot cache non-existant response for ` +
+      logger.error(`Cannot cache non-existent response for ` +
         `'${getFriendlyURL(request.url)}'.`);
     }
 
@@ -80,9 +82,15 @@ const putWrapper = async (cacheName, request, response, plugins = []) => {
       `${getFriendlyURL(request.url)}.`);
   }
 
-  // Regardless of whether or not we'll end up invoking
-  // cacheDidUpdate, wait until the cache is updated.
-  await cache.put(request, responseToCache);
+  try {
+    await cache.put(request, responseToCache);
+  } catch (error) {
+    // See https://developer.mozilla.org/en-US/docs/Web/API/DOMException#exception-QuotaExceededError
+    if (error.name === 'QuotaExceededError') {
+      await executeCallbacks();
+    }
+    throw error;
+  }
 
   for (let plugin of updatePlugins) {
     await plugin[pluginEvents.CACHE_DID_UPDATE].call(plugin, {
--- a/packages/workbox-core/_private/cacheWrapper.mjs
+++ b/packages/workbox-core/_private/cacheWrapper.mjs
@@ -18,7 +18,7 @@ import pluginEvents from '../models/pluginEvents.mjs';
 import pluginUtils from '../utils/pluginUtils.mjs';
 import {WorkboxError} from './WorkboxError.mjs';
 import {assert} from './assert.mjs';
-import {executeCallbacks} from './quota.mjs';
+import {executeCallbacks as executeQuotaErrorCallbacks} from './quota.mjs';
 import {getFriendlyURL} from './getFriendlyURL.mjs';
 import {logger} from './logger.mjs';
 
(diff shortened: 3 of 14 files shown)


raejin:
Thank you for working on this! This is very exciting, as [the cache quota reporting chromium bug](https://bugs.chromium.org/p/chromium/issues/detail?id=795701) was recently fixed. 

With this implementation, one is still possible to manually add logic to retry precaching assets when quota error callbacks are called.

packages/workbox-cache-expiration/Plugin.mjs
```diff
@@ -235,7 +242,7 @@ class Plugin {
    * There is no Workbox-specific method needed for cleanup in that case.
    */
   async deleteCacheAndMetadata() {
-    // Do this one at at a time instance of all at once via `Promise.all()` to
+    // Do this one at at a time instead of all at once via `Promise.all()` to
```
philipwalton: While we're fixing grammar, remove the duplicate "at" earlier in this line.

test/workbox-core/node/_private/test-quota.mjs
```diff
@@ -0,0 +1,58 @@
...
+
+  before(function() {
+    sandbox = sinon.sandbox.create();
+  });
+
+  afterEach(function() {
+    sandbox.restore();
+  });
+
+  devOnly.it(`registerCallback() should throw when passed a non-function`, async function() {
```
philipwalton: Can you wrap all the tests related to `registerCallback()` in a describe block and then do a separate describe block for all the tests in `executeCallback()`.

packages/workbox-core/_private.mjs
```diff
@@ -15,23 +15,28 @@
...
+  WorkboxError,
   assert,
   cacheNames,
   cacheWrapper,
   fetchWrapper,
-  WorkboxError,
-  DBWrapper,
   getFriendlyURL,
+  logger,
+  registerQuotaErrorCallback,
```
philipwalton: What's the reasoning behind exposing this?

packages/workbox-cache-expiration/Plugin.mjs
```diff
@@ -76,6 +79,10 @@ class Plugin {
     this._config = config;
     this._maxAgeSeconds = config.maxAgeSeconds;
     this._cacheExpirations = new Map();
+
+    if (config.purgeOnQuotaError) {
+      registerCallback(() => this.deleteCacheAndMetadata());
```
philipwalton: Is the plan that `registerCallback()` might/will get invoked in more places in the future? From what I can tell in this PR, this is the only place it's used, which makes the `quota.mjs` module seem like over-kill for just this.

packages/workbox-core/_private/quota.mjs
```diff
@@ -0,0 +1,74 @@
...
+      moduleName: 'workbox-core',
+      funcName: 'register',
+      paramName: 'callback',
+    });
+  }
+
+  callbacks.add(callback);
+
+  if (process.env.NODE_ENV !== 'production') {
+    logger.log('Registered a callback to respond to quota errors.', callback);
```
philipwalton: How you checked to see how these function are logged (IIRC normal functions output their `toString()` result)? Maybe it makes more sense to output `.name` value?

test/workbox-cache-expiration/node/test-Plugin.mjs
```diff
@@ -68,6 +69,30 @@ describe(`[workbox-cache-expiration] Plugin`, function() {
...
     });
+
+    it(`should register a quota error callback when purgeOnQuotaError is true`, async function() {
+      const plugin = new Plugin({
+        maxEntries: 10,
+        purgeOnQuotaError: true,
+      });
+      plugin.deleteCacheAndMetadata = sandbox.stub();
+
+      await executeCallbacks();
```
philipwalton: Same comment here as above w.r.t the name.

packages/workbox-precaching/controllers/PrecacheController.mjs
```diff
@@ -219,12 +219,14 @@ class PrecacheController {
     const tempCache = await caches.open(this._getTempCacheName());
 
     const requests = await tempCache.keys();
-    await Promise.all(requests.map(async (request) => {
+    // Process each request/response one at a time, deleting the temporary entry
+    // when done, to help avoid triggering quota errors.
+    for (const request of requests) {
```
philipwalton: +1 to this change. I think it'll make it a lot easier to respond to errors in a predictable way.

packages/workbox-core/_private/quota.mjs
```diff
@@ -0,0 +1,74 @@
...
+import '../_version.mjs';
+
+const callbacks = new Set();
+
+/**
+ * Adds a function to the set of callbacks that will be executed when there's
+ * a quota error.
+ *
+ * @param {Function} callback
+ *
```
philipwalton: Can you make the spacing here consistent with below. I'm not sure if we prefer a space or not.

packages/workbox-core/_private.mjs
jeffposnick: Just flexibility.

While the automatic cleanup of caches that use expiration is one way of dealing with quota errors, you could imagine a web app that, for example, stores a large amount of data in IndexedDB. Developers who understand how their web app stores data could use this to register a callback that cleaned up that hypothetical IndexedDB database.

packages/workbox-cache-expiration/Plugin.mjs
jeffposnick: Yes, along the lines of https://github.com/GoogleChrome/workbox/pull/1505#discussion_r194867665

In addition to giving developers a general-purpose way of registering a "cleanup" callback, we may want to make use of it elsewhere in the Workbox codebase—potentially as part of the precaching module, along the lines of #1312.

packages/workbox-core/_private/quota.mjs
jeffposnick: Done.

packages/workbox-cache-expiration/Plugin.mjs
jeffposnick: Done.

packages/workbox-core/_private/quota.mjs
jeffposnick: Yeah, it logs the code for the actual callback function. This will normally be

```js
() => this.deleteCacheAndMetadata()
```

or it might be the code for a custom callback, if you write and register your own.

I experimented with this a little bit and could go with just leaving out the logging of the function entirely, but I figured that more context for the development environment logging is better.

Since a developer could theoretically pass in whatever function they want here (in addition to using our automatically registered functions), I don't think we could count on `.name` being meaningful.

test/workbox-core/node/_private/test-quota.mjs
jeffposnick: I don't know about that—there's only one test that's just `registerCallback()` (which only runs in `devOnly` mode), and all of the rest of them rely on a combination of the two methods. That seems like overkill?

test/workbox-cache-expiration/node/test-Plugin.mjs
jeffposnick: Not sure what you mean—the name of what?

jeffposnick:
I've addressed some of the comments, and provided more context.

Another thing that I'd like to do is write up a "Guide to Handling Quota Errors" or something along those lines for https://developers.google.com/web/tools/workbox/guides/ following the next release, which will explain a bit more about opting-in to the automatic cleanup and registering your own callback.

test/workbox-cache-expiration/node/test-Plugin.mjs
philipwalton: This is definitely a nit, but I mention it because it confused me a bit when reading the code for the first time. 

Basically my point is that the name `executeCallbacks()` here has no context, so as a reader you have to go searching for where it's defined/imported to figure out what it's doing.

If it were called something like `executeQuotaErrorCallbacks()` or something like that, then it'd be more clear. Alternatively, if you imported the entire namespace as `quotaErrors`, then you'd have more context because it'd be `quotaErrors.executeCallbacks()`.

Anyway, like I said this is a nit. If you think it's fine, I'm OK with it.

philipwalton: Note, this comment primarily applies to [line 90 in `cacheWrapper`](https://github.com/GoogleChrome/workbox/pull/1505/files/b09566d00f13e1e709423d84af609dabd5d89c3d#diff-0e280af723edc4482219279065921834R90), not necessarily here.

packages/workbox-cache-expiration/Plugin.mjs
philipwalton: Yeah, I get the intention of the change, I'm just wondering if it could be somehow better integrated into the rest of workbox.

Did you consider creating a new lifecycle callback: maybe something like `cacheWriteDidFail`? And in the same way that we can register a generic fetch failure handler, we could register generic cache failure handler.

Anyway, just a thought, I don't want to block this change if a better solution isn't obvious.

test/workbox-core/node/_private/test-quota.mjs
jeffposnick: Done.

packages/workbox-core/_private/quota.mjs
jeffposnick: I'd just like to leave it as-is to make sure we're providing the full context to folks who are in debug mode and reading the Console log messages.

test/workbox-cache-expiration/node/test-Plugin.mjs
jeffposnick: Renamed to `executeQuotaErrorCallbacks()`.

packages/workbox-cache-expiration/Plugin.mjs
jeffposnick: I felt like using the lifecycle callback naming conventions would be misleading, since here you're registering a callback that has a "global" (or `workbox.core`) scope, which will be invoked regardless of which strategy triggers the quota exception. That doesn't match the behavior of the lifecycle callbacks, which are all scoped to a given strategy.

`workbox.routing.setCatchHandler()` is closer in semantics, but that only supports setting a single callback, while we need support for multiple quota exception callbacks. To me, `register...` sounds more in line with being able to create multiple callbacks than `set...`, but I could change the naming if you feel strongly.

packages/workbox-core/_private/quota.mjs
philipwalton: SG

packages/workbox-cache-expiration/Plugin.mjs
philipwalton: > I felt like using the lifecycle callback naming conventions would be misleading, since here you're registering a callback that has a "global"...

Right, I wasn't suggesting just using the name, I was suggesting we actually add a `cacheWriteDidFail` lifecycle callback. While I understand that it's not global and would have to be added per strategy, that seems like a benefit to me, not a shortcoming. And users could write their strategy code in such a way that sharing the same callbacks across all their strategies isn't so bad.

My concern is that global handlers can lead to unpredictable behavior that's harder to test, especially for folks who are using Workbox in a more modular way, or anyone building a third-party plugin.

Anyway, I don't want to block this feature being added, especially if we know folks are running into quota issues today. And you've obviously thought about the trade-offs, so I'm happy with whatever you think is best.

But I think we should revisit it later. I also think it's probably worth adding a `catchWriteDidFail` lifecycle method at some point regardless, so strategy plugin authors can handle these cases individually without potentially affecting plugins they don't own.

packages/workbox-core/_private/cacheWrapper.mjs
```diff
@@ -14,12 +14,14 @@
...
 
-import {logger} from './logger.mjs';
-import {assert} from './assert.mjs';
-import {WorkboxError} from './WorkboxError.mjs';
-import {getFriendlyURL} from '../_private/getFriendlyURL.mjs';
 import pluginEvents from '../models/pluginEvents.mjs';
 import pluginUtils from '../utils/pluginUtils.mjs';
+import {WorkboxError} from './WorkboxError.mjs';
+import {assert} from './assert.mjs';
+import {executeCallbacks} from './quota.mjs';
```
philipwalton: Can we also change this name to `executeQuotaErrorCallbacks`?

philipwalton:
Thanks Jeff! I left a few random thoughts for the future and one more comment, but overall everything LGTM.

packages/workbox-cache-expiration/Plugin.mjs
jeffposnick: I filed https://github.com/GoogleChrome/workbox/issues/1530 to capture that. To address the immediate quota issues, I'd like to go with a global approach.

tgangso:
When can we expect this to be released?
Thanks

PR #1510, opened 2018-06-01
Fix CacheExpiration / IndexedDB OpenCursor for Edge

### LINKED ISSUE #1509: Cache Expiration Logic not running in Edge due to exceptions with running 'openCursor'
**Library Affected**:
workbox-sw / workbox-core (I think this is actually from the IndexedDB Helper library used internally in workbox-core)

**Browser & Platform**:
Edge (latest)

Another browser compat bug with Edge.

In Workbox, cache timestamps are stored in IndexedDB.  When opening a cursor to iterate through the object store, one can pass optional parameters:
https://github.com/GoogleChrome/workbox/blob/118598366adfb22f3e74217ccd403713109dc277/packages/workbox-core/_private/DBWrapper.mjs#L200-L221

These optional parameters are in this block `opts.query` and `opts.direction`.

However, due to behavior in Edge, `openCursor(opts.query, opts.direction)` actually throw `DOMException: DataError`. because `opts.query` and `opts.direction` are `undefined`.

In workbox, what ends up happening is that the exception is thrown, but never handled, so the deletion logic never runs and it waits indefinitely for a `Promise` to never resolve.

I have a repro here for the cursor bug specifically:
[EdgeIndexedDBCursorUndefinedNonstandardBehavior.zip](https://github.com/GoogleChrome/workbox/files/2060114/EdgeIndexedDBCursorUndefinedNonstandardBehavior.zip)

I will also prepare a workbox specific repro app that shows the expiration logic never getting run for Edge.

This one I am not 100% sure on the spec details, I'll have to do some reading, but regardless, I think workbox will need to handle this case to expire cache entries properly on Edge.

josephliccini:

R: @jeffposnick @philipwalton

Fixes #1509

Edge Browser does not support `undefined` arguments being passed into `openCursor(...)`

I handle each case for the arguments to `openCursor(...)`

See https://developer.mozilla.org/en-US/docs/Web/API/IDBObjectStore/openCursor for more information on the API itself.

This will fix the CacheExpiration logic in Edge Browsers that depend on `openCursor(...)` in the expiration entry lookup process.

### DIFF
--- a/packages/workbox-core/_private/DBWrapper.mjs
+++ b/packages/workbox-core/_private/DBWrapper.mjs
@@ -203,7 +203,12 @@ class DBWrapper {
       const target = opts.index ? store.index(opts.index) : store;
       const results = [];
 
-      target.openCursor(opts.query, opts.direction).onsuccess = (evt) => {
+      // Passing `undefined` arguments to Edge's `openCursor(...)` causes
+      // 'DOMException: DataError'
+      // Details in issue: https://github.com/GoogleChrome/workbox/issues/1509
+      const query = opts.query || null;
+      const direction = opts.direction || 'next';
+      target.openCursor(query, direction).onsuccess = (evt) => {
         const cursor = evt.target.result;
         if (cursor) {
           const {primaryKey, key, value} = cursor;


josephliccini:
Things seem to be working in browsers, but the Mock library used doesn't like null.  I filed an issue on them, and am also discussing with Edge best approach here with `undefined`s

philipwalton:
I'm fine with these changes. In fact, the version of the DBWrapper code that we were discussing releasing separately has these exact changes (since I noticed this was failing in IE), and I was planning to upstream them.

As for the mock library, did that issue get addressed? Can you link to it? I think we're already using a fork of that anyway, so we could potentially fix it in the fork if there's no response from the maintainer.

josephliccini:
Hi @philipwalton !

Thanks for taking a look at this!

Here's the mock library issue: https://github.com/dhoulb/shelving-mock-indexeddb/issues/6

I can submit a CR to them or to your fork with the fix.

philipwalton:
If you submit a PR to https://github.com/philipwalton/shelving-mock-indexeddb I'll merge it while we wait on the master repo.

josephliccini:
@philipwalton here it is: https://github.com/philipwalton/shelving-mock-indexeddb/pull/2

Thanks!

philipwalton:
Thanks @josephliccini. The mock library has been updated, so these tests should hopefully pass now. Can you make sure to sign the CLA (as described in https://github.com/GoogleChrome/workbox/pull/1510#issuecomment-393719477) so we can merge this?

josephliccini:
Awesome, will do! Just checking internally if there's any extra process for CLA I need to be aware of.

josephliccini:
Just working through the internal process here, hopefully it won't take too much longer.  Thanks for being patient

josephliccini:
I signed it!

josephliccini:
Turns out it's OK for me to just sign the individual CLA. (I double checked, it took some time).

Should be good to go!

jeffposnick:
Thanks again, @josephliccini! We'll get this into the next Workbox release.

PR #1533, opened 2018-06-15
Use Comlink to fix Safari integration tests

### LINKED ISSUE #1429: Safari integration tests unhappy
**Library Affected**:
all

**Browser & Platform**:
Safari TP Release 53 (Safari 11.2, WebKit 13606.1.11.2)

All of the `gulp test-integration` tests are currently failing for Safari when run against the `prod` bundles, with an error like:

```sh
1) [workbox-background-sync] Load and use Background Sync
       should load a page with service worker:
     Error: the string "Regstration is not installing" was thrown, throw an Error :)
      at <anonymous>
      at process._tickDomainCallback (internal/process/next_tick.js:228:7)
```

The same tests run against the `dev` bundles complete successfully.

That originates from https://github.com/GoogleChrome/workbox/blob/685e81673d5c190209137dcaf6871f9fd5d98ce5/infra/testing/activate-sw-safari.js#L3-L8

jeffposnick:
R: @philipwalton
CC: @surma 

Fixes #1429

The latest Safari build won't let us access the Cache Storage API from the `window` context when run in Web Driver.

To work around this, I've introduced [Comlink](https://github.com/GoogleChromeLabs/comlink) and the following approach to proxying:

- Our node test harness calls methods using Web Driver the `window` context via the code in `testing/comlink/node-interface.js`.
- Our `window` context calls methods in the service work context using Comlink, via the code in`testing/comlink/window-interface.js`.
- Our service workers need to `importScripts()` our `testing/comlink/sw-interface.js` code, and then they'll be able to respond to the `window` requests.

Setting that (convoluted) system up is the bulk of the changes in this PR.

In the course of ensuring that all our integration tests pass in all browsers (on my Mac), I found a few issues with the existing test suite:

- The `maxAgeSeconds` test of the `workbox-cache-expiration` suite was just incorrect. It loaded the wrong service worker file and tested the wrong logic. I've fixed that.
- Only 2 of the 5 `workbox-strategies` tests were being run, due to the use of `describe.only()`. I've removed that, and made sure that all 5 pass in all browsers.
- Our `activate-and-control.js` logic for non-Safari browsers had a bit of a race condition that I could trigger in Firefox. It was previously fixed in the Safari variant, and we should probably look to see if we can just stick with a single `activate-and-control.js` in the future, so that we don't end up needing to fix things in two places.


infra/testing/comlink/sw-interface.js
```diff
@@ -0,0 +1,50 @@
...
+        resolve(false);
+      };
+      result.onsuccess = (event) => {
+        event.target.result.close();
+        resolve(true);
+      };
+    });
+  },
+
+  cacheContents: async (cacheName) => {
```
philipwalton: Would it make more sense to call this `cacheUrls` or `cacheRequestUrls` since it returns the request URL rather than the request body?

jeffposnick: Done.

PR #1560, opened 2018-06-28
New workbox-navigation-preload module

### LINKED ISSUE #835: Support navigationPreload in WorkboxSW
**Library Affected**:
workbox-sw

**Browser & Platform**:
Google Chrome v62 and above

Navigation preload for service workers is going to be enabled by default in Chrome 62 stable. I think workbox should provide support for enabling it through WorkboxSW class API. Thoughts?

Workbox runtime strategies that should use preloaded responses. 
CacheFirst (if not in cache)
NetworkFirst
NetworkOnly
StaleWhileRevalidate

jeffposnick:
R: @philipwalton

Fixes #835 

This PR includes:

- `workbox.navigationPreload.isSupported()`, to check whether or not the current SW environment supports it.
- `workbox.navigationPreload.enable(headerValue)`, which will automatically register an `activate` handler that will enable navigation preload (or no-op if it's not supported), using the optional `headerValue` provided for `Service-Worker-Navigation-Preload`.
- Changes to the `workbox.core.fetchWrapper` and the various built-in strategies that make use of it, so that it will detect when there's a preload response and use it whenever possible, without having to opt-in.

This should be backwards-compatible, so it could make it into `v3.4.0`.

I'll write up a module guide for https://developers.google.com/web/tools/workbox/modules/ around the same time this gets released.


packages/workbox-navigation-preload/index.mjs
```diff
@@ -0,0 +1,23 @@
...
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+
+/**
+ * @namespace workbox.rangeRequests
```
philipwalton: Is this a typo?

jeffposnick: Yup. Thanks.

PR #1566, opened 2018-07-03
Test server refactor

jeffposnick:
R: @philipwalton

This isn't a high-priority PR, but I wanted to take the chance to clean up some of the cruft that's been building up in our integration test server + test cases.

Here's a highlight of the changes:

- The code for the server has been broken out into smaller modules, making it easier to reason about what's going on.

- Instead of using a single, global request counter in our server, I added a new `RequestCounter` class which encapsulates all the request state. Our integration tests can obtain a `RequestCounter`  instance from our global server. Removing that reliance on global state opens the door to trying to run some integration tests in parallel in the future.

- I've normalized the "special" URL namespace as `/__WORKBOX/`. Everything under that is a URL that's handled by an Express route.

- Updates to the existing integration tests which should be functionally equivalent, but which accounts for all those changes.

In the process, I looked into moving off of Express and onto [koa](https://koajs.com/), but... it's pretty gnarly.


jeffposnick:
Ping, if you've got some time, @philipwalton. Should be pretty straightforward to review.

PR #1598, opened 2018-08-07
Fix empty object when adding runtimeCaching custom plugins with workbox-webpack-plugin

### LINKED ISSUE #1597: [BUG] Empty object when adding runtimeCaching custom plugins with workbox-webpack-plugin
**Library Affected**:
*workbox-webpack-plugin and workbox-build*

**Browser & Platform**:
not browser-related

When adding custom plugins to runtimeCaching option on workbox-webpack-plugin, the output sw comes with an empty object instead of the object with functions to be executed.

I think this is happening because of [this line](https://github.com/GoogleChrome/workbox/blob/096e44897e9c348f85bb51927c1113de4e339897/packages/workbox-build/src/lib/runtime-caching-converter.js#L34) once JSON.stringify removes functions. I've managed to fix the issue on a fork with [this commit](https://github.com/tsirlucas/workbox/commit/aab8affb3fb919826fdf7b32702c168661a6f40d) parsing the object to string instead (I think it's the desired behavior and it works).

My solution doesn't parse comments and once the populate-sw-template puts the plugin content inline, single-line comments can break everything. I can work on a solution for this either but I'd like to know if you guys would accept a PR and what would be the desired behavior when dealing with comments. Should them be removed or transformed into to multi-line comments? 

**Code to reproduce:**

```
new GenerateSW({
    swDest: 'sw.js',
    clientsClaim: true,
    skipWaiting: true,
    importWorkboxFrom: 'local',
    navigateFallback: '/',
    runtimeCaching: [
      {
        urlPattern: /^https:\/\/api-random\.com\/random\//,
        handler: 'cacheFirst',
        options: {
          cacheName: 'random-cache',
          expiration: {
            maxEntries: 40,
            maxAgeSeconds: 200000,
          },
          cacheableResponse: {statuses: [0, 200, 201, 206, 301, 304, 302]},
          plugins: [{
            cacheWillUpdate: ({request, response}) => {
              return caches.match(request.url, {ignoreSearch: true}).then((res) => {
                /* ignore repeated songs */
                if (res) {
                  return null;
                }
          
                /* cache only save=true param */
                if (request.url.includes('?save=true')) {
                  return response;
                }
          
                /* ignore any other thing */
                return null;
              });
            },
            cachedResponseWillBeUsed: ({request, cachedResponse}) => {
              if (cachedResponse) {
                return cachedResponse;
              }
              /* this will match same url/diff query string where the original failed */
              return caches.match(request.url, {ignoreSearch: true});
            },
          }],
        },
      },
    ],
  });

```

tsirlucas:
R: @jeffposnick @philipwalton

Fixes #1597

As discussed on the issue, I've used [this library](https://github.com/yeoman/stringify-object) by Yeoman instead of JSON.stringify because it doesn't handle functions, resulting in an empty object when adding custom plugins with workbox-webpack-plugin. Also, I've made some changes and now I'm using a regex based on [this one](http://blog.moagrius.com/javascript/javascript-regexp-to-remove-comments/) to remove comments once the object is parsed as a string.

Also added some test cases.

### DIFF
--- a/packages/workbox-build/package-lock.json
+++ b/packages/workbox-build/package-lock.json
@@ -4,6 +4,28 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "babel-extract-comments": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/babel-extract-comments/-/babel-extract-comments-1.0.0.tgz",
+      "integrity": "sha512-qWWzi4TlddohA91bFwgt6zO/J0X+io7Qp184Fw0m2JYRSTZnJbFR8+07KmzudHCZgOiKRCrjhylwv9Xd8gfhVQ==",
+      "requires": {
+        "babylon": "6.18.0"
+      }
+    },
+    "babel-plugin-syntax-object-rest-spread": {
+      "version": "6.13.0",
+      "resolved": "https://registry.npmjs.org/babel-plugin-syntax-object-rest-spread/-/babel-plugin-syntax-object-rest-spread-6.13.0.tgz",
+      "integrity": "sha1-/WU28rzhODb/o6VFjEkDpZe7O/U="
+    },
+    "babel-plugin-transform-object-rest-spread": {
+      "version": "6.26.0",
+      "resolved": "https://registry.npmjs.org/babel-plugin-transform-object-rest-spread/-/babel-plugin-transform-object-rest-spread-6.26.0.tgz",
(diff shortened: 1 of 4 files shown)


tsirlucas:
I signed it!

tsirlucas:
Hey @jeffposnick just got back to this and got some cases that the parse regex wouldn't work correctly like when parsing a function with url strings into it. I'm working on it and will make a push to the branch ASAP.

PR #1610, opened 2018-08-22
Support for offlineGoogleAnalytics config in workbox-build config

### LINKED ISSUE #1568: Option to use workbox.googleAnalytics in workbox-cli
**Library Affected**:
workbox-cli

For now, I think I could just tack onto the generated sw.js, but it would be nice to have this case handled via the CLI.

jeffposnick:
R: @philipwalton
CC: @Hoten

Fixes #1568

The name of the new boolean config property (defaulting to `false`) is `offlineGoogleAnalytics`, and it's being added to `generateSW` and `generateSWString` modes of `workbox-build`. It could also be used in the wrappers around `workbox-build`, like `workbox-cli` and `workbox-webpack-plugin`.

A custom config would still require either using `importScripts` to pull in a separate file while in `generateSW` mode, or switching to `injectManifest` mode. But for the basic functionality, where all you want is

```js
workbox.googleAnalytics.initialize();
```

this should be sufficient.

Assuming this PR looks good, we'll need to get https://developers.google.com/web/tools/workbox/modules/workbox-build#full_generatesw_config updated once we release this, as part of what will presumably be Workbox `v3.5.0`.


philipwalton:
How about you can pass `true` to include it, or (optionally) you can pass an object or a function, which gets used as the argument to `workbox.googleAnalytics.initialize()`?

jeffposnick:
It's historically been messy when converting from an `workbox-build` option that might have non-primitive values (like options set to a `function`) into the equivalent string representation.

https://github.com/GoogleChrome/workbox/pull/1598 just took care of that for one case, when folks might end up setting `runtimeCaching.options.plugins` to a `function` value. I guess we could follow that model and support usage like:

```js
offlineGoogleAnalytics: {
  hitFilter: (params) => {
    const queueTimeInSeconds = Math.round(params.get('qt') / 1000);
    params.set('cm1', queueTimeInSeconds);
  },
},
```

Do you feel like enough folks end up needing custom configs that the (slight) extra complexity is worthwhile? Might as well?

philipwalton:
I think it's worth it, if it's possible.

When not using any customization options, users have no way of knowing if their offline analytics are working, so I'd like to encourage more usage of the customization options wherever possible.

jeffposnick:
Okay, I've used the technique from #1598 and passing in a full range of config options, including functions, seems to work.

PR #1638, opened 2018-09-17
Add method option to GenerateSW validation

### LINKED ISSUE #1636: GenerateSW does not support method option in runtimeCaching
**Library Affected**:
*workbox-build, workbox-webpack-plugin*

**Browser & Platform**:
*Google Chrome Version 69.0.3497.92 (Official Build) (64-bit)*

I would like to use POST method in *runtimeCaching* options from WorkboxPlugin.GenerateSW({ }) plugin instead of GET.

It is not possible at the moment to change the default method ; it leads to a validationError

`ValidationError: child "runtimeCaching" fails because ["runtimeCaching" at position 0 fails because ["method" is not a supported parameter.]]`.

It should be possible to change default method, I can make a PR if you think this is relevant.

chrisdns:
R: @jeffposnick @philipwalton

Fixes #1636

### DIFF
--- a/packages/workbox-build/src/entry-points/options/common-generate-schema.js
+++ b/packages/workbox-build/src/entry-points/options/common-generate-schema.js
@@ -32,6 +32,7 @@ module.exports = baseSchema.keys({
   offlineGoogleAnalytics: joi.alternatives().try(joi.boolean(), joi.object())
     .default(defaults.offlineGoogleAnalytics),
   runtimeCaching: joi.array().items(joi.object().keys({
+    method: joi.string(),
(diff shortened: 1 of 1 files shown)


chrisdns:
I signed it :+1:

jeffposnick:
Would you mind adding in validation while we're at it, to match

https://github.com/GoogleChrome/workbox/blob/685fd1b383e671c73b00fa0ba4d44364c0201de1/packages/workbox-routing/utils/constants.mjs#L35-L42

It could be something like

```js
method: joi.string().valid(
  'DELETE',
  'GET',
  '...',
),
```

PR #1643, opened 2018-09-18
Import registerQuotaErrorCallback from core instead of core._private (#1642)

### LINKED ISSUE #1642: Uncaught TypeError: quota_mjs.registerQuotaErrorCallback is not a function
**Library Affected**:
*workbox-cache-expiration v3.5.0*

**Browser & Platform**:
*"all browsers"*

**Errors when using workbox.expiration.Plugin  (local npm builds) with `purgeOnQuotaError: true`**:

workbox expiration-issue

Seems to me like this error is related to [this ](https://github.com/GoogleChrome/workbox/commit/9a64fa3b7f91506137be0521dbd27b0dd8a7e93d) commit where we're exposing `registerQuotaErrorCallback` as a direct export of `core` instead of `core._private` but this change needed to also update the other packages which depend on this (e.g. chache-expiration) to import from `core` and not `core._private`

salmoro:
R: @jeffposnick @philipwalton

Fixes #1642


salmoro:
I signed it!

PR #1652, opened 2018-09-19
Update metadata for v3.6.1

philipwalton:
R: @jeffposnick


gulp-tasks/publish.js
```diff
@@ -36,7 +36,7 @@ gulp.task('publish-assets', gulp.series([
 
 gulp.task('publish', gulp.series([
   'publish:signin',
-  'test',
+  // 'test',
```
jeffposnick: We don't want to commit this 😄

PR #1657, opened 2018-09-23
Consistently reject with a WorkboxError when a strategy fails

### LINKED ISSUE #1551: networkFirst and networkOnly behave inconsistently when extending
**Library Affected**:
workbox-sw

**Browser & Platform**:
Google Chrome v67

I'm not sure if I'm doing it right, but when I try to extend a workbox strategy, networkFirst and networkOnly behave inconsistently. E.g.:

```
const customHandler = workbox.strategies.networkOnly();
// const customHandler = workbox.strategies.networkOnly();

function fallbackStrategy({event, url}) {
  return customHandler.handle({event})
    .then((response) => {
      return response || new Response('This gets called if networkFirst, no response in cache and offline');
    })
    .catch(() => {
      return new Response('This gets called if networkOnly, no response in cache and offline');
    })
}
  'my-route',
  fallbackStrategy
);
```

With NetworkFirst, if no network nor cache, I get an empty response in "then". With NetworkOnly, I get an error and handle it in catch. 

Is it intentional?

jeffposnick:
R: @philipwalton
CC: @devnook @sorendam

Fixes #1551, to be merged into the `next` branch.

Previously, some strategies would resolve with `null` instead of rejecting.

There are some updates to the docs that will end up accompanying this (once `v4` goes live). For one thing, it makes it easier to provide fallback content via `workbox.routing.setCatchHandler()`.


packages/workbox-strategies/CacheOnly.mjs
```diff
@@ -123,7 +127,10 @@ class CacheOnly {
       logger.groupEnd();
     }
 
-    return response;
+    if (response) {
```
philipwalton: Nit: to me it'd make slightly more sense to invert this condition, i.e. throw when there's not a response and default to returning the response. As it is now, it looks like throwing is the default, unconditional behavior (but maybe that's just how I read it).

packages/workbox-strategies/NetworkFirst.mjs
```diff
@@ -175,7 +179,10 @@ class NetworkFirst {
       logger.groupEnd();
     }
 
-    return response;
+    if (response) {
```
philipwalton: Same comment as above, here and presumably elsewhere.

jeffposnick:
Updated with review feedback.

Issue #1656, opened 2018-09-21
Update all package.json dependencies


jeffposnick:
In general, we should take the v4 release as an opportunity to do a sweep through our various `package.json`s and update all `dependencies` and `devDependencies` to the latest releases (unless there are one-off reasons not to update).

The PR that resolves this can probably replace https://github.com/GoogleChrome/workbox/pull/1609, which just tracks a subset of the dependencies.

abraham:
What about using something like @renovatebot to automatically open update PRs?

jeffposnick:
Sure; see https://github.com/GoogleChrome/workbox/issues/1696

Issue #1655, opened 2018-09-21
Update browser transpilation target


jeffposnick:
We currently use `@babel/preset-env` to transpile down to Chrome 51:

https://github.com/GoogleChrome/workbox/blob/361246146e5ebe38e2376ff32403dcd9e0934a02/gulp-tasks/utils/rollup-helper.js#L19-L22

This target was chosen to accommodate Samsung Internet 5.x.

As per https://developer.samsung.com/internet/android/releases, there have been two stable releases of Samsung Internet 6.x based on Chromium 56 (along with a beta 7.x release based on an even newer Chromium), so I think we can safely switch that to be the new transpilation target for Workbox v4.

CC: @poshaughnessy and @AdaRoseCannon from the Samsung Internet team as a heads-up, to see if there are any concerns about shifting to Chrome 56 as the transpilation target—I'm not sure how much of a market share Samsung Internet 5.x still has.

PR #1658, opened 2018-09-23
Updates to latest dependencies, and transpilation targets

### LINKED ISSUE #1654: Update node transpilation target
Now that [node v4 is officially deprecated](https://github.com/nodejs/Release#release-schedule), we should change our `@babel/preset-env` target to node v6:

https://github.com/GoogleChrome/workbox/blob/361246146e5ebe38e2376ff32403dcd9e0934a02/gulp-tasks/utils/build-node-package.js#L25-L26

We should also update `package.json` metadata accordingly:

https://github.com/GoogleChrome/workbox/blob/361246146e5ebe38e2376ff32403dcd9e0934a02/packages/workbox-build/package.json#L14-L16

https://github.com/GoogleChrome/workbox/blob/361246146e5ebe38e2376ff32403dcd9e0934a02/packages/workbox-cli/package.json#L20-L22

https://github.com/GoogleChrome/workbox/blob/361246146e5ebe38e2376ff32403dcd9e0934a02/packages/workbox-webpack-plugin/package.json#L27-L29

jeffposnick:
R: @philipwalton

Fixes #1654, #1655, #1656, to be merged into the `next` branch.

This has a whole host of updates to the various `dependencies` and `devDependencies`, bumping up most of them to the latest releases across all of our `package.json`.

Along with that, I've updated the transpilation targets to node 6 (which is required by some of the updated dependencies) and Chrome 56. (If it turns out that we need to stick with Chrome 51, we can change that back.)

There are a lot of whitespace changes in this PR, as the updated `eslint` config was not happy with some of our previous formatting.

The `webpack` tests are currently failing, as I need to update a whole bunch of those constants in order to get them to pass. They need to be updated whenever any of the `webpack`-related dependencies change, and I'm going to hold off on remediating everything (or scripting an update) until I make sure that we don't have any other updates.


package.json
```diff
@@ -46,68 +46,65 @@
...
-    "rollup-plugin-uglify": "^4.0.0",
+    "rollup-plugin-terser": "^3.0.0",
     "rollup-stream": "^1.24.1",
     "selenium-assistant": "^5.3.0",
-    "semver": "^5.4.1",
+    "semver": "^5.5.1",
     "serve-index": "^1.9.1",
-    "service-worker-mock": "^1.7.2",
+    "service-worker-mock": "^1.9.3",
     "shelving-mock-indexeddb": "github:philipwalton/shelving-mock-indexeddb#d2c5e52b6f9903026f5bac31cbe758cbebd98661",
```
jeffposnick: @philipwalton as far as I can tell, the change in your fork never made it to the upstream project, so we're still pinned to this.

package.json
```diff
@@ -46,68 +46,65 @@
...
   "homepage": "https://github.com/GoogleChrome/workbox#readme",
   "devDependencies": {
-    "@google-cloud/storage": "^1.7.0",
-    "@octokit/rest": "^15.8.1",
+    "@babel/core": "^7.1.0",
+    "@babel/plugin-transform-runtime": "^7.1.0",
+    "@babel/preset-env": "^7.1.0",
+    "@google-cloud/storage": "^2.0.3",
+    "@octokit/rest": "^15.12.0",
     "@std/esm": "^0.26.0",
```
jeffposnick: I tried moving to the `esm` project (which replaced `@std/esm`), but there are a lot of breaking changes and I couldn't get our unit test runner happy. So we're still stuck on this older version until we can investigate further.

.eslintrc.js
```diff
@@ -20,13 +20,14 @@ module.exports = {
   rules: {
     "jsdoc/check-types": 2,
     "jsdoc/newline-after-description": 2,
+    'indent': ['error', 2],
```
philipwalton: What's this being added for?

philipwalton: Oh, now that I'm looking at the rest of the changes, I see what you're doing. But technically this isn't consistent with Google style because in some cases Google style recommends indenting 4 spaces (e.g. line continuations and stuff like that).

The google eslint package v0.10.0 adds an indentation rule more inline with what's recommended by Google JS style. It's probably worth just upgrading to that.

gulp-tasks/utils/build-node-package.js
```diff
@@ -32,7 +32,7 @@ module.exports = (packagePath) => {
       // are only included in our Rollup bundles once, even if they're used
       // in multiple source files.
       // See https://github.com/rollup/rollup-plugin-babel#helpers
-      'transform-runtime',
+      '@babel/transform-runtime',
```
philipwalton: BTW, if we're getting rid of the regenerator runtime polyfill, we may not need this at all anymore.

jeffposnick: We still need it for the node libraries, since we're targeting node 6, and that doesn't support `async`/`await`.

jeffposnick:
I've merged in the `3.6.2` changes to this PR, just so we don't drift too far from `master`.

.eslintrc.js
jeffposnick: As discussed in person, there are going to be a lot of additional files that need whitespace adjustment if we don't include this.

We're already using `"eslint-config-google": "^0.10.0"` as part of this PR.

Let's do separate PR to remove this override and get all the whitespace consistent rather than including any more changes in this PR.

philipwalton: SGTM

package.json
philipwalton: Yeah, I'll file an issue to upstream that.

gulp-tasks/utils/rollup-helper.js
```diff
@@ -14,27 +13,20 @@ module.exports = {
...
-      externalHelpers: true,
     };
     plugins.push(babel(babelConfig));
 
     let minifyBuild = buildType === constants.BUILD_TYPES.prod;
     if (minifyBuild) {
-      const uglifyOptions = {
+      const terserOptions = {
         mangle: {
           properties: {
```
philipwalton: I discovered recently that we probably need to enable the option to make property mangling consistent across files (when we're outputting multiple files). I can file an issue to do that in a separate PR.

PR #1667, opened 2018-09-28
Fixes workbox dependencies

arcanis:
cc @jeffposnick @philipwalton (didn't open an issue since both the problem and fix are relatively straightforwards)

The `workbox-webpack-plugin` is trying to access the `babel-runtime` package, which isn't declared in its dependencies. As a result package managers offer no guarantee as to which version will be provided, or even if `babel-runtime` will be installed at all.

This diff simply adds the dependency that's missing, based on what I found in the `package-lock.json` at the root of the repository.

### DIFF
--- a/packages/workbox-webpack-plugin/package-lock.json
+++ b/packages/workbox-webpack-plugin/package-lock.json
@@ -1,7 +1,23 @@
 {
-	"requires": true,
+	"name": "workbox-webpack-plugin",
+	"version": "3.6.1",
 	"lockfileVersion": 1,
+	"requires": true,
 	"dependencies": {
+		"babel-runtime": {
+			"version": "6.26.0",
+			"resolved": "https://registry.npmjs.org/babel-runtime/-/babel-runtime-6.26.0.tgz",
+			"integrity": "sha1-llxwWGaOgrVde/4E/yM3vItWR/4=",
+			"requires": {
+				"core-js": "^2.4.0",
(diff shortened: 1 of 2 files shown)


jeffposnick:
Thanks—this is something that I caught when doing a dependency update sweep in preparation for our v4 release: https://github.com/GoogleChrome/workbox/pull/1658/files#diff-d9674ca55441128e3145e9552ce61a71R31

I believe that we were getting by because `workbox-build` was pulling in the `babel-runtime` package implicitly, but it's obviously preferable to have it explicitly listed, as it should be in that PR.

My personal preference would be to just defer to that upcoming change in v4 unless this is something that is actively causing a failure for you with the current v3 release.

arcanis:
I noticed it while working on https://github.com/facebook/create-react-app/pull/5136, which adds support for Plug'n'Play to create-react-app. Since package boundaries are enforced under this environment, it throws an error:

screen shot 2018-09-28 at 3 58 15 pm

jeffposnick:
Gotcha—Okay, we can merge this in for a `3.6.2` release given that.

jeffposnick:
Thanks—I was just going to ask about the `depcheck` 😄

jeffposnick:
`3.6.2` is out with the fix: https://github.com/GoogleChrome/workbox/releases/tag/v3.6.2

Thanks again!

PR #1676, opened 2018-10-04
Fix for the workbox-webpack-plugin tests

jeffposnick:
R: @philipwalton

This needs to be done when we change the `webpack` dependency versions, as different webpack versions tend to spit out different hashes for the same source content.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ node_js:
 
 # Read more here: https://docs.travis-ci.com/user/gui-and-headless-browsers/#Using-xvfb-to-Run-Tests-That-Require-a-GUI
 before_script:
+  - npm install -g gulp-cli
   - "export DISPLAY=:99.0"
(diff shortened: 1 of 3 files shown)


jeffposnick:
Hrm... the current failure is just in Travis (not AppVeyor) and is:

```
Error: Cannot find module 'workbox-background-sync/build/workbox-background-sync.prod.js'
      at /home/travis/build/GoogleChrome/workbox/node_modules/@std/esm/esm.js:1:10073
      at Proxy.<anonymous> (node_modules/@std/esm/esm.js:1:10083)
      at Module.r (node_modules/@std/esm/esm.js:1:148879)
      at Proxy.O (node_modules/@std/esm/esm.js:1:148643)
      at module.exports (packages/workbox-build/src/lib/copy-workbox-libraries.js:39:675)
```

That's really weird. Makes me think there might be some sort of timing issue between the build process and when the tests start. I'll look into it as part of this PR.

Issue #1677, opened 2018-10-04
301 redirect causes index.html to be stored as text/plain


kushtrimjunuzi:
Hi,

Library:
workbox-v3.5.0/workbox-sw.js

Browser:
Google Chrome: Version 69.0.3497.100 (Official Build) (64-bit)

OS:
Windows 10

I wanted to report this issue we are having recently, the issue is when a https://domain.com/index.html is requested, web server (haproxy or Caddy or any other web server) issue 301 redirect to https://domain.com, below are responses:

GET Request: https://domain.com/index.html 
Response:
HTTP/1.1 301 Moved Permanently
Date: Thu, 04 Oct 2018 18:57:49 GMT
Content-Type: text/plain; charset=utf-8
Content-Length: 0
Connection: keep-alive
Location: ./

Response after redirected to https://domain.com:
HTTP/1.1 200 OK
Date: Thu, 04 Oct 2018 18:57:49 GMT
Content-Type: text/html; charset=utf-8
Content-Length: 1283
Connection: keep-alive
Accept-Ranges: bytes
Last-Modified: Wed, 03 Oct 2018 20:12:18 GMT

as you can see first response content-type is text/plain and second response is text/html, when this gets fetched and stored into cache store workbox-precache it is stored as content-type text/plain and the content is index.html document.

![image](https://user-images.githubusercontent.com/245973/46496610-7be7e280-c7e6-11e8-955b-b92f51d733a7.png)
 
The issue occur if you do refresh F5 you get text/plain of index.html so not as html but as plain text, I'm not able to reproduce in my browser, other users are able to reproduce this issue.

Thanks,

jeffposnick:
Hmmm, I wonder if that's related to https://github.com/v8/v8.dev/issues/4

I'm looking at that redirect handling logic again and I see a bug in the code; I'm not 100% sure that fixing the bug will resolve these issues, but I'll experiment a bit.

See https://github.com/GoogleChrome/workbox/pull/1678

kushtrimjunuzi:
I tested and it is fixed, I think this is going to be critical bug for everybody who is using this library, why not release as v3.6.3?

kushtrimjunuzi:
I have merged into v3.6.2 this PR, if anybody interested can go to my fork and download the compiled workbox-* libraries. https://github.com/kushtrimjunuzi/workbox/releases/tag/v3.6.3-fork

jeffposnick:
[...]

PR #1678, opened 2018-10-04
Fix cleanRedirect logic

jeffposnick:
R: @philipwalton

While investigating https://github.com/GoogleChrome/workbox/issues/1677 I realized that there's a bug in the `Request` constructor usage inside the code that "cleans up" redirected responses.

The code in this PR just takes the [corresponding code](https://github.com/GoogleChromeLabs/sw-precache/blob/e5892c9a3cce1a8d4f6de2aa768287cb709ba1e7/lib/functions.js#L101-L105) from `sw-precache`'s implementation, which should work.

I'm filing this PR against `next` because I don't think it's an *urgent* bug to fix, and I'd rather we don't cut any additional v3 releases if we can help it.

### DIFF
--- a/infra/testing/sw-env-mocks/Headers.js
+++ b/infra/testing/sw-env-mocks/Headers.js
@@ -17,8 +17,12 @@
 // Stub missing/broken Headers API methods in `service-worker-mock`.
 // https://fetch.spec.whatwg.org/#headers-class
 class Headers {
-  constructor(obj = {}) {
-    this.obj = Object.assign({}, obj);
+  constructor(init = {}) {
+    if (init instanceof Headers) {
+      this.obj = Object.assign({}, init.obj);
+    } else {
+      this.obj = Object.assign({}, init);
+    }
   }
 
   has(key) {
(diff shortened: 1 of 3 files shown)


infra/testing/sw-env-mocks/Headers.js
```diff
@@ -17,8 +17,12 @@
 // Stub missing/broken Headers API methods in `service-worker-mock`.
 // https://fetch.spec.whatwg.org/#headers-class
 class Headers {
-  constructor(obj = {}) {
-    this.obj = Object.assign({}, obj);
+  constructor(init = {}) {
```
jeffposnick: We've diverged from `service-worker-mock` over the years, but this change is roughly https://github.com/pinterest/service-workers/blob/c0eda82a98aa46d9a6fb44a32906fa0c2a4f2b22/packages/service-worker-mock/models/Headers.js#L4-L12

jeffposnick:
> I'm filing this PR against `next` because I don't think it's an urgent bug to fix, and I'd rather we don't cut any additional v3 releases if we can help it.

PR #1680, opened 2018-10-04
Whitespace changes for eslint

jeffposnick:
R: @philipwalton

These are a bunch of whitespace-related changes in the `next` branch to adhere to the latest `eslint-config-google` rules, without having to use the `'indent': ['error', 2],` override.


jeffposnick:
You can confirm this is (almost entirely) whitespace-related via https://github.com/GoogleChrome/workbox/pull/1680/files?utf8=%E2%9C%93&diff=split&w=1

demos/functions/index.js
```diff
@@ -28,7 +28,7 @@ app.set('view engine', 'hbs');
 app.get('/', function(req, res) {
   let moduleData = workboxModules.map((moduleName) => {
     const docPath = path.join(
-      __dirname, 'views', 'demo', `${moduleName}.hbs`
+        __dirname, 'views', 'demo', `${moduleName}.hbs`
```
philipwalton: Idiomatic Google JS style wouldn't format the following like this:

```js
const docPath = path.join(
    __dirname, 'views', 'demo', `${moduleName}.hbs`
);
```

It'd do it like this:

```js
const docPath = path.join(
    __dirname, 'views', 'demo', `${moduleName}.hbs`);
```

The 4-space indentation is really intended to signify a line continuation, not for block indentation (which putting the closing parens on the next line makes this feel like).

That being said, I don't want to go though and fix all of these, so I'm fine with just doing what the autoformatting does for now and dealing with individual cases like this per-instance going forward.

PR #1682, opened 2018-10-06
Decouple Router~handleRequest from events

### LINKED ISSUE #1630: Make it possible to use router logic outside of a fetch handler
**Library Affected**:
*workbox-routing*

Similar to the issue brought up in #1395 by @jeffposnick, it's difficult to take advantage of `Router` functionality outside of a `FetchEvent` handler.

### Use case

The use case is having a list of URLs that you want to cache exactly as if they were real requests from the client. E.g. the list of URLs requested from the window (via the ResourceTiming API) prior to the service worker being activated.

If you've already written logic to match routes and strategies to URLs and put that logic in a router, it's unfortunate that you can't reuse any of it if all you have is a URL (or a `Request` instance).

### Possible solutions

When looking at the `Router#handleRequest()` method, it requires an event object, but AFAICT it does actually *need* an event object to work. It seems like the main reason it takes an event is to pass it off to the strategies, which may then need to call `event.waitUntil()`.

We could update the `handleRequest()` method to accept a url or a `Request` object (in addition to an event).

Alternatively, we could add an additional method that just returns the first registered route matching a request from a router (similar to what `_findHandlerAndParams()` does, but preferably in a way that doesn't require a dev to handle params in cases where they don't apply).

That way the developer could do:

```js
for (const url of urlListFromWindow) {
  const request = new Request(url);
  const route = router.matchRequest(request);
  
  await route.handler.makeRequest({request});
}
```

philipwalton:
R: @jeffposnick

Fixes #1630

This PR updates the routing and strategy packages in a few ways.

**Breaking changes:**

1. the `Router~handleRequest()` method from just taking an `Event` object, to taking an `options` object where `options.request` is required and `options.event` is optional. Note that technically this change is mostly backwards compatible with the previous version since fetch events have a `request` property. One thing to watch out for, though, is if people were using this method directly and passing an event, no strategies or handlers would be able to call `event.waitUntil()`. We *could* support both a `FetchEvent` argument as well an options `Object` argument -- or we could just consider this a breaking change and provide upgrade instructions.

2. The strategy classes' `handle()` methods no longer require an `options.event` parameter, and they now work essentially the same as their `makeRequest` methods. With this change, we don't really need two separate methods, so we may want to consider getting rid of one of them.

3. The `match` and `handler` callbacks are now also passed an `options.request` in addition to `options.url`, and optionally `options.event`. I'm not sure if anyone was ever using the `options.event` property, but since `Router~handleRequest()` can now be called without an event, `options.event` will now sometimes not be present and so is thus documented as an optional argument.

4. Probably more of a bug fix than a breaking change, but previous if a match callback returned a truthy string, that string would be passed as the `options.params` argument to the handler. I'm assuming this wasn't intentional, so I've now updated that logic to only pass `options.params` if the return value is a non-empty object or array.

**Non-breaking changes:**

1. I've slightly refactored the `Router~_findHandlerAndParams()` method and renamed it to `findMatchingRoute()`, meaning it's now a public method. I figure it may be useful for someone who wants to use the router without executing a handler right away.


packages/workbox-routing/NavigationRoute.mjs
```diff
@@ -79,14 +79,15 @@ class NavigationRoute extends Route {
    * Routes match handler.
    *
    * @param {Object} options
-   * @param {FetchEvent} options.event
    * @param {URL} options.url
+   * @param {Request} options.request
+   * @param {Event} [options.event]
```
jeffposnick: Remove `options.event` from the JSDoc.

packages/workbox-routing/Router.mjs
```diff
@@ -77,15 +80,10 @@ class Router {
       return;
     }
 
-    let route = null;
-    let handler = null;
-    let params = null;
-    let debugMessages = [];
+    let {params, route} = this.findMatchingRoute({url, request, event});
```
jeffposnick: Looks like all these `let`s can be changed to `const` after your refactoring.

philipwalton: `route` gets reassigned and `params` doesn't, but since one of them does, in order to use destructuring I had to do both as let...

PR #1683, opened 2018-10-06
Add methods to add fetch and cache listeners

philipwalton:
R: @jeffposnick

Adds `addFetchListener()` and `addCacheListener()` methods to `Router` in preparation for the `workbox-window` library (or any window code) that wants to send the router a list of URLs to cache.

I've updated the default router to automatically call both of these methods (previously it was running the code that became `addFetchListener()`).

I've also updated the public exports so the `DefaultRouter` class can be imported standalone -- in case developers who are building their own SW want to instantiate it directly.

Lastly, I've updated the event mocks to throw if `respondWith()` gets called twice on the same event (per spec), since I noticed my tests were doing that, and should have been failing.

**Note:** I've set this to be merged into the `routing-additions` branch so the diffs make more sense, but I think we should merge #1682 first, and then update the base branch here to `next` once the commits from #1682 are in.


jeffposnick:
👍

philipwalton:
Rebasing against `next` and will merge once the tests finish.

PR #1685, opened 2018-10-07
Change license to MIT

### LINKED ISSUE #1660: Switch Workbox over to use MIT license
We're planning on switching Workbox over from the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0) to the [MIT license](https://opensource.org/licenses/MIT) in the upcoming `v4.0.0` release.

philipwalton:
R: @jeffposnick 

Fixes #1660

In addition to updating the license, I standardized which files we're including the license headers on. Since we've historically mostly just been including headers on source files, I went ahead and removed them on files in `test`, `infra`, and `gulp-task` where it was present, just to be consistent.

/cc @willnorris


willnorris:
please keep the headers in test/**, infra/**, etc files.  Even if those files don't get included in the final deliverable, they're still source code all the same.  The goal is to ensure that the license governing the file is still clear, even if the file were separated from the rest of the repo (such as being copied and pasted into a different project).

philipwalton:
@jeffposnick it looks like adding license headers to the test files is breaking the webpack hashes. Do you still have your script to update those?

jeffposnick:
I'll commit a change to the `license` branch to clean up the tests.

philipwalton:
Thanks!

Arm08102525:
Good

PR #1688, opened 2018-10-07
Rename the workbox-google-analytics output file

### LINKED ISSUE #1646: Rename the google analytics script to avoid errors due to ad/content blockers
Forking the discussion from #1186.

We should consider renaming the built `workbox-google-analytics` script because apparently some ad/content blockers will block it, which results in an error, and thus the entire service worker install fails.

*(Note, this will probably require a breaking change.)*

/cc @dumbmatter

philipwalton:
R: @jeffposnick 

Fixes #1646

As far as I can tell this is working with the `workbox-sw` loader (and all the tests pass), but I know our e2e tests manually reference individual dependency files, so it's not clear to me if the loader logic is actually tested against built filenames.

/cc @gauntface 

**EDIT:** Actually, looks like tests *are* failing; looking into it now.


philipwalton:
Ok, it looks like the workbox-sw integration tests are actually checking against the live CDN, which makes renaming the script files tricky. Is this the intended behavior? Do I have to push a beta release to test this, or should I alter the tests to fetch local versions?

screen shot 2018-10-07 at 11 59 32 am

jeffposnick:
I don't think that was intentional; you should be able to work around it and get all the tests running against the local build output by using something similar to https://github.com/GoogleChrome/workbox/blob/59e7d4f0bcb365126b4094d0700346c99ce7a8d9/test/workbox-sw/static/sw.js#L3-L8

philipwalton:
Yeah, I can set a config for the loader, I just wanted to make sure the intent of this test wasn't to test that the CDN files were actually available.

BTW, I'm confused as to how that code you referenced even works because from my reading of the code, `WorkboxSW` doesn't take an options object (it defines a `setConfig()` method): https://github.com/GoogleChrome/workbox/blob/59e7d4f0bcb365126b4094d0700346c99ce7a8d9/packages/workbox-sw/controllers/WorkboxSW.mjs#L42-L49

jeffposnick:
That's... a good question. Okay, here's a more recently updated integration test that does the same thing in what's hopefully the correct way:

https://github.com/GoogleChrome/workbox/blob/59e7d4f0bcb365126b4094d0700346c99ce7a8d9/test/workbox-navigation-preload/static/sw-custom-header.js#L1-L4

philipwalton:
@jeffposnick, all the tests are passing now. I had to update some of the test server logic, but it looks like things are better tested now with these changes.

PR #1689, opened 2018-10-08
Ensure symbols are exported correctly

### LINKED ISSUE #1396: workbox-google-analytics doesn't expose a default export on index.mjs
**Library Affected**:
*workbox-google-analytics*

The CDN version of the module works with the default export, but it doesn't work with bundlers in node because no default export is defined in `index.mjs`.

For example, this works:

```js
import {initialize} from 'workbox-google-analytics';
```

but this doesn't work:

```js
import googleAnalytics from 'workbox-google-analytics';
```

This is not consistent with how other modules work, so we should update it.

philipwalton:
R: @jeffposnick 

Fixes #1396


jeffposnick:
👍

PR #1715, opened 2018-10-17
fixed : Local and server time are out of sync

wxungang:


Arm08102525:
Good

jeffposnick:
I don't believe that this is a solution to a problem faced by most developers. I'm going to close this out.

PR #1716, opened 2018-10-17
Fail SW installation when a precaching request results in an error response

### LINKED ISSUE #1711: Fail Installation if a precached asset returns 4XX or 5XX status code
**Library Affected**:
*workbox-sw, workbox-build, etc.*
workbox-precaching

**Browser & Platform**:
*E.g. Google Chrome v51.0.1 for Android, or "all browsers".*
all browsers

*Your request here.*
Hello!

There are cases where our CDN / service for Application Shell may be down, and return 4XX or 5XX status codes for these precached assets.  However, the Workbox precaching will still complete service worker installation if these assets return a 4XX or 5XX status code.

This can put users in a permanently bad position, where they cannot load up the app unless they clear the cache.

Is there / should there be an option to let precaching cancel service worker installation if not all assets are present?

Thanks!

jeffposnick:
R: @philipwalton
CC: @josephliccini 

Fixes #1711, to be released as part of v4.

Previously, if you precached a URL that resulted in a 4xx or 5xx response, that would not result in a failure to install the service worker, and you could end up with an active service worker that continuously responded to requests with that (potentially transient) original error.

I think the intention had always been to cause this to be a fatal situation that would then trigger another re-installation attempt, and this was an oversight. (Potentially a v3 regression? Not 100% sure.)

### DIFF
--- a/packages/workbox-core/models/messages/messages.mjs
+++ b/packages/workbox-core/models/messages/messages.mjs
@@ -242,4 +242,9 @@ export default {
     }
     return message;
   },
+
+  'bad-precaching-response': ({url, status}) => {
+    return `The precaching request for '${url}' failed with an HTTP ` +
+      `status of ${status}.`;
+  },
 };
--- a/packages/workbox-precaching/_default.mjs
+++ b/packages/workbox-precaching/_default.mjs
@@ -151,17 +151,22 @@ moduleExports.precache = (entries) => {
 
   installActivateListenersAdded = true;
   self.addEventListener('install', (event) => {
-    event.waitUntil(precacheController.install({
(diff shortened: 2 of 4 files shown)


jeffposnick:
Users in `dev` mode will see the following logged in the console when this happens:

screen shot 2018-10-17 at 1 23 02 pm

jeffposnick:
I'm going to hold off on merging this for a bit, as I'm thinking about whether we would want an opt-out setting for folks who either intentionally or unintentionally end up with their custom "not found page" HTML in their precache list, and with their server configured to respond to requests for that HTML with a 404.

PR #1736, opened 2018-10-26
Delete temporary precache after SW activation

### LINKED ISSUE #1735: Delete (rather than just empty) temporary cache after activation
**Library Affected**:
*workbox-precaching*

Since Workbox v3, `workbox-precaching` has been creating a temporary cache that's populated during the `install` phase, and cleared out during the `activate` phase. (See https://github.com/GoogleChrome/workbox/issues/1316.) Following a successful activation, the temporary cache has no entries in it, but the named `Cache` itself isn't deleted, so it shows up as empty in the cache storage debug tools, and can [cause developer confusion](https://twitter.com/simevidas/status/1055822175658229762).

We might as well delete it following activation, just to prevent that confusion. The cost of recreating the `Cache` should be negligible.

jeffposnick:
R: @philipwalton 

Fixes #1735.

### DIFF
--- a/packages/workbox-precaching/controllers/PrecacheController.mjs
+++ b/packages/workbox-precaching/controllers/PrecacheController.mjs
@@ -222,6 +222,10 @@ class PrecacheController {
       await tempCache.delete(request);
     }
 
+    // Remove the temporary Cache object, now that all the entries are copied.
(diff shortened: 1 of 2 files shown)


jeffposnick:
(Looks like our Travis CI tests are failing due to another change in the webpack hashes. I'll address that in a separate PR.)

PR #1747, opened 2018-10-31
Make Router.routes public

jeffposnick:
R: @philipwalton
CC: @westonruter

This takes a private property and exposes it publicly, as part of v4. There's going to be a (small) hit in bundle size, since that property name can't be mangled anymore.

### DIFF
--- a/packages/workbox-routing/Router.mjs
+++ b/packages/workbox-routing/Router.mjs
@@ -30,16 +30,22 @@ import './_version.mjs';
  * be used to respond to the request.
  *
  * @memberof workbox.routing
- * @property {Map<string, Array<workbox.routing.Route>>} routes A `Map` of HTTP
- * method name ('GET', etc.) to an array of all the corresponding `Route`
- * instances that are registered.
  */
 class Router {
   /**
    * Initializes a new Router.
    */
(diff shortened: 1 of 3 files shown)


philipwalton:
+1 on making it public, but I'm split between whether or not we should use a getter or just not use the underscore.

I do think we should be consistent though, and currently we're not. In the `workbox.routing.Route` class we have the public `match`, `handler`, and `method` properties. But in the `workbox.backgroundSync.Queue` class we have `get name()`, which returns `this._name`.

From a minification perspective, I'm not sure there's a obvious winner either because, on the one hand not writing a getter method makes the class definition smaller, but on the other hand being able to minify all internal instances of `._routes` also makes it smaller.

WDYT?

jeffposnick:
I guess, for something like this, we might as well go for whatever produces the smallest minified bundle?

- Currently, in v3, `workbox-routing.prod.js` is 3252 bytes.
- With this PR (making the property public), `workbox-routing.prod.js` is 3292 bytes.
- With a `get`ter, `workbox-routing.prod.js` is 3279 bytes.

I'll switch to the `get`ter and save 13 bytes (before gzip).

PR #1750, opened 2018-11-06
Updates to package-lock.json files

jeffposnick:
R: @philipwalton 

Let's see if this makes the Travis CI test happy? In particular, the `workbox-webpack-plugin` tests on my local machine are fine, and perhaps pinning the versions to what's in the updated lockfiles will lead to that being okay in Travis, too.

### DIFF
--- a/packages/workbox-build/package-lock.json
+++ b/packages/workbox-build/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "workbox-build",
-	"version": "3.6.2",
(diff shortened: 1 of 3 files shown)


jeffposnick:
That did... not fix it.

jeffposnick:
Ah, okay, I see now that Travis is running node 11, and I've been running node 10.

I've pinned node 10 to the `.travis.yml` in the latest commit, and let's see if that helps.

PR #1757, opened 2018-11-13
renaming-plugins

prateekbh:
R: @jeffposnick @philipwalton

Renames plugins for better mangling by closure compiler


philipwalton:
It looks like there's an additional issue where we have a test that ensures all files in the top level of each package are exposed (exactly as named) on the pubic namespace. I really like this convention, but we're going to have to break it to make this change.

Our options are either:

* Keep the files named `Plugin.mjs`.
* Code an exception to the test for files matching the name `*Plugin.mjs`.

Which option we choose ultimately comes down to whether we care more about

* The top-level file names matching their export names, or...
* The top-level file names matching the public property names created by `workbox-sw`.

prateekbh:
My intentions with this PR was to make this work with closure compiler `advanced` mode, however just renaming the plugins wont solve the problem. It'd need externs for multiple files like all strategies and probably some more closure annotations. I can try to investigate what all would be required to support closure compiler `advanced` mode and revert on this PR

jeffposnick:
Thanks for the context, @prateekbh. Given that, I'm going to close this PR.

PR #1763, opened 2018-11-20
adding swSrc as webpack asset capability

prateekbh:
- Adds a capability to specify `swSrc` as a webpack asset filename


prateekbh:
woops sry, I copied it from the project which caused this feature request, but I feel `this.swSrc` should be fine as it can be a filepath or the filename. Do you think the overload will be too confusing?

prateekbh:
I am closing this as swSrc should never be a webpack asset. The compilation generating the assets will mostly have webpack target as "window" but the any service worker file being produced needs to have target as worker. 

Since the service worker and rest of the assets cannot share the same compilation, the plugin can only have access to either the list of assets or the service worker file.

prateekbh:
Sorry, I did discover a valid use case for this now: "childCompiler"

Since a child compiler can generate swSrc in the same compilation with a different target, hence this change will be required

jeffposnick:
I'd love to get @developit's take on how this fits in with the other things he's looking into.

prateekbh:
@developit bumping this thread for your review, as the release of this feature is blocking integration in [preact-cli](https://github.com/developit/preact-cli/pull/674). Which btw also needs to be reviewed by you 😂

philipwalton:
@developit, can you take a look at this PR and let us know if it conflicts with any of your future plans for the webpack plugin.

This is blocking the preact cli integration, but I wouldn't want to merge this only to change the behavior in v5.

developit:
Honestly what @prateekbh has set up in his Preact CLI PR is super close to what was in my head. The only difference would be that I'd like to bundle via Webpack and import workbox as modules, which I think we can still add independently in v5?

@prateekbh do you think it would be prohibitive to require that a custom Service Worker defined via `swSrc` import workbox rather than relying on it being global?

philipwalton:
> @prateekbh do you think it would be prohibitive to require that a custom Service Worker defined via swSrc import workbox rather than relying on it being global?

Or perhaps (whether in this version or a future version), there could be an option to use either `import` or `importScripts()`?

developit:
@philipwalton yup. writing up an RFC right now.

prateekbh:
@developit is there a good reason to make it **compulsory** to not use a global workbox via importScript?

developit:
Nope, I meant leave it as an exercise for the developer. They can importScripts().

prateekbh:
I guess it'll be somewhat restrictive to say that in order to use `swSrc` you can only import Workbox module and not use importScripts

developit:
It'd also be a little counterintuitive, since Service Worker doesn't natively support modules, but does support importScripts().

developit:
Right, I am saying allow either:

```js
import workbox from 'workbox';
workbox.precaching.precacheAndRoute(self.__precacheManifest);
```

or:

```js
importScripts('https://storage.googleapis.com/workbox-cdn/releases/3.6.1/workbox-sw.js');
workbox.precaching.precacheAndRoute(self.__precacheManifest);
```

Either of these should be allowed/supported.

packages/workbox-webpack-plugin/src/inject-manifest.js
```diff
@@ -140,6 +149,13 @@ ${originalSWString}
 `;
 
     const relSwDest = relativeToOutputPath(compilation, this.config.swDest);
+    /**
+     * Delete if this file already exist,
+     * this is the case where source is also a webpack asset.
+     */
+    if (compilation.assets[relSwDest]) {
+      delete compilation.assets[relSwDest];
```
jeffposnick: Line 159 assigns a value to `compilation.assets[relSwDest]`, which should overwrite anything that's already there, so I'm not clear on what this `delete` accomplishes. Can you clarify?

prateekbh: I saw that simply overriding was somehow not overriding the content, so I added an additional delete statement which did the trick

jeffposnick: That feels kind of magic. @developit, have you ever seen anything like that?

(Also, I'm not sure if there's a reason to wrap this in an `if()`, since the `delete` will be a no-op if the property doesn't exist.)

developit: Weird, I've never had to do this. @prateekbh any chance it was some other issue? I can't think of why this would happen.

prateekbh: i'll re-check

prateekbh: well yeah it works now.. dont know what was the problem earlier..

prateekbh:
@jeffposnick added the tests

PR #1772, opened 2018-11-28
Adds a fetchDidSucceed callback

### LINKED ISSUE #788: Feature: Add fetchDidSucceed method to requestWrapper
Per this discussion https://stackoverflow.com/questions/46026849/postprocess-page-content-with-stalewhilerevalidate/

and @jeffposnick response, adding a method to allow modifying the response in all cases (ie fetched from network or fetched from cache). 

Currently, it is not possible to modify a response that was obtained from the network, only when it was obtained from the cache.

### LINKED ISSUE #1391: Mechanism to detect non-200 requests from Workbox
**Library Affected**:
*workbox-sw*

**Browser & Platform**:
*Any*

Hello again!

By Fetch API specification, HTTP statuses of any response (whether it be a 200, 300, 400, or 500) will not be considered an exception, unless the request actually fails to send (which makes sense).

Our web application on a large scale calls into various APIs and we use telemetry to monitor the health of each API we call into.  One dimension we would do this on is `statusCode`.  That way, we can see spikes / outages for various APIs.

Now, with a SW, it's possible that the API could be failing, but we serve up `stale` data (which is awesome).  However, we would like a way to continue to log the reliability of the `WhileRevalidate` part of the process, because now, any cached responses that get served will now blind our application from detecting API failures.

What is the best way to go about this?  Should this be supported in `workbox`?  Could the CatchHandler also intercept requests that have a certain response code?  I think (right now) the best way to go about it is to wrap all of our handlers in another handler that will inspect the status code of the revalidation response.  (I haven't started creating that handler yet, so I'm not sure it would even work.  Will be doing that shortly)

Thanks!

### LINKED ISSUE #1481: Plugins should have a callback for modifying network responses
Currently I can transform a network response before it is cached using `cacheWillUpdate()`.
I can even transform a cache response before it is used to respond to the fetch event using `cacheResponseWillBeUsed()`. 
But there is no callback allowing me to transform a network response before it is used to respond to the fetch event.

My use-case: 
I want a cache-first strategy for handling particular requests but I also need to transform the network response before it is used to  respond to the fetch event and I need to ensure that future cache hits return the same transformed response. I can ensure that later requests (which hit the cache) get the transformed response, but it does not seem possible to intercept network responses with the current design / implementation of CacheFirst (or any of the other strategies). 

Initially I assumed that the *result* of `cacheWillUpdate()` would be used both to update the cache and to respond to the fetch event. Testing demonstrated that the result is used to update the cache but is *not* used to respond to the fetch event. 

Then I thought I would construct a route handler using a CacheFirst handler to ensure that the cache is populated followed by a CacheOnly handler to pick up the transformed response, but the CacheFirst handler resolves before the cache is updated so CacheOnly handler fails. 

I finally solved my use-case using an async wrapper around a CacheFirst handler. The wrapper function is resolved by `cachedResponseWillBeUsed()` or `cacheWillUpdate()`. 

I suspect my use-case is common enough that it should be handled by CacheFirst with a custom-plugin instead of a custom handler that relies on inspecting the CacheFirst internals.

jeffposnick:
R: @philipwalton

Fixes #788, fixes #1391, fixes #1481

Adds a new `fetchDidSucceed` callback, which allows a plugin to modify a request that's been retrieved from the network.

This was requested by @weeblr back in the day.

We'll need to update https://developers.google.com/web/tools/workbox/guides/using-plugins once v4 goes live to reflect this new callback.

### DIFF
--- a/packages/workbox-core/_private/fetchWrapper.mjs
+++ b/packages/workbox-core/_private/fetchWrapper.mjs
@@ -113,6 +113,7 @@ const wrappedFetch = async ({
       if (pluginEvents.FETCH_DID_SUCCEED in plugin) {
         fetchResponse = await plugin[pluginEvents.FETCH_DID_SUCCEED]
             .call(plugin, {
+              event,
               request: pluginFilteredRequest,
               response: fetchResponse,
             });
--- a/packages/workbox-core/_private/fetchWrapper.mjs
+++ b/packages/workbox-core/_private/fetchWrapper.mjs
@@ -102,12 +102,33 @@ const wrappedFetch = async ({
   const pluginFilteredRequest = request.clone();
 
   try {
-    const fetchResponse = await fetch(request, fetchOptions);
+    let fetchResponse = await fetch(request, fetchOptions);
     if (process.env.NODE_ENV !== 'production') {
       logger.debug(`Network request for `+
       `'${getFriendlyURL(request.url)}' returned a response with ` +
       `status '${fetchResponse.status}'.`);
     }
+
+    for (const plugin of plugins) {
+      if (pluginEvents.FETCH_DID_SUCCEED in plugin) {
+        fetchResponse = await plugin[pluginEvents.FETCH_DID_SUCCEED]
+            .call(plugin, {
+              request: pluginFilteredRequest,
+              response: fetchResponse,
+            });
+
+        if (process.env.NODE_ENV !== 'production') {
+          if (fetchResponse) {
+            assert.isInstance(fetchResponse, Response, {
+              moduleName: 'Plugin',
+              funcName: pluginEvents.FETCH_DID_SUCCEED,
+              isReturnValueProblem: true,
+            });
+          }
+        }
+      }
+    }
+
     return fetchResponse;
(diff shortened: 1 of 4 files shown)


packages/workbox-core/_private/fetchWrapper.mjs
```diff
@@ -102,20 +102,41 @@ const wrappedFetch = async ({
...
       logger.debug(`Network request for `+
       `'${getFriendlyURL(request.url)}' returned a response with ` +
       `status '${fetchResponse.status}'.`);
     }
+
+    for (const plugin of plugins) {
+      if (pluginEvents.FETCH_DID_SUCCEED in plugin) {
+        fetchResponse = await plugin[pluginEvents.FETCH_DID_SUCCEED]
+            .call(plugin, {
+              request: pluginFilteredRequest,
```
philipwalton: Can you also add `event`, so it's consistent with the changes in https://github.com/GoogleChrome/workbox/pull/1640?

jeffposnick:
PTAL.

PR #1791, opened 2018-12-13
Preserve the temporary cache

### LINKED ISSUE #1783: Precaching is in a bad state following a failed activation in Firefox
**Library Affected**:
*workbox-v3.6.3*

**Browser & Platform**:
Firefox 63.0.3 Windows

Basically the same as https://github.com/GoogleChrome/workbox/issues/1528, but it would be cool if the problem would be solved in the library itself.

I am using the InjectManifest plugin (its the same with GenerateSW). 

After a deploy the service woker claims to have updates. When closing the browser and opening it again one js file is deleted form the cache. The index.html is still on the old version that depends on this js file. So the page can't load anymore.

Here the config (i am using https://www.npmjs.com/package/@vue/cli-plugin-pwa)

vue.config.js
```
process.env.VUE_APP_VERSION = (new Date()).toISOString();

module.exports = {
    devServer: {
        proxy: 'http://localhost:8000'
    },
    // ...other vue-cli plugin options...
    pwa: {
        name: 'My App',
        themeColor: '#4DBA87',
        msTileColor: '#000000',
        appleMobileWebAppCapable: 'yes',
        appleMobileWebAppStatusBarStyle: 'black',
        workboxPluginMode: 'InjectManifest',
        workboxOptions: {
            //https://developers.google.com/web/tools/workbox/modules/workbox-webpack-plugin
            swSrc: __dirname + '/public/service-worker.js',
            dontCacheBustUrlsMatching: /\.[0-9a-f]{8}\./,
            importWorkboxFrom: 'local',
        }
    }
}
```
__dirname + '/public/service-worker.js'
```
workbox.setConfig({
    debug: true
});
workbox.core.setLogLevel(workbox.core.LOG_LEVELS.debug);


workbox.core.setCacheNameDetails({prefix: "my-garden"});

/**
 * See https://goo.gl/S9QRab
 */
//workbox.precaching.suppressWarnings();

workbox.routing.registerNavigationRoute("/index.html", {
  
  blacklist: [/^\/(api|admin|vendor|storage)/,/\.[a-z]{2,5}$/i],
});
```

The server is configured to not serve 404, but to fallback to index.html instead.

jeffposnick:
R: @philipwalton

Fixes #1783

Based on my manual testing, this should ensure that precaching in both Chrome and Firefox are able to recover from a crash or unexpected shutdown that prevents the `activate` handler from completing. The actual `Response` entries in the temporary cache still get cleaned up once each one is moved over to the permanent cache in a successful `activate` handler.

Unfortunately:

- I haven't been able to add in an integration test to confirm this behavior, since you can't cause activation to fail by, e.g., throwing an exception or passing a rejected promise to `event.waitUntil()`. Activation failure isn't really *supposed* to happen, and it only appears to happen in the wild due to browser bugs.

- This does not account for the Edge is supposed to behave when activation fails, which is not to re-run either the `install` or `activate` event handlers at all. I'm not sure *how* we could actually recover from that, and have an open question at https://github.com/w3c/ServiceWorker/issues/1372#issuecomment-446766388 to see whether there's any advice. But this PR should help Firefox, at least.

- Some folks wanted the `-temp` cache to disappear entirely following a successful activation, and I had previously committed something to the `next` branch that would do that. In the course of investigating this PR, I realize why we have to keep the `-temp` cache around—see https://github.com/GoogleChrome/workbox/pull/1372 for more context. Given that, this PR also reverts the previous PR (https://github.com/GoogleChrome/workbox/pull/1736) which deleted the cache.

### DIFF
--- a/packages/workbox-precaching/controllers/PrecacheController.mjs
+++ b/packages/workbox-precaching/controllers/PrecacheController.mjs
@@ -156,16 +156,6 @@ class PrecacheController {
       }
     }
 
-    // Empty the temporary cache.
-    // NOTE: We remove all entries instead of calling caches.delete(), as the
-    // cache may be marked for deletion but still exist.
-    // See https://github.com/GoogleChrome/workbox/issues/1368
-    const tempCache = await caches.open(this._getTempCacheName());
-    const requests = await tempCache.keys();
-    await Promise.all(requests.map((request) => {
-      return tempCache.delete(request);
-    }));
-
     const entriesToPrecache = [];
     const entriesAlreadyPrecached = [];
 
@@ -221,10 +211,6 @@ class PrecacheController {
       });
       await tempCache.delete(request);
     }
-
-    // Remove the temporary Cache object, now that all the entries are copied.
-    // See https://github.com/GoogleChrome/workbox/issues/1735
-    await caches.delete(this._getTempCacheName());
 
     return this._cleanup();
   }
--- a/test/workbox-precaching/integration/precache-and-update.js
+++ b/test/workbox-precaching/integration/precache-and-update.js
@@ -43,6 +43,7 @@ describe(`[workbox-precaching] Precache and Update`, function() {
     const keys = await runInSW('cachesKeys');
     expect(keys).to.eql([
       'workbox-precache-http://localhost:3004/test/workbox-precaching/static/precache-and-update/',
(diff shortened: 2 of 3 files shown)


jeffposnick:
This was a stopgap to work around an edge case, but it sounds like there's a path forward in https://github.com/GoogleChrome/workbox/issues/1793, and I'm thinking that we could just close this.

(I am super-frustrated that there's an integration test failure in Travis CI that I can't reproduce locally, too...)

PR #1820, opened 2019-01-08
workbox-precaching rewrite to avoid dependency on activate handler

### LINKED ISSUE #1783: Precaching is in a bad state following a failed activation in Firefox

### LINKED ISSUE #1793: Rethink precaching's activate behavior
(CC: @josephliccini)

**Library Affected**:
*workbox-precaching*

## Background

- https://github.com/GoogleChrome/workbox/issues/1316 is the original issue that led to Workbox v3's current behavior: during `install`, `Response`s corresponding to the set of new assets listed in the precache manifest are written to a "temporary" cache, and during `activate`, they're copied over to the "main" cache. `workbox-precaching` will never serve `Resposne`s directly from the temporary cache. Unless the `activate` handler completes successfully, assets will end up "stuck" in the temporary cache indefinitely.

- https://github.com/w3c/ServiceWorker/issues/1372 is a discussion about how various browser's deal with errors that prevent `activate` handlers from completing. As per the service worker spec, an `activate` handler is effectively not allowed to fail, but bugs and edge cases can lead to them failing in practice. The general takeaway is that Edge and Chrome do not make any effort to retry the `activate` handler when an error leads to failure (with some provisions in place to delay activation when it's likely to fail, e.g. when the browser is shutting down).

- https://github.com/GoogleChrome/workbox/pull/1791 is a PR to deal with a particular issue observed (presumably due to a browser bug) in Firefox, where the `activate` handler doesn't end up completing, leading to Firefox re-triggering the `install` event. Firefox's current behavior goes against the service worker specification, but at the same time, we've now seen it in the wild, so a fix seemed prudent.

## Proposed changes

`workbox-precaching`'s current reliance on the `activate` event running to completion can lead to cache inconsistencies in any browser that follows the service worker spec if the `activate` event terminates prior to completion. This is apparently easy-ish(?) to trigger in Edge if you shut down the browser while a service worker activation is happening.

As far as I can tell, the best way to deal with this is to ensure that the *only* things that happen during the `activate` event is purging unneeded cache entries—the sort of thing that we want to do, but which won't lead to cache inconsistencies if it didn't complete.

I can think of a couple of ways to move to that model:

### sw-precache's approach

The earlier `sw-precache` library [handled this scenario](https://github.com/GoogleChromeLabs/sw-precache/blob/e5892c9a3cce1a8d4f6de2aa768287cb709ba1e7/service-worker.tmpl#L51-L60) by using a single cache whose entries including a synthetic `_sw-precache=<revision>` query parameter appended to the end of the URL. (URLs that were known to already include revision information did not have this query parameter appended.) These cache entries could safely be written to the single cache during the `install` event without any risk of overwriting entries still in use by the previously activated service worker, since updated entries would have updated URLs.

The `activate` handler, in that approach, [just needs to do cleanup](https://github.com/GoogleChromeLabs/sw-precache/blob/e5892c9a3cce1a8d4f6de2aa768287cb709ba1e7/service-worker.tmpl#L106-L126).

One tangential benefit of this approach is that it's not necessary to use IndexedDB to maintain out-of-band metadata keeping track of the active mapping of URLs to revision info, which we have to do in Workbox. Adopting this approach would mean eliminating the dependencies on the IndexedDB runtime code from the `workbox-precaching` bundle. The mapping of "real" URL to the correct URL to use as a cache key can be maintained via a variable in the service worker's scope.

The primary downside of this approach is that developers who expect their precached resources to use specific URLs as their cache keys will end up confused; the cached entry for `/index.html`, for instance, might have a cache key of `/index.html?_workbox=abcd1234`. Attempts to read the cache via, e.g., `caches.match('/index.html')` will fail.

To work around this, developers would have to do one of two things, each with drawbacks:

1) Rely on a function we'd expose in `workbox-precaching` that would map a "real" URL into a cache key (c.f. [this in `sw-precache`](https://github.com/GoogleChromeLabs/sw-precache/blob/e5892c9a3cce1a8d4f6de2aa768287cb709ba1e7/service-worker.tmpl#L51-L60)). It would be hard to get this info from inside the `window` context, since it's relies on data that's only available inside the service worker's scope.

2) Use `caches.match('/index.html', {ignoreSearch: true})`, with the risk that the incorrect entry is returned if multiple URLs match the same path and only vary based on query parameters.

### Double-cache everything

A less radical departure from `workbox-precaching`'s current behavior would be to create a new logical cache each time the `install` handler is run, presumably using some information from inside the precache manifest to include a unique versioned string in the name. The new cache would contain a complete copy of every resource listed in the current precache manifest, even if that resource has already been previously precached.

In this scenario, the `activate` handler would just call `caches.delete()` on any out-of-date logical caches.

This approach would not lead to extra data being transferred over the network, since we can use `cache.match()` + `cache.put()` to make copies of existing entries locally.

The major downside of this approach is that we'd risk triggering quota errors by maintaining two full copies of all precached resources during the interval between `install` and cleanup in `activate`. It's not clear how we could automatically recover from quota errors encountered during the `install` handler, and it's easy to envision failure scenarios in which `install` is re-attempted repeatedly, failing each time due to quota errors, leaving users stuck with a previous set of assets indefinitely.

(This is actually an existing failure scenario that applies today, and our current approach to avoiding it is just to try not to duplicate anything when we can help it; see https://github.com/GoogleChrome/workbox/issues/1312 for better ideas.)

A tangential downside is that we would have to continue using IndexedDB inside of `workbox-precaching`, leading to a larger runtime bundle and greater code complexity than we'd otherwise have.

### Other approaches?

I'm not aware of any, but I'm interested in hearing other ideas.

jeffposnick:
R: @philipwalton

Fixes #1793, fixes #1783

This is a pretty significant rewrite of the `workbox-precaching` core behavior. The goal is to ensure that only cleanup work is done in the `activate` event. Previously, the `activate` event was used to move precached `Response`s over from a `-temp-` cache into the active cache. Now, we just add all `Response`s directly in the activate cache in the `install` handler, using an extra URL query parameter to store any `revision` metadata that's provided. (This is the same model that `sw-precache` previously used.)

Moving to this model means that we no longer have a dependency on IndexedDB within `workbox-precaching`, which should reduce the overall bundle size and lead to what's hopefully less complex code.

In addition to that change, there are two new functions exposed on `workbox.precaching` as part of this PR:

- `workbox.precaching.getCacheKeyForUrl(url)`, which will return the key that corresponds to the cache entry for `url`. This takes into account the fact that the key might have an extra URL query parameter appended to it. It's useful if you need to manually call `cache.match(key)` to retrieve a precached `Response`, and you need to know the specific `key` to use.

- `workbox.precaching.cleanupOutdatedCaches()` is not called by default, but developers can try it out by explicitly calling it. It will iterate through all of the caches in scope for the current service worker, and attempt to identify and delete "out of date" precaches. Developers upgrading from earlier versions of Workbox to Workbox v4 might find this useful, since we have a breaking change in precache compatibility in v4, and a brand-new precache is created as part of the upgrade. Calling `workbox.precaching.cleanupOutdatedCaches()` means that the old, Workbox v1-v3 precached entries will be deleted after the new Workbox v4 service worker activates.


packages/workbox-precaching/controllers/PrecacheController.mjs
```diff
@@ -156,88 +89,61 @@ class PrecacheController {
...
-      notUpdatedEntries: entriesAlreadyPrecached,
+      updatedUrls: urlsToPrecache,
+      notUpdatedUrls: urlsAlreadyPrecached,
     };
   }
 
   /**
-   * Takes the current set of temporary files and moves them to the final
-   * cache, deleting the temporary cache once copying is complete.
+   * Call this method from the service worker activate event to delete assets
```
philipwalton: Same comment as above (put description before recommendation).

packages/workbox-precaching/controllers/PrecacheController.mjs
```diff
@@ -299,119 +201,41 @@ class PrecacheController {
...
 
   /**
-   * Goes through all entries in indexedDB and removes any that are outdated.
-   *
-   * @private
-   * @param {Array<string>} expectedCacheUrls Array of URLs that are
-   * expected to be cached.
-   * @return {Promise<Array<string>>} Resolves to an array of URLs removed
-   * from indexedDB.
+   * @return {Array<string>} All of the URLs that have been precached.
```
philipwalton: Is this method public? If so it should probably have a description for the docs; if it's not then it should start with an underscore so it can be minified.

packages/workbox-precaching/utils/cleanupOutdatedCaches.mjs
```diff
@@ -0,0 +1,45 @@
...
+  license that can be found in the LICENSE file or at
+  https://opensource.org/licenses/MIT.
+*/
+
+import '../_version.mjs';
+
+const SUBSTRING_TO_FIND = '-precache-';
+
+/**
+ * Can be called as part of the `activate` event to clean up incompatible
```
philipwalton: Same comment as above (put description before recommendation).

packages/workbox-precaching/utils/cleanupOutdatedCaches.mjs
```diff
@@ -0,0 +1,45 @@
...
+ * @param {string} currentPrecacheName The cache name currently in use for
+ * precaching. This cache won't be deleted.
+ * @param {string} [substringToFind='-precache-'] Cache names which include this
+ * substring will be deleted (excluding `currentPrecacheName`).
+ * @return {Array<string>} A list of all the cache names that were deleted.
+ *
+ * @private
+ * @memberof module:workbox-precaching
+ */
+export default async function(currentPrecacheName,
```
philipwalton: Can we make this a named export, so it's consistent with future plans to not have default exports?

packages/workbox-precaching/utils/createCacheKey.mjs
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Converts a manifest entry into a versioned URL suitable for precaching.
+ *
+ * @param {Object} entry
+ * @return {string} A URL with versioning info.
+ *
+ * @private
+ * @memberof module:workbox-precaching
+ */
+export default function(entry) {
```
philipwalton: Same comment as above, can we make this a named export?

packages/workbox-precaching/utils/createCacheKey.mjs
```diff
@@ -0,0 +1,64 @@
...
+  license that can be found in the LICENSE file or at
+  https://opensource.org/licenses/MIT.
+*/
+
+import {WorkboxError} from 'workbox-core/_private/WorkboxError.mjs';
+
+import '../_version.mjs';
+
+// Name of the search parameter used to store revision info.
+const REVISION_SEARCH_PARAM = '_wbRevision';
```
philipwalton: Do you think there's any value in making this customizable? My initial thought is *no*, since it would make upgrading harder, but what do you think?

Also, I wonder if it makes sense to make it something that's even less likely to collide and be more obvious what it is when looking at devtools, e.g. `__WB_REVISION__`.

packages/workbox-precaching/utils/generateUrlVariations.mjs
```diff
@@ -0,0 +1,55 @@
...
+ * Generator function that yields possible variations on the original URL to
+ * check, one at a time.
+ *
+ * @param {string} url
+ * @param {Object} options
+ *
+ * @private
+ * @memberof module:workbox-precaching
+ */
+export default function* (url, {
```
philipwalton: Same comment as above, can we make this a named export?

packages/workbox-precaching/utils/printCleanupDetails.mjs
```diff
@@ -7,47 +7,32 @@
...
  *
  * @private
- * @memberof module:workbox-precachig
+ * @memberof module:workbox-precaching
  */
-export default (deletedCacheRequests, deletedRevisionDetails) => {
-  if (deletedCacheRequests.length === 0 &&
-    deletedRevisionDetails.length === 0) {
-    return;
+export default (deletedUrls) => {
```
philipwalton: Same comment as above, can we make this a named export?

packages/workbox-precaching/utils/printCleanupDetails.mjs
```diff
@@ -7,47 +7,32 @@
...
+  const deletionCount = deletedUrls.length;
+  if (deletionCount > 0) {
```
philipwalton: Should this all be inside a `NODE_ENV !== 'production'` conditional?

packages/workbox-precaching/utils/printInstallDetails.mjs
```diff
@@ -7,57 +7,53 @@
...
-
-  const precachedCount = entriesToPrecache.length;
-  const alreadyPrecachedCount = alreadyPrecachedEntries.length;
-  let printText =
-    `Precaching ${precachedCount} file${precachedCount === 1 ? '' : 's'}.`;
-  if (alreadyPrecachedCount > 0) {
-    printText += ` ${alreadyPrecachedCount} ` +
-      `file${alreadyPrecachedCount === 1 ? ' is' : 's are'} already cached.`;
-  }
+export default (urlsToPrecache, urlsAlreadyPrecached) => {
```
philipwalton: Same comment as above, can we make this a named export?

packages/workbox-precaching/utils/printInstallDetails.mjs
```diff
@@ -7,57 +7,53 @@
...
 /**
  * @param {string} groupTitle
- * @param {Array<PrecacheEntry>} entries
+ * @param {Array<string>} urls
  *
  * @private
  */
-const _nestedGroup = (groupTitle, entries) => {
-  if (entries.length === 0) {
+function _nestedGroup(groupTitle, urls) {
```
philipwalton: Why change away from `const`?

packages/workbox-precaching/utils/printInstallDetails.mjs
```diff
@@ -7,57 +7,53 @@
...
-  _nestedGroup(
-      `View URLs that were already precached.`,
-      alreadyPrecachedEntries);
-  logger.groupEnd();
+    if (alreadyPrecachedCount > 0) {
+      message += ` ${alreadyPrecachedCount} ` +
+        `file${alreadyPrecachedCount === 1 ? ' is' : 's are'} already cached.`;
+    }
+
+    logger.groupCollapsed(message);
```
philipwalton: Should this all be inside a `NODE_ENV !== 'production'` conditional?

packages/workbox-precaching/utils/removeIgnoredSearchParams.mjs
```diff
@@ -0,0 +1,33 @@
...
+ * @param {URL} urlObject The original URL.
+ * @param {Array<RegExp>} ignoreUrlParametersMatching RegExps to test against
+ * each search parameter name. Matches mean that the search parameter should be
+ * ignored.
+ * @return {URL} The URL with any ignored search parameters removed.
+ *
+ * @private
+ * @memberof module:workbox-precaching
+ */
+export default function(urlObject, ignoreUrlParametersMatching) {
```
philipwalton: Same comment as above, can we make this a named export?

test/workbox-precaching/integration/cleanup-outdated-caches.js
```diff
@@ -0,0 +1,44 @@
+/*
+  Copyright 2018 Google LLC
```
philipwalton: Nit: s/2018/2019

test/workbox-precaching/node/utils/test-cleanupOutdatedCaches.mjs
```diff
@@ -0,0 +1,98 @@
+/*
+  Copyright 2018 Google LLC
```
philipwalton: Nit: s/2018/2019/

test/workbox-precaching/static/cleanup-outdated-caches/sw.js
```diff
@@ -0,0 +1,22 @@
+/*
+  Copyright 2018 Google LLC
```
philipwalton: Nit: s/2018/2019/

packages/workbox-core/_private/cacheNames.mjs
```diff
@@ -9,11 +9,11 @@
 import '../_version.mjs';
 
 const _cacheNameDetails = {
-  prefix: 'workbox',
-  suffix: self.registration.scope,
   googleAnalytics: 'googleAnalytics',
-  precache: 'precache',
+  precache: 'precache-v4',
```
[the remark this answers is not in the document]
jeffposnick: Right. Another option is `'precache-v2'`, implying that it's the second precache version, rather than using the version number associated with Workbox.

Would you prefer that?

packages/workbox-precaching/_default.mjs
```diff
@@ -276,27 +214,54 @@ moduleExports.precacheAndRoute = (entries, options) => {
...
  *
- * @param {Array<Object>} newPlugins
+ * @alias workbox.precaching.cleanupOutdatedCaches
+ */
+moduleExports.cleanupOutdatedCaches = () => {
+  self.addEventListener('activate', (event) => {
+    event.waitUntil(cleanupOutdatedCaches(cacheName).then((cachesDeleted) => {
+      if (cachesDeleted.length > 0) {
+        logger.log(`The following out-of-date precaches were cleaned up ` +
+        `automatically ${cachesDeleted}`);
```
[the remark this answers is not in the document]
jeffposnick: I'll remove the implicit string conversion.

packages/workbox-precaching/controllers/PrecacheController.mjs
```diff
@@ -55,97 +54,31 @@ class PrecacheController {
...
-        firstEntry: existingEntry._originalInput,
-        secondEntry: entryToAdd._originalInput,
-      });
-    }
-  }
-
-  /**
-   * Call this method from a service work install event to start
-   * precaching assets.
+   * Call this method from the service worker install event to precache new and
```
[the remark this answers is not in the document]
jeffposnick: Changed.

packages/workbox-precaching/utils/cleanupOutdatedCaches.mjs
jeffposnick: Changed.

packages/workbox-precaching/utils/cleanupOutdatedCaches.mjs
jeffposnick: Changed to use the `export async function cleanupOutdatedCaches` syntax. I think that's okay? Let me know if I need to switch to defining the `cleanupOutdatedCaches` function and then later doing `export {cleanupOutdatedCaches}` instead.

packages/workbox-precaching/utils/createCacheKey.mjs
jeffposnick: Changed.

packages/workbox-precaching/utils/createCacheKey.mjs
jeffposnick: I don't see there being a risk of query parameter collision, so I'd rather not make it customizable.

I can change to `'__WB_REVISION__'`.

packages/workbox-precaching/utils/generateUrlVariations.mjs
jeffposnick: Changed.

packages/workbox-precaching/utils/printCleanupDetails.mjs
jeffposnick: Changed.

packages/workbox-precaching/utils/printCleanupDetails.mjs
jeffposnick: The code that calls this module is inside a `NODE_ENV` check.

packages/workbox-precaching/utils/printInstallDetails.mjs
jeffposnick: Changed.

packages/workbox-precaching/utils/printInstallDetails.mjs
jeffposnick: Just for consistency's sake, since the other function uses `function`.

packages/workbox-precaching/utils/removeIgnoredSearchParams.mjs
jeffposnick: The code that calls this module is inside a `NODE_ENV` check.

packages/workbox-precaching/utils/printInstallDetails.mjs
jeffposnick: The code that calls this module is inside a `NODE_ENV` check.

packages/workbox-precaching/utils/removeIgnoredSearchParams.mjs
jeffposnick: Changed.

test/workbox-precaching/integration/cleanup-outdated-caches.js
jeffposnick: Changed.

test/workbox-precaching/node/utils/test-cleanupOutdatedCaches.mjs
jeffposnick: Changed.

test/workbox-precaching/static/cleanup-outdated-caches/sw.js
jeffposnick: Changed.

packages/workbox-precaching/controllers/PrecacheController.mjs
```diff
@@ -55,97 +54,31 @@ class PrecacheController {
...
+   * updated assets.
    *
    * @param {Object} options
-   * @param {boolean} [options.suppressWarnings] Suppress warning messages.
    * @param {Event} [options.event] The install event (if needed).
    * @param {Array<Object>} [options.plugins] Plugins to be used for fetching
-   *     and caching during install.
+   * and caching during install.
```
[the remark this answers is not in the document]
philipwalton: SG. We can make them consistent in another PR sometime in the future.

packages/workbox-precaching/utils/printInstallDetails.mjs
philipwalton: But uglify may not be able to figure that out. Can you confirm that it doesn't appear in the minified file?

jeffposnick: http://storage.googleapis.com/workbox-cdn/releases/4.0.0-beta.0/workbox-precaching.prod.js suggests that it's not in the `prod` bundle, due to what I guess is tree-shaking.

This PR changes some of the logging logic in this file, but shouldn't lead to changes in how things are bundled up.

packages/workbox-core/_private/cacheNames.mjs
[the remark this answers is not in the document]
jeffposnick: If you're following the guidance at https://developers.google.com/web/tools/workbox/guides/configure-workbox#configure_cache_names and changing `prefix`, then you'll automatically get the new `'precache-v2'` as part of your cache name.

If you're bypassing that cache name mechanism entirely (which... I'm assuming not too many people are, since you have to use `PrecacheController` directly vs. using `workbox.precaching`) then your cache name will not end up changing. As for how bad a problem that is... I *think* the worst case scenario is ending up with extra entries in the named cache that would never be cleaned up, because the new `workbox-precaching` code isn't tracking them.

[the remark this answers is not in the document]
philipwalton: Cool. Probably worth mentioning in the upgrade notes that anyone in that latter case should probably choose a new name for v4.

PR #1827, opened 2019-01-09
Add a workbox-window package

philipwalton:
This PR is an initial implementation of the [`workbox-window` proposal](https://docs.google.com/document/d/16BJDNhfqMjIvJ3_kR4lklpbliCGpi1dFhCdf7kHOsL8/edit?usp=sharing).

Most of the implementation is purely additive, but a few things have changed in the build process. Mainly, an additional `build-window-packages` gulp task has been added that builds `workbox-window` since its build format is ES modules rather than IIFEs and browser globals.

I've also added a new `templates` route, that allow me to render content dynamically via nunjucks if the route matches the pattern `*.tmp.*`. This is how I test dynamic SW updates for scripts with the same URL in my integration tests.

R: @jeffposnick


packages/workbox-window/Workbox.mjs
```diff
@@ -0,0 +1,417 @@
...
+const urlsMatch = (url1, url2) => {
+  return new URL(url1, location).href === new URL(url2, location).href;
+};
+
+
+/**
+ * A class to aid in handling service worker registration, updates, and
+ * reacting to service worker lifecycle events.
+ *
+
```
jeffposnick: Empty line?

packages/workbox-window/Workbox.mjs
```diff
@@ -0,0 +1,417 @@
...
+export class Workbox extends EventTargetShim {
+  /**
+   * Creates a new Workbox instance with a script URL and service worker
+   * options. The script URL and options are the same as those used when
+   * calling `navigator.serviceWorker.register(scriptUrl, options)`. See:
+   * https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerContainer/register
+   *
+   * @param {string} scriptUrl The service worker script associated with this
+   *     instance.
+   * @param {Object} options The service worker options associated with this
```
jeffposnick: `[options]` to make it clear that they're... optional?

packages/workbox-window/Workbox.mjs
```diff
@@ -0,0 +1,417 @@
...
+   *     instance.
+   */
+  constructor(scriptUrl, options = {}) {
+    super();
+
+    this._scriptUrl = scriptUrl;
+    this._swOptions = options;
+
+    // this._registration = null;
```
jeffposnick: Any reason to leave these here, commented out? If so, can you put in a `TODO:` or something similar for context?

packages/workbox-window/Workbox.mjs
```diff
@@ -0,0 +1,417 @@
...
+            'been registered. Create a new instance instead.');
+        return;
+      }
+    }
+
+    if (!immediate && document.readyState !== 'complete') {
+      await new Promise((res) => addEventListener('load', res));
+    }
+
+    // Create a local reference for better minification/
```
jeffposnick: Remove trailing `/`

packages/workbox-window/Workbox.mjs
```diff
@@ -0,0 +1,417 @@
...
+
+      if (this._controllingSw) {
+        if (urlsMatch(this._controllingSw.scriptURL, this._scriptUrl)) {
+          logger.debug('A service worker with the same script URL is ' +
+              'already controlling this page. This service worker will ' +
+              'remain active unless an update is found.');
+        } else {
+          logger.debug('A service worker with a different script URL is ' +
+              'currently controlling the page.');
+          logger.debug('Fetching the new script...');
```
jeffposnick: What does this refer to? A request that our library is making, or the fact that the browser is implicitly fetching the new SW script for us?

packages/workbox-window/Workbox.mjs
```diff
@@ -0,0 +1,417 @@
...
+      this._broadcastChannel.addEventListener('message', this._onMessage);
+    }
+    navigator.serviceWorker.addEventListener('message', this._onMessage);
+
+    // Keep track of when registration happened, so it can be used in the
+    // `this._onUpdateFound` heuristic. Also use the presence of this
+    // property as a way to see if `.register()` has been called.
+    this._registrationTime = performance.now();
+
+    // Expose and return the registration object.
```
jeffposnick: Drop "and return", since nothing is being `return`ed.

packages/workbox-window/Workbox.mjs
```diff
@@ -0,0 +1,417 @@
...
+        // install event is to observe that the time spent in the waiting phase
+        // is very short.)
+        // NOTE: we don't need separate timeouts for the new and updated SWs
+        // since they can't go through these phases at the same time.
+        this._waitingTimeout = setTimeout(() => {
+          // Ensure the SW is still waiting (it may now be redundant).
+          if (state === 'installed' && this._registration.waiting === sw) {
+            onWaiting();
+          }
+        }, 100);
```
jeffposnick: Can you put this `100` into a constant somewhere and document what it's used for? It's kind of magic as it is.

gulp-tasks/utils/build-window-bundle.js
```diff
@@ -0,0 +1,106 @@
...
+  const outputFilename = `${packageName}.${buildType.slice(0, 4)}.mjs`;
+  const outputDirectory = path.join(
+      packagePath, constants.PACKAGE_BUILD_DIRNAME);
+
+  logHelper.log(oneLine`
+    Building Window Bundle: ${logHelper.highlight(outputFilename)}`);
+
+  const plugins = rollupHelper.getDefaultPlugins(buildType, {module: true});
+
+  const validateModules = () => {
```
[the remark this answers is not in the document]
philipwalton: Sure, it was kind of a reminder for how to write these plugins, but I can always look it up again.

gulp-tasks/utils/build-window-bundle.js
```diff
@@ -0,0 +1,106 @@
...
+      // The final builds should have no warnings.
+      if (warning.code && warning.message) {
+        throw new Error(`Unhandled Rollup Warning: [${warning.code}] ` +
+          `${warning.message}`);
+      } else {
+        throw new Error(`Unhandled Rollup Warning: ${warning}`);
+      }
+    },
+  }).on('error', (err) => {
+    const args = [];
```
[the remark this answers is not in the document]
philipwalton: Yeah, this was copied directly from here: https://github.com/GoogleChrome/workbox/blob/v3.6.3/gulp-tasks/utils/build-browser-bundle.js#L175-L182

I could change both?

test/workbox-window/integration/test.js
```diff
@@ -0,0 +1,332 @@
...
+
+/**
+ * Executes the passed function (and args) async and logs any errors that
+ * occur. Errors are assumed to be passed to the callback as an object
+ * with the `error` property.
+ *
+ * @param {...*} args
+ * @return {*}
+ */
+const executeAsyncAndCatch = async (...args) => {
```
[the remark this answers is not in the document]
philipwalton: Yeah, I was going to, but I also wasn't super happy with it. I wish there was a better way to run `executeAsyncScript()` and have it automatically report errors, but since we can't pass helpers from the page scope in, it's really tricky.

packages/workbox-window/Workbox.mjs
philipwalton: I was using it as a way to keep all instance variables visible from the constructor (closure compiler actually requires this, which is where I got it from, but uglify doesn't remove them, so I commented them out).

gulp-tasks/utils/build-window-bundle.js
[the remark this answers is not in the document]
jeffposnick: I wouldn't go through too much effort to scrub old code, but I figured I'd flag it for incoming new code.

packages/workbox-window/Workbox.mjs
philipwalton: The latter. I could make it more obvious by saying "Fetching the new service worker script...". WDYT?

packages/workbox-window/Workbox.mjs
philipwalton: Good catch. I was originally returning the registration, but then opted not to do that.

packages/workbox-window/Workbox.mjs
philipwalton: Yeah, good call.

test/workbox-window/integration/test.js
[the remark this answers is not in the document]
jeffposnick: I'm fine with the actual code. It was mainly a comment about putting code that might be reused in multiple tests into its own standalone file in our `/infra/` directory.

Is there something about the way the integration test environment is set up that prevents that?

packages/workbox-window/Workbox.mjs
jeffposnick: How about `'The browser is fetching the new service worker script...'`. The main thing to get across is that `workbox-window` is just logging to provide info about something that the browser is doing anyway.

test/workbox-window/integration/test.js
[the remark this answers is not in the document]
philipwalton: Nope, it was just that I wasn't happy with it, so I didn't want to generalize it yet, but I certainly can.

packages/workbox-window/Workbox.mjs
philipwalton: I added a comment to make this clear. I can remove it later though if we don't like it.

test/workbox-window/integration/test.js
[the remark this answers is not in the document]
philipwalton: Added all the helpers to a new `/infra/testing/webdriver` directory.

philipwalton:
Ok, I think I've addressed all your comments. PTAL.

frlinw:
A `controlled` promise (@jeffposnick proposition from the past) would be a nice addition to avoid some boilerplate https://github.com/w3c/ServiceWorker/issues/799#issuecomment-165499718

jadjoubran:
I noticed while testing that the `workbox copyLibraries build/` wouldn't copy the *workbox-window* package. I personally don't need this, but let me know if you think other people would need it so I open an issue

philipwalton:
> I noticed while testing that the workbox copyLibraries build/ wouldn't copy the workbox-window package...if you think other people would need it so I open an issue

@jadjoubran yes, please do!

frlinw:
> it would only resolve on the page load that installed the SW for the very first time.

@philipwalton what do you think about that ?
```js
function controlled () {
  return new Promise((resolve) => {
    // resolve immediately if a service worker is already in control (every page loads after the first one)
    if (navigator.serviceWorker.controller) {
       return resolve()
    }
    // Or wait for the service worker to be in control (first page load)
    const controllerChangeHandler = (event) => {
      navigator.serviceWorker.removeEventListener('controllerchange', controllerChangeHandler)
      resolve()
    }
    navigator.serviceWorker.addEventListener('controllerchange', controllerChangeHandler)
  })
}
```
A variation (not sure about this one)
```js
async function doSomethingOnFirstInstallation(something) {
    if (navigator.serviceWorker.controller) {
       return
    }

    const controllerChangeHandler = (event) => {
      navigator.serviceWorker.removeEventListener('controllerchange', controllerChangeHandler)
      something()
    }
    navigator.serviceWorker.addEventListener('controllerchange', controllerChangeHandler)
  })
}
```

philipwalton:
Correct me if I'm wrong, but from your use cases, it sounds like you're primarily interested in knowing when a SW is in control the very first time. It doesn't sound like you're wanting to run any code *every* page load that a particular SW is in control over the page.

To do that in `workbox-window` currently, you'd do something like this:

```js
const wb = new Workbox('sw.js');
await wb.register();

wb.addEventListener('controlling', () => {
  // Run code as soon as the registered SW is in control of the page.
  // On subsequent page loads, this event handler will never run.
});

// Do other SW stuff here...
```

If we make this a `controlling` promise, it really won't change the code much:

```js
const wb = new Workbox('sw.js');
await wb.register();

wb.controlling.then(() => {
  // Run code as soon as the registered SW is in control of the page.
  // On subsequent page loads, this event handler will never run.
});

// Do other SW stuff here...
```

The only benefit I see to a promise is:

- async/await, but I always explains why I think that's problematic, since it may block subsequent code from running ever
- Promise.all() on multiple events (but I've yet to hear anyone ask for that).

frlinw:
Consider a web app (SPA) which requires some data relative to its global context (translation map, select list, w/e). These data are requested on every page load.
SPA just needs to be sure the fetch event will be handled by the service worker (to fetch these data from a local cache (indexeddb,...) instead of over the network).

philipwalton:
> SPA just needs to be sure the fetch event will be handled by the service worker (to fetch these data from a local cache (indexeddb,...) instead of over the network).

Sure, but consider all cases after the first time a user installs a SW for your origin (which most of the future of the web will entail). In these cases a `controlling` promise (as you've proposed) would often resolve to the previous app version's SW. Is that what you want?

I think the 99% use case is:
**is the SW that my app code was released with currently controlling the page?**

And in my experience, you only need a callback for that once it starts controlling. Future page loads can assume that the controlling code already ran (either via a previous page load, or via the SW's `activate` event). And if you really need code that runs on every page load but only if you're version of the SW is controlling the page, I'm not aware of any reliable way to do that without storing a SW version and sending a message to the SW to confirm the versions match. But I don't think Workbox could possibly do that (outside of a build tool).

Am I'm ignoring another use case?

frlinw:
> Is that what you want?
Yes, if I need the new one, I can handle service worker update and force a reload (or ask the user)

> Am I'm ignoring another use case?
I don't know :( still waiting for a 1%-man to decrease the 99% use case to 98%!

Just kidding, you are probably right about the 99% use case but I'm sure workbox-window ambition is not to abstract 10 loc of a basic "register & check update"

Issue #1830, opened 2019-01-11
Bundling changes/improvements (proposal)


philipwalton:
While it *is* possible to use Workbox with JavaScript bundlers, the process for doing so is not currently documented nor is it intuitive.

To make the experience better, I'd like to propose the following changes/improvement, which I've captured in this doc: https://docs.google.com/document/d/1xInIqp1IZZ8NG1H0Tbm1av28ej8N1obkMXnZ9ZPXWPc/edit?usp=sharing

If anyone has any suggestions or feedback on the proposal, please leave comments either here or in the doc.

philipwalton:
Progress so far:

- [x] Convert the properties of any default export objects to named exports in their own files.
- [x] Combine the _public.mjs, _default.mjs, index.mjs, and browser.mjs files into just index.mjs.
- [x] Update the build process to assign all named exports of a package to the respective workbox namespace object
- [x] Update `workbox-build` to not require the dot prefix on `.precacheAndRoute([])` in `injectManifesat()`.
- [ ] Rename npm packages to be consistent with workbox browser namespaces
- [x] Remove custom logging code from all production builds
- [x] Lazily instantiate the default router and precache controller
- [x] Remove the `skipWaiting` and `clientsClaim` helpers from workbox-sw

PR #1831, opened 2019-01-11
Bundling changes/improvements

philipwalton:
As a first step toward the changes proposed in #1830, I've updated the `workbox-routing` and `workbox-core` packages to give you an idea of what I'm thinking. If no one has any serious objections, I'll go ahead and update the rest of the packages as well.

The primary things I've done here are:

- Remove all default exports
- Move default export methods into their own files (exported individually)
- Removed `_public.mjs`, `_default.mjs`, and `browser.mjs` files
- Removed the log level code
- Updated the build script with the changes
- Moved the `skipWaiting()` and `clientClaim()` methods from `workbox-sw` to `workbox-core` (see [here](https://docs.google.com/document/d/1xInIqp1IZZ8NG1H0Tbm1av28ej8N1obkMXnZ9ZPXWPc/edit#heading=h.d9sslf54v5vc) for rationale)
- [Deprecated the strategy functions](https://docs.google.com/document/d/1xInIqp1IZZ8NG1H0Tbm1av28ej8N1obkMXnZ9ZPXWPc/edit#heading=h.f84urgepz8yo) in favor of using strategy classes (similar to how we do plugins).

R: @jeffposnick @gauntface


philipwalton:
I finished updating the rest of the packages, and I updated all the tests accordingly.

I know this is a lot of changes to review, but overall I think these changes make our package structure much simpler. Also, it looks like it reduced our overall code size by more than 1KB (gzipped) from the previous PR. We're now in the 8KB range! 🎉

packages/workbox-precaching/utils/deleteOutdatedCaches.mjs
```diff
@@ -28,7 +28,7 @@ const SUBSTRING_TO_FIND = '-precache-';
  * @private
  * @memberof module:workbox-precaching
  */
-export async function cleanupOutdatedCaches(currentPrecacheName,
+export async function deleteOutdatedCaches(currentPrecacheName,
```
jeffposnick: Do you want to convert this to `const deleteOutdatedCaches = (...) => {}` too? It looks like this PR includes a number of similar conversions.

packages/workbox-precaching/utils/getOrCreatePrecacheController.mjs
```diff
@@ -0,0 +1,24 @@
...
+
+let precacheController;
+
+/**
+ * @return {PrecacheController}
+ * @private
+ */
+export const getOrCreatePrecacheController = () => {
+  if (!precacheController) {
+    precacheController = new PrecacheController();
```
jeffposnick: Are we losing the ability to customize the precache name?

C.f. https://github.com/GoogleChrome/workbox/blob/next/packages/workbox-precaching/_default.mjs#L29-L30 which will pass in the result of `cacheNames.getPrecacheName()` to the `PrecacheController` constructor.

Or maybe there's some other mechanism of customizing the name now?

packages/workbox-precaching/utils/precachePlugins.mjs
```diff
@@ -0,0 +1,30 @@
...
+
+  Use of this source code is governed by an MIT-style
+  license that can be found in the LICENSE file or at
+  https://opensource.org/licenses/MIT.
+*/
+
+import '../_version.mjs';
+
+
+let plugins = [];
```
jeffposnick: This could be a `const`.

philipwalton:
> Marking workbox.strategies.<strategyName> deprecated in favor of new workbox.strategies.StrategyName() is also a pretty big change that got rolled this PR, which wasn't included in the description. 

Ahh, sorry, I updated that in the [proposal doc](https://docs.google.com/document/d/1xInIqp1IZZ8NG1H0Tbm1av28ej8N1obkMXnZ9ZPXWPc/edit#heading=h.f84urgepz8yo), but I forgot to also add it to the bullet points in this issue.

FWIW, I'm open to other options as well. Here are a few I think are find in a rough order of preference:

1. Switch to using constructors (e.g. `new CacheFirst()`) everywhere (like we do with plugins), but continue to support the old strategy functions for one major version and log a deprecation warning (what this PR does).
2. Keep the strategy functions and rename the strategy classes to something like `CacheFirstStrategy` (note we can't put them in a sub-folder because it would break the convention of keeping everything public in the top-level directory).
3. Consider the strategy classes private and only expose the strategy functions. (Someone would still extend a strategy class, they'd just have to instantiate one first, which is a bit awkward).

packages/workbox-precaching/utils/deleteOutdatedCaches.mjs
philipwalton: Sure.

BTW, I've personally always preferred inline `export` declarations, but I discovered this weekend that [they don't place nicely with jsdocs](https://github.com/jsdoc3/jsdoc/issues/1272), so I think it makes sense to always use an export block at the bottom of a file.

philipwalton: Ahh, now I see you were talking more about function declarations vs assigning an arrow function to a const.

packages/workbox-precaching/utils/getOrCreatePrecacheController.mjs
philipwalton: I don't think we're losing the ability to customize the name. Previously the code that instantiated the default `PrecacheController` looked like [this](https://github.com/GoogleChrome/workbox/blob/bab6be5/packages/workbox-precaching/_default.mjs#L29-L30), which passed in the current (or default) cache name, and the first line in `PrecacheController()` is [this](https://github.com/GoogleChrome/workbox/blob/bab6be502b3499e4d9bd12225716a298c6268395/packages/workbox-precaching/controllers/PrecacheController.mjs#L35), which uses the same method if no cache name was passed.

I believe in both before and after this PR, the way to change the cache name is still `workbox.core.setCacheNameDetails()`, and that needs to be done before using any `workbox-precaching` methods in both cases.

packages/workbox-precaching/utils/precachePlugins.mjs
philipwalton: Good call.

packages/workbox-precaching/utils/getOrCreatePrecacheController.mjs
philipwalton: Looking a bit closer at the code and the docs, it doesn't look like this is well documented or well tested. It's probably worth updating [this docs](https://developers.google.com/web/tools/workbox/guides/configure-workbox) to mention that setting of cache names needs to be done before loading any other modules.

jeffposnick:
I'm cool with moving to `new workbox.strategies.StrategyName()` in general. Having both ways of creating strategies has been a source of confusion in the past. My comment was mainly that we need to remember to call out that change, as it's going to be very visible to users.

https://github.com/GoogleChrome/workbox/blob/b8eb5a25a5b6f8c03dc3b7b58ed2d38a9f37ab96/packages/workbox-build/src/lib/runtime-caching-converter.js#L144-L145 should also get updated to use the constructor syntax. This PR is already large enough, so that could be deferred until after this PR is merged.

philipwalton:
Ok, I can address that in a new PR. The rest of the stuff you mentioned has been updated.

PR #1839, opened 2019-01-14
Update module versioning technique

philipwalton:
Our module versioning technique assigns a version property to a global variable on `self.workbox.v`. For normal CDN usage, this is fine because that variable exists, but in the case of people building their own service worker with Workbox, they're unlikely to import the `workbox-core/index.mjs` file (which defines that variable), and that means all of these try/catch blocks will fail.

The changes in this PR use a different technique to create an expression that will survive minification, and this one doesn't do any assignments, so the chances of side effects are very low. The only chance for side-effects is if there's already something defined on `self['workbox:v4:package-name']`, which is very unlikely, and even in the event that it happens, it'll execute a function that's also most likely undefined (so no side effects).

While no technique will be perfect, I think this is a little better than our current technique for the module use case, and it's a few bytes shorter.

Another slightly safer (but also longer) technique is as follows, which banks on the fact that a `typeof` expression will never result in a string like `'workbox:window:4.0.0-beta.0'`:

```js
try{typeof _=='workbox:window:4.0.0-beta.0'&&_()}catch(e){}
```

Thoughts?


prateekbh:
This is fine, what was stopping me from using CC in advanced mode was the absence of closure types, it needs externs and more to actually be able to work in advanced mode.

The above change doesnt really help there but should be good otherwise

philipwalton:
> The above change doesnt really help there but should be good otherwise

When i run this code through closure compiler I don't get any errors or warnings. Are you sure it's still an issue?

prateekbh:
My issue was that CC in advanced mode was deleting code from plugin code because the lifecycle methods were not directly referenced anywhere. 
But this is completely unrelated to this change.

PR #1841, opened 2019-01-15
Url to URL renaming for build interfaces

jeffposnick:
R: @philipwalton

Renaming of `'Url'` to `'URL'` for `workbox-build` and the associated interfaces.


packages/workbox-build/src/lib/checkForDeprecatedOptions.js
```diff
@@ -0,0 +1,26 @@
+/*
+  Copyright 2018 Google LLC
```
philipwalton: s/2018/2019/

(I'm not sure if we actually need to care about this?)

jeffposnick: I'll change it.

PR #1843, opened 2019-01-15
Support mjs files being published

philipwalton:
R: @jeffposnick 

Includes `.mjs` files in the set of build files that get updated to the CDN.


philipwalton:
Note, I'm not sure if `cdn-details.json` and `package-lock.json` need to be updatd, but I included them here because they were updated when I ran the `publish` task.

Issue #1856, opened 2019-01-23
Should workbox-window use promises instead of events?


philipwalton:
**Library Affected**:
*workbox-window*

All service worker [lifecycle events](https://docs.google.com/document/d/16BJDNhfqMjIvJ3_kR4lklpbliCGpi1dFhCdf7kHOsL8/edit#heading=h.dciabll4y3mv) dispatched by the `Workbox` object in the `workbox-window` package are only ever dispatched once (per `Workbox` instance). This design is intentional and meant to simplify the mental model of service worker registration/updates for most developer use cases (see the [`workbox-window` design doc](https://docs.google.com/document/d/16BJDNhfqMjIvJ3_kR4lklpbliCGpi1dFhCdf7kHOsL8/edit?usp=sharing) for more details).

But based on this [thread with @jakearchibald](https://twitter.com/philwalton/status/1086392135622877185), I'm considering converting these single-fire events to promises.

In the original design of `workbox-window`, these actually *were* promises, but I switched them to events because I was concerned that, with the rising popularity of async/await, developers might `await` promises that will never resolve.

For example, if the `Workbox` object's `activated` event were instead an `.activated` promise, it would only resolve 1) on page loads that register a particular SW for the very first time, or 2) page loads where the SW has updated. **However**, that means on many page loads it will never resolve, so it could easily lead to problems like this: 

```js
const wb = new Workbox('/sw.js');
wb.register();

await wb.activated;

// This line of code is intended to run on every page load, but because
// the above will never resolve on some page loads, this won't work.
wb.messageSW(...);
```

But when using `.addEventListener()`, that mistake is much less likely to happen.

```js
const wb = new Workbox('/sw.js');
wb.register();

wb.addEventListener('activated', () => {
  // Code that should only run when a new/updated SW has activated.
});

// This line will now run on every page load.
wb.messageSW(...);
```

Granted, you could use `.activated.then(...)` in every case where you could use `.addEventListener('activated', ...)`, but my concern is that the use of `.then()` seems to be going out of fashion.

Anyway, now that the initial version of the `workbox-window` library is available to try out, I'm curious to hear whether developers like the events API or whether they'd prefer a promises API. If we do switch, I'd prefer to do it before our initial release, so I'm hoping to get feedback on this quickly.

/cc @jakearchibald, @jeffposnick, @addyosmani, @developit, @prateekbh, @westonruter,  @jadjoubran, @gauntface (in case y'all have strong opinions)

[...]

PR #1861, opened 2019-01-28
Versioning features

philipwalton:
R: @jeffposnick

Adds the changes proposed in https://github.com/GoogleChrome/workbox/issues/1856#issuecomment-458010806.

*(Note, these are the initial changes needed for the proposal, but I have a few more ideas to make the bundle smaller before we merge.)*


gulp-tasks/utils/rollup-helper.js
```diff
@@ -18,7 +19,7 @@ module.exports = {
   // plugin set up and used to ensure as consist set of tests
   // as possible.
   getDefaultPlugins: (buildType, {module = false} = {}) => {
-    const plugins = [];
+    const plugins = [resolve()];
```
jeffposnick: Is this an intentional change? If so, what's it used for?

philipwalton: Yeah, this is because I'm loading `logger.mjs` and `Deferred.mjs` from `workbox-core` now, and since that's a `node_module` dependency, I need to use `rollup-plugin-node-resolve` or Rollup will error.

The reason this isn't needed for other Workbox packages is all SW packages [treat other package dependencies as external](https://github.com/GoogleChrome/workbox/blob/07bef20/gulp-tasks/utils/build-browser-bundle.js#L29-L90), thus there's nothing to resolve.

philipwalton:
Yup, I think I fixed the failing tests.

Issue #1864, opened 2019-01-29
Flaky [workbox-window] integration test


jeffposnick:
I'm seeing the following failure when run locally on OS X:

```
[Workbox]: Running Integration test on test/workbox-window/integration with NODE_ENV 'dev' and browser 'Google Chrome Beta'

<snip>

  1) [workbox-window]
       passes all unit tests:
     Error: async () => {
      return await webdriver.executeScript(() => self.mochaResults);
    } didn't return true after 120 retries.
      at module.exports (infra/testing/wait-until.js:17:9)
      at <anonymous>
```

PR #1876, opened 2019-01-31
Remove the scriptVersion option

philipwalton:
R: @jeffposnick

Based on in-person discussion, this PR removes the `scriptVersion` option from `Workbox` and reverts the constructor function signature to how it was before beta 2.


philipwalton:
Also, this most likely fixes #1864.

Issue #1435, opened 2018-04-16
Samsung Internet browser throws an error when cached resources get requested


Schepp:
**Library Affected**:
workbox-sw 3.0.0

**Browser & Platform**:
Samsung Internet 6.4.10.5

**Issue or Feature Request Description**:
I created a Service Worker with the help of Workbox. This is the SW's source code:

https://gist.github.com/Schepp/adbc1579e8d1a38087e087a40b205322

This is the page in which context it is running:
https://templates.park.works
User: templates
PW: gera16

Upon page load, I perform a SW precaching of HTML pages, by searching all `<a>` elements for URLs and by prefetching them via `<link rel="prefetch">` element. When idle the browser fetches these URLs and the SW then caches them in the background (see source code above).

This works well on Chrome and Safari. On Samsung Internet though, the page only works upon the first visit. Once I reload the page, or navigate to a linked page, this is what I get:

![image](https://user-images.githubusercontent.com/135141/38805935-f7e1f5b0-4177-11e8-9d0b-a391ad636003.png)

When I look into Samsung Internet's `internet://serviceworker-internals` I spot these error messages:

```
Console: {"lineNumber":0,"message":"The FetchEvent for \"https://templates.park.works/\" resulted in a network error response: the promise was rejected.","message_level":2,"sourceIdentifier":1,"sourceURL":""}
Console: {"lineNumber":1,"message":"Uncaught (in promise) InvalidStateError: Failed to execute 'waitUntil' on 'ExtendableEvent': The event handler is already finished.","message_level":3,"sourceIdentifier":1,"sourceURL":"https://storage.googleapis.com/workbox-cdn/releases/3.0.0/workbox-strategies.prod.js"}
Console: {"lineNumber":0,"message":"The FetchEvent for \"https://templates.park.works/\" resulted in a network error response: the promise was rejected.","message_level":2,"sourceIdentifier":1,"sourceURL":""}
Console: {"lineNumber":1,"message":"Uncaught (in promise) InvalidStateError: Failed to execute 'waitUntil' on 'ExtendableEvent': The event handler is already finished.","message_level":3,"sourceIdentifier":1,"sourceURL":"https://storage.googleapis.com/workbox-cdn/releases/3.0.0/workbox-strategies.prod.js"}
[...]

PR #1883, opened 2019-02-04
Update the workbox-expiration IDB data model

### LINKED ISSUE #1398: workbox-precaching and workbox-cache-expiration could use the same IDB database name.
**Library Affected**:
*workbox-precaching, workbox-cache-expiration*

Both of these modules use the cache name as the IDB name, but this means if I pass the cache expiration plugin to `precache.activate()`, it errors because the names are the same.

I think all packages that create IDB databases should (wherever possible) use the package name as the IDB database name. If more than one database is needed, then we can use the package name as a prefix. This will eliminate the possibility of naming collisions.

In the case of the cache expiration plugin, only one database is needed. Instead of creating multiple databases, we can create a single database with a single object store and use the cache name as an indexed property in the object store. We take this same approach with tag names in `workbox-background-sync`.

### LINKED ISSUE #1400: Cache expiration should have an option to base maxEntries on last accessed time
**Library Affected**:
*workbox-cache-expiration*

Currently the cache expiration plugin takes a `maxEntries` option that removes old entries based on their added date. I think it'd be a nice addition to add an option to remove entires based on their "last accessed" date.

The use cases is a static assets cache where JS or CSS entries could quickly become obsolete, but image or SVG assets might have been originally added long in the past but are still currently referenced on the page.

Expiring entries based on their "last accessed" date would help provide a convenient way to auto expire assets that are no longer referenced.

philipwalton:
R: @jeffposnick 

Fixes #1398, Fixes #1400. 

This PR updates the IDB data model used by `workbox-expiration`, switching it to from multiple databases (named the same as the cache name), to  a single database (named `workbox-expiration`) where the entries in the database have a `cacheName` property.

I've also updated how expiration happens to make it more efficient. Previously the entire contents of the object store would be read out, the purgeable entires were determined, and then in a second transaction those entries would be removed. In this PR I've moved the logic that determines what needs to be expired into `CacheTimestampsModel`, so it can all happen inside a single transaction.

Lastly, I had to update our IDB mocks yet again due to bugs I found when running tests. I've manually confirmed that the logic in this PR does work in Chrome, FF, and Safari's IDB implementations.

### DIFF
--- a/packages/workbox-expiration/Plugin.mjs
+++ b/packages/workbox-expiration/Plugin.mjs
@@ -10,7 +10,8 @@ import {CacheExpiration} from './CacheExpiration.mjs';
 import {WorkboxError} from 'workbox-core/_private/WorkboxError.mjs';
 import {assert} from 'workbox-core/_private/assert.mjs';
 import {cacheNames} from 'workbox-core/_private/cacheNames.mjs';
-import {registerQuotaErrorCallback} from 'workbox-core/index.mjs';
+import {registerQuotaErrorCallback}
+  from 'workbox-core/registerQuotaErrorCallback.mjs';
 
 import './_version.mjs';
 
@@ -104,7 +105,7 @@ class Plugin {
 
   /**
    * A "lifecycle" callback that will be triggered automatically by the
-   * `workbox.runtimeCaching` handlers when a `Response` is about to be returned
+   * `workbox.strategies` handlers when a `Response` is about to be returned
    * from a [Cache](https://developer.mozilla.org/en-US/docs/Web/API/Cache) to
    * the handler. It allows the `Response` to be inspected for freshness and
    * prevents it from being used if the `Response`'s `Date` header value is
@@ -119,7 +120,7 @@ class Plugin {
    *
    * @private
    */
-  cachedResponseWillBeUsed({cacheName, cachedResponse}) {
+  cachedResponseWillBeUsed({event, request, cacheName, cachedResponse}) {
     if (!cachedResponse) {
       return null;
     }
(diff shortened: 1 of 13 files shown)


packages/workbox-expiration/Plugin.mjs
```diff
@@ -131,6 +132,13 @@ class Plugin {
     const cacheExpiration = this._getCacheExpiration(cacheName);
     cacheExpiration.expireEntries();
 
+    // Update the metadata for the request URL to the current timestamp,
+    // but don't `await` it as we don't want to block the response.
+    const updateTimestampDone = cacheExpiration.updateTimestamp(request.url);
+    if (event) {
+      event.waitUntil(updateTimestampDone);
```
jeffposnick: You're going to need to put this in a `try`/`catch`, à la https://github.com/GoogleChrome/workbox/pull/1457/files

[Samsung Internet 6.4](https://developer.samsung.com/internet/android/releases), which I believe we're still trying to support, breaks when you call `event.waitUntil()` asynchronously: https://github.com/GoogleChrome/workbox/issues/1435#issuecomment-385690454

packages/workbox-expiration/Plugin.mjs
```diff
@@ -155,7 +163,7 @@ class Plugin {
       return true;
     }
 
-    // If we have a valid headerTime, then our response is fresh iff the
+    // If we have a valid headerTime, then our response is fresh if the
```
jeffposnick: https://en.wikipedia.org/wiki/If_and_only_if

philipwalton: Ahh. I just saw the red underline in my editor and changed it without thinking.

philipwalton: Good catch. Updated.

PR #1906, opened 2019-02-18
Add a cache listener message response

philipwalton:
R: @jeffposnick @philipwalton

When writing the `workbow-window` documentation, I realized that the logic in `Router#addCacheListener` includes a check for `meta === 'workbox-window'`, but we shouldn't require developers to be use `workbox-window` to cache URLs (or ask them to set that meta value themselves, which wouldn't make sense).

This PR removes that check. It also adds a response to the message event if it contains a message channel (which `workbox-window` message do by default). This allows users to await successful caching of all URLs, similar to what you get with precaching.

Note: I did not add a message response in the event of an error, but I potentially could? WDYT?

### DIFF
--- a/packages/workbox-routing/Router.mjs
+++ b/packages/workbox-routing/Router.mjs
@@ -74,7 +74,6 @@ class Router {
    * ```
    * {
    *   type: 'CACHE_URLS',
-   *   meta: 'workbox-window',
    *   payload: {
    *     urlsToCache: [
    *       './script1.js',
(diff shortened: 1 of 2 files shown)


packages/workbox-routing/Router.mjs
```diff
@@ -86,21 +85,27 @@ class Router {
...
+        const requestPromises = payload.urlsToCache.map((entry) => {
           if (typeof entry === 'string') {
             entry = [entry];
           }
 
           const request = new Request(...entry);
-          this.handleRequest({request, event});
+          return this.handleRequest({request, event});
+        });
+
```
jeffposnick: I realize now that there should be an `event.waitUntil(requestPromises)` here (wrapped in a `try`/`catch` to handle browsers that don't support async `waitUntil`).

https://github.com/w3c/ServiceWorker/issues/669#issuecomment-402779578

jeffposnick: Nevermind! `handleRequest()` already calls `waitUntil()`.

philipwalton: I believe all the built-in strategy handlers already add their own try/catch logic, so I don't think this is necessary here.

PR #1928, opened 2019-02-28
Call bind() to get the right this in precaching's plugins

### LINKED ISSUE #1927: Unable to use `cacheableResponse.Plugin` with precaching.addPlugins()
There are reports of issues in v4 using code along the lines of 

```
workbox.precaching.addPlugins([
  new workbox.cacheableResponse.Plugin({
    statuses: [200],
  }),
]);
```

It leads to

```
Uncaught (in promise) TypeError: Cannot read property '_cacheableResponse' of undefined
    at cacheWillUpdate (Plugin.mjs:45)
    at PrecacheController._addURLToCache (PrecacheController.mjs:186)
cacheWillUpdate @ Plugin.mjs:45
_addURLToCache @ PrecacheController.mjs:186
```

The code in question is https://github.com/GoogleChrome/workbox/blob/d76e77066e9a23cb33f99a5dde796fa81bc0ea93/packages/workbox-cacheable-response/Plugin.mjs#L45

I believe the underlying issue is not properly binding `this` when the plugin is called.

jeffposnick:
R: @philipwalton 

Fixes #1927 by properly `bind()`ing to the `Plugin` instance.

This also adds to the integration test for `workbox-precaching` to ensure that the `addPlugins()` flow has better test coverage.

### DIFF
--- a/test/workbox-precaching/node/test-PrecacheController.mjs
+++ b/test/workbox-precaching/node/test-PrecacheController.mjs
@@ -442,8 +442,6 @@ describe(`[workbox-precaching] PrecacheController`, function() {
     });
 
     it(`it should use the proper 'this' when calling a cacheWillUpdate plugin`, async function() {
-      sandbox.spy(cacheWrapper, 'put');
-
       const precacheController = new PrecacheController();
       const cacheList = [
         '/index.1234.html',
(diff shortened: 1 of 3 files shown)


jeffposnick:
Sure, I just added one.

test/workbox-precaching/node/test-PrecacheController.mjs
```diff
@@ -441,6 +441,31 @@ describe(`[workbox-precaching] PrecacheController`, function() {
...
+
+      const precacheController = new PrecacheController();
+      const cacheList = [
+        '/index.1234.html',
+      ];
+      precacheController.addToCacheList(cacheList);
+
+      class TestPlugin {
+        cacheWillUpdate({response}) {
+          expect(this).to.be.an.instanceof(TestPlugin);
```
philipwalton: Can you use make this a spy and then at the bottom of the test use `spy.thisValues` or `spy.calledOn()` to assert the correct context?

As it is now, if you changed the file to not call `cacheWillUpdate` at all, this test would still pass.

jeffposnick: Oh, I didn't know about `thisValues`.

PR #1932, opened 2019-03-02
Bg sync array buffer

### LINKED ISSUE #1908: StorableRequest should not store blobs, as they don't always work in Safari
**Library Affected**:
*workbox-background-sync*

As part of my ongoing work to get all our unit tests running in real browsers, I've discovered that in some cases Safari will fail to store a `Blob` in IndexedDB, which is needed for requests with a body in `workbox-background-sync`.

We can get around this by using an `ArrayBuffer` instead, which should be a fairly straightforward change.

philipwalton:
R: @jeffposnick

Fixes #1908. This PR mostly cherry-picks commits from the WIP [unit-test-updates](https://github.com/GoogleChrome/workbox/tree/unit-test-updates) branch (the first two commits), which means there's some extra stuff in here that might not seem relevant, but it was easier to just cherry-pick existing work that to redo everything. Also that should make merging those changes cleaner in the future.

I kept the tests mostly verbatim from the `unit-tests-updates` branch  since I knew they were passing in all browsers, then I updated the mocks in this branch until I could get them to pass here (I opted to not change the tests since I know they were passing and run correctly in browsers).

The most relevant changes are to the [StorableRequest class here](https://github.com/GoogleChrome/workbox/compare/bg-sync-array-buffer?expand=1#diff-e8886ed06b9cceabf23569b770bed554). The rest of the changes include several small compat issues (e.g. some stuff in the DBWrapper library) I found and fixed in the other branch.


.eslintrc.js
```diff
@@ -21,15 +21,15 @@ module.exports = {
...
+    "jsdoc/check-types": 2,
+    "jsdoc/newline-after-description": 2,
+    "max-len": [2, {
       code: 80,
       tabWidth: 2,
       ignoreComments: true,
-      ignoreUrls: true,
-      ignorePattern: '^\\s*import',
-    }],
+      "ignorePattern": "^\\s*import",
```
jeffposnick: These don't need quotes (it's a JS file, not JSON).

And elsewhere where there are quotes, could you switch them back to single and not double?

philipwalton: I ran eslint or the `.eslintrc.js` file (not sure why it wasn't already doing that?).

philipwalton:
I'd rather not put any effort into using their mocks instead of ours since I plan (as the very next thing I work on) to move off the mocks entirely and switch to testing in real browsers.

We could let them know about these changes, or submit a PR with them though.

PR #1963, opened 2019-03-15
Allow wb.removeEventListener to work without an exception

### LINKED ISSUE #1962: wb.removeEventListener throws an exception
**Library Affected**:
workbox-window

**Browser & Platform**:
all browsers

I'm getting a "remove is not a function" when I do a `wb.removeEventListener` where `wb` is the results of doing a `new Workbox("service-worker.js");`

My actual use case is much more complicated (I'm adding and removing from a `useEffect` hook in React), but for the sake of reproducing the error, I've boiled is down to this:
```js
const wb = new Workbox('service-worker.js');
wb.addEventListener('waiting', handler);
wb.removeEventListener('waiting', handler); // <= this is where I get the exception
```

Obviously, you would never actually turn around an immediately remove a listener once it was added.

**complete log**
```
workbox-window.prod.es5.mjs:4 Uncaught TypeError: this.M(...).remove is not a function
    at v.t.removeEventListener (workbox-window.prod.es5.mjs:4)
    at eval (useServiceWorkerUpdate.js:33)
    at commitHookEffectList (react-dom.development.js:17231)
    at commitPassiveHookEffects (react-dom.development.js:17260)
    at HTMLUnknownElement.callCallback (react-dom.development.js:149)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:199)
    at invokeGuardedCallback (react-dom.development.js:256)
    at commitPassiveEffects (react-dom.development.js:18728)
    at wrapped (scheduler-tracing.development.js:207)
    at flushFirstCallback (scheduler.development.js:107)
```

donavon:
R: @jeffposnick @philipwalton

Fixes #1962 

`Set` does not have a `remove` method. It should be `delete`.

### DIFF
--- a/packages/workbox-window/utils/EventTargetShim.mjs
+++ b/packages/workbox-window/utils/EventTargetShim.mjs
@@ -36,7 +36,7 @@ class EventTargetShim {
    * @param {Function} listener
    */
   removeEventListener(type, listener) {
-    this._getEventListenersByType(type).remove(listener);
+    this._getEventListenersByType(type).delete(listener);
   }
 
   /**
--- a/test/workbox-window/unit/test-Workbox.mjs
+++ b/test/workbox-window/unit/test-Workbox.mjs
@@ -868,5 +868,30 @@ describe(`[workbox-window] Workbox`, function() {
         expect(externalActivated2Spy.callCount).to.equal(0);
       });
     });
+
+    describe(`removeEventListener()`, function() {
+      it(`will register and then unregister event listeners of a given type`, function() {
+        const eventType = 'testEventType';
+        const event = {type: eventType};
+        const eventListener1 = sandbox.stub();
+        const eventListener2 = sandbox.stub();
+
+        const wb = new Workbox(uniq('sw-clients-claim.tmp.js'));
+        wb.addEventListener(eventType, eventListener1);
+        wb.addEventListener(eventType, eventListener2);
+
+        wb.dispatchEvent(event);
+        expect(eventListener1.calledOnceWith(event)).to.be.true;
+        expect(eventListener2.calledOnceWith(event)).to.be.true;
+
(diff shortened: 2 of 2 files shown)


jeffposnick:
Thanks for the catch, @donavon!

One thing that this points to is that the `removeEventListener()` code path is currently untested. Would you mind if I pushed an additional commit to this PR that added a test to ensure we will catch any regressions?

donavon:
please do

jeffposnick:
Merging, unless @philipwalton chimes in with an objection related to the code for the new test...

philipwalton:
Thanks @donavon! And yeah, embarrassing that we weren't testing the `removeEventListener` flow.

@jeffposnick I'm fine with merging this as is, but I may refactor the test later and move it into a file that specifically tests the `EventTargetShim` module

donavon:
No prob. Please let me know when and what version I can expect an npm package.

philipwalton:
We have a few bug fixes that might warrant a patch version push. @jeffposnick what do you think? I'm fine with releasing this and the `getSW()` fix right away unless you have a reason to hold off?

jeffposnick:
No reason to hold off from my side of things.

philipwalton:
Ok, I'll do a `4.1.1` release later this afternoon then.

philipwalton:
`4.1.1` has been published 🎉

donavon:
Thx!

I also found this on your docs site. It needs updated as well.

https://developers.google.com/web/tools/workbox/reference-docs/latest/workbox-window_utils_EventTargetShim.mjs

philipwalton:
Thanks, that file actually shouldn't even be in the docs since it's private. I've filed an issue to fix that: https://github.com/GoogleChrome/workbox/issues/1964

PR #1965, opened 2019-03-19
Run background-sync tests in a SW environment

philipwalton:
R: @jeffposnick 

This PR is the first step in moving our unit tests from running in node to running in a real service worker environment. In this PR the `workbox-background-sync` unit tests previously in the `/tests/workbox-background-sync/node` directory and been moved to the `/tests/workbox-background-sync/sw` directory and all node-based imports have been removed.

To make this work, I've added a few templates and routes, so that you can start the test server (`gulp test-server`) and navigate to the following URL and see the tests run:
http://localhost:3004/test/workbox-background-sync/sw/

When navigating to this URL, three main things are happening:

- The [`templates/test-sw.html.njk`](https://github.com/GoogleChrome/workbox/compare/sw-testing-changes?expand=1#diff-dec6373b990b9ec6066570e312d8833f) template is rendered...
- which registers a service worker file at [`templates/test-sw-runner.js.njk`](https://github.com/GoogleChrome/workbox/compare/sw-testing-changes?expand=1#diff-c4b637a2b5db377fb853778d7e1e4cab)...
- and that imports the test bundle via the [`routes/sw-bundle.js`](https://github.com/GoogleChrome/workbox/compare/sw-testing-changes?expand=1#diff-e675c7bf88d939a991887adab052f169) endpoint

The tests are currently being run via webdriver in the integration tests. In the future we might want to consider separating this out into its own `gulp test-unit` tasks, but while this is still a work in progress, running them via the integration tests tasks seems fine.

Note: this setup required very few changes to the test files themselves, so hopefully it's a rather painless process to switch them all over.


infra/testing/server/routes/test-sw.js
```diff
@@ -0,0 +1,37 @@
...
+const globby = require('globby');
+const path = require('path');
+const templateData = require('../template-data');
+
+
+const match = '/test/:packageName/sw/';
+
+async function handler(req, res) {
+  const {packageName} = req.params;
+  const testFiles = await globby(`test/${packageName}/sw/**/test-*.mjs`) || [];
```
jeffposnick: Any reason to introduce `globby` here? We're using `glob` elsewhere in the project. Adding extra `devDependencies` isn't a big deal, but maybe we should consider migrating elsewhere if `globby` is better.

infra/testing/server/templates/test-sw-runner.js.njk
```diff
@@ -0,0 +1,86 @@
...
+
+        reports.push({
+          name: flattenTitles(test),
+          result: false,
+          message: err.message,
+          stack: err.stack,
+        });
+      });
+
+      runner.on('end', async (...args) => {
```
jeffposnick: Remove `...args` if they're not being used.

infra/testing/server/templates/test-sw-runner.js.njk
```diff
@@ -0,0 +1,86 @@
...
+
+mocha.setup({
+  ui: 'bdd',
+  timeout: 5000,
+  reporter: null,
+});
+
+addEventListener('message', (event) => {
+  if (event.data && event.data.type === 'RUN_TESTS') {
+    const testsComplete = new Promise((resolve, reject) => {
```
jeffposnick: You never call `reject`, right? I'd remove it from the argument list in that case.

infra/testing/server/templates/test-sw.html.njk
```diff
@@ -0,0 +1,59 @@
...
+  </ul>
+  <p>Open the console to see progress...</p>
+  <p><strong id="results"></strong></p>
+
+  <script type="module">
+    import {Workbox} from '/__WORKBOX/buildFile/workbox-window';
+
+    (async () => {
+      // // Randomize the test URL so every test run forces a new SW install.
+      const wb = new Workbox('test-sw-runner.js.njk?v=' + Math.random());
```
jeffposnick: How would you feel about using `${Date.now()}-${global.swCount++}` instead of a random number? I could see it being sorta useful from a debugging perspective if you take a look at two SW registrations and can tell in which order they were made, but also... I tend to avoid random numbers in tests?

test/workbox-background-sync/sw/test-Queue.mjs
```diff
@@ -22,52 +19,51 @@ const getObjectStoreEntries = async () => {
...
-//   for (const store of Array.from(db.db.objectStoreNames)) {
-//     await db.clear(store);
-//   }
-//   await db.close();
-// };
+const clearIndexedDBEntries = async () => {
+  // Open a conection to the database (at whatever version exists) and
+  // clear out all object stores. This strategy is used because deleting
+  // databases inside service worker is flaky in FF and Safari.
+  // TODO(philipwalton): the version is not needed in real browsers, so it
```
jeffposnick: That TODO suggests that we could remove this code, right? Although this whole block was itself commented out... so I'm confused.

test/workbox-background-sync/sw/test-Queue.mjs
```diff
@@ -22,52 +19,51 @@ const getObjectStoreEntries = async () => {
...
+  // can be removed when we move away from running tests in node.
+  const db = await new DBWrapper('workbox-background-sync').open();
+
+  // Edge cannot convert a DOMStringList to an array via `[...list]`.
```
jeffposnick: Have you tried this new testing setup Edge, by the way?

package.json
```diff
@@ -105,6 +107,18 @@
     "sinon": "^6.3.4",
     "tempy": "^0.2.1",
     "url-search-params": "^1.1.0",
-    "webpack": "^4.19.1"
+    "webpack": "^4.19.1",
+    "workbox-background-sync": "file:./packages/workbox-background-sync",
```
jeffposnick: You mentioned out of band that this was required for Rollup to bundle everything, but I am not sure how nicely this is going to play with Lerna moving forward. (To be clear, I'm not sure one way or another.)

Can you get away with, e.g., using the [following option](https://github.com/rollup/rollup-plugin-node-resolve) in `rollup-plugin-node-resolve` instead?

```
// Any additional options that should be passed through
// to node-resolve
customResolveOptions: {
  moduleDirectory: 'js_modules'
}
```

infra/testing/server/routes/build-file.js
```diff
@@ -7,35 +7,39 @@
...
+  const pkgDir = path.resolve('packages', pkg.name);
+
+
+  let file;
+  if (packageFile.includes( '.')) {
+    file = path.join(pkgDir, 'build', packageFile);
+  } else {
+    const pkg = outputFilenameToPkgMap[packageFile];
+
+    // If the pkg.module references something in the build directory, use
```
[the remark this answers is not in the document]
philipwalton: We don't currently build modules versions of our SW libraries, but I actually do think we should. The guidance in [this section of our bundlers guide](https://developers.google.com/web/tools/workbox/guides/using-bundlers#keeping_dev-only_code_out_of_the_bundle) could be simplified if we did.

infra/testing/server/routes/sw-bundle.js
```diff
@@ -0,0 +1,44 @@
...
+      multiEntry(),
+      resolve(),
+      replace({
+        'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV),
+      }),
+    ],
+    cache,
+  });
+
+  // Update the cache so subsequent bundles are faster.
```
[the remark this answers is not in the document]
philipwalton: We definitely don't need to kill the test server each time. I confirmed that it both does pick up the changes as well as *seems* to build faster the second time.

It's not clear to me how much it's saving -- you can look at the cache object and see what it's storing, and it's not storing things like last modified time, so it may still be reading the file again.

infra/testing/server/templates/test-sw.html.njk
philipwalton: SG

package.json
philipwalton: Seems to work fine, I'll switch to this.

test/workbox-background-sync/sw/test-Queue.mjs
philipwalton: I tried with my previous setup, but not with this one yet. I will try though.

infra/testing/server/templates/test-sw.html.njk
philipwalton: Added a `uniqueID()` function to the template data, which just increments a counter.

test/workbox-background-sync/sw/test-Queue.mjs
philipwalton: At some point I had the DB version hard-coded in there (which was required by the mocking library). Since that's not there anymore I think I can just remove the TODO comment.

PR #1970, opened 2019-03-20
Migrate core tests to run in a SW environment

philipwalton:
R: @jeffposnick

Continues the test-conversion work started in #1965.

One note here: I added `rollup-plugin-commonjs` because the `workbox-core` tests include lots of helpers and I didn't feel like adding them all to the global scope like I'd done previously.

Once all this migration work is done, we can probably do a pass to clean up these helpers and put them in one common location, but for now I'd rather the diffs be as minimal as possible while migrating, so it's clear what changes were needed for the tests to pass in browsers.


packages/workbox-core/_private/cacheWrapper.mjs
```diff
@@ -64,10 +64,10 @@ const putWrapper = async ({
   }
 
   if (process.env.NODE_ENV !== 'production') {
-    if (responseToCache.method && responseToCache.method !== 'GET') {
+    if (request.method && request.method !== 'GET') {
```
jeffposnick: Good catch. My assumption is that this is not a breaking change, as the `cache.put()` later on would have rejected if the `request`'s `mode` wasn't already `'GET'`.

infra/testing/server/routes/sw-bundle.js
```diff
@@ -10,12 +10,16 @@ const {rollup} = require('rollup');
 const replace = require('rollup-plugin-replace');
 const resolve = require('rollup-plugin-node-resolve');
 const multiEntry = require('rollup-plugin-multi-entry');
+const commonjs = require('rollup-plugin-commonjs');
 
 
 const match = '/test/:package/*/sw-bundle.js';
-let cache;
+
+let caches = {};
```
jeffposnick: Nit, but this can be `const` now.

PR #1981, opened 2019-03-22
Support for navigationPreload in workbox-build

### LINKED ISSUE #1915: Configure navigationPreload via workbox-webpack-plugin
**Library Affected**:
workbox-webpack-plugin

**Browser & Platform**:
all browsers (that support navigation preload)

It doesn't appear that `workbox-webpack-plugin` exposes a configuration option to enable `navigationPreload`, forcing a fallback to `importScripts` and `workbox.navigationPreload.enable();`. Did I miss something? And if not, is this something that could be added?

merrywhether:
R: @jeffposnick @philipwalton

Fixes #1915

Adds `navigationPreload` config property (defaulting to false) to `workbox-build`'s `generateSW` and `generateSWString` modes, which would also expose it to the wrappers like `workbox-cli` and `workbox-webpack-plugin`.

`gulp lint test` passes locally. I've been added to my company's CCLA group.


packages/workbox-build/src/templates/sw-template.js
```diff
@@ -19,6 +19,8 @@ module.exports = `/**
  * See https://goo.gl/2aRDsh
  */
 
+<% if (navigationPreload) { %>workbox.navigationPreload.enable();<% } %>
```
merrywhether: I wasn't entirely sure where this should go in the template order-wise. From the docs, it looked like it should go ahead of strategies and routing config.

merrywhether:
Travis fails on:

>[Workbox]: Running Integration test on test/workbox-expiration/integration with NODE_ENV 'production' and browser 'Firefox Beta'
>  [workbox-expiration]
>1 test failure(s):
>Name     :  CacheExpiration expireEntries() should expire and delete entries beyond maximum entries
>Message  :  expected [ Array(2) ] to deeply equal [ 'http://localhost:3004/first' ]

I didn't see this locally. Is this me?

jeffposnick:
Thanks for the contribution! I'll give it a review soon.

packages/workbox-build/src/templates/sw-template.js
jeffposnick: Under the hood, this just installs an `activate` event listener, which won't run until that lifecycle event is fired. So it doesn't matter too much where the code is located.

merrywhether:
@jeffposnick We're actually starting slow with service workers and testing them out only on images (for which we have defined `runtimeCaching`) to begin with while we verify minimal performance regressions from cold starts, etc. As we fully SSR our pages (and they update rather often), we were intending to use `navigationPreload` to allow the browser to fetch the html immediately and are (at least for now) implicitly falling back to the `NetworkOnly` strategy for handling these requests. It sounds like we're potentially wrong about the behavior and would need to explicitly list a `NetworkOnly` strategy for our non-image routes in order for navigation preload to work? Especially if that's the case, ensuring at least something exists in `runtimeCaching` seems like a great way to help people towards proper configuration.

jeffposnick:
So if you enable navigation preload but don't actually make use of the response in a `fetch` handler, you'll end up seeing something like the following logged:

> The service worker navigation preload request was cancelled before 'preloadResponse' settled. If you intend to use 'preloadResponse', use `waitUntil()` or `respondWith()` to wait for the promise to settle.

There's no implicit usage of the preloaded responses when you don't respond to the `fetch` event.

You need to do something like https://developers.google.com/web/tools/workbox/modules/workbox-navigation-preload#basic_usage, i.e. at the minimum, set up a `NetworkOnly` strategy that will match your navigations and end up using the responses.

In the `GenerateSW` use case, setting that up would require using `runtimeCaching`, along the lines of:

```
{
  runtimeCaching: [{
    // urlPattern supports matchCallback functions in addition to RegExps...
    urlPattern: ({event}) => event.request.mode === 'navigate',
    handler: 'NetworkOnly',
  }]
}
```

It sounds like using `runtimeCaching` along with `navigatePreload: true` is a worthwhile restriction. I'm happy to modify your PR to add in that additional validation rule, or you could if you feel comfortable with `joi`'s syntax.

merrywhether:
@jeffposnick Thanks for pointers, makes the PR effort even more useful to us!

I went ahead and updated the `joi` validation logic and updated the tests to match.

test/workbox-build/node/entry-points/generate-sw-string.js
```diff
@@ -521,6 +522,57 @@ describe(`[workbox-build] entry-points/generate-sw-string.js (End to End)`, func
...
+      try {
+        await generateSWString(options);
+        throw new Error('Unexpected success.');
+      } catch (error) {
+        expect(error.name).to.eql('ValidationError');
+        expect(error.details[0].context.key).to.eql('runtimeCaching');
+      }
+    });
+
+    it(`should reject with a ValidationError when 'navigationPreload' is true and 'runtimeCaching' is malformed`, async function() {
```
merrywhether: This test wasn't strictly necessary but added to ensure that `joi`'s `when` syntax didn't clobber the underlying validation and figured it couldn't hurt to keep.

jeffposnick: 👍

PR #1985, opened 2019-03-26
Migrate google analytics tests to run in a SW environment

philipwalton:
R: @jeffposnick

Continues the test-conversion work started in #1965. I also added the `extendable-event-utils.mjs` helpers, which makes it easier to test dispatching events and awaiting their extend lifetime promises.


test/workbox-google-analytics/sw/test-initialize.mjs
```diff
@@ -271,35 +283,37 @@ describe(`[workbox-google-analytics] initialize`, function() {
...
-
-    await eventsDoneWaiting();
+    // Manually trigger the `onSync` callback in both sync and non-sync
+    // supporting browsers.
+    if ('sync' in registration) {
+      await dispatchAndWaitUntilDone(new SyncEvent('sync', {
+        tag: `workbox-background-sync:workbox-google-analytics`,
+      }));
+    } else {
+      const queue = Queue.prototype.pushRequest.thisValues[0];
```
jeffposnick: Nit: I found this line to be confusing, and had to trace things back to earlier in the function to figure out that there was a `Spy` set up for `Queue.prototype.pushRequest`, and that `thisValues[0]` was a way of reading info from the `Spy`. 

Would you mind changing this (and the one other usage elsewhere) to something more explicit, like:

```
const pushRequestSpy = sandbox.spy(Queue.prototype, 'pushRequest');
// Later...
const queue = pushRequestSpy.thisValues[0];
```

philipwalton: Done.

PR #1989, opened 2019-03-28
Migrate precaching tests to run in a SW environment

philipwalton:
R: @jeffposnick

Continues the test-conversion work started in #1965. I also found a mistake in my previous [navigation-preload PR](#1987) where the tests weren't being run. I've fixed that in this PR.


philipwalton:
Yeah, unfortunately the diffs here aren't super helpful since the file names changed, so it's not showing them side-by-side (most of the logic stayed the same).

PR #1990, opened 2019-03-28
Adds cacheKeyWillBeUsed lifecycle callback

### LINKED ISSUE #1709: Support for overriding URL used when writing to the cache

**Library Affected**:
*workbox-sw*

**Browser & Platform**:
*Google Chrome v69.0.3497.100 for mac*

*I set matchOptions:{ignoreSearch: true} in workbox.strategies.networkFirst. It indicates I think the response of special page doesn't be affected by query string or hash. ex. When I request /another.html, /another.html?query string, /another.html#hash,  I expect it has only one cache for /another.html. How can I do?*

jeffposnick:
R: @philipwalton

Fixes #1709

This adds a new `cacheKeyWillBeUsed` lifecycle callback, following the interface described at https://github.com/GoogleChrome/workbox/issues/1709#issuecomment-477354428. This allows developers to override the default cache key for reads or writes (or both), for requests that pass through our `cacheWrapper` interface.

Once it's released, we'll need to update https://developers.google.com/web/tools/workbox/guides/using-plugins with info on usage.


packages/workbox-core/_private/cacheWrapper.mjs
```diff
@@ -41,52 +41,56 @@ const putWrapper = async ({
...
     throw new WorkboxError('cache-put-with-no-response', {
-      url: getFriendlyURL(request.url),
+      url: getFriendlyURL(effectiveRequest.url),
     });
   }
 
-  let responseToCache =
-      await _isResponseSafeToCache({request, response, event, plugins});
+  let responseToCache = await _isResponseSafeToCache({
+    response, event, plugins, request: effectiveRequest});
```
philipwalton: Nit: I think the old way is technically more correct according to Google JS style.

packages/workbox-core/_private/cacheWrapper.mjs
```diff
@@ -226,6 +234,46 @@ const _isResponseSafeToCache = async ({request, response, event, plugins}) => {
   return responseToCache ? responseToCache : null;
 };
 
+/**
+ * This checks the list of plugins for the cacheKeyWillBeUsed callback, and
```
philipwalton: s/This checks/Checks/

test/workbox-core/sw/_private/test-cacheWrapper.mjs
```diff
@@ -221,6 +221,83 @@ describe(`cacheWrapper`, function() {
       }))).to.be.true;
     });
 
+    it(`should call cacheKeyWillBeUsed`, async function() {
+      const cacheName = 'cacheKeyWillBeUsed-test-cache';
+      const cache = await caches.open(cacheName);
+      sandbox.stub(self.caches, 'open').resolves(cache);
```
philipwalton: You don't really need `self` here.

test/workbox-core/sw/_private/test-cacheWrapper.mjs
```diff
@@ -221,6 +221,83 @@ describe(`cacheWrapper`, function() {
...
+          request,
+          response,
+          plugins: [plugin],
+        });
+      }, 'incorrect-class');
+
+      expect(spy.calledOnceWith({
+        request,
+        mode: 'write',
+      }));
```
philipwalton: Add `.to.be.true`, otherwise this won't check anything.

test/workbox-core/sw/_private/test-cacheWrapper.mjs
```diff
@@ -366,5 +443,78 @@ describe(`cacheWrapper`, function() {
       expect(spyOne.callCount).to.equal(1);
       expect(spyTwo.callCount).to.equal(1);
     });
+
+    it(`should call cacheKeyWillBeUsed`, async function() {
```
philipwalton: Repeat code? Looks like these two tests are the same as the ones above.

packages/workbox-core/_private/cacheWrapper.mjs
jeffposnick: The standalone stuff after `await ...` doesn't fit on a single line anymore, so the old way wouldn't work either.

I'll reformat to put each object property on its own line instead.

test/workbox-core/sw/_private/test-cacheWrapper.mjs
jeffposnick: Ugh.

test/workbox-core/sw/_private/test-cacheWrapper.mjs
jeffposnick: They're in different `describe()` blocks—one for `.put()` and the other for `.match()`. The actual test cases are different, too.

packages/workbox-core/_private/cacheWrapper.mjs
philipwalton: Yeah, like I said it's a nit, but if you want it on two lines it should be indented 4 spaces (to show it's a line continuation). And if you want to indent 2 spaces it should be expanded (to be a block indent).

The style guide is weird that way, and the eslint plugin doesn't do a good job of handling it.

test/workbox-core/sw/_private/test-cacheWrapper.mjs
jeffposnick: In particular, the tests confirm that the correct `mode` value (`'read'` or `'write'`) is passed in to the plugin, and that the underlying `cache.match()` or `cache.put()` operation is actually called with the plugin's return value.

philipwalton: Ahh, I see. I couldn't easily see the outer blocks from the diff.

PR #1996, opened 2019-04-01
Migrate strategies tests to run in a SW environment 

philipwalton:
R: @jeffposnick

Continues the test-conversion work started in #1965. One catch with these tests is it's very hard (perhaps impossible?) to stub an opaque response in browsers, for two reasons:

- You can stub the `status` property after you create the response, but if you clone it, the clone will have the original response status.
- You can work around the above by *also* stubbing the `clone()` method, but if you put that response object in the cache and the retrieve it later, it'll revert back to the original status.

My solution to these issues was just to create a real opaque response by making a cross-origin request once. We can then clone that response for all future tests, which avoids the problem of having network latency affect the test duration.

**NOTE:** the diffs shown by GitHub in this PR are pretty terrible since file renames were involved, but if you only look at the [diffs from the second commit](https://github.com/GoogleChrome/workbox/pull/1996/commits/90756b4a5a5460e9cdb55fd5311cb3a6202630c8), it'll give you a much better idea of what changed.


infra/testing/helpers/generateOpaqueResponse.mjs
```diff
@@ -0,0 +1,18 @@
...
+  https://opensource.org/licenses/MIT.
+*/
+
+
+// Cache a resonse value and clone it instead of re-fetching every time.
+let response;
+
+export const generateOpaqueResponse = async () => {
+  if (!response) {
+    response = await fetch('https://google.com', {mode: 'no-cors'});
```
jeffposnick: I *thought* we used to start up two test servers on different ports, and then use the second when we wanted to simulate a cross-origin request. I am having a hard time tracking down when that changed, though.

Doing that again would be another option, and one that would allow us to keep everything "in house".

philipwalton: Yeah, I think we could do this, but I'll wait to do it in a separate PR.

PR #2004, opened 2019-04-03
4.2.0 metadata

jeffposnick:
R: @philipwalton


jeffposnick:
I think we know that CI is unhappy at the moment, so I'm just going to merge these metadata changes...

PR #2005, opened 2019-04-03
Remove our custom SW mocks and esm from tests

philipwalton:
R: @jeffposnick

This PR removes the outdated `@std/esm` package, our custom SW mocks, and my forked `shelving-mock-indexeddb` package from Workbox.

The primary changes in this PR are:
- Converting `import` to `require` in the `/tests/all` directory.
- Using `acorn` to test the package exports since `@std/esm` isn't available.


test/all/node/test-exports.mjs
```diff
@@ -30,40 +33,44 @@ describe(`[all] Window and SW packages`, function() {
...
       });
-
-      // Ensure there's a public export for every top-level file in the package.
-      for (const publicFile of publicFiles) {
-        const expectedExportName = path.basename(publicFile, '.mjs');
-
-        if (!(expectedExportName in moduleExports)) {
-          throw new Error(`Unable to find export '${expectedExportName}' in ${indexFile}`);
+      for (const node of indexAST.body) {
+        if (node.type === 'ExportDefaultDeclaration"') {
```
jeffposnick: Is there an extra `"` character in your string?

philipwalton: Ah, good catch! (In this case it doesn't affect things because we don't have default exports)

PR #2007, opened 2019-04-03
Ensure workbox is available before referencing it

### LINKED ISSUE #1915: Configure navigationPreload via workbox-webpack-plugin

merrywhether:
R: @jeffposnick @philipwalton

Fixes #1915 / #1981 

Sorry to follow up with a PR immediately after the publish, but it turns out order in the template does matter, as otherwise `workbox` might not be defined yet (total facepalm on my part)! I feel like I probably missed implementing a test somewhere in the initial pass but I'm not sure where that should've gone, so I at least added a warning comment for now. Let me know if there's something better I could add.

### DIFF
--- a/packages/workbox-build/src/templates/sw-template.js
+++ b/packages/workbox-build/src/templates/sw-template.js
@@ -6,6 +6,8 @@
   https://opensource.org/licenses/MIT.
 */
 
+// All workbox references must come after the workboxSWImport/importScripts block
+
 module.exports = `/**
  * Welcome to your Workbox-powered service worker!
  *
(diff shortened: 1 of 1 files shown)


derekdowling:
Hey @jeffposnick any ETA on a new minor version release? We'd love to start utilizing this.

jeffposnick:
We just pushed out `v4.3.0` which should include this fix.

EduardoRicardez:
I can't find this in v5.0.0, how can i use it?

PR #2014, opened 2019-04-05
Fix a background sync reply bug

### LINKED ISSUE #1984: queue.replayRequests() fails with TypeError: Failed to execute 'clone' on 'Request': Request body is already used
**Library Affected**:
workbox.backgroundSync.Queue. (Workbox 4.1.1)

**Browser & Platform**:
Google Chrome v73.0.3683.86 macOS

When `queue.replayRequests()` is called and the server is still down, it throws the following error and removes one item from the queue even if it wasn't synced.

```
sw.js:132 TypeError: Failed to execute 'clone' on 'Request': Request body is already used
    at d.g (Queue.mjs:176)
    at d.unshiftRequest (Queue.mjs:140)
    at d.replayRequests (Queue.mjs:246)
```

This is the code to register a route for background sync:

```
  const queue = new workbox.backgroundSync.Queue('myQueue')

  const addEventRoute = new workbox.routing.Route(
    ({ url }) => url.pathname === '/api/add',
    async args => {
      try {
        await fetch(args.event.request.clone())
      } catch (err) {
        console.error(err)
        queue.pushRequest(args.event)
      }
    },
    'POST'
  )

  workbox.routing.registerRoute(addEventRoute)

  self.addEventListener('sync', async ev => {
    try {
      await queue.replayRequests()
    } catch (e) {
      console.error(e)
    }
  })
```

When I try to manually trigger the sync from the UI with: 

```
<Button
   variant="contained"
   onClick={() => navigator.serviceWorker.ready.then(reg => reg.sync.register('syncEvents'))}
>
sync
</Button>
```

it fails and removes the oldest entry from the queue.
The sync works well, when the server is reachable again but not when it is still down.

philipwalton:
R: @jeffposnick 

Fixes #1984

The issue in #1984 was the request wasn't being cloned in the re-`fetch()` attempt, which in some cases (depending on the request method and body) was causes issues when attempting to clone it later when re-adding it to the queue. 

This PR fixes the issue and updates the tests to manually read the request when re-fetching to ensure this kind of thing would be caught in the future.

### DIFF
--- a/packages/workbox-background-sync/Queue.mjs
+++ b/packages/workbox-background-sync/Queue.mjs
@@ -238,7 +238,7 @@ class Queue {
     let entry;
     while (entry = await this.shiftRequest()) {
       try {
-        await fetch(entry.request);
+        await fetch(entry.request.clone());
 
         if (process.env.NODE_ENV !== 'production') {
           logger.log(`Request for '${getFriendlyURL(entry.request.url)}'` +
--- a/test/workbox-background-sync/sw/test-Queue.mjs
+++ b/test/workbox-background-sync/sw/test-Queue.mjs
@@ -119,9 +119,15 @@ describe(`Queue`, function() {
       if (!('sync' in registration)) this.skip();
 
       const onSync = sandbox.stub().callsFake(async ({queue}) => {
-        await queue.pushRequest({request: new Request('/one')});
-        await queue.pushRequest({request: new Request('/two')});
-        await queue.pushRequest({request: new Request('/three')});
+        await queue.pushRequest({
+          request: new Request('/one', {method: 'POST', body: '...'}),
+        });
+        await queue.pushRequest({
+          request: new Request('/two', {method: 'POST', body: '...'}),
+        });
+        await queue.pushRequest({
+          request: new Request('/three', {method: 'POST', body: '...'}),
+        });
       });
 
       const queue = new Queue('foo', {onSync});
@@ -138,7 +144,9 @@ describe(`Queue`, function() {
       if (!('sync' in registration)) this.skip();
 
       const onSync = async ({queue}) => {
-        await queue.pushRequest({request: new Request('/one')});
+        await queue.pushRequest({
+          request: new Request('/one', {method: 'POST', body: '...'}),
+        });
         throw new Error('sync failed');
       };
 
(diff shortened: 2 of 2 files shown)


bligny:
Hi 
@philipwalton 

After upgrading to version 4.3, we are facing a new issue (that we did not have before) which is apparently related to the new cloning mechanism of replayed requests. 

Here is the encountered error:

```
workbox-strategies.dev.js:829 Uncaught (in promise) no-response: The strategy could not generate a response for 'http://reese.ec.homeshore.be:8380/api/v1/touring/tours/795/assignments/chanquez'. The underlying error is DataCloneError: Failed to execute 'add' on 'IDBObjectStore': async replayRequests() {
      let entry;

      while (entry = await this.shiftRequest()) {
        try {
    ...<omitted>... } could not be cloned..
    at NetworkOnly.makeRequest (http://localhost:8100/workbox-4.x/workbox-strategies.dev.js:829:15)
WorkboxError @ workbox-core.dev.js:382
makeRequest @ workbox-strategies.dev.js:829
```

Can this be a direct consequence of commit [f165ca0](https://github.com/GoogleChrome/workbox/commit/f165ca0ae5eda5c0e78f2a2e9ab407b4e4446c4b) ?

philipwalton:
> For the sake of completeness, here is the request:

@bligny it's not clear to me where the code is running, as I'm not familiar with the APIs being used. Is that server-side code running in node?

> and the applied WB strategy:

One thing that might be the problem is you're creating both a `Queue` instance and then you're creating a `Plugin` instance from that queue instance (which I don't think will work). The `Plugin` constructor [takes the same arguments](https://developers.google.com/web/tools/workbox/reference-docs/latest/workbox.backgroundSync.Plugin) as the `Queue` constructor, and I don't think you should need both.

Try this (which just creates a `Plugin` instance):

```js
const queuePlugin = new workbox.backgroundSync.Plugin('tp-upload-v1', {
  maxRetentionTime: Infinity
});

  new RegExp('.*/api/v1/touring/.*'),
  new workbox.strategies.NetworkOnly({
    fetchOptions: {
      credentials: 'include',
    },
    plugins: [queuePlugin]
  }),
  'PUT'
);
```

bligny:
@philipwalton thx for spotting the mistake in my code.
The strange thing is that invoking the `Plugin` constructor with a Queue object, although incorrect, does not throw any error... 
I'm pretty sure I've seen (and copy/pasted) this kind of construction somewhere, but I do not remember where...
Anyway, the strange behaviour is now explained. Thanks again

bligny:
@philipwalton 
FYI : I've just created #2042 in reaction to your comments

philipwalton:
Thanks, that seems like a reasonable feature request.

PR #2017, opened 2019-04-08
Fixes opts passed to the BCU Plugin.

### LINKED ISSUE #2016: workbox-build generates incorrect channelName options
workbox-build@4.1.1 generates service-worker.js with code like this:

```js
new workbox.broadcastUpdate.Plugin('api-updates')
```

which does not work. Using the following does:

```js
new workbox.broadcastUpdate.Plugin({ channelName: 'api-updates' })
```

**Library Affected**:
workbox-build@4.1.1, workbox-webpack-plugin@4.1.1

jeffposnick:
Fixes #2016, by accounting for the fact that the broadcast update plugin is now passed the channel name as a named parameter in the options object.


philipwalton:
Looks like this is failing on a lint error, but that lint error has been fixed upstream, so I'll just merge this.

PR #2018, opened 2019-04-08
Add a size method to the Queue class

### LINKED ISSUE #2011: BackgroundSync - get queue length (WB-V4)
We would like to regularly display the number of pending requests that are parked into workbox  queue.

Yes I can use a custom "onSync" function in order to follow the play/replay of the queued items (and thereby increase/decrease an applicative counter), but the problem is that I need to initialize this counter on application/session startup.

I there a "workbox way" to query the workbox queue to get its length ?
(workbox way = not using indexeddb API to access the "request" store of the  storage named "workbox-background-sync").

Thx in advance!

philipwalton:
R: @jeffposnick

Fixes #2011

This PR adds a method to count the number of entries in the queue store for the given queue name.

***

**UPDATE:** Given https://github.com/GoogleChrome/workbox/issues/2011#issuecomment-480731157 and the fact that the initial changes in the PR didn't consider that a simple IndexedDB count operation wouldn't account for entries that have expired, I decided to just add a single `getAll()` method that returns an array of unexpired entries. Users who need a count can access that array's length.

### DIFF
--- a/packages/workbox-background-sync/Queue.mjs
+++ b/packages/workbox-background-sync/Queue.mjs
@@ -172,7 +172,7 @@ class Queue {
     const allEntries = await this._queueStore.getAll();
     const now = Date.now();
 
-    const unexpirtedEntries = [];
+    const unexpiredEntries = [];
     for (const entry of allEntries) {
       // Ignore requests older than maxRetentionTime. Call this function
       // recursively until an unexpired request is found.
@@ -180,11 +180,11 @@ class Queue {
       if (now - entry.timestamp > maxRetentionTimeInMs) {
         await this._queueStore.deleteEntry(entry.id);
       } else {
-        unexpirtedEntries.push(convertEntry(entry));
+        unexpiredEntries.push(convertEntry(entry));
       }
     }
 
-    return unexpirtedEntries;
+    return unexpiredEntries;
   }
 
 
--- a/packages/workbox-background-sync/Queue.mjs
+++ b/packages/workbox-background-sync/Queue.mjs
@@ -163,6 +163,32 @@ class Queue {
   }
 
   /**
+   * Returns all the entries that have not expired (per `maxRetentionTime`).
+   * Any expired entries are removed from the queue.
+   *
+   * @return {Promise<Array<Object>>}
+   */
+  async getAll() {
+    const allEntries = await this._queueStore.getAll();
+    const now = Date.now();
+
+    const unexpirtedEntries = [];
+    for (const entry of allEntries) {
+      // Ignore requests older than maxRetentionTime. Call this function
+      // recursively until an unexpired request is found.
+      const maxRetentionTimeInMs = this._maxRetentionTime * 60 * 1000;
+      if (now - entry.timestamp > maxRetentionTimeInMs) {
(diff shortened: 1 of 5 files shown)


philipwalton:
There's not currently any deduping, so it would report the total number of entries regardless of whether the underlying requests were the same (or not).

philipwalton:
@jeffposnick I force-pushed the changes we talked about in chat. Let me know if you have any comments on them.

bligny:
I understand that adding a single `getAll()` method also allows to get the queue size by taking the array length, but it's a bit resource-consuming just to get the size. It would be much more efficient to have - in addition to `getAll()` - a separate and dedicated `size()` method using the `count()` method of the `IDBObjectStore`.
My two cents
BTW: a small typo in your commit ("unexpir*t*edEntries")

philipwalton:
@bligny thanks for taking a look. Your suggestion is actually how I originally implemented it, but then I realized that a `size()` method using the `count()` IDB call would return potentially expired entries. It would be weird/confusing if count return 10, but then you iterated through the queue and only got 8 entries, no?

We could potentially add a `size()` method that uses a cursor in IDB to count without returning the entire list, but that would still involved reading the data into memory.

The best solution (in my opinion) is unfortunately not available to us at the moment, and that's to use a multi-key index on our IDB object store, so we can have the entries sorted by `queueName` *and* `timestamp`. That would allow us to easily count all entries above a certain point without having to iterate through each. Unfortunately this isn't supported in Edge.

So, given all this context, what do you think is the best option to support your use case?

philipwalton:
Also cc: @Fitzchev

bligny:
@philipwalton thanks for your quick reply. I do agree that the `size()` and `getAll()` responses should be consistent.  There might yet be some room for an intermediate solution, e.g:

```
async getAll(includeExpiredEntries = true) {
 ...
}
```
The next question is what would be the best (most logical) default value for  the optional `includeExpiredEntries` parameter. On the one hand, `true` is nice because consistent with size(), but on the other hand `false` is probably much more interesting. 

I let you take the final decision, I'm already happy with the existing improvement :-)
Thanks again for the follow-up.

Fitzchev:
@philipwalton thanks a lot, it definitively covers my use case. 

Reading @bligny comment and  looking at the "automatic purge" of expired items, it makes me think that it could be interesting to know that there are expired items left in the queue. What about having a global parameter _automatic_purge = yes/no_ - if yes, keep the same behavior, if no rise an event workbox_expired_items so that the user can decide what to do with them (rise an alert, purge anyway...)

philipwalton:
Thanks everyone for the comments. I think for the moment I'm going to keep this PR as is.

Once we can drop Edge 18 support, we can rework the data model to use multi-key indexes and add a standalone `size()` method that can take advantage of the underlying IndexedDB count method.

Also, once we get to the next major release we can re-evaluate whether or not it makes sense to have entries silently expire (or maybe change the default, but that would be a breaking change). At the moment, if you don't like the silent expire behavior, you can alway set the `maxRetentionTime` configuration option to `Infinity` and expire entries manually (i.e. just don't re-fetch them after removing them from the queue).

Issue #1854, opened 2019-01-22
Proposal/RFC: workbox-webpack-plugin changes


developit:
👋  Hi all!

I put together a proposal for how we might modify `workbox-webpack-plugin` to allow for a more bundler-centric development approach, using Workbox as a modular library and Webpack to bundle the resulting ServiceWorker.

You can view the proposal here:
https://docs.google.com/document/d/1qyzxo5MBZ5BHva2GaAPk_Hw-jkVk1-NXcqfAIlvHgkc

Comments welcome!

mcshaman:
Cool. Questions:
1. In your proposal are you suggesting doing away with the webpackbootstrap and just using the importScript to import chunks?
2. You only seem to mention modules from workbox. What about other modules either relative or in the node_modules folder? Will they be bundled under this proposal and if so can they be bundled into a single service worker file or is the proposition to have them as a seperate chunk?

frlinw:
My wishes for v5:
- `importWorkboxFrom` values are enum only ['local', 'disabled', 'cdn'], no more chunks
- `importScripts` supports chunks
- import order should be: 
  - `importWorkboxFrom` (if not disabled)
  - precache-manifest.js
  - `importScripts` list

This gives full control over precache-manifest assets to the developper.
Breaking change: if you want to change `__precacheManifest` it has to be explicit
```js
__precacheManifest = __precacheManifest.concat([{
   url: 'nonAssetScript.js',
   revision: 'abcdefg'
}])
```

> The solution was to “teach” Webpack to detect Service Workers by introducing parser logic for navigator.serviceWorker.register()

Parser seems to match simple cases https://github.com/GoogleChromeLabs/squoosh/blob/master/config/auto-sw-plugin.js#L73
what about:
```js
const n = navigator.serviceWorker
n.register()
```
> Should this be a new plugin?

It can't be just a flag ?

developit:
@mcshaman:

> In your proposal are you suggesting doing away with the webpackbootstrap and just using the importScript to import chunks?

No - webpack's bootstrapping code uses importScripts() when the [output target](https://webpack.js.org/configuration/target/) is set to webworker.

> You only seem to mention modules from workbox. What about other modules either relative or in the node_modules folder? Will they be bundled under this proposal and if so can they be bundled into a single service worker file or is the proposition to have them as a seperate chunk?

[...]

PR #2026, opened 2019-04-11
feat: add minification for the cli and webpack-plugin

### LINKED ISSUE #900: workbox-build: generate minified output?
**Library Affected**:
workbox-build, workbox-webpack-plugin

**Browser & Platform**:
all browsers

Hello, I would like a way to emit minified output from workbox. I see the libraries already have a minified production version, but it seems there's no option to minify the output from workbox-build.

I'm using the Workbox Webpack plugin specifically, but I believe this option could be useful for the whole ecosystem.

wood1986:
R: @jeffposnick @philipwalton

fix #900

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2357,12 +2357,6 @@
         "array-from": "^2.1.1"
       }
     },
-    "@std/esm": {
-      "version": "0.26.0",
-      "resolved": "https://registry.npmjs.org/@std/esm/-/esm-0.26.0.tgz",
-      "integrity": "sha512-g3RDuosSa5fZOzENtrZdx7Gevb3zabfn8qglug2aCJIVz/4woFpKoqm1yD3mG2RD0zJEZRnkkuPHsmNglKGl7g==",
-      "dev": true
-    },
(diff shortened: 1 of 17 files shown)


jeffposnick:
Thanks for the contribution, @wood1986!

We're in the process of rewriting the `workbox-webpack-plugin` along the lines of the proposal outlined at https://github.com/GoogleChrome/workbox/issues/1854 (Work on this has actually begun!)

Given that, I don't think it makes sense to move forward with this PR at this time. The rewrite to the plugin will result in a compiled service worker that takes advantage of minification.

wood1986:
@jeffposnick how does it take to release v5? if it is longer than 3 months, is it possbile to accept the merge?

jeffposnick:
There were some open questions about staffing this effort that have been resolved, and work on the `workbox-webpack-plugin` v5 has begun. Unfortunately, I can't offer any guarantees around timelines for the release.

I do not believe it makes sense to make fundamental changes to the plugin's behavior in v4 while that effort is underway.

PR #2050, opened 2019-04-30
Ensure all private modules are exported correctly

### LINKED ISSUE #2045: navigation requests w/ StaleWhileRevalidate + broadcastUpdate = TypeError: Deferred_mjs.Deferred is not a constructor
**Library Affected**:
workbox-broadcast-cache-update

**Browser & Platform**:
Chrome

When trying to use workbox-broadcast-cache-update for navigation requests using the StaleWhileRevalidate for which there is an update (using the Last-Modified header), I get the following error: 

```
Uncaught (in promise) TypeError: Deferred_mjs.Deferred is not a constructor
    at BroadcastCacheUpdate._windowReadyOrTimeout (BroadcastCacheUpdate.mjs:154)
    at sendUpdate (BroadcastCacheUpdate.mjs:106)
    at BroadcastCacheUpdate.notifyIfUpdated (BroadcastCacheUpdate.mjs:112)
    at Plugin.cacheDidUpdate (Plugin.mjs:80)
    at Object.putWrapper [as put] (cacheWrapper.mjs:99)
```

I am already using workbox-broadcast-cache-update with native Fetch API requests on loaded pages and it's working great. Something to do here with the page not being loaded? Is there a workaround or is there is no way to use BroadcastCacheUpdate for navigation requests?

### LINKED ISSUE #2049: Remove checkSWFileCacheHeaders
I don't think we need [`checkSWFileCacheHeaders`](https://github.com/GoogleChrome/workbox/blob/94d34d687cf4ab68f76cca938dba590ba3d7b630/packages/workbox-core/_private/checkSWFileCacheHeaders.mjs) anymore, which is used in our dev builds at https://github.com/GoogleChrome/workbox/blob/94d34d687cf4ab68f76cca938dba590ba3d7b630/packages/workbox-core/utils/welcome.mjs#L36-L38

The changes described in https://developers.google.com/web/updates/2018/06/fresher-sw mean that setting `Cache-Control` on your service worker file won't do any harm by default.

philipwalton:
R: @jeffposnick

Fixes #2049, fixes #2045.

This PR ensures all files in `_private` directory are exported via the top level `_private.mjs` file. It also removes exports that are no longer needed.

The `quota.mjs` file exported two separate symbols, so it initially failed the new private-exports test, but since it's not used outside of `workbox-core` it was fine to just move it to the `utils` folder instead.


philipwalton:
Ahh, yes, so given that `workbox-expiration` uses it, I split the file into to separate modules. One for registering and one for executing.

PR #2058, opened 2019-05-24
Convert workbox-core to Typescript

philipwalton:
R: @jeffposnick

This PR is an initial attempt at converting our packages to TypeScript (and doing so in a backwards compatible way). Here's a summary of what I've done:

- All `.mjs` source files have been moved inside `/src` and renamed to `.ts`
- Type definitions have been added to the `.ts` files, and a global `type-overrides.d.ts` was added for some missing IDB and SW types.
- A new `transpile-typescript` gulp task has been added that converts `.ts` files in `/src` to `.mjs` files in the root directory (mimicking the current structure).
- All other build tasks have been updated to run `transpile-typescript` prior to them doing their thing.


gulp-tasks/build-packages.js
```diff
@@ -6,17 +6,33 @@
...
 
-const cleanPackage = (packagePath) => {
-  const outputDirectory = path.join(packagePath,
-      constants.PACKAGE_BUILD_DIRNAME);
-  return fse.remove(outputDirectory);
+const cleanPackage = async (packagePath) => {
+  // Delete generate files from the the TypeScript transpile.
+  if (await fs.pathExists(path.join(packagePath, 'src', 'index.ts'))) {
+    const deletedPaths = await del([
+      path.join(packagePath, '**/*.+(mjs|d.ts)'),
```
jeffposnick: You probably want `path.posix.join()` here, since I believe these patterns should always use `/`, event when run on Windows.

gulp-tasks/build-packages.js
```diff
@@ -6,17 +6,33 @@
...
+      // Don't delete files in node_modules
+      '!**/node_modules', '!**/node_modules/**/*',
+    ]);
+
+    const directoriesToDelete = new Set();
```
jeffposnick: I don't understand what this bit of code is deleting.

gulp-tasks/transpile-typescript.js
```diff
@@ -0,0 +1,173 @@
...
+ * https://github.com/rollup/rollup/issues/2847
+ *
+ * @param {string} oldExt
+ * @param {string} newExt
+ */
+const renameImportExtensionsPlugin = (oldExt, newExt) => {
+  return {
+    generateBundle: (options, bundle) => {
+      const importPattern =
+          new RegExp(`((?:import|export)[^']+')([^']+?)\\${oldExt}';`, 'g');
```
jeffposnick: Should we include `"` along with `'` as the possible quote characters?

gulp-tasks/transpile-typescript.js
```diff
@@ -0,0 +1,173 @@
...
+};
+
+/**
+ * @type {RollupCache}
+ */
+let moduleBundleCache;
+
+/**
+ * Transpiles a package's source TypeScript files to `.mjs` JavaScript files
+ * in the root package directory, along with the correspdonging definition
```
jeffposnick: correspdonging => corresponding

gulp-tasks/build-packages.js
```diff
@@ -6,17 +6,33 @@
...
 
 const constants = require('./utils/constants');
 const packageRunnner = require('./utils/package-runner');
 
-const cleanPackage = (packagePath) => {
-  const outputDirectory = path.join(packagePath,
-      constants.PACKAGE_BUILD_DIRNAME);
-  return fse.remove(outputDirectory);
+const cleanPackage = async (packagePath) => {
+  // Delete generate files from the the TypeScript transpile.
```
jeffposnick: generate => generated

packages/workbox-core/src/utils/pluginUtils.ts
```diff
@@ -0,0 +1,25 @@
...
+  Copyright 2018 Google LLC
+
+  Use of this source code is governed by an MIT-style
+  license that can be found in the LICENSE file or at
+  https://opensource.org/licenses/MIT.
+*/
+
+import '../_version';
+
+export interface Plugin {
```
jeffposnick: Could we add in some TypeScript definitions for the signatures of these various callbacks, and then say that each of those properties maps to the corresponding callback signature, instead of a generic `Function`?

(I'm happy to figure out the syntax for doing that if you're not sure, as I don't want to create more work for you.)

gulp-tasks/build-packages.js
philipwalton: I *think* you might be right. I know that's true for `node-glob`, but I believe `del` uses `globby`, so it may be different?

gulp-tasks/transpile-typescript.js
philipwalton: As long as we stick to our style guide, we shouldn't ever have double quotes. And since we're only matching our files, I think this is fine for now. Though I suppose if we add any third-party libraries (like maybe `idb`) then yeah, we'd have to change this.

gulp-tasks/build-packages.js
philipwalton: When TypeScript generates `.mjs` and `.d.ts` files, it does so keeping the same directory structure as was in `/src`, so any sub-directories will get copied to the root-level package directory, and this code deletes those.

I've added a comment to help better explain what this does.

gulp-tasks/transpile-typescript.js
philipwalton: Done.

gulp-tasks/build-packages.js
philipwalton: Done.

packages/workbox-core/src/utils/pluginUtils.ts
philipwalton: I definitely think we can, and consider this initial conversion the first step in what will likely be more changes as I learn more about Typescript definition capabilities.

My expectation is as we convert more packages over, it'll start to become clear how we can update these plugin definitions to be more useful for consumers of them.

philipwalton:
@jeffposnick I updated this PR with the changes we talked about today.

Now the `transpile-typescript` task will create a `.js` files for each source file, and it will also create a corresponding `.mjs` file that just re-exports any exports from the respective `.js` file:

```js
// moduleName.mjs
export * from './moduleName.js';
```

This seemed like a good way to keep backwards compatibility with existing `.mjs` files while we wait for Typescript to officially support `.mjs`.

philipwalton:
Hmmm, looks like appveyor is failing with some weird `npm install` error. Any idea why that might be?

jeffposnick:
It looks like it's due to 

```
    "jsdoc-baseline": "https://github.com/hegemonic/baseline/tarball/master",
```

in our `package.json`. I don't remember the history behind that, but do you want to try swapping that out for the latest release of https://www.npmjs.com/package/jsdoc-baseline? It was pushed out 21 days ago, so presumably it has whatever we previously needed to get from the GitHub source.

philipwalton:
🤦 why didn't I think of that!

PR #2060, opened 2019-05-25
Copy handleMake from serviceworker-webpack-plugin

### LINKED ISSUE #1513: InjectManifest Plugin doesn't compile the serviceworker with webpack

ALiangLiang:

R: @jeffposnick @philipwalton

Fixes #1513

Compile the serviceworker with webpack.

### DIFF
--- a/packages/workbox-webpack-plugin/src/inject-manifest.js
+++ b/packages/workbox-webpack-plugin/src/inject-manifest.js
@@ -9,6 +9,7 @@
 const assert = require('assert');
 const path = require('path');
 const {getManifest} = require('workbox-build');
+const SingleEntryPlugin = require('webpack/lib/SingleEntryPlugin')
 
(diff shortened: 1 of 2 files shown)


jeffposnick:
Hello!

I appreciate the fact that this is something that many folks in the community would like. I'm actively working towards bringing this functionality to Workbox v5—part of this is outlined in https://github.com/GoogleChrome/workbox/issues/1854, but there are also some underlying changes being made to the `workbox-build` library that will factor in to the rewrite as well.

Because of this, I'd much rather address this use case as part of the official v5 revised interface. I'm more than happy that folks from the community who are not satisfied with the v4 behavior have found their own solutions in the meantime, but I do not want to incorporate them into the official v4 release piecemeal.

PR #2068, opened 2019-06-04
Rollup bundles in generateSW, and more for v5

### LINKED ISSUE #1495: Compared the resolved swSrc and swDest paths in InjectManfiest to ensure they're different

**Library Affected**:
*workbox-sw, workbox-build, etc.*

`workbox-build`

**Browser & Platform**:
*E.g. Google Chrome v51.0.1 for Android, or "all browsers".*

`all browsers`

*Your request here.*

Whenever you are injecting precache routes into an existing sw, you should be able to save the results into the file with the same name and pathname.

Usecase:
* Build your project with a bundler (e.g. parcel) using workbox-build to inject precache routes into a sw. The sw could import different libraries, hence you can' t just generate a new worker since it should contain imported modules.
* Instead, you could specify swSrc and srDest to be the bundeled version of the sw, so that workbox-build replaces RegEx inside that bundeled version retaining all of the dependencies.
* Instead of throwing 
> Service worker injection failed: Error: 'swSrc' and 'swDest' should not be set to the samefile. Please use a different file path for
'swDest'.

It could be turned into a warning.

### LINKED ISSUE #1809: 'useless-glob-pattern' should be a warning, not a fatal error
**Library Affected**:
*workbox-webpack-plugin version 3.3.1

**Browser & Platform**:
*node version 10.13.0

*Related issue: #1353 
*I am just wondering from a development side: Why is the glob handling at (globPatterns): https://github.com/GoogleChrome/workbox/blob/912080a1bf3255c61151ca3d0ebd0895aaf377e2/packages/workbox-build/src/lib/get-file-details.js#L45 so strict - is there some risk/danger involved with this option? In my project I had to add an extra glob filter step (https://github.com/witneyjs/witney/commit/34450b41e0078485ac05710a2e10d74f8529701d#diff-9f705cad1d5e127e633dda9ec3641b84R95) as a workaround, just to make sure, that workbox will not throw an error :D. 
If I was guessing, I think every developer using the globPatterns, has to add such a filter logic, because its the nature of globs that they might have an empty result which workbox doesn't accept ;). If there is a hidden reasoning behind this error, I recommend that the reason should be added to the error message, so that developers as I don't just add workarounds to their code :).

P.S. thanks for the awesome work on this project <3!

### LINKED ISSUE #2059: Simpler InjectManifest replacement behavior
Currently, `InjectManifest` mode in Workbox's build tools works by first generating a [precache manifest](https://developers.google.com/web/tools/workbox/modules/workbox-precaching#explanation_of_the_precache_list), and then using a regular expression to find the appropriate empty array to swap out from your source service worker file.

The end result is that something like `workbox.precaching.precacheAndRoute([])` in your source file is transformed to `workbox.precaching.precacheAndRoute([{url: '...', revision: '...'}, ...])`

The regular expression used to find just the `[]` that's used as a parameter to the precaching method is a bit gnarly, and while it is possible to customize it, I don't know that in practice it ends up being overridden much. (I know that I personally have to look up the syntax and go through a lot of trial-and-error to get a custom pattern working.)

I'd like to update this behavior in v5 to be more straightforward, and rely on simple string replacement, looking for `self.__WB_MANIFEST` in the service worker source file, and replacing it with the corresponding precache manifest as part of the build process.

That means that you could write `workbox.precaching.precacheAndRoute(self.__WB_MANIFEST)` in your source file, and things would work as before. Since `self.__WB_MANIFEST` will just be `undefined` prior to building, this should result in the unbuild service worker file being valid JavaScript.

Developers who need to manual control over adding extra manifest entries could do something like `workbox.precaching.precacheAndRoute([{url: '...', revision: '...'}, ...].concat(self.__WB_MANIFEST))`, which is a lot more difficult to do with the old regular expression replacement. Similarly, this would make it easier to post-process the precache manifest via runtime (as opposed to build time) JavaScript, like

```js
const processedManifest = (self.__WB_MANIFEST || []).map((entry => {
  // return some transformation
});
workbox.precaching.precacheAndRoute(processedManifest);
```

(I don't see runtime JavaScript manipulation as being a replacement for build-time manifest manipulation, but as something that would augment what you could do with build-time logic.)

I'm leaving this issue open for discussion in case there are any use cases I'm not aware of that this would end up breaking.

### LINKED ISSUE #2063: Remove generateSWString from workbox-build
I'm considering removing the `generateSWString` mode from the `workbox-build` build tool from Workbox v5.

I don't _believe_ that this is a widely used mode (compared to `injectManifest` and `generateSW`), and one of the reasons it was kept around is because it was used under the hood in the `workbox-webpack-plugin`. That's going to change in Workbox v5 as well.

If you have a strong use case for `generateSWString` that can't be filled by `generateSW`, please let me know.

### LINKED ISSUE #2064: generateSW uses custom Workbox runtime bundles, not CDN
My current plan for Workbox v5's `generateSW` mode is to produce output that uses a custom bundle of the Workbox runtime, which includes code for only the features of Workbox that are used by the configuration you provide.

This will effectively automate the process described in "[Using Bundlers (webpack/Rollup) with Workbox](https://developers.google.com/web/tools/workbox/guides/using-bundlers)" for folks who want the simplicity of `generateSW` mode. (It will use Rollup under the hood for the actual bundling and code elimination.)

Because the underlying template used in `generateSW` mode looks very different when the Workbox runtime is referenced via a custom bundle vs. when the Workbox runtime is loaded from the CDN via `workbox-sw`, I plan on removing the ability to use the CDN copy of Workbox with `generateSW` mode, along with the `importWorkboxFrom` configuration option.

For the time being, we will still publish bundles of the Workbox runtime to the CDN, but my expectation is that it will be only be useful for `injectManifest` users. Longer-term, we may phase out the CDN entirely, and encourage all developers to go through a flow that produces a Workbox runtime bundle that's customized for their specific usage.

## Parameter changes

As mentioned above, `importWorkboxFrom` will no longer be supported in `generateSW` mode.

New configuration parameters that I anticipate adding include:

- `inlineWorkboxRuntime`, defaulting to `false`, which controls whether the Workbox runtime code is inlined into a single service worker file, or whether the runtime code for the Workbox libraries are output into a separate, versioned file that is automatically pulled in to the main service worker via `importScripts()`.

- `mode`, defaulting to either `process.env.NODE_ENV || 'production'`, which controls whether `terser` is used, and whether Workbox's option code for, e.g., enhanced logging is enabled or disabled in the generated bundle.

- `sourcemap`, defaulting to `true`, which controls whether sourcemaps are also created for the generated file(s).

- `babelPresetEnvTargets`, defaulting to `['chrome >= 56']`, which controls what transpilation is performed via `@babel/preset-env` when creating the generated file(s). `['chrome >= 56']` matches the setting currently used when we produce our CDN-hosted bundles.

jeffposnick:
R: @philipwalton

Fixes #2063, fixes #2059, fixes #2064, fixes #1809, fixes #1495

This implements the first of a set of planned changes for the build tools in v5. 

The focus here is on `workbox-build` (and by extension, `workbox-cli`). It moves the `generateSW` output away from using a CDN copy of the Workbox runtime and on to a custom generated Rollup bundle that includes only the code that's actually used in the underlying service worker.

It checks either a `mode` config option or the current `NODE_ENV` environment variable, and will enable/disable minification and the additional development logging depending on that value. `inlineWorkboxRuntime`, `sourcemaps`, and `babelPresetEnvTargets` can also be set if developers need more control over the Rollup output.

There are also a smaller number of changes to `injectManifest` included in this PR, and a removal of the `generateSWString` mode.

Still to come are refinements to how we load the external Workbox runtime chunk (see https://github.com/GoogleChromeLabs/proxx/issues/470), and then a sigifnicant number of changes to `workbox-webpack-plugin`. This seemed like a good stopping/review point before moving forward, though.


packages/workbox-build/src/entry-points/options/defaults.js
```diff
@@ -7,18 +7,22 @@
...
+  babelPresetEnvTargets: ['chrome >= 56'],
   cleanupOutdatedCaches: false,
   clientsClaim: false,
   globFollow: true,
   globIgnores: ['**/node_modules/**/*'],
   globPatterns: ['**/*.{js,css,html}'],
   globStrict: true,
   importWorkboxFrom: 'cdn',
-  injectionPointRegexp: /(precacheAndRoute\()\s*\[\s*\]\s*(\)|,)/,
+  injectionPoint: 'self.__WB_MANIFEST',
```
philipwalton: Maybe `__WB_PRECACHE_MANIFEST`? So it's a little more descriptive. WDYT?

packages/workbox-build/src/lib/bundle.js
```diff
@@ -0,0 +1,109 @@
+/*
+  Copyright 2018 Google LLC
```
philipwalton: Super nit: I think for new files we should use the current year, 2019 (here and for other files as well).

packages/workbox-build/src/lib/bundle.js
```diff
@@ -0,0 +1,109 @@
...
+        },
+      },
+    }));
+  }
+
+  const rollupConfig = {
+    plugins,
+    input: temporaryFile,
+    treeshake: {
+      moduleSideEffects: false,
```
philipwalton: Didn't you determine that this removes the versioning code?

packages/workbox-build/src/entry-points/options/defaults.js
```diff
@@ -7,18 +7,22 @@
...
+  inlineWorkboxRuntime: false,
```
philipwalton: I don't care much about this as long as there's an option, but since inlining everything is the default in Rollup (and not inlining requires loadz0r + amd, etc.), it feels to me like the default should be `true`.

I know not inlining potentially produces more cacheable builds, but I'm skeptical as to how much of a difference that'll actually make in practice since we update workbox frequently, and any configuration/logic change made by the user would likely change the workbox bundle anyway.

packages/workbox-build/src/entry-points/options/defaults.js
```diff
@@ -7,18 +7,22 @@
...
 
 module.exports = {
+  babelPresetEnvTargets: ['chrome >= 56'],
   cleanupOutdatedCaches: false,
   clientsClaim: false,
   globFollow: true,
   globIgnores: ['**/node_modules/**/*'],
   globPatterns: ['**/*.{js,css,html}'],
   globStrict: true,
   importWorkboxFrom: 'cdn',
```
philipwalton: Does this need to be changed too (now that we're generating a user-specific bundle)?

packages/workbox-build/src/entry-points/options/defaults.js
```diff
@@ -7,18 +7,22 @@
...
-  injectionPointRegexp: /(precacheAndRoute\()\s*\[\s*\]\s*(\)|,)/,
+  injectionPoint: 'self.__WB_MANIFEST',
+  inlineWorkboxRuntime: false,
   maximumFileSizeToCacheInBytes: 2 * 1024 * 1024,
+  mode: 'production',
   navigateFallback: undefined,
   navigationPreload: false,
   offlineGoogleAnalytics: false,
-  purgeOnQuotaError: false,
+  purgeOnQuotaError: true,
```
philipwalton: +1

packages/workbox-build/src/entry-points/options/defaults.js
jeffposnick: We need some sort of default for the custom bundles, and `['chrome >= 56']` seemed like the one most in keeping with our overall compatibility goals.

At least folks have straightforward control over this now via configuring an alternative, which wasn't the case with our pre-built bundles.

packages/workbox-build/src/entry-points/options/defaults.js
jeffposnick: As per our offline conversation, I *think* this is the right default, given that an entry in the precache manifest changing is a fairly likely event for any actively developed web app.

Paying the cost of re-downloading the Workbox runtime following every re-deploy doesn't seem like the right default if we could avoid it.

We could revisit the default if developers push back for some reason.

packages/workbox-build/src/lib/bundle.js
jeffposnick: Yeah, it does.

I think I'll prevent the "too many versioning strings in the final bundle" issue by conditionally toggling the various `import` statements in the service worker source based on whether they're actually needed, instead of relying on dead code elimination.

packages/workbox-build/src/entry-points/options/defaults.js
jeffposnick: How about `self.__WB_INJECTED_MANIFEST` to match the `injectManifest` naming?

infra/testing/validator/service-worker-runtime.js
```diff
@@ -101,6 +101,39 @@ function setupSpiesAndContext() {
...
+
+  const workboxContext = {
+    CacheFirst: sinon.stub().returns({name: 'CacheFirst'}),
+    clientsClaim: sinon.spy(),
+    enable: sinon.spy(),
+    getCacheKeyForURL: sinon.stub().returns('/urlWithCacheKey'),
+    initialize: sinon.spy(),
+    NetworkFirst: sinon.stub().returns({name: 'NetworkFirst'}),
+    Plugin: sinon.spy(),
+    Plugin$1: sinon.spy(),
```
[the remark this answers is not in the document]
jeffposnick: https://github.com/GoogleChrome/workbox/issues/2072

packages/workbox-build/src/lib/bundle.js
jeffposnick: I'm now dynamically generating the list of `import` statements based on actual usage. It's a little funky, but it means that we aren't reliant on dead code elimination.

packages/workbox-build/src/entry-points/options/defaults.js
philipwalton: Ahh, I wasn't clear. This comment is referring to the `importWorkboxFrom` field.

packages/workbox-build/src/entry-points/options/defaults.js
philipwalton: Hmmm, the past tense seems a bit weird to me since whenever you're reading it, it hasn't been injected yet.

packages/workbox-build/src/entry-points/options/defaults.js
jeffposnick: Ah, good catch. `importWorkboxFrom` was removed as a supported option, so I can remove it from the default list here.

packages/workbox-build/src/entry-points/options/defaults.js
jeffposnick: Okay, I'll just switch to` __WB_PRECACHE_MANIFEST` for now, and we can revisit if needed.

packages/workbox-build/src/entry-points/options/defaults.js
```diff
@@ -7,18 +7,21 @@
...
+  babelPresetEnvTargets: ['chrome >= 56'],
   cleanupOutdatedCaches: false,
   clientsClaim: false,
   globFollow: true,
   globIgnores: ['**/node_modules/**/*'],
   globPatterns: ['**/*.{js,css,html}'],
   globStrict: true,
-  importWorkboxFrom: 'cdn',
-  injectionPointRegexp: /(precacheAndRoute\()\s*\[\s*\]\s*(\)|,)/,
+  injectionPoint: 'self.__WB_PRECACHE_MANIFEST',
```
philipwalton: Sorry to keep bikeshedding this, but after playing with this locally a bit and seeing all the other uses of the term "manifest" without any mention of "precache", I think it'd be fine to keep this `self.__WB_MANIFEST` if that's your preference.

packages/workbox-build/src/lib/create-module-imports.js
```diff
@@ -0,0 +1,144 @@
...
+  offlineAnalyticsConfigString,
+  runtimeCaching,
+  skipWaiting,
+}) => {
+  const nodeModulesPath = path.posix.resolve(
+      __dirname, '..', '..', 'node_modules');
+
+  const modulesUsed = new Map();
+
+  // precacheAndRoute is always included.
```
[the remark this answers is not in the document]
jeffposnick: I opened https://github.com/GoogleChrome/workbox/issues/2078 to track this. Let's deal with that in a separate PR as there's a decent amount of follow-on changes that would need to be made in order to enable that usage.

packages/workbox-build/src/entry-points/options/defaults.js
jeffposnick: Okay, switching back.

jeffposnick:
> Where `ensureImported(pgkName, moduleName)` would both add the module to the import list and return the local name.

I think the problem with "add the module to the import list" is that it implies there's going to be some sort of pre-processing step, where the template is evaluated by EJS once to get a list of imported modules, and then a second time to spit out the `import` statements along with the actual usage statement.

Something along the lines of what you're suggesting would be if

```
<% if (navigationPreload) { %>
  <%= ensureImported('workbox-navigation-preload', 'enable') %>();
<% } %>
```

evaluated to

```
import {enable as workbox_navigationPreload_enable} from 'path/to/workbox-navigation-preload/enable.mjs;'
workbox_navigationPreload_enable();
```

(i.e. it spit out the import statement right before the usage). But in that case, I don't think we could guard against the same module being imported twice, which I think is a syntax error. That could happen in, e.g., the `runtimeCaching` section, where multiple routes might each use the `workbox-expiration` module.

philipwalton:
Right, I was imagining it'd be a two-step process:

1. First, you'd generate the logic
2. Then, you'd prepend the necessary imports

It's arguably not any cleaner, but the benefit is it determines the imports from actual usage rather than repeated logic in a separate file )which I could see getting out of sync rather easily).

jeffposnick:
I think what you're describing would be similar to:

```
<% if (navigationPreload) { %>
  <%= workbox.navigationPreload.enable %>();
<% } %>
```

and then the `workbox` object that gets passed in to EJS looks like:

```
const workbox = {
  navigationPreload: {
    enable: constants.navigationPreload.enable,
  },
  ... etc., for all of the other things that might be used.
};
```

And then the import statement would end up looking like:

```
imports.push(`import {enable as ${constants.navigationPreload.enable}} from 'path/to/navigation-preload/enable.mjs';`);
```

How does that sound?

jeffposnick:
(And actually, we could probably cut out the `constants` module in that case, and just use a module that establishes the `workbox` object as the source of truth.)

jeffposnick:
To clarify:

```
// In workbox-namespace.js
module.exports = {
  navigationPreload: {
    enable: 'workbox_navigationPreload_enable',
  },
 // ...etc...
};
```

```
// In create-module-imports.js
const workboxNamespace = require('./workbox-namespace);

if (navigationPreload) {
  modulesUsed.set(workboxNamespace.navigationPreload.enable, {
    moduleName: 'enable',
    pkg: 'workbox-navigation-preload',
  });
}
```

```
// In the EJS template, with workboxNamespace
// passed in to the template's global data.
<% if (navigationPreload) { %>
<%= workboxNamespace.navigationPreload.enable %>();
<% } %>
```

packages/workbox-build/src/lib/module-registry.js
```diff
@@ -0,0 +1,72 @@
...
+        '${this.nodeModulesPath}/${pkg}/${moduleName}.mjs';`;
+
+      workboxModuleImports.push(importStatement);
+    }
+
+    return workboxModuleImports;
+  }
+
+  /**
+   * This function returns the corresponding local variable name for a given
```
philipwalton: The "this function" part seems superfluous to me. Just saying "Returns the corresponding..." is fine. (Here and for other functions as well.)

test/workbox-build/node/lib/module-registry.js
```diff
@@ -0,0 +1,55 @@
...
+          resolve: () => '/path/to/node_modules',
+        },
+      },
+    });
+    moduleRegistry = new ModuleRegistry();
+  });
+
+  it(`getImportStatements() should return [] when nothing is used`, function() {
+    expect(moduleRegistry.getImportStatements()).to.be.empty;
+  });
```
philipwalton: Can you nest this (and the following blocks) within a `describe()` block?

So like this:

```js
describe('getImportStatements()', function() {
  it(`should return [] when nothing is used`, function() {
    expect(moduleRegistry.getImportStatements()).to.be.empty;
  });
});
```

Viktor286hearts:
@jeffposnick custom Workbox bundle via generateSW is a great feature, thank you for adding it! 

Excuse me for repeating my [question from this issue](https://github.com/GoogleChrome/workbox/issues/2064#issuecomment-637850957), but is there a way to setup destination path for "custom Workbox runtime bundle" or it always saves in the same directory as main generated `sw.js` file (`swDest` property)?

jeffposnick:
It's always alongside the `swDest` file. The `importScripts()` statements inside of the generated service worker will always look for the external runtime in the same directory.

PR #2071, opened 2019-06-06
Add ${pathnameAndSearch} info within _match() logging

nuragic:
Given that I struggled to find these console logs because I was using console filtering (which is a quite common practice), I think it would be nice to add the `${pathnameAndSearch}` within the `_match()` logging, in order to discover issues like this more quickly: https://github.com/facebook/create-react-app/pull/7176

Cc @jeffposnick

Thanks!

### DIFF
--- a/packages/workbox-routing/NavigationRoute.mjs
+++ b/packages/workbox-routing/NavigationRoute.mjs
@@ -88,8 +88,9 @@ class NavigationRoute extends Route {
     for (const regExp of this._blacklist) {
       if (regExp.test(pathnameAndSearch)) {
         if (process.env.NODE_ENV !== 'production') {
-          logger.log(`The navigation route is not being used, since the ` +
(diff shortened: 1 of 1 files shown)


nuragic:
Ups sorry! I'll fix that.

nuragic:
Fixed. What do you think about using e.g. https://github.com/declandewet/common-tags#oneline to avoid the `+` concatenations? 😄

jeffposnick:
We deliberately avoid using external dependencies inside of the code that runs in the service worker and/or in the `window` context.

(We [do use](https://github.com/GoogleChrome/workbox/blob/master/packages/workbox-cli/src/app.js#L10) `commen-tags` in our node-based tools, however.)

nuragic:
Oh, ok it's great that you don't have external deps! 👌

PR #2085, opened 2019-07-01
Convert workbox-window to Typescript

philipwalton:
R: @jeffposnick

This PR continues the migration to Typescript (started in #2058) and adds type definitions for `workbox-window`.

Also, as part of the migration of the first non-core package, I had to make a few changes to the TS build and transpile process to get things working with cross-package dependencies. Here is a list of all changes

* The `.js` extension was added to all import statements (which is necessary until https://github.com/microsoft/TypeScript/issues/16577 is resolved).
* The build now uses `tsc` instead of `rollup-plugin-typescript2` and leverages its [project references](https://www.typescriptlang.org/docs/handbook/project-references.html) to support faster incremental builds


packages/workbox-window/src/Workbox.ts
```diff
@@ -40,7 +42,25 @@ const REGISTRATION_TIMEOUT_DURATION = 60000;
  *
  * @memberof module:workbox-window
  */
-class Workbox extends EventTargetShim {
+class Workbox extends WorkboxEventTarget {
+  private _scriptURL: 'string';
```
jeffposnick: What's with the `''` around `'string'`? As far as I can tell, the type is `string`, without quotes: https://www.typescriptlang.org/docs/handbook/basic-types.html#string

packages/workbox-window/src/Workbox.ts
```diff
@@ -40,7 +42,25 @@ const REGISTRATION_TIMEOUT_DURATION = 60000;
...
+  private _registerOptions: RegistrationOptions = {};
+  private _updateFoundCount: number = 0;
+
+  // Deferreds we can resolve later.
+  private _swDeferred: Deferred<ServiceWorker> = new Deferred();
+  private _activeDeferred: Deferred<ServiceWorker> = new Deferred();
+  private _controllingDeferred: Deferred<ServiceWorker> = new Deferred();
+
+  private _registrationTime: DOMHighResTimeStamp = 0;
+  private _isUpdate?: Boolean;
```
jeffposnick: Similar question about `Boolean` vs. `boolean`, with the lower-case version being what I see at https://www.typescriptlang.org/docs/handbook/basic-types.html#boolean

Is there a linting rule that might catch this sort of thing, to ensure consistency if both are supported?

packages/workbox-window/src/Workbox.ts
```diff
@@ -250,7 +258,7 @@ class Workbox extends EventTargetShim {
    * @param {Object} data An object to send to the service worker
    * @return {Promise<Object>}
    */
-  async messageSW(data) {
+  async messageSW(data: {}) {
```
jeffposnick: Similar question for `{}` vs. `object` here: https://www.typescriptlang.org/docs/handbook/basic-types.html#object

packages/workbox-window/src/Workbox.ts
philipwalton: Good catch! Fixed.

packages/workbox-window/src/Workbox.ts
philipwalton: Changed to `boolean`. As for linting, my plan was to add that to eslint at some point, but I haven't looked into it yet.

packages/workbox-window/src/Workbox.ts
philipwalton: Updated to `object`.

I'd originally thought the Google TS style guide preferred `{}` over `object`, but upon further reading it says this:

> There are a few types related to JavaScript primitives that should never be
used: - `String`, `Boolean`, and `Number` have slightly different meaning from
the corresponding primitive types `string`, `boolean`, and `number`. Always use
the lowercase version. - `Object` has similarities to both `{}` and `object`,
but is slightly looser. Use `{}` for a type that include everything except
`null` and `undefined`, or lowercase `object` to further exclude the other
primitive types (the three mentioned above, plus `symbol` and `bigint`).

Issue #1373, opened 2018-03-16
`dontCacheBustUrlsMatching` ignored in Workbox v3?


freezy:
**Library Affected**:
workbox-webpack-plugin

**Browser & Platform**:
Webpack 4.1.1, Workbox 3.0.0, Node 8.9.1, Windows 10 x64

**Issue or Feature Request Description**:
When migrating to v3 I noticed that most of my assets are cache-busted when they used not to be with v2.

You can reproduce using this repo:

```bash
$ git clone https://github.com/vpdb/website
$ cd website && git checkout workbox3
$ npm install && npm run build:prod
```
For example, the manifest contains:
```js
{
    "revision": "ad84f24427599d1c1b955914cfd9a4fa",
    "url": "https://vpdb.io/profile/settings/profile.settings-a1VkNzRv.html"
}
```
whereas v2 of Workbox produces:
```js
{
    "url": "https://vpdb.io/profile/settings/profile.settings-a1VkNzRv.html"
}
```

The [config](https://github.com/vpdb/website/blob/workbox3/config/webpack.prod.js#L26) uses this value:

    dontCacheBustUrlsMatching: /-([0-9a-f]{12}|[0-9A-Za-z]{8})\./,

I've also noticed that `globIgnores` has the same problem, while `exclude` works fine. Furthermore, chunks never seem to be cache-busted, even when `dontCacheBustUrlsMatching` isn't set.

jeffposnick:
See https://github.com/GoogleChrome/workbox/issues/1380 and https://developers.google.com/web/tools/workbox/modules/workbox-webpack-plugin#cache_additional_non-webpack_assets

We need to do a better job of explaining this to folks.

In general, if you've got `webpack` configured to include hash value in an asset name, the `workbox-webpack-plugin` attempts to detect this automatically and excludes the revision field by default.

The detection of which assets include `[hash]` or `[chunkhash]` or something related to that is unfortunately not 100% accurate. See https://stackoverflow.com/questions/47658134

freezy:
Well the glob thing was just something I noticed. All my assets run through Webpack, so this doesn't seem to be an issue (`exclude` covers this just fine).

When you say you're detecting hashes in assets automatically, does that mean that the `dontCacheBustUrlsMatching` option is now only applied to non-Webpack assets? That would be a breaking change, right?

tmchng:
@jeffposnick 

[...]

PR #2087, opened 2019-07-02
webpack plugin updates for v5

### LINKED ISSUE #1227: [workbox-webpack-plugin] v3 Entries array passed to manifestTransforms is empty
**Library Affected**:
workbox-webpack-plugin@3.0.0-alpha.6

**Browser & Platform**:
node v8.9.1, npm v5.5.1, Mac OSX 10.13.2

Title describes the issue, the manifestEntries array passed as an argument into manifestTransform functions is empty.

I've created a repo to illustrate the problem.  Here's a link to the branch with a minimal example reproducing the issue:

https://github.com/akash5474/workbox_ex/tree/manifestTransforms

### LINKED ISSUE #1287: Fail the Webpack build if a chunk doesn't exist

### LINKED ISSUE #1338: importWorkboxFrom: 'local' bypasses the webpack asset pipeline
**Library Affected**:
*workbox-webpack-plugin*

Moving over @raejin's [comment](https://github.com/GoogleChrome/workbox/pull/1290#issuecomment-369423980) into a standalone issue:

> After testing out v3.0.0-beta.1, I think this is not compatible with static asset that gets served from a different domain. By tracing the source a bit, we are copying the file on the disk instead of copying the file to the Webpack compilation.
> 
> https://github.com/raejin/workbox/blob/17c38de990585d7076b03ecca55d33b3166e53e9/packages/workbox-build/src/lib/copy-workbox-libraries.js#L78-L79
> 
> What do you guys think of a different implementation on copying on disk versus Webpack compilation?

### LINKED ISSUE #1363: Warn when swDest is set to the webpack output directory
**Library Affected**:
workbox-webpack-plugin

**Browser & Platform**:
macOS High Sierra

I'm using `workbox-webpack-plugin@^3.0.0` and trying to generate the service worker with this plugin setup:

```
new WorkboxPlugin.GenerateSW({
      chunks: Object.keys(base.entry),
      include: [/\.css$/, /\.js$/],
      swDest: path.resolve(__dirname, 'dist')
    })
```
and I'm getting this error when running `webpack-dev-server`, looks like the module is not able to write the file in the file system.

```
Error
    at MemoryFileSystem.writeFileSync (/node_modules/memory-fs/lib/MemoryFileSystem.js:211:9)
    at MemoryFileSystem.writeFile (/node_modules/memory-fs/lib/MemoryFileSystem.js:324:8)
    at writeOut (/node_modules/webpack/lib/Compiler.js:339:28)
    at Immediate.<anonymous> (/node_modules/memory-fs/lib/MemoryFileSystem.js:288:4)
    at runCallback (timers.js:800:20)
    at tryOnImmediate (timers.js:762:5)
    at processImmediate [as _immediateCallback] (timers.js:733:5)
```

### LINKED ISSUE #1454: CSS with contenthash should not be revisioned
**Library Affected**: workbox-sw, workbox-webpack-plugin

**Browser & Platform**:
"all browsers"

I am using workbox-webpack-plugin with InjectManifest plugin. I use [mini-css-extract-plugin](https://github.com/webpack-contrib/mini-css-extract-plugin) to extract my css into a separate file. For long term caching, I use `contenthash` in my filename but looks like workbox is not able to detect it and creates revision field for it. Thereby, re-downloading the css everytime any of my js changes.

image

### LINKED ISSUE #1485: GenerateManifest in workbox-webpack-plugin
**Library Affected**:
*workbox-webpack-plugin*

Like `workbox-webpack-plugin~InjectManifest`, can we have  `workbox-webpack-plugin~GenerateManifest`? 

It should have one callback function with following arguments.
1. webpack plugin callbacks arguments (`compiler`, `compilation`)
2. `generated-precahce-manifest-filename`

Now developer can inject precahce-manifest-filename in compiled service worker file according to his choice.

### LINKED ISSUE #1493: The workbox-webpack-plugin generate an error url in the precache-manifest.js file
**Library Affected**:
workbox-webpack-plugin  

**Browser & Platform**:
windows 10 x64  
node.js 8.11.1  
webpack 4.8.3  
workbox-webpack-plugin 3.2.0  
html-webpack-plugin 3.2.0  

Use ```webpack``` to generate the ```js``` file in the ```/dist/assets/``` directory.  
Use ```html-webpack-plugin``` to generate the ```html``` file in the ```/dist/``` directory.  
The ```workbox-webpack-plugin``` will generate an error ```url``` in the ```precache-manifest.js``` file.  

Here is an example https://github.com/pfdgithub/workbox-webpack-plugin-demo  

webpack.config.js  
```
const path = require('path');
const HtmlWebpackPlugin = require('html-webpack-plugin');
const WorkboxPlugin = require('workbox-webpack-plugin');

module.exports = {
  mode: 'development',
  entry: {
    index: './src/index.js'
  },
  output: {
    path: path.resolve(__dirname, 'dist', 'assets'),
    filename: '[name].[chunkhash].js',
    publicPath: '/assets/',
  },
  plugins: [
    new HtmlWebpackPlugin({
      filename: path.resolve(__dirname, 'dist', 'index.html')
    }),
    new WorkboxPlugin.GenerateSW({
      swDest: path.resolve(__dirname, 'dist', 'service-worker.js')
    })
  ]
};
```

Build output
```
Hash: 6f0da71c3b43dd24faf2
Version: webpack 4.8.3
Time: 598ms
Built at: 2018-05-24 11:18:28
                                                Asset       Size  Chunks             Chunk Names
                        index.0c3568d41c904e638ac1.js   3.14 KiB   index  [emitted]  index
                                        ..\index.html  210 bytes          [emitted]
precache-manifest.1b5cad419bfeb0a74d42b688531989eb.js  186 bytes          [emitted]
                                 ..\service-worker.js  915 bytes          [emitted]
Entrypoint index = index.0c3568d41c904e638ac1.js
[./src/index.js] 309 bytes {index} [built]
Child html-webpack-plugin for "..\index.html":
     1 asset
    Entrypoint undefined = ..\index.html
    [./node_modules/webpack/buildin/global.js] (webpack)/buildin/global.js 509 bytes {0} [built]
    [./node_modules/webpack/buildin/module.js] (webpack)/buildin/module.js 519 bytes {0} [built]
        + 2 hidden modules
```

precache-manifest.js  
```
self.__precacheManifest = [
  {
    "url": "/assets/index.0c3568d41c904e638ac1.js"
  },
  {
    "revision": "1e49e2609410a3c4b3a3f57f23d50519",
    "url": "/assets/..\\index.html"
  }
];
```

### LINKED ISSUE #1513: InjectManifest Plugin doesn't compile the serviceworker with webpack

### LINKED ISSUE #1529: Precache manifest inaccurate for ExtractTextPlugin output
(This issue was moved over from https://stackoverflow.com/questions/50836611)

**Library Affected**:
*workbox-webpack-plugin*

I'm using Angular (custom Webpack build procedure) together with the Workbox v.3.2.0 with workbox-webpack-plugin v.3.2.0 (InjectManifest plugin) for the precaching purposes and have difficulties with revisioning of the style.css file that is generated by the extract-text-webpack-plugin.

If I change the appropriate CSS files, Webpack generates updated style.css file, but its revision is not changed in the precache manifest.

As a result, I end up with the file that is not updated in the service worker cache.

Here are the related parts of the Webpack configuration (full config is too big):

```js
{
  entry: {
    'polyfills': './src/polyfills.ts',
    'main': AOT ? './src/main.browser.aot.ts' : './src/main.browser.ts',
    'style': './src/main.scss'
  },
  resolve: {
    extensions: ['.ts', '.js', 'json', '.scss', '.html'],
    modules: [helpers.root('src'), helpers.root('node_modules')]
  }, 
  module: {
    rules: [
      ...
      {
        test: /\.scss$/,
        use: ExtractTextPlugin.extract(['css-loader?sourceMap', 'postcss-loader', 'sass-loader?sourceMap']),
        exclude: [helpers.root('src', 'styles')]
      },
      ...
    ]
  },
  plugins: [
    ...
    new ExtractTextPlugin("[name].css"),
    new InjectManifest({
      swSrc: './service-worker.js',
      include: [
        /\.html$/,
        /\.js$/,
        /\.css$/,
        /\.ico$/,
        /\.json$/,
        /\.png/,
        /\.svg$/,
        /\.gif$/,
        /\.woff$/,
        /\.ttf$/,
        /\/workbox.*\/.*$/,
        /css.*\.css$/,
        /fonts.*\.(svg|eot|ttf|woff)$/,
        /i18n.*\.json$/,
        /img.*\.(svg|png|gif|jpeg|jpg)$/,
      ],
      exclude: [
        /poc\/.*$/,
        /service-worker.js$/,
        /icons-sprite/,
      ],
      importWorkboxFrom: 'local',
    }),
    ...
  ]
}
```

Please, note, that I can't use somethig like `new ExtractTextPlugin("[name].[contenthash].css")` due to some issues with caching the outdated version of the index.html referencing files from the previous releases (not existing anymore). That's why we use anti-cache URL params instead of embedding the revision in the file name.

For any other Webpack-generated file the revision is updated without any issues (even for the assets that do not have revision embedded into the filename)

Could anyone please help me out here? Thanks :)

### LINKED ISSUE #1543: workbox-webpack-plugin - InjectManifest links to manifest relative to webpack root, not the swDest
**Library Affected**:
workbox-webpack-plugin

If I use swDest option in webpack plugin, InjectManifest creates a link in injectScripts in SW that is relative to webpack root, not the new swDest. Therefore, the injectScripts call in SW cannot find the manifest file. 

Webpack config:

```
const { InjectManifest } = require('workbox-webpack-plugin');

module.exports = {
  entry: {
    main: './src/client/js/app.js',
    ...
  },
  output: {
    filename: 'js/[name].js',
    chunkFilename: 'js/chunk-[name].js',
    path: path.resolve(__dirname, 'dist/static/js'),
  },
  plugins: [
    new InjectManifest({
      swSrc: './src/sw.js',
      swDest: '../sw.js',
    })
  ]
};

```
Outcome dir structure:

```
|--static
...... |-sw.js 
...... |--js
............ |--main.js
............ |--precache-manifest.XXX.js
```

Content of sw.js:

`importScripts("precache-manifest.XXX.js", ...);   //<- the js subdir is missing
`
I can solve it by changing the whole webpack to output to /static, but then swDest is obsolete.

### LINKED ISSUE #1556: workbox-webpack-plugin - `chunks` option in conflict with precaching `index.html`
**Library Affected**:
*workbox-webpack-plugin*

**Browser & Platform**:
*all browsers*

Background here: we are using the `webpack-html-plugin` to generate `index.html` and `workbox-webpack-plugin` to inject the `__precacheManifest` into our service worker file.

When `chunks` option is not specified, `InjecInjectManifest`will generate a list of all the files compiled by `webpack` (including all the static assets like html, images etc). However this is not really an option for us, as we have 100+ JS files. It's not wise to precache all of them upfront.

So we are using the `chunks` option to only precache the application shell related code. However, since the `webpack-html-plugin` generated html will have no associated chunks. The `index.html` will be filtered out from the `precacheManifest`

I tried to work around the issue by using the old `glob` way. However it's meant only for external files, not the webpack compilations. 

I dig into the source code and seems no way to work around this issue. I'm pretty sure this is a common use case for large websites like us, am I missing anything here or this is a feature by     ####design?

### LINKED ISSUE #1591: workbox-webpack-plugin v5 interface
# Background

The team is planning on addressing the various issues and pain points raised against the `workbox-webpack-plugin` with a number of breaking changes to the interface and behavior in an upcoming v4.0.0 release.

We're looking to gather feedback from the community prior to starting on the implementation, so please feel free to reply if you've got any concerns.

# Open issues

These are categorized, but not prioritized, based on the [open issues](https://github.com/GoogleChrome/workbox/issues?q=is%3Aissue+is%3Aopen+label%3Aworkbox-webpack-plugin). It gives a good overview of the general areas in which feedback has been concentrated.

We won't be able to address *all* of these issues in the initial v4.0.0 release. Our goal is to address many of them that require an incompatible API surface or a significant change in behavior in at once, to minimize the number of major releases with breaking changes.

## Filesystem paths

*   [workbox-webpack-plugin - InjectManifest links to manifest relative to webpack root, not the swDest](https://github.com/GoogleChrome/workbox/issues/1543) (#1543)
*   [Warn when swDest is set to the webpack output directory](https://github.com/GoogleChrome/workbox/issues/1363) (#1363)

## URL paths

*   [InjectManifest Plugin: Generate manifest url based on config instead of webpack public path](https://github.com/GoogleChrome/workbox/issues/1565) (#1565)
*   [manifest file generated by workbox-webpack-plugin has wrong public path when using file-loader](https://github.com/GoogleChrome/workbox/issues/1564) (#1564)

## Asset pipeline

*   [workbox-webpack-plugin - `chunks` option in conflict with precaching `index.html`](https://github.com/GoogleChrome/workbox/issues/1556) (#1556)
*   [Precache manifest inaccurate for ExtractTextPlugin output](https://github.com/GoogleChrome/workbox/issues/1529) (#1529)
*   [CSS with contenthash should not be revisioned](https://github.com/GoogleChrome/workbox/issues/1454) (#1454)
*   [importWorkboxFrom: 'local' bypasses the webpack asset pipeline](https://github.com/GoogleChrome/workbox/issues/1338) (#1338)
*   [Fail the Webpack build if a chunk doesn't exist](https://github.com/GoogleChrome/workbox/issues/1287) (#1287)
*   [Precaching the generated WebpackPwaManifest manifest asset](https://github.com/GoogleChrome/workbox/issues/1588) (#1588)
* [Precache manifest missing entries when recompiled via webpack --watch](https://github.com/GoogleChrome/workbox/issues/1790) (#1790)

## Parity with "glob" configuration

*   [templatedUrls option fails if no globDirectory provided](https://github.com/GoogleChrome/workbox/issues/1418) (#1418)
*   [[workbox-webpack-plugin] proposal to add support Webpack assets in `templatedUrls` option](https://github.com/GoogleChrome/workbox/issues/1384) (#1384)
*   [`dontCacheBustUrlsMatching` ignored in Workbox v3?](https://github.com/GoogleChrome/workbox/issues/1373) (#1373)
*   [[workbox-webpack-plugin] v3 Entries array passed to manifestTransforms is empty](https://github.com/GoogleChrome/workbox/issues/1227) (#1227)

## Developer experience

*   [InjectManifest Plugin doesn't compile the serviceworker with webpack](https://github.com/GoogleChrome/workbox/issues/1513) (#1513)
*   [[Feature] Add ability to specify [hash] in navigationFallback](https://github.com/GoogleChrome/workbox/issues/1506) (#1506)
*   [The workbox-webpack-plugin generate an error url in the precache-manifest.js file](https://github.com/GoogleChrome/workbox/issues/1493) (#1493)
*   [GenerateManifest in workbox-webpack-plugin](https://github.com/GoogleChrome/workbox/issues/1485) (#1485)
*   [Webpack plugin causes massive prolonged spike in CPU and Memory](https://github.com/GoogleChrome/workbox/issues/1475) (#1475)
*   [New webpack plugin config option to specify hash length](https://github.com/GoogleChrome/workbox/issues/1458) (#1458)
*   [Use webpack's performance.hints for reporting non-fatal errors](https://github.com/GoogleChrome/workbox/issues/1380) (#1380)
*   [Integration with the webpack DllPlugin](https://github.com/GoogleChrome/workbox/issues/1348) (#1348)
*   [accept chunk names in importScripts](https://github.com/GoogleChrome/workbox/issues/1752) (#1752)

# Proposed v5 plugin interface

This interface will apply to both `InjectManifest` and `GenerateSW`. (`swSrc` is the only option that's limited to `InjectManifest`.)

I'll keep this example up-to-date as folks add feedback to the issue thread.

```js
new InjectManifest or GenerateSW ({
  // Same as in v3:
  swDest,
  importWorkboxFrom,
  importsDirectory,
  precacheManifestFilename,
  skipWaiting,
  clientsClaim,
  runtimeCaching,
  navigateFallback,
  navigateFallbackBlacklist,
  navigateFallbackWhitelist,
  importScripts,
  ignoreUrlParametersMatching,
  directoryIndex,
  cacheId,
  maximumFileSizeToCacheInBytes,
  // InjectManifest only:
  swSrc,

  // Newly changed in v4:
  webpackAssets: {
    // Moved to be under a webpackAssets parent:
    chunks,
    excludeChunks,
    include,
    exclude,

    // The next two are the same as the v3 options,
    // but now only apply to webpack assets.
    modifyUrlPrefix,
    manifestTransforms,

    // Maps a static URL to one or more webpack assets. E.g.
    // {
    //   '/shell': {
    //     chunks: ['chunk1'],
    //     include: [/*.js/],
    //   }
    // }
    templatedUrls,

    // Array defaulting to ['[hash]'].
    // You can add any RegExp, or options along the lines of
    // https://github.com/webpack/loader-utils#interpolatename
    dontCacheBustUrlsMatching,
  },

  fileSystemAssets: {
    // Same as in v3, but moved to be under a fileSystemAssets parent.
    globDirectory,
    globFollow,
    globIgnores,
    globPatterns,
    globStrict,

    // The next four behave the same as the v3 options,
    // but explicitly only apply to file system assets.
    modifyUrlPrefix,
    manifestTransforms,
    templatedUrls,
    dontCacheBustUrlsMatching,
  },

  // Removed as top-level options in v4:
  /*
    dontCacheBustUrlsMatching,
    templatedUrls,
    globDirectory,
    globFollow,
    globIgnores,
    globPatterns,
    globStrict,
    modifyUrlPrefix,
    manifestTransforms,
    chunks,
    excludeChunks,
    include,
    exclude,
  */
});
```

### LINKED ISSUE #1774: Inject generated precache-manifest directly inside the service worker instead of creating a precache-manifest file.

I'm using `workbox-webpack-plugin`'s `GenerateSW`. But it creates a precache manifest file resulting to the following code:

```js
/**
 * Welcome to your Workbox-powered service worker!
 *
 * You'll need to register this file in your web app and you should
 * disable HTTP caching for this file too.
 * See https://goo.gl/nhQhGp
 *
 * The rest of the code is auto-generated. Please don't update this file
 * directly; instead, make changes to your Workbox build configuration
 * and re-run your build process.
 * See https://goo.gl/2aRDsh
 */

importScripts("workbox-v3.6.3/workbox-sw.js");
workbox.setConfig({modulePathPrefix: "workbox-v3.6.3"});

importScripts(
  "precache-manifest.4d330e4e1240ee5970dbef2d1c1a1fd7.js"
);

/**
 * See https://goo.gl/S9QRab
 */
workbox.precaching.suppressWarnings();
```

What I want is to remove `importScripts("precache-manifest.4147d0eae38244e105572db91f23fa97.js");` and totally inject the contents of `precache-manifest` inside of the service worker resulting to the following code:

```js
/**
 * Welcome to your Workbox-powered service worker!
 *
 * You'll need to register this file in your web app and you should
 * disable HTTP caching for this file too.
 * See https://goo.gl/nhQhGp
 *
 * The rest of the code is auto-generated. Please don't update this file
 * directly; instead, make changes to your Workbox build configuration
 * and re-run your build process.
 * See https://goo.gl/2aRDsh
 */

importScripts("workbox-v3.6.3/workbox-sw.js");
workbox.setConfig({modulePathPrefix: "workbox-v3.6.3"});

self.__precacheManifest = [
  {
    "url": "css/file2.d39cd1945dcec8bba59a.css"
  },
  {
    "url": "js/file.d39cd1945dcec8bba59a.js"
  }
];

/**
 * See https://goo.gl/S9QRab
 */
workbox.precaching.suppressWarnings();
```

I'm not sure if it already exists, I'm looking at the issues to see if someone else has opened a similar issue but I also couldn't search for one.

Here's my config btw:

```js
new workboxWebpackPlugin.GenerateSW({
  swDest: 'serviceWorker.js',
  importWorkboxFrom: 'local',
  excludeChunks: ['runtime', 'main.css']
})
```

### LINKED ISSUE #1854: Proposal/RFC: workbox-webpack-plugin changes

### LINKED ISSUE #1859: webpack plugin does not respect `splitChunks`
**Library Affected**:
workbox-webpack-plugin

I have enabled [webpack's `splitChunks` option](https://webpack.js.org/plugins/split-chunks-plugin/). This splits my chunks so that "vendor" imports are stored in a separate asset. For example, the `main` chunk will result in two assets: `main.js` *and* `vendors~main.js`.

When I use `InjectManifest`'s `chunks` option to whitelist the `main` chunk, I expect it to include all the assets relating to this chunk. In this case, that would be `main.js` and `vendors~main.js`.

However, it does not seem to respect the vendor asset created by `splitChunks`—it only includes `main.js`.

It is not feasible for me to list these manually, because webpack may create any number of vendor assets, depending on the common module imports between all my chunks.

### LINKED ISSUE #2012: Is there a way to override publicPath?
**Library Affected**:
No libraries affected

**Browser & Platform**:
All

It's a feature request.
Currently, when I'm using GenerateSW from workbox-webpack-plugin@4.2.0 

`const {GenerateSW} = require('workbox-webpack-plugin');`

it generates  precache-manifest.js with the list of assets created during the webpack build phase. It works perfect, but it uses **publicPath** property from configuration in webpack 

`  
output: {
    publicPath: publicPaths.staticServer
  },`

and there is no way to override it. 
For example, for our case we need this, because we use different static servers for different types of assets.

Do you have plans to support this feature. Or can I just  implement this feature and create a pull request for it?

### LINKED ISSUE #2021: Runtime manifest doesn't respect [contenthash]
**Library Affected**:
workbox-webpack-plugin@4.2.0

**Browser & Platform**:
all browsers

Webpack provides ability to use [contenthash] in outputnames (see https://webpack.js.org/guides/caching/) In this case, when I'm using GenerateSW from workbox-webpack-plugin library, my generated precache-manifest.js doesn't respect this option and generates it's own revision number, after that the resources could be downloaded from the network two times,  if they are in the critical rendering path for the app and in the manifest.

```
self.__precacheManifest = (self.__precacheManifest || []).concat([

  {
    "revision": "f62b7714d7f5a2ad8629",
    "url": "/resources/publisher.c537d714c8ba995c06cc.css"
  },
]);

### LINKED ISSUE #2039: Webpack plugin: Only create precache manifest for a specified entrypoint
**Library Affected**:
workbox-webpack-plugin

We’ve a Webpack configuration with several entry points which represent (somewhat) isolated parts. I’d like to build the precache manifest JSON file only for the files which are being used in **one** entry point of the project -- files and assets (managed via `file-loader`) from other entry points should **not** appear in the generated manifest.

I tried using both, the `chunks` and the `excludeChunks` settings, but neither give me the desired result:

* when I use `chunks` and set it to the relevant entry point (e.g. `entrypoint1`), my manifest does not contain the assets (e.g. images) which are referenced in the entry point

* when I use `excludeChunks`, the manifest contains **all** assets (e.g. all images from the entire project, no matter whether in which entrypoint they're being referenced)

A minimal example project which demonstrates my issue can be found here:
https://github.com/qqilihq/workbox-chunk-issue

**Steps to reproduce:**

1. Clone the repo
2. `yarn`
3. `yarn build`
4. Check `dist/precache-manifest.af8924052142092476d2b7a9665c1d2e.js` -- I'd expect it to contain an image beside the JS file
5. Modify `webpack.config.js` and toggle comments in line 18 and 19 to use `excludeChunks`
6. Check `dist/precache-manifest.89bef461327312c73acadd63bd01a11a.js` -- I'd epxect it to contain **one** image (the one from `entrypoint1.js`), however it also contains an image from `entrypoint2.js`

### LINKED ISSUE #2059: Simpler InjectManifest replacement behavior

### LINKED ISSUE #2064: generateSW uses custom Workbox runtime bundles, not CDN

### PR DESCRIPTION
**Pull request by jeffposnick**
R: @philipwalton 

This includes a pretty large refactoring of both of the webpack plugins, to match the proposed design and behavior for v5.

In general, a lot of the webpack code has gotten simpler, and we're sharing more with `workbox-build`, which should lead to better consistency.

I have not yet updated any of the tests, and I'll be working on that, but I wanted to give you a chance to look at the code changes first.

Additionally, I think a refactoring of the code under https://github.com/GoogleChrome/workbox/tree/wepback-v5/packages/workbox-build/src/entry-points/options is probably in order, as that's now a grab bag of things used directly by both `workbox-build` and `workbox-webpack-plugin`. I'd like to save that for a different PR, though.

Fixes #2064, fixes #2059, fixes #2039, fixes #1859, fixes #2021, fixes #2012, fixes #1854, fixes #1774, fixes #1591, fixes #1543, fixes #1363, fixes #1556, fixes #1529, fixes #1454, fixes #1338, fixes #1287, fixes, #1373, fixes #1227, fixes #1513, fixes #1493, fixes #1485

### CODE REVIEW

File: packages/workbox-build/src/entry-points/options/webpack-common.js [hunk 1]
```diff
@@ -0,0 +1,21 @@
+/*
+  Copyright 2018 Google LLC
```
philipwalton: s/2018/2019/

packages/workbox-build/src/entry-points/options/webpack-inject-manifest-schema.js
```diff
@@ -0,0 +1,25 @@
+/*
+  Copyright 2018 Google LLC
```
philipwalton: s/2018/2019/

packages/workbox-build/src/lib/loadz0r-loader.js
```diff
@@ -0,0 +1,69 @@
+/*
+  Copyright 2018 Google LLC
```
philipwalton: s/2018/2019/

jeffposnick:
Okay, the tests are updated.

Post-merge, I think we should be good to cut the first v5 alpha and start getting feedback on the new build tools.

PR #2088, opened 2019-07-02
Convert workbox-background-sync to Typescript

philipwalton:
R: @jeffposnick

This PR continues the migration to Typescript (started in #2058) and adds type definitions for `workbox-background-sync`.


packages/workbox-background-sync/src/Queue.ts
```diff
@@ -86,15 +107,15 @@ class Queue {
    *     for you (defaulting to `Date.now()`), but you can update it if you
    *     don't want particular requests to expire.
    */
-  async pushRequest(entry) {
+  async pushRequest(entry: QueueEntry) {
     if (process.env.NODE_ENV !== 'production') {
-      assert.isType(entry, 'object', {
+      assert!.isType(entry, 'object', {
```
jeffposnick: Just checking—this is an alternative to `assert && assert.is...()` that you've used in previous PRs?

packages/workbox-background-sync/src/Queue.ts
```diff
@@ -211,7 +235,8 @@ class Queue {
       entry.metadata = metadata;
     }
 
-    await this._queueStore[`${operation}Entry`](entry);
+    await this._queueStore[
+        `${operation}Entry` as 'pushEntry' | 'unshiftEntry'](entry);
```
jeffposnick: I don't think I've seen this syntax before—this is how you tell TypeScript that `operation` might be either `push` or `unshift`?

packages/workbox-background-sync/src/lib/StorableRequest.ts
```diff
@@ -22,6 +25,13 @@ const serializableProperties = [
   'keepalive',
 ];
 
+export interface RequestData {
+  url: string;
+  headers: {[headerName: string]: any};
```
jeffposnick: Should this be typed as that official `Headers` interface you manually added? Or is this not actually a real `Headers`?

packages/workbox-background-sync/src/Queue.ts
philipwalton: The method name called was being dynamically generated as a way to reduce total code size. I could add an `if` or `switch` statement (which I'm fine with as it's not *that* much more code), but specifying the possible method names seemed to make Typescript happy, so I opted to do that rather than refactoring -- at least for now.

packages/workbox-background-sync/src/lib/StorableRequest.ts
philipwalton: It's an object of key/value pairs. I suppose it may make more sense to rework this as a `HeadersInit` object (i.e. `string[][]`) and reuse the existing typing, but yeah, it's definitely not a `Headers` object because it needs to be structured cloneable.

jeffposnick: Gotcha.

packages/workbox-background-sync/src/Queue.ts
jeffposnick: It's fine as-is.

packages/workbox-background-sync/src/Queue.ts
philipwalton: Hmmmm, I don't remember doing that, but looks like I did.

Yeah, the `!` operator is just a way of telling Typescript that you know a value in some context is definitely not `null` or `undefined`. In this case it's because we're in the `process.env.NODE_ENV !== 'production'` block.

I'll change those previous cases to the more idiomatic form in the next PR.

PR #2100, opened 2019-07-10
Convert workbox-routing to Typescript 

philipwalton:
R: @jeffposnick

This PR continues the migration to Typescript (started in #2058) and adds type definitions for workbox-routing.


infra/type-overrides.d.ts
```diff
@@ -11,6 +11,12 @@ interface IDBObjectStore {
   openKeyCursor(query?: IDBValidKey | IDBKeyRange | null, direction?: IDBCursorDirection): IDBRequest<IDBCursor | null>;
 }
 
+// Service Worker
+// ------------------------------------------------------------------------- //
+interface CacheQueryOptions {
```
jeffposnick: There's a [few more](https://developers.google.com/web/updates/2016/09/cache-query-options): `ignoreSearch`, `cacheName`, `ignoreMethod`, and `ignoreVary`.

I guess only `cacheName` is needed for this PR, but might as well add them all?

packages/workbox-routing/src/Router.ts
```diff
@@ -241,28 +263,32 @@ class Router {
    * @param {Object} options
    * @param {URL} options.url
    * @param {Request} options.request The request to match.
-   * @param {FetchEvent} [options.event] The corresponding event (unless N/A).
+   * @param {Event} [options.event] The corresponding event (unless N/A).
```
jeffposnick: Should this be `ExtendableEvent` to match the argument type?

packages/workbox-routing/src/_types.ts
```diff
@@ -6,7 +6,17 @@
   https://opensource.org/licenses/MIT.
 */
 
-import './_version.mjs';
+import './_version.js';
+
+export interface MatchCallbackOptions {
+  url: URL;
+  request?: Request;
```
jeffposnick: I *think* `request` is meant to be required. Is there a place in the codebase that we use `MatchCallbackOptions` where `request` isn't set?

packages/workbox-routing/src/registerRoute.ts
```diff
@@ -38,11 +40,14 @@ import './_version.mjs';
  *
  * @alias workbox.routing.registerRoute
  */
-export const registerRoute = (capture, handler, method = 'GET') => {
+export const registerRoute = (
+    capture: RegExp | string | matchCallback | Route,
+    handler: handlerCallback,
```
jeffposnick: Both `handler` and `method` are optional the first parameter is a `Route`.

(This method is terribly overloaded.)

infra/type-overrides.d.ts
philipwalton: The built-in TS types includes all those. The only one it was missing was `cacheName` for some reason (perhaps because it shares the same type between `caches.match()` and `cache.match()`?).

packages/workbox-routing/src/Router.ts
philipwalton: I could, but elsewhere I've been keeping the JSDoc comments simple by using `Event`. I can't use `Event` in TS because it'll say things like `event.request` doesn't exist, so I have to be specific.

packages/workbox-routing/src/_types.ts
philipwalton: Not that I can see, but in our current docs [it's listed as being optional](https://developers.google.com/web/tools/workbox/reference-docs/latest/workbox.routing.Route#~matchCallback).

jeffposnick: I think it should be required. Let's switch and see if anyone ends up complaining...

philipwalton: Looks like [`request` is required in `findMatchingRoute()`](https://developers.google.com/web/tools/workbox/reference-docs/latest/workbox.routing.Router#findMatchingRoute), so at least in terms of our code it seems to be always used.

I suppose there's no harm in changing it to say it's now always passed. If someone wasn't using it before it won't hurt.

packages/workbox-routing/src/registerRoute.ts
philipwalton: Gotcha. Fixed.

PR #2105, opened 2019-07-11
Convert workbox-streams to Typescript

philipwalton:
R: @jeffposnick

This PR continues the migration to Typescript (started in #2058) and adds type definitions for workbox-streams.

Also, this PR changes some of the logic to make Typescript happy because we say the `StreamSource` type includes `BodyInit`, which technically includes `FormData` and `URLSearchParams` objects, but the old logic would break if those types were used.

To see just the edits (not the file renames), you can review this diff: https://github.com/GoogleChrome/workbox/pull/2105/commits/0824b5ae989a10ffe104be4c89c4dde816cf2e64.


packages/workbox-streams/src/strategy.ts
```diff
@@ -7,22 +7,26 @@
...
+
 /**
  * A shortcut to create a strategy that could be dropped-in to Workbox's router.
  *
  * On browsers that do not support constructing new `ReadableStream`s, this
  * strategy will automatically wait for all the `sourceFunctions` to complete,
  * and create a final response that concatenates their values together.
  *
- * @param {
- *   Array<function(workbox.routing.Route~handlerCallback)>} sourceFunctions
```
philipwalton: @jeffposnick I believe this is incorrect currently, so I've changed it in this PR. But just wanted you to double check to confirm.

jeffposnick: Here's an example of `workbox.streams.strategy` being used in the wild:

https://github.com/GoogleChromeLabs/so-pwa/blob/9385b059bb5265735525e4133eb3c4424ecba725/src/service-worker.mjs#L75-L92

I think based on that, yeah, `Array<handlerCallback>` is the appropriate type.

philipwalton: Yeah, and here's another example: https://github.com/philipwalton/blog/blob/23a69017694e8b63332f0cd42cefd92a7f8b287c/assets/sw/routes/page.js#L25-L29.

packages/workbox-streams/src/concatenate.ts
```diff
@@ -63,15 +62,10 @@ function concatenate(sourcePromises) {
     });
   });
 
-  let fullyStreamedResolve;
-  let fullyStreamedReject;
-  const done = new Promise((resolve, reject) => {
-    fullyStreamedResolve = resolve;
-    fullyStreamedReject = reject;
-  });
+  const streamDeferred:Deferred<void> = new Deferred();
```
jeffposnick: Nice reuse of `Deferred`.

PR #2108, opened 2019-07-12
Convert workbox-google-analytics to Typescript

philipwalton:
R: @jeffposnick

This PR wraps up the initial migration to Typescript (started in #2058) and adds type definitions for workbox-google-analytics.

It also consolidates all the common/shared type definitions in `workbox-core`, including all the route handler and match callback types as well as all the plugin callbacks.

Lastly, this PR adds project references to all workbox package dependencies for each package. This is needed, e.g. if you need to re-transpile a package and you've update type definitions in `workbox-core`.


tsconfig.json
```diff
@@ -32,5 +33,5 @@
     { "path": "./packages/workbox-strategies/" },
     { "path": "./packages/workbox-streams/" },
     { "path": "./packages/workbox-window/" }
-  ]
+  ],
```
jeffposnick: Trailing commas are invalid in JSON, right?

PR #2117, opened 2019-07-16
Use a .js extension when generating a default swDest

jeffposnick:
R: @philipwalton

Our webpack plugin will generate a default output service worker filename (`swDest`) using the basename of the required `swSrc` parameter when `swDest` isn't explicitly provided.

If the `swSrc` file is TypeScript, then chances are it has a `.ts` extension, and we don't want that to be used as the extension in `swDest`. (Ideally, the developer will have set up `ts-loader` to automatically compile the TypeScript file in their build process, so the output will be JavaScript.)

This change forces the `.js` file extension whenever we're calculating a default `swDest` value.


jeffposnick:
They could set `swDest: 'app.sw'` and get the behavior you describe. This change only applies when there's no `swDest` and we have to infer it based on the `swSrc` value.

philipwalton:
Ahh, gotcha, SGTM!

PR #2119, opened 2019-07-16
Define WorkboxGlobalScope in workbox-precaching

jeffposnick:
R: @philipwalton 

I'm not 100% sure that this is the right way to do things, so please review this with skepticism, but this definition is the cleanest way I could come up with to support using `self.__WB_MANIFEST` inside of a TypeScript `swSrc` file.

The usage ends up looking like:

```typescript
import {precacheAndRoute, WorkboxGlobalScope} from 'workbox-precaching/precacheAndRoute';

declare const self: WorkboxGlobalScope;

precacheAndRoute(self.__WB_MANIFEST);
```

Maybe we should have a separate `WorkboxGlobalScope` declaration inside of `workbox-core` instead, which would presumably make it cleaner if we ever added in custom global properties that weren't related to `workbox-precaching`?

### DIFF
--- a/packages/workbox-precaching/src/precache.ts
+++ b/packages/workbox-precaching/src/precache.ts
@@ -12,6 +12,9 @@ import {precachePlugins} from './utils/precachePlugins.js';
 import {PrecacheEntry} from './_types.js';
 import './_version.js';
 
+export interface WorkboxGlobalScope extends ServiceWorkerGlobalScope {
+  '__WB_MANIFEST': Array<PrecacheEntry|string>;
+}
 
 const installListener = (event: ExtendableEvent) => {
   const precacheController = getOrCreatePrecacheController();
--- a/packages/workbox-precaching/src/precache.ts
+++ b/packages/workbox-precaching/src/precache.ts
@@ -13,8 +13,8 @@ import {PrecacheEntry} from './_types.js';
 import './_version.js';
 
 declare global {
-  interface ServiceWorkerGlobalScope {
-    '__WB_MANIFEST': Array<PrecacheEntry|string>;
+  interface WorkerGlobalScope {
+    __WB_MANIFEST: Array<PrecacheEntry|string>;
   }
 }
 
--- a/packages/workbox-precaching/src/precache.ts
+++ b/packages/workbox-precaching/src/precache.ts
@@ -12,8 +12,10 @@ import {precachePlugins} from './utils/precachePlugins.js';
 import {PrecacheEntry} from './_types.js';
 import './_version.js';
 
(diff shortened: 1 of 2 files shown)


philipwalton:
E.g. `lit-html` does something similar, but I believe that's only working because they're referencing the property in their source: https://github.com/Polymer/lit-html/blob/47a8af4805d775614fe075975eba2c3f6e521a04/src/lit-html.ts#L50-L54

jeffposnick:
Thanks for the tip regarding `lit-html`'s usage. I think the magic here is the `declare global {}` wrapper around the `interface`.

With this latest commit, it looks like it just adds in support for the `__WB_MANIFEST` property as a side-effect of importing one of the precaching modules, and there's not requirement that I set `self.__WB_MANIFEST` to any value in the TypeScript source.

Given that this approach works, I'm thinking it makes more sense to keep it scoped tightly to the actual modules in which the additional properties will actually be used, rather than in `workbox-core`. If we need to extend `ServiceWorkerGlobalScope` with new properties in multiple places I believe that they're all considered additive, rather than overwriting each other.

philipwalton:
I see, so you're thinking of doing this but still expecting users to have to do this in their own code, correct?

```ts
declare var self: ServiceWorkerGlobalScope;
```

That seems fine to me, but note it's probably the case that many Workbox users wouldn't otherwise have to do `declare var self: ServiceWorkerGlobalScope` since they're likely not using `self` anywhere (they're using Workbox APIs).

Given that I wonder if it makes sense to also extend `WorkerGlobalScope`?

packages/workbox-precaching/src/precacheAndRoute.ts
```diff
@@ -12,6 +12,11 @@ import {precache} from './precache.js';
 import {PrecacheEntry} from './_types.js';
 import './_version.js';
 
+declare global {
+  interface ServiceWorkerGlobalScope {
+    '__WB_MANIFEST': Array<PrecacheEntry|string>;
```
philipwalton: You don't need quotes around `__WB_MANIFEST`.

packages/workbox-precaching/src/precache.ts
```diff
@@ -12,6 +12,11 @@ import {precachePlugins} from './utils/precachePlugins.js';
 import {PrecacheEntry} from './_types.js';
 import './_version.js';
 
+declare global {
+  interface ServiceWorkerGlobalScope {
+    '__WB_MANIFEST': Array<PrecacheEntry|string>;
```
philipwalton: You don't need quotes around `__WB_MANIFEST`.

jeffposnick:
I've switched to extending `WorkerGlobalScope`.

PR #2120, opened 2019-07-17
Adds the private TS annotation to _methods()

jeffposnick:
R: @philipwalton

When testing out the TypeScript functionality, I noticed that there were a bunch of methods on `workbox-window` that started with `_` and were `@private` in JSDoc, but weren't marked `private` in TypeScript.

I wonder if this is something we could catch via linting or some sort of post-TypeScript-compilation test?

### DIFF
--- a/packages/workbox-background-sync/src/Queue.ts
+++ b/packages/workbox-background-sync/src/Queue.ts
@@ -339,7 +339,7 @@ class Queue {
    *
    * @private
    */
-  _addSyncListener() {
+  private _addSyncListener() {
     if ('sync' in self.registration) {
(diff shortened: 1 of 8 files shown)


jeffposnick:
Want me to push more changes to this PR that cleans up other instances?

philipwalton:
If you want, but it'll like be a ton of files, so if you want to do it in a separate PR, that's fine too.

jeffposnick:
I did a quick search for the general pattern and found a few more obvious ones. We can include them in this PR, and then if I missed any, feel free to pick them up later.

PR #2128, opened 2019-07-19
Consolidate all stream-detection code in workbox-core

### LINKED ISSUE #2023: The cleanRedirect logic fails in Edge 18
**Library Affected**:
*workbox-precaching*

**Browser & Platform**:
*Edge 18 (pre Chromium)*

The conditional we use in `workbox-precaching` [`cleanRedirect()`](https://github.com/GoogleChrome/workbox/blob/v4.2.0/packages/workbox-precaching/utils/cleanRedirect.mjs#L21-L25) fails in Edge 18 because, while it supports `response.body`, it does not not support creating a new response from a body stream.

The best solution is to probably use the same check for streams support that we use in `workbox-streams` [`isSupported()`](https://github.com/GoogleChrome/workbox/blob/v4.2.0/packages/workbox-streams/isSupported.mjs#L26-L33), though we'd probably want to move that into `workbox-core`, so folks using `workbox-precaching` don't also have to download `workbox-streams` just for this one method.

We may also want to generalize the `cleanRedirect()` logic into a utility function that copies a response while allowing you to override headers. I think that would be generally useful to developers in many cases.

philipwalton:
R: @jeffposnick

Fixes #2023

This PR moves all the stream-detection code into `workbox-core` so it can be reused in multiple package. I opted not to have `workbox-precaching` depend on `workbox-streams` because I didn't want to arbitrarily inflate the usage count of `workbox-streams` just for feature detection (and not any _actual_ usage).

Note: I also had to `learn clean && lerna bootstrap` to get some type checking to update across packages. That's the reason for the `package-lock` updates in this PR.


packages/workbox-precaching/src/utils/cleanRedirect.ts
```diff
@@ -20,9 +21,9 @@ export async function cleanRedirect(response: Response) {
 
   // Not all browsers support the Response.body stream, so fall back
   // to reading the entire body into memory as a blob.
-  const bodyPromise = 'body' in clonedResponse ?
+  const bodyPromise = browserSupportsStreams() ?
```
jeffposnick: I'm concerned that this isn't checking the same thing—it's at least possible to imagine a browser in which `new ReadableStream({start() {}});` executes properly, but in which `clonedResponse` doesn't contain a `body` property.

Can you change this to `'body' in clonedResponse && browserSupportsStreams()` just to be safe?

philipwalton: What about having that check in the `browserSupportsStreams()` core function? Putting all the detection logic there would allow us to only use streams if they're fully supported for all our use cases.

Or do you think there may be situations where we're OK with partial support?

jeffposnick: I believe that these are all separate, potentially independently implemented features:

- Support for constructing a new `ReadableStream` from scratch.
- The `body` property of an existing `Response` can be read as a stream.
- Using a `ReadableStream` as the `BodyInit` when constructing a new `Response`.

CC: @wanderview who might be able to provide some more insight there, because I think that Firefox (at one time, at least) only supported a subset of those features, and I remember being bit by making assumptions in the past.

With this whole area, being explicit about checking for what we need (and only what we need) in order to enable a specific type of functionality seems the safest. We could end up regressing if we impose checks that are too stringent.

wanderview: At least for Firefox it was just that there was a separate flag to enable the constructor-vs-fetch-integrations.  So this:

* Support for constructing a new ReadableStream from scratch.

Might be true, but these might be false:

* The body property of an existing Response can be read as a stream.
* Using a ReadableStream as the BodyInit when constructing a new Response.

But these later two would either both be false or true at the same time.

At this point I believe FF has shipped all three of these and no shipping FF had any of the flags flipped with other configurations.

Hope that helps.

philipwalton: > With this whole area, being explicit about checking for what we need (and only what we need) in order to enable a specific type of functionality seems the safest.

But that's exactly what we did and what caused this bug in Edge. We assumed that the presence of `response.body` meant we could make a new `Response` from it, and that's what was failing in Edge.

Given that all browsers other than Edge currently support all of these stream-related features, and given that we already have a bug (this one) from making assumptions about partial stream support, I think it makes sense for us to have a robust `areStreamsFullySupported()` function that does a full check, and then we only use stream-related features if they all pass.

> We could end up regressing if we impose checks that are too stringent.

I can see how it might be hypothetically possible for a browser to support a subset of these features in a way that doing what I describe might not be the 100% most efficient way for some partial-support use case, but from looking at our code I don't think that is the case for us (the uses in the PR are the only streams-related cases AFAIK).

How about doing what I suggest above and then changing to a more granular check if we discover a browser with partial support that would benefit from it?

jeffposnick:
This was mentioned in-person, but capturing here for the record: the concept of a generic `browserSupportsStreams()` method is problematic just from a naming perspective, since there are (as far as I know) no browser that fully support everything detailed in https://streams.spec.whatwg.org

What that hypothetical `browserSupportsStreams()` method would test for is really just a subset of functionality that encompasses all the ways Workbox is currently using `ReadableStream`s.

My argument is that instead of putting all of our detection methods in a single function whose naming would be non-intuitive and would need to evolve over time, we instead do feature detection locally at each point in the codebase where we actually use a given feature, and test *only* for the feature(s) in the Streams API that we actually care about for that particular usage.

philipwalton:
Closing in favor of #2193.

PR #2131, opened 2019-07-19
Adds importScriptsViaChunks option to webpack plugin

### LINKED ISSUE #1752: workbox-webpack-plugin: accept chunk names in importScripts
**Library Affected**:
workbox-webpack-plugin

**Browser & Platform**:
All browsers.

Background: I want to use the generated service worker from `WorkboxWebpackPlugin.GenerateSW` to handle precache of files but I want to handle `push` events from another script.
Right now imported scripts in service worker are not updating if changed, like the service worker itself does.

It would be very helpful if I could put the other script in a chunk and specify that chunk name in the `importScripts` option (right now you can only specify static file names.)
This way the service worker itself would update on any change in the push handling script (because of fingerprint) and I would avoid above issue.

(btw `sw-precache-webpack-plugin` does provide this capability)

jeffposnick:
R: @philipwalton

Fixes #1752

This new option in `workbox-webpack-plugin`'s `GenerateSW` mode allows you to use the name of chunks from your `webpack` compilation as the source for `importScripts()`. It also takes steps to exclude whatever's imported via `importScripts()` from the precache manifest, since the service worker maintains its own cache of imported scripts.

It's important that developers use this with versioned filenames (i.e. including a hash), since the current service worker implementations in browsers won't check for updates to imported scripts unless their URL changes.

In the course of writing tests for this, I also fixed some of the validation logic around calls to `importScripts`, making it so that we can check to make sure that the Workbox runtime is imported without having to hardcode the exact filename. This should lead to less test case churn resulting from small changes in the Workbox runtime code leading to new hashed filenames.

### DIFF
--- a/infra/testing/validator/service-worker-runtime.js
+++ b/infra/testing/validator/service-worker-runtime.js
@@ -106,6 +106,7 @@ function setupSpiesAndContextForGenerateSW() {
   const importScripts = sinon.spy();
 
   const workboxContext = {
+    importScripts,
     CacheFirst: sinon.stub().returns({name: 'CacheFirst'}),
     clientsClaim: sinon.spy(),
     enable: sinon.spy(),
@@ -137,6 +138,16 @@ function setupSpiesAndContextForGenerateSW() {
 function validateMethodCalls({methodsToSpies, expectedMethodCalls}) {
   for (const [method, spy] of Object.entries(methodsToSpies)) {
     if (spy.called) {
+      // Special-case handling for importScripts(), as the first call may be
+      // to load in the workbox runtime. We don't want to have to hardcode the
+      // unique hash associated with the runtime into our test suite.
+      if (method === 'importScripts') {
+        if (spy.args[0] && spy.args[0][0].startsWith('./workbox-')) {
+          // Remove the first call.
+          spy.args.shift();
+        }
+      }
+
       const args = spy.args.map(
           (arg) => Array.isArray(arg) ? stringifyFunctionsInArray(arg) : arg);
       expect(args).to.deep.equal(expectedMethodCalls[method],
--- a/packages/workbox-build/src/entry-points/options/webpack-generate-sw-schema.js
+++ b/packages/workbox-build/src/entry-points/options/webpack-generate-sw-schema.js
@@ -13,5 +13,6 @@ const defaults = require('./defaults');
 const webpackCommon = require('./webpack-common');
 
 module.exports = commonGenerateSchema.keys(Object.assign({
+  importScriptsViaChunks: joi.array().items(joi.string()),
   swDest: joi.string().default(defaults.swDestFilename),
 }, webpackCommon));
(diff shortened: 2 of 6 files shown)


frlinw:
The current order `importScripts` > `importScriptsViaChunk` fits my use case but a single `importScripts` property with chunk support could be more generic.
```js
{
  importScripts: ['./prechunk-script.js', 'chunk', './postchunk-script.js']
}
```

jeffposnick:
I am not a fan of the overloaded nature of passing in `chunk` names mixed in with script names, and looking at how [`sw-precache-webpack-plugin` implemented it](https://github.com/goldhand/sw-precache-webpack-plugin#configuration), it's also overly complex:

> - importScripts: [Array<String|Object>]
>   - When importScripts array item is a String:
>     -  Converts to object format { filename: '<publicPath>/my-script.js'}
>   - When importScripts array item is an Object:
>     - Looks for chunkName property.
>     - Looks for filename property.
>     - If a chunkName is specified, it will override the accompanied value for filename.
> 

I guess I'm hoping that folks who *really* need a specific order will be able to use either `importScripts` or `importScriptsViaChunks`, rather than both? We'll see how well that works out based on alpha feedback.

frlinw:
One last thing, I used InjectManifest until now but I see `importScriptsViaChunk` has been added to `GenerateSW` only so I tried it in v4:
image
The boilerplate (yellow) generate by `GenerateSW` is already part of my custom imported chunk, so is it possible to disable this boilerplate in v5? If not, could you add `importScriptsViaChunk` to `InjectManifest`?

jeffposnick:
If you're using `InjectManifest` in v5, then your `swSrc` should go through a full compilation. Can you try explicitly importing (using a syntax that `webpack` natively supports) the code from the desired chunk inside of your `swSrc` file, and allow `webpack` to inline that chunk's code, rather than relying on `importScriptsViaChunks`?

frlinw:
@jeffposnick I guess I can if `swSrc` can be a lodash template and if there is a way to access the list of chunks, like in `html-webpack-plugin` but I didn't find any details about this in the doc, could you help?

frlinw:
tested in v5.

webpack config
```js
new WorkboxPlugin.InjectManifest({
  swSrc: `${rootPath}/src/sw.js`,
  swDest: 'sw.js',
  include: [
    /\.html$/,
    /\.js$/,
    /\.css$/,
    /\.woff2$/,
    /\.jpg$/,
    /\.png$/
  ]
})
```
sw.js
```js
importScripts('/env.js')

import './sw/main.js' // entry point
```
In dev & prod mode, `importScripts('/env.js')` is put after the generated code so script evaluation fails (it contains global vars used in the main script)
image
If you have an idea...

jeffposnick:
(The comments of a closed PR isn't a great place to discuss this—can you open a new issue if you're still running into problems?)

My suggestion was that _instead_ of using `importScripts('abc.js')` in your `swSrc` file, you can use `import ./abc.js` (or some other syntax that webpack supports for pulling module code).

In v5, `swSrc` should end up being compiled by webpack just like any other source file, so the `import ./abc.js` should also end up being compiled. You may need to confirm a specific set of loaders to get exactly the behavior you want during that compilation, but that's a problem that always exists for webpack.

frlinw:
@jeffposnick Thanks for the reply, I added `Edit` note in my previous comment with the solution. Everything is good 👍

PR #2134, opened 2019-07-19
Pass more matchCallback return values through to handlerCallback

### LINKED ISSUE #2079: Primitive return value from matchCb not passed as 'params' into handlerCb
**Library Affected**:
workbox-sw

**Browser & Platform**:
all browsers

The documentation of [workbox-routing](https://developers.google.com/web/tools/workbox/modules/workbox-routing) indicates 

> The params value is the value returned by the "match" function.

This does not work if the match function returns non-Object and non-Array values.

E.g.:

```
    ({event}) => event.request.headers.get('x-cache-name'),
    ({event, params}) => {
        const strategy = new workbox.strategies.NetworkFirst({cacheName: CACHE_PREFIX + params});
        return strategy.makeRequest({request: event.request});
    }
);
```

Passes `undefined` as params into the handler function even though x-cache-name is present on the request - since it is a string.

If you instead return `{cacheName: event.request.headers.get('x-cache-name')}` from matchCb, it is passed to handlerCb as expected.

jeffposnick:
R: @philipwalton

Fixes #2079

This change results in more return values from the `matchCallback` being passed through to the `handlerCallback` as the `params` option.

(There is still some special-case logic for `{}` and `[]` being treated as `{params: undefined}`, matching the existing v4 behavior. It's implemented this way to make it easier to deal with a `matchCallback` that uses a `RegExp` with capture groups.)

### DIFF
--- a/packages/workbox-routing/src/Router.ts
+++ b/packages/workbox-routing/src/Router.ts
@@ -293,13 +293,15 @@ class Router {
       let params;
       let matchResult = route.match({url, request, event});
       if (matchResult) {
-        if (Array.isArray(matchResult) && matchResult.length > 0) {
+        // See https://github.com/GoogleChrome/workbox/issues/2079
+        params = matchResult;
+        if (Array.isArray(matchResult) && matchResult.length === 0) {
           // Instead of passing an empty array in as params, use undefined.
-          params = matchResult;
+          params = undefined;
         } else if ((matchResult.constructor === Object &&
-            Object.keys(matchResult).length > 0)) {
+            Object.keys(matchResult).length === 0)) {
           // Instead of passing an empty object in as params, use undefined.
-          params = matchResult;
+          params = undefined;
         }
 
         // Return early if have a match.
--- a/packages/workbox-routing/src/Router.ts
+++ b/packages/workbox-routing/src/Router.ts
@@ -303,7 +303,7 @@ class Router {
           // Instead of passing an empty object in as params, use undefined.
           params = undefined;
         } else if (typeof matchResult === 'boolean') {
-          // For thee boolean value true (rather than just something truth-y),
+          // For the boolean value true (rather than just something truth-y),
           // don't set params.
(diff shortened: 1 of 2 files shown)


philipwalton:
I'm mostly OK with this change, but (and I remember thinking about this when converting the code to TypeScript) I think the intention of the match callback is to be boolean-ish. That is, I think in most cases the function answers the question: _Does this route match? Yes or no._

With this change a return value of `true` will get passed as the params object, which feels like it goes against the intention of the params property (though it probably won't have any actual consequences).

I do see a use base for returning a string and having that passed to `params` though.

jeffposnick:
> With this change a return value of true will get passed as the params object, which feels like it goes against the intention of the params property (though it probably won't have any actual consequences).

Would you rather I change it so that `true` doesn't get passed through, or are you just cool with this PR as-is?

philipwalton:
I think so. Let's make is so that:
- Any falsy value means no match.
- A strict `true` value means a match but no params.
- Any non-`true`, truthy value means params.

I'm also fine with dropping the support for removing empty array/objects, but maybe there's good reason to keep that?

> It's implemented this way to make it easier to deal with a matchCallback that uses a RegExp with capture groups.

I don't really understand why this is the case. Wouldn't `RegExp.exec()` and friends return falsy when there's no match? In what case is there a match but empty capture groups that we want to ignore?

jeffposnick:
Okay, please take another look with the latest changes.

packages/workbox-routing/src/Router.ts
```diff
@@ -302,6 +302,11 @@ class Router {
             Object.keys(matchResult).length === 0)) {
           // Instead of passing an empty object in as params, use undefined.
           params = undefined;
+        } else if (typeof matchResult === 'boolean') {
+          // For thee boolean value true (rather than just something truth-y),
```
philipwalton: s/thee/the

jeffposnick: Stupid keyboard...

PR #2135, opened 2019-07-19
Fixing typos

gauntface:
R: @jeffposnick @philipwalton 👋 

This fixes a few typos in comments. Nothing too crazy.


packages/workbox-google-analytics/src/initialize.ts
```diff
@@ -68,7 +68,7 @@ const createOnSyncCallback = (config: GoogleAnalyticsInitializeOptions) => {
         // Set the qt param prior to applying hitFilter or parameterOverrides.
         params.set('qt', String(queueTime));
 
-        // Apply `paramterOverrideds`, if set.
+        // Apply `parameterOverrideds`, if set.
```
philipwalton: Oh man, looks like I had two typos in this one. Can you remove the last "d"?

jeffposnick: I took the liberty of fixing this so that we could close out the PR.

PR #2136, opened 2019-07-20
feat: workbox-window method to trigger SW update check 

### LINKED ISSUE #2130: workbox-window method to trigger SW update check
**Library Affected**:
workbox-sw, workbox-window

**Browser & Platform**:
all browsers

If I try to instantiate workbox a second time no events are fired. What I basically want to achieve is that the user presses a button,  the service worker should reload and thus he will check for updates. It seems that currently the service worker can only be instantiated once. Appending a random timestamp in the file name causes the service worker to always download everything again.

There should be a method that just checks for updates of the service worker itself.

```
  const workbox = new Workbox(`/service-worker.js`, {
    scope: '/',
  })
  await workbox.register()
```

jaulz:
R: @jeffposnick @philipwalton

Fixes #2130

This PR introduces an `update` method on the `Workbox` instance in order to check for updates after the page has loaded. This is useful for long running applications that are seldom closed (e.g. mail applications) and where you need to check for updates on regular time interval.

### DIFF
--- a/packages/workbox-window/src/Workbox.ts
+++ b/packages/workbox-window/src/Workbox.ts
@@ -196,6 +196,27 @@ class Workbox extends WorkboxEventTarget {
   }
 
   /**
+   * Checks for updates of the registered service worker.
+   */
+  async update() {
+      if (!this._registration) {
+        if (process.env.NODE_ENV !== 'production') {
+        logger.error('Cannot update a Workbox instance without ' +
+            'being registered. Register the Workbox instance first.');
+          }
+        return;
+      }
+
+    // Reset the registration time and update count so it's not treated as 
+    // external in the `this._onUpdateFound` heuristic.
+    this._registrationTime = performance.now();
+    this._updateFoundCount = 0;
+
+    // Try to update registration
+    await this._registration.update();
+  }
+
+  /**
    * Resolves to the service worker registered by this instance as soon as it
    * is active. If a service worker was already controlling at registration
    * time then it will resolve to that if the script URLs (and optionally
--- a/packages/workbox-window/src/Workbox.ts
+++ b/packages/workbox-window/src/Workbox.ts
@@ -199,18 +199,13 @@ class Workbox extends WorkboxEventTarget {
    * Checks for updates of the registered service worker.
    */
   async update() {
-      if (!this._registration) {
-        if (process.env.NODE_ENV !== 'production') {
+    if (!this._registration) {
+      if (process.env.NODE_ENV !== 'production') {
         logger.error('Cannot update a Workbox instance without ' +
-            'being registered. Register the Workbox instance first.');
-          }
-        return;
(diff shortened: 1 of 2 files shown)


jaulz:
I signed it!

packages/workbox-window/src/Workbox.ts
```diff
@@ -195,6 +195,27 @@ class Workbox extends WorkboxEventTarget {
...
+      if (!this._registration) {
+        if (process.env.NODE_ENV !== 'production') {
+        logger.error('Cannot update a Workbox instance without ' +
+            'being registered. Register the Workbox instance first.');
+          }
+        return;
+      }
+
+    // Reset the registration time and update count so it's not treated as 
+    // external in the `this._onUpdateFound` heuristic.
```
philipwalton: I'm not sure resetting these here is what we want to do. Based on how I describe [the intention of the "external" events in the workbox-window guide](https://developers.google.com/web/tools/workbox/modules/workbox-window#when_an_unexpected_version_of_the_service_worker_is_found), an update found via calling `update()` is (for all intents and purposes) the same as an update found via the app being updated in a new tab.

Do you disagree? Based on what you said in this comment I thought you wanted to treat them the same way: https://github.com/GoogleChrome/workbox/issues/2130#issuecomment-513350467

jaulz: I changed it now but I don't understand why an update via `update()` is the same as an update via a new tab? Isn't the update eventually the same as a `register()` which is called after the page has been loaded?

philipwalton: > Isn't the update eventually the same as a `register()` which is called after the page has been loaded?

Not quite. The difference is `register()` is always called first, and after the `register()` call you either have an update or you don't. But when you have an update from the `register()` call (in the same tab), it's always for a SW that is older (or the same age) as the code running on the window.

But if an update is found when calling `update()` (or `.register()` in another tab in the future), it means the updated SW is newer than the code running on the page, which means there could potentially be conflicts, depending on how old the code running on the page is.

philipwalton:
Thanks, this looks good now. Do you want to try adding a test to https://github.com/GoogleChrome/workbox/blob/ba31cb87236ac4d3b1cd9c1b6580d518a6d47b73/test/workbox-window/window/test-Workbox.mjs?

You can debug the window tests by running `gulp test-server` in your terminal, and then visiting http://localhost:3004/test/workbox-window/window/ in your browser to see the results.

If you're not sure how to do that (or have problems getting the tests to run), I can merge your changes into another branch and add the tests myself when I get a chance.

jaulz:
Okay, I added some tests... though, I couldn't find out how to actually replicate an update (i.e. a file change). Any idea or do you think these tests are enough?

philipwalton:
Thanks! (Sorry for the relayed reply)

PR #2140, opened 2019-07-23
Honor subdirectory prefixes as part of swDest

### LINKED ISSUE #2138: swDest with relative folder prefix not supported.
**Library Affected**:
workbox-webpack-plugin

**Browser & Platform**:
OSX

As part of upgrading next-offline to v5 Alpha, we [noticed](https://github.com/hanford/next-offline/pull/158#issuecomment-513893814) that specifying a folder prefixed `swDest` value such as `/static/service-worker.js` is ignored and the SW file is output at the default `service-worker.js` destination.

It appears that [this line](https://github.com/GoogleChrome/workbox/blob/master/packages/workbox-build/src/lib/bundle.js#L30) in workbox-build/src/lib/bundle.js causes this due to its use of `path.basename` which ignores the provided folder prefix. For example:
```
path.basename('/foo/bar/baz/asdf/quux.html');
// Returns: 'quux.html'

path.basename('/foo/bar/baz/asdf/quux.html', '.html');
// Returns: 'quux'
```

Reproduction Steps:
See https://github.com/hanford/next-offline/pull/158#issuecomment-513647451 and use the code from that Pull Request.

jeffposnick:
R: @philipwalton
CC: @derekdowling

Fixes #2138

In the course of fixing the bug that prevented relative `swDest` path prefixes from working, I took the opportunity to normalize all of our `path` usage in the `node` tools to `upath`. That saved some effort in adding in some Windows-specific logic to this PR, and making the change universally seemed like the safest thing to do in the long run, as we've already made some piecemeal changes.


test/workbox-webpack-plugin/node/generate-sw.js
```diff
@@ -1091,14 +1091,48 @@ describe(`[workbox-webpack-plugin] GenerateSW (End to End)`, function() {
         }
       });
     });
+
+    it(`should support using a swDest that includes a subdirectory`, function(done) {
```
jeffposnick: FYI—GitHub auto-hides this diff by default, but here's the new test case.

PR #2141, opened 2019-07-24
Support for supplying subresource integrity for precaching requests

### LINKED ISSUE #2073: Subresource Integrity via Request.integrity for Pre-Caching
**Library Affected**:
*workbox-precaching*

**Browser & Platform**:
*All Browsers*

We have ended rolling our own pre-caching implementation, but are contemplating moving back into the Workbox mechanism, now that v4 has fixed up a few bugs.

One feature we have added that I don't see supported in Workbox Precaching is Subresource Integrity enforcement.

It wouldn't be too hard to generate the SRI hashes since you already are doing some hashing for Workbox Build anyways.

I think this would be a useful feature for Workbox Precaching, as it would prevent corrupt or compromised CDN Assets from being served and stored offline for PWAs to run on subsequent activation.

The implementation we rolled utilized: https://developer.mozilla.org/en-US/docs/Web/API/Request/integrity

On an unrelated note, it would be awesome to get SRI support into `importScripts()` too, but I think that's out of the scope of this package.

jeffposnick:
R: @jeffposnick 
CC: @josephliccini 

Fixes #2073 

This adds support for a new, optional field in `PrecacheEntry` called `integrity`. If provided, it should be a value that conforms to the [subresource integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity#Using_Subresource_Integrity) format, and it will be passed as-is as part of the the `RequestInit` when creating the `Request` used to populate the precache.

Actually generating the `integrity` fields is left as an exercise to the developer (at least for the time being—we *might* want to add in the ability to have the build tools generate them at some point in a future release).

The most likely way that developers would add them in is via a custom `manifestTransform` function that can operate on each entry and optionally add `integrity` properties to those that need them, like so:

```
const integrityManifestTransform = (originalManifest) => {
  const warnings = [];
  const manifest = originalManifest.map((entry) => {
    // If some criteria match:
    if (entry.url.startsWith('...')) {
      // This has to be a synchronous function call:
      entry.integrity = generateSRIInfo(entry.url);

      // Push a message to warnings if needed.
    }
    return entry;
  });

  return {warnings, manifest};
};
```


packages/workbox-precaching/src/PrecacheController.ts
```diff
@@ -62,13 +63,23 @@ class PrecacheController {
...
+
       if (this._urlsToCacheKeys.has(url) &&
           this._urlsToCacheKeys.get(url) !== cacheKey) {
         throw new WorkboxError('add-to-cache-list-conflicting-entries', {
           firstEntry: this._urlsToCacheKeys.get(url),
           secondEntry: cacheKey,
         });
       }
+
+      if (typeof entry !== 'string' && entry.integrity) {
```
jeffposnick: I'm not 100% sure that this is the most TypeScript-y way of performing this conditional logic.

packages/workbox-precaching/src/PrecacheController.ts
```diff
@@ -15,12 +15,11 @@ import {WorkboxPlugin} from 'workbox-core/types.js';
 
 import {cleanRedirect} from './utils/cleanRedirect.js';
 import {createCacheKey} from './utils/createCacheKey.js';
+import {PrecacheEntry} from './_types.js';
```
philipwalton: Super nit, but BTW my preference has been to order imports alphabetically by their filename rather than their import names. My reasoning: when you import multiple symbols (or import X as Y), it's no longer possible to maintain an alphabetical ordering by import name.

packages/workbox-precaching/src/PrecacheController.ts
philipwalton: I think this is fine. Does TypeScript complain if you just check `entry.integrity`?

packages/workbox-precaching/src/PrecacheController.ts
```diff
@@ -62,13 +63,23 @@ class PrecacheController {
...
           secondEntry: cacheKey,
         });
       }
+
+      if (typeof entry !== 'string' && entry.integrity) {
+        this._urlsToIntegrity.set(cacheKey, entry.integrity);
+      } else {
+        // If we don't have an integrity value, then clear things out, just in
+        // case addToCacheList() was called multiple times: last call wins.
+        this._urlsToIntegrity.delete(cacheKey);
```
philipwalton: We throw in the case of duplicate cache keys, why wouldn't we just do the same in the case of duplicate URLs with different integrity values? I don't see the value in support last call wins here.

test/workbox-precaching/sw/test-PrecacheController.mjs
```diff
@@ -486,6 +486,56 @@ describe(`PrecacheController`, function() {
...
+      expect(fetchSpy.secondCall.args[0].request.integrity).to.eql('sha256-second');
+    });
+
+    it(`it should use the last-specified integrity value when making requests`, async function() {
+      const fetchSpy = sandbox.spy(fetchWrapper, 'fetch');
+
+      const precacheController = new PrecacheController();
+      const cacheList = [
+        {url: '/test', integrity: 'sha256-one'},
+        {url: '/test', integrity: 'sha256-two'},
```
philipwalton: As mentioned above, I would expect this to be an error. What's the rationale for supporting this use case?

packages/workbox-precaching/src/PrecacheController.ts
```diff
@@ -40,6 +40,7 @@ class PrecacheController {
   constructor(cacheName?: string) {
     this._cacheName = cacheNames.getPrecacheName(cacheName);
     this._urlsToCacheKeys = new Map();
+    this._urlsToIntegrity = new Map();
```
philipwalton: Why not just have a mapping between URLs and an object with `{cacheKey, integrity}` properties?

packages/workbox-precaching/src/PrecacheController.ts
jeffposnick: Yeah, without the first condition I end up with:

```
Property 'integrity' does not exist on type 'string | PrecacheEntry'.
  Property 'integrity' does not exist on type 'string'.ts(2339)
```

packages/workbox-precaching/src/PrecacheController.ts
jeffposnick: I'm fine with that, but I'd prefer to see the `import './_version.js';` stand on its own, and not mix in "meaningful" imports with it.

So the `'./_types.js';` import would show up first in the list of "local" imports, separate from the versioning import.

I just installed https://marketplace.visualstudio.com/items?itemName=Snowcoders.sci-sort-import-groups to help automate this in the future.

packages/workbox-precaching/src/PrecacheController.ts
jeffposnick: Yeah, I got the naming wrong. This is effectively a `cacheKey` => `integrity` mapping, not a `url` => `integrity` mapping. (At least according to the distinction between `url` and `cacheKey` that's used elsewhere.)

I'll rename, and hopefully that explains why we can't just use a single `Map` with tuple values.

packages/workbox-precaching/src/PrecacheController.ts
jeffposnick: Okay, fair enough. I just went with that as my first instinct, but I'm happy to change.

jeffposnick: (I realize that one drawback is that this is going to add probably a couple of hundred extra bytes to the `PrecacheController` size for the error message and detailed check, all for a feature that is not going to be used by 99% of developers. Not sure if that should factor in to how we handle it.)

test/workbox-precaching/sw/test-PrecacheController.mjs
jeffposnick: I've changed the test to expect this to fail.

packages/workbox-precaching/src/PrecacheController.ts
philipwalton: Yeah, that's fine. I think it was out of order before as well anyway...

packages/workbox-precaching/src/PrecacheController.ts
philipwalton: Yeah, I think that's OK. In this case I don't think it's enough bytes to really worry about.

PR #2142, opened 2019-07-25
WorkboxEvent extends WorkboxEventProps

azizhk:
Just a small change for this Typescript error:
Screenshot 2019-07-25 at 8 05 38 PM

### DIFF
--- a/packages/workbox-window/src/utils/WorkboxEvent.ts
+++ b/packages/workbox-window/src/utils/WorkboxEvent.ts
@@ -19,6 +19,8 @@ export interface WorkboxEventProps {
   wasWaitingBeforeRegister?: boolean;
 }
 
+export interface WorkboxEvent extends WorkboxEventProps {}
+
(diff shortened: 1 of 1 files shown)


philipwalton:
Thanks, I merged this into a new branch so I could add some additional commits to it before moving to master. I decided to break apart the type definition for the `WorkboxEvent` types since many of them have unrelated properties (e.g. the `message` and `installed` events).

azizhk:
Oh cool, no problem.
Just added this so that it correctly represents the Object.assign statement:
https://github.com/GoogleChrome/workbox/blob/fec46efbc3ed02eaf0e10d3f65ef11662e9e0a4e/packages/workbox-window/src/utils/WorkboxEvent.ts#L31-L33

 I don't know why Typescript did not mention anything that the keys of WorkboxEventProps don't exist in WorkboxEvent at the Object.assign statement.

PR #2143, opened 2019-07-25
Remove registerNavigationRoute, add createHandlerForURL

### LINKED ISSUE #2095: Deprecate workbox-routing's registerNavigationRoute()
`workbox.routing.registerNavigationRoute()` is a recurring source of confusion for developers (c.f. https://github.com/GoogleChrome/workbox/issues/1988 recently, as well as this [archive of past issues](https://github.com/GoogleChrome/workbox/issues?utf8=✓&q=is%3Aissue+registerNavigationRoute) in general).

`workbox.routing.registerNavigationRoute()` appears to be a general-purpose method of registering a route that matches navigation requests, when in fact, it's implemented as a way of bridging the gap between `workbox-routing` and `workbox-precaching`, making it easier to handle navigations using precached URLs.

Not everyone _wants_ to handle navigations with precached URLs, though.

I think in v5, we might be able to get away with moving most of the logic in `registerNavigationRoute()` from `workbox-routing` to `workbox-precaching`, and have it just return logic that could be used as a `handler` callback. This `handler` callback could then be used when constructing a route.

If we moved this logic to `workbox-precaching`  we could also avoid having to explicitly call `workbox.precaching.getCacheKeyForURL()`, which has been _another_ source of confusion in Workbox v4.

```
// v4:
workbox.routing.registerNavigationRoute(
  workbox.precaching.getCacheKeyForURL('/app-shell.html')
);

// v5:
const handler = workbox.precaching.createHandlerForURL('/app-shell.html');
const navigationRoute = new workbox.routing.NavigationRoute(handler);
workbox.routing.regsiterRoute(navigationRoute);
```

CC: @philipwalton who was looking into the feasibility of using `workbox-strategies` under the hood for the handler logic in `workbox-precaching`, as this proposed change might impact that plan.

jeffposnick:
R: @philipwalton

Fixes #2095

This includes the removal of `registerNavigationRoute()`, the addition of `createHandlerForURL()`, and changes to the `generateSW` template to preserve the equivalent behavior when `navigateFallback` is set.


packages/workbox-precaching/src/PrecacheController.ts
```diff
@@ -260,6 +262,60 @@ class PrecacheController {
...
+        // an exception, then attempt to fulfill the navigation request with
+        // a response from the network rather than leaving the user with a
+        // failed navigation.
+        if (process.env.NODE_ENV !== 'production') {
+          logger.debug(`Unable to respond to navigation request with ` +
+              `cached response. Falling back to network.`, error);
+        }
+  
+        // This might still fail if the browser is offline...
+        return fetch(cacheKey);
```
philipwalton: Do you want to fetch the `cacheKey` or the `url` in the fallback case?

packages/workbox-routing/src/index.ts
```diff
@@ -30,7 +29,6 @@ if (process.env.NODE_ENV !== 'production') {
 export {
   NavigationRoute,
   RegExpRoute,
-  registerNavigationRoute,
```
philipwalton: 🎉

test/workbox-precaching/sw/test-createHandlerForURL.mjs
```diff
@@ -0,0 +1,77 @@
...
+    expect(fetchStub.notCalled).to.be.true;
+    expect(await response2.text()).to.eql('response 2');
+
+    const handler3 = createHandlerForURL('/url3');
+    const response3 = await handler3();
+
+    expect(matchStub.calledThrice).to.be.true;
+    expect(matchStub.thirdCall.args).to.eql([`${location.origin}/url3`]);
+    expect(fetchStub.calledOnce).to.be.true;
+    expect(fetchStub.firstCall.args).to.eql([`${location.origin}/url3`]);
```
philipwalton: Similar to my question above, do you also want to test the case with the fetch fallback for the cache key (assuming that's correct)?

packages/workbox-precaching/src/PrecacheController.ts
jeffposnick: `cacheKey` makes the most sense, I think. Whenever `workbox-precaching` makes a network request, it's for the cache key (which includes the `__WB_REVISION` query parameter, as needed). Using the `cacheKey` here is consistent.

test/workbox-precaching/sw/test-createHandlerForURL.mjs
jeffposnick: I can add that, sure.

jeffposnick:
PTAL.

packages/workbox-precaching/src/PrecacheController.ts
philipwalton: Interesting. Doesn't that potentially bypass the HTTP cache in a way that may be unnecessary (or is that the point)?

jeffposnick: It's more likely to _hit_ the HTTP cache, since `/something.html?__WB_REVISION=1234` will likely have been requested by the service worker previously (in order to populate the precache), while `/something.html` hasn't necessarily been requested.

Getting to this point in the code is, by definition, a failure scenario, so I'm not sure what the expectation is for a developer. We're just making a best-case effort to (temporarily) return something when this happens.

philipwalton: I see, my question was based on the assumption that this was different from the initial precache fetch, but it sounds like my assumption was incorrect. As long as the same URL is used in both cases, caching shouldn't be an issue.

I will say that I'm surprised the cache key is used in the initial precaching fetch (hadn't noticed that before because the `_addURLToCache()` method always references the `url` variable, which is not the same as the `url` variable used in `addToCacheList()`).

Anyway, if this is what's been happening for a while and no one has complained, I'm sure it's fine. My concern would have been that adding a query param to the URL that the developer may not be expecting might have unintended side effects -- depending on how their server is set up.

jeffposnick: Right, I think there's a decent amount of refactoring and renaming that could be done to the `workbox-precaching` codebase that would clarify some of those points.

You could visit a site that uses `workbox-precaching` in an Incognito window to confirm that `__WB_REVISION=...` is included in the outgoing request URLs.

PR #2151, opened 2019-07-29
Upgrade test suite to use selenium-webdriver v4

jeffposnick:
This PR updates our dependency on `selenium-assistant` to v6.0.0, which supports a version of the underlying `selenium-webdriver` project that will work with Safari.

It also changes our Travis CI config to use an OS X container instead of Linux, which will result in our Safari tests being run as part of our CI. (This unfortunately takes a long-ish time to complete.)

To work around some issues related to `chromedriver` (and `geckodriver`) versioning, it removes them from `package.json` and instead installs them fresh before every `gulp test-integration`.

(There's a somewhat mysterious increase in size in the `packages/workbox-window/build/workbox-window.dev.umd.js` bundle reported by the `PR-Bot Size Plugin`, and I'm not sure what's up with that? It might be because the various Babel-related dependencies that were locked to earlier versions in our old `package-lock.json` have been updated?)


.travis.yml
```diff
@@ -1,25 +1,29 @@
...
+os:
+  - osx
+
+# See https://docs.travis-ci.com/user/reference/osx/#macos-version
+# This implies macOS 10.14
+osx_image: xcode11
+
 cache:
   directories:
-    - "$HOME/.npm"
```
philipwalton: What's the story here?

jeffposnick: As per https://docs.travis-ci.com/user/caching/#npm-cache, it should happen automatically now.

jeffposnick:
(Here's the before-and-after of `workbox-window.dev.umd.js`: https://gist.github.com/jeffposnick/1168ba18561b94b82233d64f5d5521ac/revisions?diff=split. They're too big for the built-in `gist` diff viewer, so you need to download them.)

jeffposnick:
Based on eyeballing the diff locally, it looks like the bulk of the changes in the `workbox-window` `dev` UMD bundle is due to the inclusion of some JSDocs that were not included in the previous bundle, related to events; e.g.:

```
   /**
     * The `externalactivated` event is dispatched if the state of an
     * [external service worker]{@link https://developers.google.com/web/tools/workbox/modules/workbox-precaching#def-external-sw}
     * changes to `activated`.
     *
     * @event module:workbox-window.Workbox#externalactivated
     * @type {WorkboxEvent}
     * @property {ServiceWorker} sw The service worker instance.
     * @property {Event} originalEvent The original [`statechange`]{@link https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorker/onstatechange}
     *     event.
     * @property {string} type `externalactivated`.
     * @property {Workbox} target The `Workbox` instance.
     */
```

Issue #2144, opened 2019-07-25
[lodash.template vulnerability] CVE-2019-10744 v4.3.1


hc-12:
Welcome! Please use this template for reporting bugs or requesting features. For questions about using Workbox, the best place to ask is Stack Overflow, tagged with `[workbox]`: https://stackoverflow.com/questions/ask?tags=workbox

**Library Affected**:
workbox-webpack-plugin -> workbox-build

```
npm ls lodash.template

└─┬ workbox-webpack-plugin@4.3.1
  └─┬ workbox-build@4.3.1
    └── lodash.template@4.4.0
```

**Browser & Platform**:
"all browsers".*

**Issue or Feature Request Description**:
see [CVE-2019-10744](https://github.com/lodash/lodash/pull/4336) for more details.

The fix currently in v5 alpha but should be back ported to current released version.

upgrade `lodash.template` to v4.5 would fix the issue.

*When reporting bugs, please include relevant JavaScript Console logs and links to public URLs at which the issue could be reproduced.*

shennan:
Any intention to backport this fix to the current stable release? Would be useful.

jeffposnick:
Apologies; https://github.com/GoogleChrome/workbox/releases/tag/v5.0.0 is out. We're not going to be backporting this to the v4.x branch.

shennan:
It was released 16 days ago; this ticket was opened July 2019.

PR #2156, opened 2019-08-01
Workbox event types

philipwalton:
R: @jeffposnick 

This PR adds some additional type information to address #2144. Primarily, it adds additional event callback types as well as a `WorkboxEventMap` interface so callbacks can declare their event callback argument types based on the event type passed to `addEventListener()`.

This means developer code doesn't have to import the `WorkboxEvent` type and annotate their own callbacks -- the types are inferred from the event type listened to. (Note: I took this pattern from the built-in [`addEventListener()`](https://github.com/microsoft/TypeScript/blob/3f75d2c8790b9e2d5d99b9d2b830a4e6a624151a/src/lib/dom.generated.d.ts#L18602) declaration, so I assume it's how these types of APIs are supposed to be typed.)

/cc @azizhk

### DIFF
--- a/packages/workbox-window/src/Workbox.ts
+++ b/packages/workbox-window/src/Workbox.ts
@@ -11,10 +11,8 @@ import {logger} from 'workbox-core/_private/logger.js';
 import {messageSW} from './messageSW.js';
 import {WorkboxEventTarget} from './utils/WorkboxEventTarget.js';
 import {urlsMatch} from './utils/urlsMatch.js';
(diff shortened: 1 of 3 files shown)


azizhk:
@googlebot I consent.

azizhk:
> I took this pattern from the built-in addEventListener() declaration, so I assume it's how these types of APIs are supposed to be typed.

Wow, I did not know about this. I applied it to props as well. https://github.com/GoogleChrome/workbox/pull/2157. Makes it a little verbose, but up to you.

PR #2157, opened 2019-08-01
WorkboxEvent props also should be strictly typed

azizhk:

R: @jeffposnick @philipwalton

Fixes #*issue number*

### DIFF
--- a/packages/workbox-window/src/utils/WorkboxEvent.ts
+++ b/packages/workbox-window/src/utils/WorkboxEvent.ts
@@ -9,6 +9,7 @@
 import {WorkboxEventTarget} from './WorkboxEventTarget.js';
 import '../_version.js';
 
+type PropExtractor<WorkboxEventInterface> = Omit<WorkboxEventInterface, "type">;
 
 /**
  * A minimal `Event` subclass shim.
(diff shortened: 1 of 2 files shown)


packages/workbox-window/src/utils/WorkboxEvent.ts
```diff
@@ -9,36 +9,34 @@
...
+export interface WorkboxMessageEvent extends WorkboxEvent<"message"> {
   originalEvent: Event;
   data: any;
 }
 
-export interface WorkboxLifecycleEvent extends WorkboxEvent {
-  type: keyof WorkboxLifecycleEventMap;
-  originalEvent: Event;
+export interface WorkboxLifecycleEvent extends WorkboxEvent<keyof WorkboxLifecycleEventMap> {
+  originalEvent?: Event;
```
azizhk: I had to add ? because we don't pass originalEvent here:
https://github.com/GoogleChrome/workbox/blob/fec46efbc3ed02eaf0e10d3f65ef11662e9e0a4e/packages/workbox-window/src/Workbox.ts#L142-L145

philipwalton:
Thanks! I was trying to figure out how to do something like this (you probably noticed I removed the "props" interfaces), and I think this works well.

PR #2164, opened 2019-08-02
importScriptsViaChunks checks for .js extension

### LINKED ISSUE #2161: workbox-webpack-plugin: importScriptsViaChunks also imports the sourcemap
Hi,

i´m using the new `importScriptsViaChunks` feature of the GenerateSW plugin in 5.0.0-alpha.2. 

```
new GenerateSW({
            swDest: join(OUTPUT_PATH, 'service-worker.js'),
            importScriptsViaChunks: ["sw"],
            inlineWorkboxRuntime: true
        }),
```

This results in the following snippet in the generated service worker. The chunk import looks fine, however the source map is also imported.
```
importScripts("js/sw.61bca7cccf48f0bfc891.js", "js/sw.61bca7cccf48f0bfc891.js.map");
```

Because of the source map is also imported, it fails during runtime with
```
service-worker.js:4144 Refused to execute script from '.../js/sw.8bea19874efa62089cab.js.map' because its MIME type ('application/octet-stream') is not executable
```

jeffposnick:
R: @philipwalton

Fixes #2161

I think this is a reasonable restriction, and allowing developers to customize things to allow for non-`.js` extensions seems like overkill.

(`importScripts()` will never play nicely with ES modules, so `.mjs` is a non-issue.)

### DIFF
--- a/packages/workbox-webpack-plugin/src/lib/get-script-files-for-chunks.js
+++ b/packages/workbox-webpack-plugin/src/lib/get-script-files-for-chunks.js
@@ -6,6 +6,8 @@
   https://opensource.org/licenses/MIT.
 */
 
+const upath = require('upath');
+
 const resolveWebpackURL = require('./resolve-webpack-url');
 
 module.exports = (compilation, chunkNames) => {
@@ -17,7 +19,10 @@ module.exports = (compilation, chunkNames) => {
     const chunk = chunks.find((chunk) => chunk.names.includes(chunkName));
     if (chunk) {
       for (const file of chunk.files) {
-        scriptFiles.add(resolveWebpackURL(publicPath, file));
+        // See https://github.com/GoogleChrome/workbox/issues/2161
+        if (upath.extname(file) === '.js') {
+          scriptFiles.add(resolveWebpackURL(publicPath, file));
+        }
       }
     } else {
       compilation.warnings.push(`${chunkName} was provided to ` +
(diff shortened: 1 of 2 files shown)


packages/workbox-webpack-plugin/src/lib/get-script-files-for-chunks.js
```diff
@@ -17,7 +19,10 @@ module.exports = (compilation, chunkNames) => {
     const chunk = chunks.find((chunk) => chunk.names.includes(chunkName));
     if (chunk) {
       for (const file of chunk.files) {
-        scriptFiles.add(resolveWebpackURL(publicPath, file));
+        // See https://github.com/GoogleChrome/workbox/issues/2161
+        if (upath.extname(file) === '.js') {
```
philipwalton: I doubt non-js extensions ever happen with webpack chunks, but I wonder if we want to be a bit more forgiving here and accept things like `.mjs`. What do you think?

jeffposnick: See my comment in the PR description 😄 

> (importScripts() will never play nicely with ES modules, so `.mjs` is a non-issue.)

philipwalton: Gah, sorry, missed that!

PR #2167, opened 2019-08-05
Fix for multiple compilations/webpack-dev-server

### LINKED ISSUE #2158: [5.0.0-alpha.2] GenerateSW Hot reload error
**Library Affected**:
"workbox-webpack-plugin": "^5.0.0-alpha.2"
"webpack": "^4.38.0",
"webpack-dev-server": "^3.7.2",

I am trying the latest alpha and everything looks to be working great however when I trigger a hot reload by making a code change then the plugin presents the following error:

ERROR in Please check your GenerateSW plugin configuration:
"manifestEntries" is not a supported parameter.

Config:
```
 new WorkboxPlugin.GenerateSW({
        swDest: 'sw.js',
        maximumFileSizeToCacheInBytes: 10485760,
        clientsClaim: true,
        skipWaiting: true,
        cleanupOutdatedCaches: true,
        exclude: [/ckeditor/],
      }),

```

jeffposnick:
R: @philipwalton

Fixes #2158

The nuance about `emit` getting called multiple times on the same plugin instance was lost in the v5 rewrite. This makes sure that we don't modify the `this.config` member variable in our `emit` callback, allowing the callback to work when called multiple times via, e.g., `webpack-dev-server`.

### DIFF
--- a/packages/workbox-webpack-plugin/src/generate-sw.js
+++ b/packages/workbox-webpack-plugin/src/generate-sw.js
@@ -76,37 +76,41 @@ class GenerateSW {
    * @private
    */
   async handleEmit(compilation) {
+    let config;
     try {
-      this.config = validate(this.config, webpackGenerateSWSchema);
+      // emit might be called multiple times; instead of modifying this.config,
+      // use a validated copy.
+      // See https://github.com/GoogleChrome/workbox/issues/2158
+      config = validate(this.config, webpackGenerateSWSchema);
     } catch (error) {
       throw new Error(`Please check your ${this.constructor.name} plugin ` +
         `configuration:\n${error.message}`);
     }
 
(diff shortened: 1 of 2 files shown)


simonsinclair:
@jeffposnick, thanks for fixing this. Is there a way I can test this change in my project today?

jeffposnick:
We should be cutting a new alpha release next week, but if you need the fix sooner, checking out the `https://github.com/GoogleChrome/workbox` repo and running `npm install && gulp build` should give you a local copy of everything.

They you can reference `workbox-webpack-plugin` from `<git-clone-root>/packages/workbox-webpack-plugin` inside of your `webpack` config.

simonsinclair:
That worked, thanks. Looking forward to the new alpha when it lands.

PR #2169, opened 2019-08-05
Reduce GA integration test flakiness

### LINKED ISSUE #2168: Failing google-analytics integration test (Chrome/Travis CI)
I've seen the [following failure](https://travis-ci.org/GoogleChrome/workbox/builds/567988253#L2679) at least a couple of times since we've moved to the OS X Travis CI environment. It happens in Chrome Stable:

```
  1) [workbox-google-analytics] initialize
2680       replay failed Google Analytics hits:
2681
2682      AssertionError: expected 'pixel' to equal 'beacon'
2683      + expected - actual
2684
2685      -pixel
2686      +beacon
2687      
2688      at Context.<anonymous> (test/workbox-google-analytics/integration/test-all.js:149:38)
2689      at <anonymous>
```

jeffposnick:
R: @philipwalton

Fixes #2168

I'm not super-familiar with this test, but I could plausibly see how sending both of the GA events back-to-back could cause the replay order to get out of whack. So I'm taking the coward's way out and adding a short timeout in between each.

### DIFF
--- a/test/workbox-google-analytics/integration/test-all.js
+++ b/test/workbox-google-analytics/integration/test-all.js
@@ -96,16 +96,17 @@ describe(`[workbox-google-analytics] initialize`, function() {
       action: 'clear-spied-requests',
     });
 
-    // Check the "simulate offline" checkbox and make some requests.
-    await simulateOfflineEl.click();
+    // Switch the service worker into "offline" mode.
+    await webdriver.executeAsyncScript(messageSW, {
+      action: 'simulate-offline',
+      value: true,
+    });
 
     await webdriver.executeAsyncScript((done) => {
       window.gtag('event', 'beacon', {
         transport_type: 'beacon',
         event_label: Date.now(),
-        event_callback: () => {
-          setTimeout(done, 50);
-        },
+        event_callback: done,
       });
     });
 
@@ -113,9 +114,7 @@ describe(`[workbox-google-analytics] initialize`, function() {
       window.gtag('event', 'pixel', {
         transport_type: 'image',
         event_label: Date.now(),
-        event_callback: () => {
-          setTimeout(done, 50);
-        },
+        event_callback: done,
       });
     });
 
@@ -131,8 +130,13 @@ describe(`[workbox-google-analytics] initialize`, function() {
     });
     expect(requests).to.have.lengthOf(0);
 
-    // Uncheck the "simulate offline" checkbox and then trigger a sync.
-    await simulateOfflineEl.click();
+    // Switch the service worker into "online" mode.
+    await webdriver.executeAsyncScript(messageSW, {
+      action: 'simulate-offline',
+      value: false,
+    });
+
+    // Dispatch a sync event.
     await webdriver.executeAsyncScript(messageSW, {
       action: 'dispatch-sync-event',
     });
@@ -160,6 +164,8 @@ describe(`[workbox-google-analytics] initialize`, function() {
     // and ensure those values reflect the original order of the hits.
     expect(requests[0].params.qt > 0).to.be.true;
     expect(requests[1].params.qt > 0).to.be.true;
-    expect(requests[0].originalTime < requests[1].originalTime).to.be.true;
+    // This should be <= instead of less than, since a ms didn't necessarily
+    // pass in between the two events.
+    expect(requests[0].originalTime <= requests[1].originalTime).to.be.true;
   });
 });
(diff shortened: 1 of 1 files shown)


test/workbox-google-analytics/integration/test-all.js
```diff
@@ -152,7 +159,7 @@ describe(`[workbox-google-analytics] initialize`, function() {
     // Ensure the hit's qt params were present and greater than 0,
     // and ensure those values reflect the original order of the hits.
     expect(requests[0].params.qt > 0).to.be.true;
-    expect(requests[0].params.qt > 0).to.be.true;
+    expect(requests[1].params.qt > 0).to.be.true;
```
jeffposnick: This was just an error in the original test logic.

philipwalton:
Hmmm, I would have thought the fact that we were `await`-ing it already would have been enough.

Looking at the code, I think it's more likely that the race condition is due to the fact that we're just clicking a button on this line and not `await`-ing the `postMessage` call: https://github.com/GoogleChrome/workbox/blob/39cd5b5a3be58deb133113392350fecb9f114bdf/test/workbox-google-analytics/integration/test-all.js#L128

If we triggered the `postMessage()` in the test (as we do elsewhere) we could await it.

test/workbox-google-analytics/integration/test-all.js
```diff
@@ -152,7 +159,9 @@ describe(`[workbox-google-analytics] initialize`, function() {
     // Ensure the hit's qt params were present and greater than 0,
     // and ensure those values reflect the original order of the hits.
     expect(requests[0].params.qt > 0).to.be.true;
-    expect(requests[0].params.qt > 0).to.be.true;
-    expect(requests[0].originalTime < requests[1].originalTime).to.be.true;
+    expect(requests[1].params.qt > 0).to.be.true;
+    // This should be <= instead of less than, since a ms didn't necessarily
+    // pass in between the two events.
+    expect(requests[0].originalTime <= requests[1].originalTime).to.be.true;
```
jeffposnick: I feel like this should be `<=` instead of `<`, because I'm really not sure how much time has elapsed in between the two GA events.

jeffposnick:
Okay, I've switched to explicitly talking to the SW instead of simulating clicks on the page's buttons. Let's see if that helps, too.

(Travis CI was happy with my original change, but it wasn't failing 100% time prior to this PR, so that might have just been luck.)

test/workbox-google-analytics/integration/test-all.js
philipwalton: Hmmm, but then I think the timestamp update logic could be removed and the test would still pass.

I'd rather sleep for 1ms and continue to assert `<` (though I'm pretty sure 1ms will also elapse since we're waiting for a postMessage to succeed -- but I get your point).

jeffposnick:
Explicitly talking to the SW instead of simulating the click didn't fix the issue (but it's probably a good idea): https://travis-ci.org/GoogleChrome/workbox/builds/568035312#L2469-L2480

I'm going to add back the timeouts and confirm that they *do* fix the issue a second time.

test/workbox-google-analytics/integration/test-all.js
jeffposnick: I've added back the timeouts, so I'm fine with using `<` now (since actual time will definitely have passed).

jeffposnick:
So there are now two successful test runs with the extra timeout logic in place, @philipwalton. It's a small sample size, but I think it did help. Can you give this PR another review?

test/workbox-google-analytics/integration/test-all.js
```diff
@@ -96,22 +92,32 @@ describe(`[workbox-google-analytics] initialize`, function() {
...
       });
     });
+
     await webdriver.executeAsyncScript((done) => {
       window.gtag('event', 'pixel', {
         transport_type: 'image',
         event_label: Date.now(),
-        event_callback: () => done(),
+        event_callback: () => {
+          setTimeout(done, 50);
```
philipwalton: Can you add a comment that links to https://github.com/GoogleChrome/workbox/issues/2168 here?

test/workbox-google-analytics/integration/test-all.js
```diff
@@ -96,22 +92,32 @@ describe(`[workbox-google-analytics] initialize`, function() {
...
+      value: true,
+    });
+
     await webdriver.executeAsyncScript((done) => {
       window.gtag('event', 'beacon', {
         transport_type: 'beacon',
         event_label: Date.now(),
-        event_callback: () => done(),
+        event_callback: () => {
+          setTimeout(done, 50);
```
philipwalton: Can you add a comment that links to https://github.com/GoogleChrome/workbox/issues/2168 here?

PR #2176, opened 2019-08-07
Set cache: 'reload' on precaching requests

### LINKED ISSUE #2146: Use cache='reload' for precaching unversioned URLs
I'm spinning this off of https://github.com/GoogleChrome/workbox/pull/2143#discussion_r307891796

In Workbox v3, we used code like https://github.com/GoogleChrome/workbox/blob/d27aafbdf164f051a883965058e6eb4c0df3a052/packages/workbox-precaching/models/PrecacheEntry.mjs#L54-L77

to cache-bust an outgoing request to populate the precache via setting the [`cache` property to `'reload'`](https://fetch.spec.whatwg.org/#concept-request-cache-mode), with some feature detection to fall back to setting a cache-bust query parameter if needed.

In v4, we stopped that (partly because we end up using that query parameter as the actual cache key, so it's somewhat simpler to also use it in the outgoing request.)

We can change that, though, and potentially just set `cache = 'reload'` again. I did a quick check and it's supported in all modern browsers; you can check with:

```
fetch('https://httpbin.org/headers', {cache: 'reload'}).then(r => r.text()).then(console.log);
```

This might be done as part of a refactoring that would clear up some of our naming in `workbox-precache` around `cacheKey` and `url`.

CC: @philipwalton

jeffposnick:
R: @philipwalton

Fixes #2146

This PR sets `cache: 'reload'` as an alternative to using the `__WB_REVISION=...` URL query parameter on the outgoing network requests used to populate the precache.

I also renamed some variables to clarify the distinction between `cacheKey` & `url` within the `install()` method.

The underlying precache keys still use the `__WB_REVISION=...` query parameters, as we continue to use those parameters as versioning metadata.

There's a slight breaking change in the `updatedURLs` & `notUpdatedURLs` values that are returned by `install()`: the previously would include `__WB_REVISION` query parameters, and now they won't. Since those return values are only used for logging/display purposes, this change is probably an improvement.

### DIFF
--- a/packages/workbox-precaching/src/PrecacheController.ts
+++ b/packages/workbox-precaching/src/PrecacheController.ts
@@ -113,8 +113,8 @@ class PrecacheController {
       }
     }
 
-    const toBePrecached = [];
-    const alreadyPrecached = [];
+    const toBePrecached: {cacheKey: string, url: string}[] = [];
+    const alreadyPrecached: string[] = [];
 
     const cache = await caches.open(this._cacheName);
     const alreadyCachedRequests = await cache.keys();
--- a/packages/workbox-precaching/src/PrecacheController.ts
+++ b/packages/workbox-precaching/src/PrecacheController.ts
@@ -113,35 +113,37 @@ class PrecacheController {
       }
     }
 
-    const urlsToPrecache = [];
-    const urlsAlreadyPrecached = [];
+    const toBePrecached = [];
+    const alreadyPrecached = [];
 
     const cache = await caches.open(this._cacheName);
     const alreadyCachedRequests = await cache.keys();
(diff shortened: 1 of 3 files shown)


packages/workbox-precaching/src/PrecacheController.ts
```diff
@@ -113,35 +113,37 @@ class PrecacheController {
       }
     }
 
-    const urlsToPrecache = [];
-    const urlsAlreadyPrecached = [];
+    const toBePrecached = [];
+    const alreadyPrecached = [];
```
philipwalton: Looks like these two arrays take different types of objects. Maybe it'd be easier to understand what's happening if you add type information here in the declarations?

jeffposnick: Done.

PR #2187, opened 2019-08-15
Rename all Plugin classes to be package-specific

### LINKED ISSUE #2072: Rename Plugin classes to include the module name
We currently have a bunch of classes named `Plugin` spread across various packages.

That's clear enough when you're referencing them via the `workbox-sw`-created namespaces, like `workbox.expiration.Plugin`, or `workbox.cacheableResponse.Plugin`.

Since we're moving to a model that involves more bundling of the Workbox runtime via ES imports, this ends up feeling a bit awkward, as you have to write

```
import {Plugin as CacheableResponsePlugin} from 'workbox-cacheable-response/Plugin.mjs';
import {Plugin as ExpirationPlugin} from 'workbox-expiration/Plugin.mjs';
```

etc.

We could go back to how things were in Workbox v2, where each of the classes included the name of the package, so that explicit aliasing isn't needed.

We could keep the interface exposed via the `workbox-sw` namespaces the same, so this would not be a huge hit in terms of backwards compatibility unless you're creating a custom bundle.

philipwalton:
R: @jeffposnick @philipwalton

Fixes #2072

@jeffposnick I opted not to include any kind of deprecation warning or fallback. Let me know if you think that's needed as I guess this might require a lot of people to change their code.

### DIFF
--- a/demos/functions/views/demo/workbox-background-sync/sw.hbs
+++ b/demos/functions/views/demo/workbox-background-sync/sw.hbs
@@ -1,6 +1,6 @@
 {{{WORKBOX_SW_IMPORT}}}
 
-const bgSyncPlugin = new workbox.backgroundSync.Plugin('myQueueName');
+const bgSyncPlugin = new workbox.backgroundSync.BackgroundSyncPlugin('myQueueName');
 
 workbox.routing.registerRoute(
   ({url}) => url.pathname === '/demo/workbox-background-sync/example.txt',
--- a/demos/functions/views/demo/workbox-range-requests/sw.hbs
+++ b/demos/functions/views/demo/workbox-range-requests/sw.hbs
@@ -1,13 +1,13 @@
 {{{WORKBOX_SW_IMPORT}}}
 
-console.log(new workbox.rangeRequests.Plugin());
+console.log(new workbox.rangeRequests.RangeRequestsPlugin());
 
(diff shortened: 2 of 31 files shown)


jeffposnick:
In terms of a potential fallback, would that entail changing each package's `index.ts` so that both `PackageNamePlugin` and `Plugin` were `export`ed, with both names referencing the same underlying class?

That... doesn't sound too bad, and I'm assuming wouldn't result in a meaningful change to bundle sizes. But I guess it might lead to confusion as to which name to use if developers see both listed in an IDE's autocomplete (assuming they're importing from the `index` and not a named module).

I'm probably in favor of helping folks who are still using `workbox-sw` not have to change things, but I could be convinced otherwise.

philipwalton:
> In terms of a potential fallback, would that entail changing each package's `index.ts` so that both `PackageNamePlugin` and `Plugin` were `export`ed, with both names referencing the same underlying class?

If we want to go deprecation route, I think it be better to make `Plugin` classes that extend the renamed classes, log a warning, and then call `super()`. That would be a bit more code, but then at least the code would continue to work and there'd be a warning.

jeffposnick:
That's more overhead than I was hoping for. We can just go with the approach in this PR (and make sure that we update our docs/samples).

philipwalton:
SG, if we hear feedback we can always add later.

PR #2191, opened 2019-08-19
Build tool options rework

### LINKED ISSUE #2078: generateSW with runtimeCaching & no precaching
There previously wasn't a great reason to use `generateSW` in the build tools without also using precaching—if you just wanted a service worker that performed runtime caching, writing it by hand (loading the Workbox runtime via `workbox-sw`) and not doing anything during the build process was fine.

In Workbox v5, the build tools will output a customized, local Workbox runtime bundle that includes only the modules being used in the configuration. I can see folks wanting to use `generateSW` to get that bundle even if they're just using `runtimeCaching`, so we can relax the requirement that folks specify patterns matching files to precache, etc.

This will require some additional parameter validation rules to make sure  that, e.g., `navigateFallback` isn't used unless precaching takes place.

jeffposnick:
R: @philipwalton

This is a general refactor of the directory structure and inheritance for our `joi`-based options validation. In the process, I loosened one of the restrictions so that developers using `workbox-build`'s `generateSW` mode could use `runtimeCaching` could be used without precaching (fixes #2078).

The PR got pretty large, given all the files moved around, so I'm going to submit a follow-up PR that deals with JSDocs for our options, which I think is ultimately a more sustainable documentation setup than what we currently use. (C.f. https://github.com/google/WebFundamentals/tree/master/src/content/en/tools/workbox/_shared/config)


packages/workbox-build/src/options/schema/generate-sw.js
```diff
@@ -0,0 +1,19 @@
+/*
+  Copyright 2018 Google LLC
```
philipwalton: s/2018/2019/

(here and throughout for new files)

Issue #2129, opened 2019-07-19
Add a utility function to copy a responses and modify its headers


philipwalton:
Forking this feature request from #2023: copying a response and modifying its headers is probably a fairly common thing to want to do, but actually implementing it in a performant and cross-browser way is non-trivial.

I think we could provide a utility function to do this in `workbox-core`. Sample usage might look something like this:

```js
const newResponse = await copyResponseAndModifyHeaders(oldResponse, (headers) => {
  // Any modifications to the `headers` object get copied to the new
  headers.append('X-Cache-Hit', '1');
})
```

In this example, the `X-Cache-Hit` header is added so analytics on the main page can know that this response came from the cache and not the network.

I don't 100% understand they history behind when a similar copy technique was used for `cleanRedirect()`, but #2023 shows that doing it right, and cross-browser is tricker than it might initially seem.

nhoizey:
I never had this need yet, but it looks useful.

What about this:

```javascript
const newResponse = await copyResponseAndModifyBody(oldResponse, (body) => {
  // Any modifications to the `headers` object get copied to the new
  body.replace('src="https://my-domain.com', 'src="https://images-cdn.com');
})
```

Just a probably dumb example, but is the idea useful?

philipwalton:
Hmmm, in my suggestion the copied response would share the body stream of the original response, so modifying it wouldn't be possible with a `replace()` function like you have there.

I'm not sure if there'd be an easy way to make body modification work in both the streaming and non-streaming case, or if it's worth doing that considering Edge is moving to Chromium in the near future, and then all SW supporting browsers should also support streams (I believe).

nhoizey:
You're right, I didn't think about streams, I'm not yet used to it.

philipwalton:
Closed via #2144

PR #2193, opened 2019-08-21
Add copyResponse and more granular stream detects

### LINKED ISSUE #2023: The cleanRedirect logic fails in Edge 18

philipwalton:
R: @jeffposnick

Fixes #2023

This PR replaces #2128 and implements the resolution from the discussion there. It adds two new, more granular feature detects for streams to `workbox-core`:

- `canConstructReadableStream()`
- `canConstructResponseFromBodyStream()`

And these features are used by other stream-related code to detect support.

Also, based on the feature request proposed in #2023, it adds a `copyResponse()` function that allows you to copy a response and modify any of the values settable via the `ResponseInit` object (`headers`, `status`, `statusText`).

The API I've landed on in this PR (totally open to suggestions though) is a function that takes a response and a `modifier()` function, here's an example (you can also see [other examples in the test file](https://github.com/GoogleChrome/workbox/blob/46f51a8fc5660c07bf919182d3bf500e26793ebb/test/workbox-core/sw/test-copyResponse.mjs#L32-L80)):

```js
const newResponse = await copyResponse(oldResponse, (init) => {
  init.headers.set('Content-Type', 'text/plain');
  init.headers.set('X-One', '1');
  init.status = 203;
  init.statusText = 'Really?';

  return init;
});
```

In addition, as currently implement, the `modifier()` function is optional, and if not passed then no modifications are made. This is equivalent to the old `cleanRedirect()` function that this PR removes. (I'm not sure if it's weird to have the `modifier()` function be optional. If you think it is, I can make it required and just have it be a no-op in the redirect case.)


philipwalton:
Also fixes #2129

PR #2198, opened 2019-08-22
Update Rollup OMT plugin

surma:
This updates `rollup-plugin-off-main-thread` to the latest version (no breaking changes) and to the new package name.

I couldn’t quite figure out how to run the build system, so I am not 100% sure that nothing is broken. PTAL :)


jeffposnick:
Thanks, @surma! Our CI tests are unhappy because there needs to be a corresponding change to `package-lock.json`, but I'll force-push that to your branch if you don't mind.

surma:
Oh how did I miss that file??! Either way, please go ahead :)

jeffposnick:
(There was also a couple of usages that needed to be updated, but I've taken care of that.)

PR #2200, opened 2019-08-23
Add redundant to WorkboxLifecycleEventMap

### LINKED ISSUE #2199: "redundant" event getting triggered.
**Library Affected**:
*workbox-sw* 5.0.0-alpha.2

**Browser & Platform**:
*Google Chrome Version 76.0.3809.100 (Official Build) (64-bit)*

So this is a two part issue.
1. redundant is not part of:
https://github.com/GoogleChrome/workbox/blob/fc38ffb6656edb7941b8116cfb1ff7dc71f289fa/packages/workbox-window/src/utils/WorkboxEvent.ts#L42-L54
So it might need to be added there.

2. Redundant is getting triggered for me in some weird scenarios.
So here is a reproduction https://github.com/azizhk/workbox-route-bug
I register my service worker only on `/a` route.

**Case A:**
1. Go to `/a` route -> SW gets registered.
2. Create a new build (`yarn build`)
3. Refresh on `/a`.
4. Click on Reload button which triggers skipWaiting:
```
registration.waiting.postMessage({
  type: "SKIP_WAITING"
});
```
Then "activated" event is triggered.

**Case B:**
1. Go to `/a` route -> SW get registered.
2. Go to `/` route.
3. Create new build (`yarn build`)
4. Refresh on `/`
5. Navigate to `/a` which registers SW and finds `registration.waiting` so shows Reload button
6. Click on reload button which triggers skipWaiting
Then "redundant" event is triggered.

Over here: https://developers.google.com/web/fundamentals/primers/service-workers/lifecycle
It mentions "redundant" is triggered when **SW is discarded. Either failed install, or it's been replaced by a newer version**

Here it hasn't erred or failed. It is replaced by a newer version but even in "activated" case the Service Worker is replaced by a newer version right? So why is "redundant" event different from "activated".

I have the "Update on reload" unchecked.

azizhk:
R: @jeffposnick @philipwalton
Fixes #2199 

Should I add externalredundant as well? I don't think that exists right?

### DIFF
--- a/packages/workbox-window/src/utils/WorkboxEvent.ts
+++ b/packages/workbox-window/src/utils/WorkboxEvent.ts
@@ -51,6 +51,7 @@ export interface WorkboxLifecycleEventMap {
   'externalwaiting': WorkboxLifecycleWaitingEvent;
   'externalactivating': WorkboxLifecycleEvent;
   'externalactivated': WorkboxLifecycleEvent;
+  'redundant': WorkboxLifecycleEvent;
 }
 
 export interface WorkboxEventMap extends WorkboxLifecycleEventMap {


philipwalton:
> Should I add externalredundant as well? I don't think that exists right?

I think it's possible that it could be emitted by a `Workbox` (given the right circumstances), but it's not an event we want to promote people listening to, so I we don't document it and I don't think we need to add typings for it.

In general, as soon as there's any `external*` events fired, that's the point at which the user should probably refresh the page, as it usually means the version of the page they're running is out of date and things may break.

azizhk:
> In general, as soon as there's any external* events fired, that's the point at which the user should probably refresh the page"
So I don't expect the user to reload, I do `window.location.reload()` explicitly.

In the example that I have shared, before triggering `skipWaiting` for a SW which has multiple clients connected, I double check with the user that I will be reloading multiple tabs and he might lose data.

Screenshot 2019-08-28 at 2 21 30 PM

So similarly, in that case, I would like to have `"externalredundant"` event getting triggered to reload the page.

But I'll come back to this requirement based on the answer to https://github.com/GoogleChrome/workbox/issues/2199#issuecomment-524839518

PR #2209, opened 2019-08-31
Listen statechange on waiting as well

### LINKED ISSUE #2199: "redundant" event getting triggered.

azizhk:
Fixes #2199 
R: @jeffposnick @philipwalton

Changes as described in https://github.com/GoogleChrome/workbox/issues/2199#issuecomment-526790836
Writing tests with my scenario.


packages/workbox-window/src/Workbox.ts
```diff
@@ -148,6 +145,8 @@ class Workbox extends WorkboxEventTarget {
 
     // If an "own" SW is already set, resolve the deferred.
     if (this._sw) {
+      this._sw.addEventListener(
+          'statechange', this._onStateChange, {once: true});
```
philipwalton: You can remove the `{once: true}` here, as that only applied in the case where the next expected event was `redundant`. And since no lifecycle events are ever going to fire after `redundant` anyway, I think it's OK to just let that be the mechanism for stopping future events.

azizhk:
Hi @philipwalton,
I don't think this solution works. I'm still missing something. I applied the [patch](https://github.com/azizhk/workbox-route-bug/blob/master/patches/workbox-window%2B5.0.0-alpha.2.patch) to my bug reproduction repo, but it did not fix there.
Now not getting any events getting triggered.

azizhk:
So I realized that our problem would also occur for `installing` service workers as well.
I wrote some tests but I am having some problems getting them working on all browsers.

philipwalton:
Hmmm, regarding triggering an `installing` event with `wasInstallingBeforeRegiser:true`, I'm not sure how much that makes sense. I agree it can happen in the case in your example in #2199, but in order for it to happen you have to delay calling `register()`, and in general I'm not sure if that's a pattern we want to promote or support with `workbox-window`.

Can you elaborate as to why you want to do this?

@jeffposnick what are your thoughts here? Are you aware of reasons (for or against) delaying the `register()` call? If we choose to support this use case then we'd have to make a lot of changes to the `Workbox` class, because at the moment it doesn't add any lifecycle listeners until `register()` is called, which means updates triggered by the browser's [soft update](https://w3c.github.io/ServiceWorker/#soft-update-algorithm) checks would be missed.

azizhk:
Well I don't necessarily need to trigger `installing`. I just added it to keep parity with `waiting`.

But I would like to add `statechange` listener on the `installingSW`.
This is because the user/developer can call register at any point. (After requestIdleCallBack, etc.)
And this can cause race conditions with browser's soft update.

I'm ok with not adding any lifecycle listeners until `register()` is called. It's just that documentation needs to mention that one can't expect all events to be triggered in sequence.
They can expect workbox to trigger "waiting" & "activated" events (as that covers majority use case)
But for "installing" & "installed" they should check registration.installing as well as listen for these events.

azizhk:
Also I've add `statechange` listeners on both `waiting` & `installing` SWs if both exist.
I think we should add them on all `this._ownSWs` that you are adding in #2211 [diff](https://github.com/GoogleChrome/workbox/pull/2211/files#diff-3c13b193f7b64e2ef506486a17173077R58)

philipwalton:
Hey @azizhk after discussing with @jeffposnick I've filed #2216 to track support for issues and API changes surrounding conditional registration.

Can we keep this PR focused on the original problem where service workers that were waiting before register do not receive any additional lifecycle events?

Then we can discuss in #2216 how best to update `workbox-window` to handle conditional/deferred registration.

azizhk:
Sure, I've reverted the code. But am not sure why tests are failing on "Safari Stable"
I'll also isolate my other failing test in a different PR.

Loboistaken (approved):

tomayac:
Hi there,

The Workbox team

PR #2210, opened 2019-09-01
Add a check to wait for the resulting client ID

### LINKED ISSUE #2203: For navigation requests BroadcastCacheUpdate messages are sent too early
**Library Affected**:
*workbox-broadcast-update*

The switch to using postMessage in #2186 was an incomplete solution to the issue in #1399. In addition to using postMessage, we also have to wait to send the message until the new window client exists.**

In #2184 I stated that [`postMessage()` buffering](https://www.chromestatus.com/features/5751307839209472) had been implemented in all browsers, but it looks like that isn't actually true. It *appeared* to be true because, in the test I was running, I could see the message being logged to the console. However, upon further inspecting, those messages were actually being received by the old window client that is *unloading*, not the new window client being navigated to.

Specifically, here's the code that gets a list of all window clients: 
https://github.com/GoogleChrome/workbox/blob/ae80e3958c79cb93f4c21bca63cd6e81f7878536/packages/workbox-broadcast-update/src/BroadcastCacheUpdate.ts#L143-L146

When this code is run, the window client being navigated to is not listed in the matches because the response has not yet been sent. However, the unloading window *is* listed in the matches because it still exists at this point, and (depending on how fast the message is sent) the unloading window can receive the message and run the message callback. This is what was happening in my original tests that made it seem like message buffering was working in all browsers (in reality the callback was being run by the unloading window).

A solution to this problem would be to wait for a window client matching the [`FetchEvent.resultingClientId`](https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent/resultingClientId) in supporting browsers, and in browsers that don't support the `resultingClientId`, we can wait until a window client exists with an ID that doesn't match any of the existing client IDs (pre-response). Once the matching window client is found, we send the message (or we stop waiting if a certain amount of time passes). In my testing this seems to work well.

### LINKED ISSUE #2204: Safari does not implement postMessage buffering
**Library Affected**:
*workbox-broadcast-update*

The tests I ran in #2184 where I concluded that all browsers support `postMessage()` buffering turned out to actually be incorrect (more details in #2203).

Safari does not support `postMessage()` buffering, despite supporting the `ServiceWorkerContainer.startMessages()` API (which means there's also no reliable way to feature detect it): https://bugs.webkit.org/show_bug.cgi?id=201169

But given that both Chrome and Firefox support message buffering, and given that the [previous solution](#1673) required `workbox-window` to be effective, I think instead of reverting to that method, we should just add a timeout in Safari to increase the probability that the message will be received.

As an alternative to UA sniffing for Safari, since the fix for #2203 also relies on feature detecting `resultingClientId`, it may make sense to use that as the branch point for using a timeout vs sending the message as soon as the resulting window client exists (the only problem there would be if Safari implements `resultingClientId` before implementing message buffering).

philipwalton:
R: @jeffposnick

Fixes #2203, fixes #2204 

To fix #2203, this PR adds a `resultingClientExists()` method that accepts a `resultingClientId` value and awaits the existence of the client. It can also resolve to `undefined` in one of two conditions:

- The passed `resultingClientId` value is `undefined` (as will happen in a browser that doesn't support it).
- The function tries for 2 seconds and still does not find a matching client.

Then, to fix #2204, in the `BroadcastCacheUpdate` class, I've added a `timeout` that waits for an additional 3.5 seconds ([explanation here](https://github.com/GoogleChrome/workbox/compare/resulting-client?expand=1#diff-c2fe184c1aabeb6e92b91e9c5fe67b9aR165-R168)) in Safari or if the `resultingClientExists()` function doesn't find a matching client. *(Note: I've added a TODO to remove the Safari check once [this bug](https://bugs.webkit.org/show_bug.cgi?id=201169) has been released.)*

I've also updated the integrations to be more comprehensive. Previously there was only a test for resource requests, so I've added:

- A test verifying a message gets sent on navigation requests *and* is actually received by the resulting page
- A test verifying the message is sent to all window clients


philipwalton:
For some reason the `workbox-window` integration tests consistently fail for this PR, despite no changes being made to that package and other PRs with the same base branch running just fine.

I think let's merge a few of the other open `workbox-window`-related PRs and then rebase this on top of them. I suspect that might magically fix things.

test/workbox-window/window/test-Workbox.mjs
```diff
@@ -301,24 +301,27 @@ describe(`[workbox-window] Workbox`, function() {
...
     });
 
-    it(`triggers an updatefound event if the SW was udpated`, async function() {
+    it(`triggers an updatefound event if the SW was updated`, async function() {
       const scriptURL = navigator.serviceWorker.controller.scriptURL;
 
       const wb = new Workbox(scriptURL);
+
       const reg = await wb.register();
+      const updatefoundPromise = new Promise((resolve) => {
```
jeffposnick: This test from yesterday needed a bit more cleaning up, as there was a race condition in which the `updatefound` listener might not have been registered in time to catch the update. Registering the listener early and then waiting later on for the promise resolution should fix that.

philipwalton:
Updated to skip the failing test in Safari. @jeffposnick can you PTAL.

Issue #2122, opened 2019-07-17
Documentation changes for v5


philipwalton:
This is a master tracking issue for documentation changes we need to make along with releasing v5. We can update this issue as new breaking changes are introduced.

- [x] The `makeRequest` method is not longer supported on strategy classes [(docs page)](https://developers.google.com/web/tools/workbox/guides/advanced-recipes#make-requests])
- [x] `registerNavigationRoute` is gone, and [this docs page](https://developers.google.com/web/tools/workbox/modules/workbox-routing#how_to_register_a_navigation_route) needs to be updated, along with a new recipe geared to the App Shell use case.
- [x] A ton of changes to the documentation page for the various build tools' options.
- [x] There's a new, optional `integrity` field in the precache manifest.
- [x] Update the `workbox-broadcast-update` docs to point out it no longer uses the `BroadcastChannel` API (and that it works in all browsers).
- [x] Update all Plugin docs to use the renamed names.
- [x] Document a "disable logging" flag [#1619]
- [x] Talk about using `revision: null` and the eventual deprecation of `string` parameters in the `workbox-precaching` [guide](https://developers.google.com/web/tools/workbox/modules/workbox-precaching#explanation_of_the_precache_list).
- [x] Update Getting Started guide(s) to prefer module imports syntax, with links to `workbox-sw` module page as an alternative.
- [x] Include a simpler standalone offline HTML fallback recipe that uses `workbox-navigation-preload`.
- [x] Remove existing include-based build cool config docs on https://developers.google.com/web/tools/workbox/modules/workbox-build and https://developers.google.com/web/tools/workbox/modules/workbox-webpack-plugin, and link directly to the JSDoc-based docs instead.
- [x] Update the Rollup bundling guide to mention https://www.npmjs.com/package/rollup-plugin-workbox and https://www.npmjs.com/package/rollup-plugin-workbox-inject
- [ ] Update the code labs to prefer module imports syntax.

pscheit:
can i see the docs somewhere online?

jeffposnick:
During the pre-release process, the best way to stay informed is to read each release's notes: https://github.com/GoogleChrome/workbox/releases

We're not going to update the official docs until the final release of v5.

nhoizey:
@jeffposnick do you plan to work on migration documentation, now that rc0 is out?

I would like to try v5, but I'm not sure to understand what needs to be changed from [my v4 code](https://nicolas-hoizey.com/sw.js).

[...]

PR #2212, opened 2019-09-02
Fix workbox-broadcast-update link

abraham:
R: @jeffposnick @philipwalton

Current URL 404s.

### DIFF
--- a/packages/workbox-broadcast-update/README.md
+++ b/packages/workbox-broadcast-update/README.md
@@ -1,1 +1,1 @@
-This module's documentation can be found at https://developers.google.com/web/tools/workbox/modules/workbox-broadcast-update
+This module's documentation can be found at https://developers.google.com/web/tools/workbox/modules/workbox-broadcast-cache-update


philipwalton:
Hmmm, I think the fix for the problem here is our docs site needs to be updated. Specifically the file name of this file as well as some of its titles. @abraham if you're able to make that update and cc me, I can approve and merge.

We also have a TODO to update some of the content there in line with v5 changes: https://github.com/GoogleChrome/workbox/issues/2122

philipwalton:
Hmm, probably makes sense to also add a redirect in this file: https://github.com/google/WebFundamentals/blob/c3570fd9de43d40f91fcf9b3b22e7e64d227538b/src/content/en/tools/_redirects.yaml

abraham:
https://github.com/google/WebFundamentals/pull/8069

philipwalton:
I've published the demos, and the page on developers.google.com/web/ should be updated too, so everything should work now. Thanks!

PR #2214, opened 2019-09-02
Fix incorrect Workbox#update() test logic

### LINKED ISSUE #2213: Incorrect Workbox#update() test logic
**Library Affected**:
*workbox-window*

While making some updates to the `Workbox` tests, I noticed that the test added in #2136 for `update()` do not actually test the update behavior (I missed this in the PR review): https://github.com/GoogleChrome/workbox/blob/ae80e3958c79cb93f4c21bca63cd6e81f7878536/test/workbox-window/window/test-Workbox.mjs#L292-L301

There should only ever be one registration instance per scope, so this test shouldn't be asserting calling `update()` returns a new object. It's only passing here because `update()` is returning `Promise<void>` rather than a registration.

philipwalton:
R: @jeffposnick

Fixes #2213

This PR fixes the incorrect test logic in #2213 and adds more comprehensive tests in general around the `update()` method (handling both own and external updates).

### DIFF
--- a/test/workbox-window/window/test-Workbox.mjs
+++ b/test/workbox-window/window/test-Workbox.mjs
@@ -311,13 +311,14 @@ describe(`[workbox-window] Workbox`, function() {
       // Update the SW after so an update check triggers an update.
       await updateVersion('2.0.0', scriptURL);
 
-      const updatefoundSpy = sandbox.spy();
-      reg.addEventListener('updatefound', updatefoundSpy);
+      wb.update();
 
-      await wb.update();
-      await waitUntil(() => updatefoundSpy.callCount === 1);
-
-      expect(reg.installing).to.not.equal(navigator.serviceWorker.controller);
+      await new Promise((resolve) => {
+        reg.addEventListener('updatefound', () => {
+          expect(reg.installing).to.not.equal(navigator.serviceWorker.controller);
+          resolve();
+        });
+      });
     });
 
     describe(`logs in development-only`, function() {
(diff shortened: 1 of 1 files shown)


test/workbox-window/window/test-Workbox.mjs
```diff
@@ -289,15 +289,35 @@ describe(`[workbox-window] Workbox`, function() {
...
+      await updateVersion('2.0.0', scriptURL);
 
-      expect(reg).not.to.equal(newReg);
+      const updatefoundSpy = sandbox.spy();
+      reg.addEventListener('updatefound', updatefoundSpy);
+
+      await wb.update();
+      await waitUntil(() => updatefoundSpy.callCount === 1);
+
+      expect(reg.installing).to.not.equal(navigator.serviceWorker.controller);
```
jeffposnick: Is there a race condition here, where `reg.wating` (or `reg.active`; not sure if `skipWaiting()` is being called) might be set instead of `reg.installing` if the updated service worker's `install` handler finishes quickly?

I would feel better if there were an assertion making sure that `reg.installing` is not `null`.

Or alternatively, it looks like you work around this elsewhere with

```
reg.addEventListener('updatefound', () => {
  updatedSW = reg.installing;
});
```

philipwalton: How about I make the assertion in the `updatefound` event listener?

jeffposnick: I am pretty sure that when `updatefound` is triggered, `reg.installing` will be set: https://w3c.github.io/ServiceWorker/#service-worker-registration-updatefound-event

So stashing away a reference via `updatedSW = reg.installing` should be fine, and then you could change your assertion in this line to

```
expect(updatedSW).to.not.equal(navigator.serviceWorker.controller);
```

philipwalton: Updated. PTAL.

Issue #2216, opened 2019-09-03
Add support for deferred/conditional registration in workbox-window


philipwalton:
**Library Affected**:
*workbox-window*

After discussion with @azizhk in https://github.com/GoogleChrome/workbox/pull/2209#issuecomment-527223148 and then with @jeffposnick. We think it makes sense to update `workbox-window` to better support use cases around deferred or conditional registration.

Currently, the library runs most of its initialization logic in the `register()` method (including adding lifecycle event listeners), so if a developer doesn't call `register()` right away, they may miss lifecycle events when an updated version of the service worker is found.

With a slight refactoring and putting most of the initialization logic in the constructor, we should be able to support this.

azizhk:
I don't think moving the logic from register to constructor would help.
In my example I would create the workbox instance also quite late.

tomayac:
Hi there,

Workbox is [moving to a new engineering team within Google](https://github.com/GoogleChrome/workbox?tab=readme-ov-file#maintenance-update:~:text=Maintenance%20update,owners%20of%20Workbox.). As part of this move, we're declaring a partial bug bankruptcy to allow the new team to start fresh. We realize this isn't optimal, but realistically, this is the only way we see it working. For transparency, here're the criteria we applied:

- For PRs, we closed [everything that doesn't target the current `v7` branch](https://github.com/GoogleChrome/workbox/pulls?q=is%3Apr+is%3Aclosed+-base%3Av7).
- For Issues, we closed [everything that was created before 2024](https://github.com/GoogleChrome/workbox/issues?q=is%3Aissue+is%3Aclosed+created%3A%3C2024-01-01+). 

Thanks, and we hope for your understanding!
The Workbox team

PR #2218, opened 2019-09-04
Installing Race Condition Test

azizhk:
Based on discussion in #2209 #2216 
This is a test case trying to recreate a race condition where the service worker is in `"installing"` state when the user/developer calls `wb.register()`.

Workbox does not add listeners to pre-existing `registration.installing` service worker but only when `updatefound` event is triggered.

R: @jeffposnick @philipwalton


azizhk:
I hope it's clear that my aim is not to defer "installing" but actually to programmatically create a condition where there exists a `registration.installing` SW when register is called.

tomayac:
Hi there,

The Workbox team

PR #2232, opened 2019-09-24
Ignore both swSrc and swDest when globbing for the manifest

### LINKED ISSUE #2229: Ensure that swSrc is automatically added to globIgnores
**Library Affected**:
*workbox-build*

When using `workbox-build` in `injectManifest` mode, `swDest` is automatically added to the `globIgnores` config. We should do this for `swSrc` as well.

jeffposnick:
R: @philipwalton

Fixes #2229

We previously had some code that ensured that `swDest` was omitted from the `glob` matches when generating the precache manifest.

This extends that to also ignore `swSrc` in `injectManifest` mode.

I also improved the that generates the `globIgnores` value so that instead of just matching anything that has the same basename as `swSrc`/`swDest`, it actually uses a full path to the file, relative to `globDirectory`. This should reduce the chance of a false positive.

### DIFF
--- a/packages/workbox-build/src/lib/rebase-path.js
+++ b/packages/workbox-build/src/lib/rebase-path.js
@@ -1,0 +1,22 @@
+/*
+  Copyright 2018 Google LLC
+
+  Use of this source code is governed by an MIT-style
+  license that can be found in the LICENSE file or at
+  https://opensource.org/licenses/MIT.
+*/
+
+const upath = require('upath');
+
+module.exports = ({baseDirectory, file}) => {
+  // file starts our relative to the current directory, so make it absolute.
+  const absolutePath = upath.resolve(file);
+
+  // Convert the absolute path so that it's relative to the baseDirectory.
+  const relativePath = upath.relative(baseDirectory, absolutePath);
+
+  // Remove any leading ./ as it won't work in a glob pattern.
+  const normalizedPath = upath.normalize(relativePath);
+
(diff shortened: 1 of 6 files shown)


jeffposnick:
Actually, let me extend this to ignore the `workbox-*.js` runtime files that are generated, too.

packages/workbox-build/src/lib/rebase-path.js
```diff
@@ -0,0 +1,22 @@
+/*
+  Copyright 2018 Google LLC
```
philipwalton: s/2018/2019/

packages/workbox-build/src/lib/rebase-path.js
```diff
@@ -0,0 +1,22 @@
...
+
+  Use of this source code is governed by an MIT-style
+  license that can be found in the LICENSE file or at
+  https://opensource.org/licenses/MIT.
+*/
+
+const upath = require('upath');
+
+module.exports = ({baseDirectory, file}) => {
+  // file starts our relative to the current directory, so make it absolute.
```
philipwalton: s/file/File/

I'm not really sure what you mean by "file starts". Maybe reword as "Files are relative to the..."?

jeffposnick: Sorry, that should have been "out" instead of "our".

But I'll rephrase.

PR #2237, opened 2019-09-25
Fixes calculateEffectiveBoundaries bug when start=0

### LINKED ISSUE #2236: calculateEffectiveBoundaries improperly handles start=0
**Library Affected**:
*workbox-range-requests*

I remember a discussion around this coming up during the TypeScript migration review, but after testing out v5's `RangeRequestsPlugin` in actual application, I found that it's behavior is buggy when `start` (and presumably `end`, but that's less likely) is set to `0`:

https://github.com/GoogleChrome/workbox/blob/bfd5c718a81af2add2801883de1b18a85d072529/packages/workbox-range-requests/src/utils/calculateEffectiveBoundaries.ts#L47-L57

Those statements need to be a comparison against `undefined` instead of truthiness, since `0` needs to be treated as a valid value.

jeffposnick:
R: @philipwalton

Fixes #2236

### DIFF
--- a/packages/workbox-range-requests/src/utils/calculateEffectiveBoundaries.ts
+++ b/packages/workbox-range-requests/src/utils/calculateEffectiveBoundaries.ts
@@ -44,14 +44,15 @@ function calculateEffectiveBoundaries(
   let effectiveStart: number;
   let effectiveEnd: number;
 
-  if (start && end) {
+  // != and == are used so both null and undefined are treated the same.
+  if (start != undefined && end != undefined) {
     effectiveStart = start;
     // Range values are inclusive, so add 1 to the value.
     effectiveEnd = end + 1;
-  } else if (start && !end) {
+  } else if (start != undefined && end == undefined) {
     effectiveStart = start;
     effectiveEnd = blobSize;
-  } else if (end && !start) {
+  } else if (end != undefined && start == undefined) {
     effectiveStart = blobSize - end;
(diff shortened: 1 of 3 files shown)


packages/workbox-range-requests/src/utils/calculateEffectiveBoundaries.ts
```diff
@@ -44,14 +44,15 @@ function calculateEffectiveBoundaries(
   let effectiveStart: number;
   let effectiveEnd: number;
 
-  if (start && end) {
+  // != and == are used so both null and undefined are treated the same.
```
jeffposnick: If there's a more TypeScript-y way of doing this, let me know.

philipwalton: From my reading of [the code for the `parseRangeHeader()` function](https://github.com/GoogleChrome/workbox/blob/bfd5c718a81af2add2801883de1b18a85d072529/packages/workbox-range-requests/src/utils/parseRangeHeader.ts#L50-L51), only `undefined` can be returned, so I don't think it's necessary to account for `null`.

jeffposnick: Okay, yeah, the 4.3.1 version returned `null` in that scenario, so I thought without checking to closely that `null` was still possible.

PR #2239, opened 2019-09-27
Account for injectManifest replacements in sourcemaps

### LINKED ISSUE #2235: Webpack InjectManifest plugin produces incorrect source maps in Webpack production mode
**Library Affected**:
*workbox-webpack-plugin v5 beta*

**Browser & Platform**:
*all browsers*

When using the `workbox-webpack-plugin` `InjectManifest` plugin with Webpack in production mode, the service worker source maps are corrupted/incorrect.

- Repro repo: [ezzatron/webpack-v5-injectmanifest-sourcemap-bug-repro](https://github.com/ezzatron/webpack-v5-injectmanifest-sourcemap-bug-repro)
- Repro steps:
  1. `yarn install`
  2. `webpack`
  3. Feed `dist/sw.js` and `dist/sw.js.map` into a [source map visualization tool] and see the [buggy output]

[source map visualization tool]: https://sokra.github.io/source-map-visualization/
[buggy output]: https://git.io/JeGZN

The issue seems to be that the source map is generated before the precache manifest is added, because manually removing the precache manifest from the compiled service worker makes the source map line up with the intended positions in the code.

jeffposnick:
R: @philipwalton

Fixes #2235

This was quite a rabbit hole to go down.

Ultimately, I *think* the code that I ended up with to modify an existing sourcemap to account for the injected manifest does the job, but I would not be shocked if there are some edge cases that I'm not aware of.


packages/workbox-build/src/inject-manifest.js
```diff
@@ -78,23 +78,60 @@ async function injectManifest(config) {
...
+
+  const url = sourceMapURL.getFrom(swFileContents);
+  // If our swSrc file contains a sourcemap, we would invalidate that
+  // mapping if we just replaced injectionPoint with the stringified manifest.
+  // Instead, we need to update the swDest contents as well as the sourcemap
+  // at the same time.
+  // See https://github.com/GoogleChrome/workbox/issues/2235
+  if (url) {
+    const sourcemapSrcPath = upath.resolve(upath.dirname(options.swSrc), url);
+    const sourcemapDestPath = upath.resolve(upath.dirname(options.swDest), url);
```
steverep, line 92: This will always keep the same source map filename, and if `swSrc` and `swDest` are in the same folder, it'll overwrite the original source map file.

Instead, this should just append ".map" to `swDest`.

packages/workbox-build/src/lib/replace-and-update-source-map.js
```diff
@@ -0,0 +1,108 @@
...
+  const updatedSourceMap = Object.assign(JSON.parse(generator.toString()), {
+    names: originalMap.names,
+    sourceRoot: originalMap.sourceRoot,
+    sources: originalMap.sources,
+    sourcesContent: originalMap.sourcesContent,
+  });
+
+  return {
+    map: JSON.stringify(updatedSourceMap),
+    source: src,
```
steverep, line 104: This is returned without changing the URL of the source map to reflect the filename in `swDest`.

steverep (commented):
I filed #3233 thinking there was no support in `injectManifest`, but I now see the root issues per my comments below.  In short, the source map file and reference URL in the new source file will only be correct when `swSrc` and `swDest` are in different directories with the same filename, with their corresponding maps in the same folders.

PR #2244, opened 2019-10-04
Strict TypeScript cleanup

### LINKED ISSUE #2242: Use strict: true tsc option
#2241 enables `strictPropertyInitialization`, but we would have had that enabled already if we went with the `"strict": true` compiler option.

It's probably a good idea to flip that on and remediate all the problems (in a separate PR).

jeffposnick:
R: @philipwalton

Fixes #2242 

This cleans up a number of problems that were found by switching to using `strict: true` in the TypeScript compiler options.

There are some... questionable?... casts going on, so I'll address them via comments on the specific lines to see whether there are better approaches.

### DIFF
--- a/packages/workbox-core/src/_private/DBWrapper.ts
+++ b/packages/workbox-core/src/_private/DBWrapper.ts
@@ -266,7 +266,9 @@ export class DBWrapper {
     ...args: any[]
   ) {
     const callback = (txn: IDBTransaction, done: Function) => {
-      const objStore = txn.objectStore(storeName)
+      const objStore = txn.objectStore(storeName);
+      // TODO(philipwalton): Fix this underlying TS2684 error.
+      // @ts-ignore
       const request = <IDBRequest> objStore[method].apply(objStore, args);
 
       request.onsuccess = () => done(request.result);
--- a/packages/workbox-precaching/src/cleanupOutdatedCaches.ts
+++ b/packages/workbox-precaching/src/cleanupOutdatedCaches.ts
@@ -19,7 +19,8 @@ import './_version.js';
  * @alias workbox.precaching.cleanupOutdatedCaches
  */
 export const cleanupOutdatedCaches = () => {
-  addEventListener('activate', (event: ExtendableEvent) => {
+  // See https://github.com/Microsoft/TypeScript/issues/28357#issuecomment-436484705
+  addEventListener('activate', ((event: ExtendableEvent) => {
     const cacheName = cacheNames.getPrecacheName();
 
     event.waitUntil(deleteOutdatedCaches(cacheName).then((cachesDeleted) => {
@@ -30,5 +31,5 @@ export const cleanupOutdatedCaches = () => {
         }
       }
     }));
-  });
+  }) as EventListener);
 };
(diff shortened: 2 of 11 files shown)


packages/workbox-core/src/_private/DBWrapper.ts
```diff
@@ -266,7 +266,9 @@ export class DBWrapper {
     ...args: any[]
   ) {
     const callback = (txn: IDBTransaction, done: Function) => {
-      const objStore = txn.objectStore(storeName)
+      const objStore = txn.objectStore(storeName);
+      // TODO(philipwalton): Fix this underlying TS2684 error.
+      // @ts-ignore
```
jeffposnick: This is the biggest issue, and I just punted on it as I don't have any idea how to resolve it. @philipwalton, if you can look into it as part of either another PR or push an update to this PR, that would be great.

packages/workbox-precaching/src/cleanupOutdatedCaches.ts
```diff
@@ -19,7 +19,8 @@ import './_version.js';
  * @alias workbox.precaching.cleanupOutdatedCaches
  */
 export const cleanupOutdatedCaches = () => {
-  addEventListener('activate', (event: ExtendableEvent) => {
+  // See https://github.com/Microsoft/TypeScript/issues/28357#issuecomment-436484705
```
jeffposnick: There are a lot of similar changes to this one, due to TypeScript being unhappy about the callback taking a parameter of type `ExtendableEvent`.

Declaring these functions as being of type `EventListener` seems fairly "safe", but perhaps the more official route would be to override the `addEventListener` definition so that it supports a listener function that takes in an `ExtendableEvent` parameter, instead of just an `Event` parameter.

packages/workbox-routing/src/Router.ts
```diff
@@ -128,11 +130,10 @@ class Router {
 
         // If a MessageChannel was used, reply to the message on success.
         if (event.ports && event.ports[0]) {
-          await requestPromises;
-          event.ports[0].postMessage(true);
+          requestPromises.then(() => event.ports[0].postMessage(true));
```
jeffposnick: Declaring the listener function as being an `EventListener` wouldn't work if it was `async`, since `EventListener`s aren't supposed to return `Promise<void>`.

So I just took the one `await` and converted it into a `then()`, and dropped the `async` from the listener function declaration.

packages/workbox-window/src/utils/WorkboxEventTarget.ts
```diff
@@ -27,7 +27,7 @@ export class WorkboxEventTarget {
    */
   addEventListener<K extends keyof WorkboxEventMap>(type: K, listener: (event: WorkboxEventMap[K]) => any) {
     const foo = this._getEventListenersByType(type)
-    foo.add(listener);
+    foo.add(listener as ListenerCallback);
```
jeffposnick: This and the other place where I declare `listener` as being a `ListenerCallback` might be a little sketchy, so your thoughts are appreciated.

The underlying error is because some varieties of `listener` might have a `data` property (the one that wraps a message event), and TypeScript isn't happy about that.

There might be a better remediation here, like changing the `_getEventListenersByType()` definition so that it returns a `Set` with  different signature.

Issue #1790, opened 2018-12-13
Precache manifest missing entries when recompiled via webpack --watch


lakesare:
**Library Affected**:
workbox-webpack-plugin

**Browser & Platform**:
Webpack 2.3.3, workbox-webpack-plugin 3.6.3

**Issue Description**:
In the `webpack --watch` mode, `precache-manifest.hash.js` file at first generates a perfect list of files, but on subsequent updates it misses some files.
For example, files generated with `CopyWebpackPlugin` are missing (but including `copyUnmodified: true` in options helps).
Another kind of files that are missing, is `fontawesome-webfont.eot?hash` files (importes via this process https://medium.com/@chanonroy/webpack-2-and-font-awesome-icon-importing-59df3364f35c), this I don't yet know how to fix.

Here is a similar issue for another manifest plugin: https://github.com/danethurber/webpack-manifest-plugin/issues/144.

jeffposnick:
Sorry for letting this drop until now.

### workboox-webpack-plugin background

`workbox-webpack-plugin` generates its precache manifest based on the contents of the current compilation's `assets`, as determined in a compiler hook into the [`emit`](https://webpack.js.org/api/compiler-hooks/#emit) event. I _think_ that's fairly standard for a webpack plugin.

### webpack --watch mode behavior

After running some local experiments with `--watch` mode, I've found that only brand new assets, and modifications to existing assets, show up in the compilation's `assets` available during each subsequent `emit` hook invocation.

I believe some other plugins (like `webpack-manifest-plugin`, as referenced in https://github.com/danethurber/webpack-manifest-plugin/issues/144) maintain state information about all of the previous compilation's `assets` and make additions or changes to hashes to that state information upon each subsequent compilation. They then create their output based on the cumulative total of all the assets that have been created across multiple compilations. (This seems to be the purpose of the [`seed` parameter](https://github.com/danethurber/webpack-manifest-plugin#optionsseed) that can be passed to `webpack-manifest-plugin`.)

### Problems with cumulative state

`workbox-webpack-plugin` could do something similar, and generate its precache manifest based on a cumulative list of assets across multiple compilations.

The concern that I have, though, is that there is no way for a webpack plugin to detect that a given asset has been **removed** following a recompilation—the absence of an asset from any particular compilation can't be used as a signal that it was removed, since that scenario is identical to what would happen if it was still present but unchanged.

[...]

PR #2245, opened 2019-10-04
Warn during multi-compilation scenarios

jeffposnick:
R: @philipwalton

This addresses some of #1790 by adding in a `webpack` compilation warning message letting developers know that they shouldn't rely on a consistent set of assets in the precache manifest when the same plugin instance is used across multiple compilations (such as in `--watch` scenarios).


packages/workbox-webpack-plugin/src/generate-sw.js
```diff
@@ -80,6 +81,17 @@ class GenerateSW {
    * @private
    */
   async handleEmit(compilation) {
+    // See https://github.com/GoogleChrome/workbox/issues/1790
+    if (this.alreadyCalled) {
```
jeffposnick: I was thinking about putting this logic in a reusable module, but it a) relies on instance properties of the plugin and b) relies on the `compilation` object being available. It ended up being cleaner just duplicating this logic in both plugin classes, but if there are strong feelings, I could refactor.

PR #2248, opened 2019-10-05
New feature : Skip waiting response

### LINKED ISSUE #2247: [Feature Request] Skip waiting response
**Library Affected**:
*workbox-build*

**Browser & Platform**:
all browsers

It would be useful to allow browser to provide a `MessageChannel` `port` in additionnal to the `postMessage` required to trigger the `skipWaiting`.

**Exemple code**
```js
const updated = (registration) => {
    const { waiting } = registration
    if (!waiting) {
        return
    }
    const channel = new MessageChannel()
    channel.port1.onmessage = (event) => {
        const { data } = event
        if (data.error) {
            console.error('[sw] update failed : ' + data.error)
        } else {
            location.reload(true)
        }
    }
    waiting.postMessage({type: 'skip-waiting'}, [channel.port2])
}
```

Thanks for your great job

DaedalusDev:
R: @jeffposnick @philipwalton

Fixes #2247

Add a response message when the `skipWaiting` resolve.

### DIFF
--- a/packages/workbox-build/src/templates/sw-template.js
+++ b/packages/workbox-build/src/templates/sw-template.js
@@ -32,8 +32,8 @@ importScripts(
 <%= use('workbox-core', 'skipWaiting') %>();
 <% } else { %>
 self.addEventListener('message', (event) => {
-  const replyPort = event.ports[0]
-  const data = event.data
+  const replyPort = event.ports[0];
+  const data = event.data;
   if (data && data.type === 'SKIP_WAITING') {
     event.waitUntil(
       self.skipWaiting()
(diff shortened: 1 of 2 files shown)


DaedalusDev:
@googlebot I signed it!

philipwalton:
If your end goad is to have a promise that resolves when the service worker transitions from waiting to activating, you can already do that from the window via `statechange` events.

The code you've written here would resolve after the `skipWaiting()` call has been made in the service worker, but that may or may not correspond to when that service worker starts activating.

I *think* it's better for you to use `statechange` events to observe that a `SKIP_WAITING` message was successful, but perhaps I'm missing a use case you're wanting to solve?

DaedalusDev:
@philipwalton 👍 
I will try this approach (it's also more simple)
Thanks.

PR #2254, opened 2019-10-15
New precache helper functions + removal of createHandlerForURL

### LINKED ISSUE #2249: PrecacheController helpers for retrieving responses from the cache
**Library Affected**:
*workbox-precaching.*

@philipwalton had previously pinged me about wanting this, but I didn't realize how bad the current ergonomics were until I tried doing something similar using TypeScript.

I've ended up writing this a helper function in my own code:

```
async function getPrecachedResponse(url: string) {
  const cacheKey = getCacheKeyForURL(url);
  if (!cacheKey) {
    throw new Error(`${url} is not in the precache manifest.`);
  }

  const cache = await caches.open(cacheNames.precache);
  const cachedResponse = await cache.match(cacheKey);
  if (!cachedResponse) {
    throw new Error(`${url} is not precached.`);
  }

  return cachedResponse;
}
```

It would make sense to add this as `PrecacheController.getPrecachedResponse(url: string)`, though probably returning `null` instead of throwing an error when there's a cache miss would be more in keeping with the behavior of the Cache Storage API in general. (I'm already handling exceptions in my own code.)

We can *probably* keep `PrecacheController.createHandlerForURL(url: string)` as-is, but I would also be open to renaming it. (We need to keep it around under some name, since it's needed for handling navigation routes in App Shell-like scenarios.)

jeffposnick:
R: @philipwalton

Fixes #2249

Exposes three new functions:

- `async function matchPrecache(request: string|Request): Promise<Response|null>`
- `function createHandler(fallbackToNetwork = true): RouteHandlerCallback`
- `function createHandlerBoundToURL(url: string, fallbackToNetwork = true): RouteHandlerCallback`

`createHandlerBoundToURL` replaces `createHandlerForURL`, and serves the same purpose.

### DIFF
--- a/packages/workbox-precaching/src/PrecacheController.ts
+++ b/packages/workbox-precaching/src/PrecacheController.ts
@@ -359,8 +359,10 @@ class PrecacheController {
 
         // This shouldn't normally happen, but there are edge cases:
         // https://github.com/GoogleChrome/workbox/issues/1441
-        throw new Error(`The cache ${this._cacheName} did not have an entry ` +
-            `for ${request.url}.`);
+        throw new WorkboxError('missing-precache-entry', {
+          cacheName: this._cacheName,
+          url: request.url,
+        });
       } catch (error) {
         if (fallbackToNetwork) {
           if (process.env.NODE_ENV !== 'production') {
--- a/packages/workbox-precaching/src/PrecacheController.ts
+++ b/packages/workbox-precaching/src/PrecacheController.ts
@@ -1,5 +1,5 @@
 /*
-  Copyright 2018 Google LLC
+  Copyright 2019 Google LLC
 
   Use of this source code is governed by an MIT-style
   license that can be found in the LICENSE file or at
@@ -13,7 +13,8 @@ import {fetchWrapper} from 'workbox-core/_private/fetchWrapper.js';
 import {logger} from 'workbox-core/_private/logger.js';
 import {WorkboxError} from 'workbox-core/_private/WorkboxError.js';
 import {copyResponse} from 'workbox-core/copyResponse.js';
-import {RouteHandlerCallback} from 'workbox-core/types.js';
+import {RouteHandlerCallback, RouteHandlerCallbackOptions}
+    from 'workbox-core/types.js';
 import {WorkboxPlugin} from 'workbox-core/types.js';
 
 import {PrecacheEntry} from './_types.js';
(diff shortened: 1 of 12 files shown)


packages/workbox-precaching/src/PrecacheController.ts
```diff
@@ -310,58 +311,94 @@ class PrecacheController {
...
+  async matchPrecache(request: string|Request): Promise<Response|undefined> {
+    const url = request instanceof Request ? request.url : request;
     const cacheKey = this.getCacheKeyForURL(url);
-    if (!cacheKey) {
-      throw new WorkboxError('non-precached-url', {url});
+    if (cacheKey) {
+      const cache = await caches.open(this._cacheName);
+      return cache.match(cacheKey);
     }
+    return undefined;
```
philipwalton: Is this line needed?

packages/workbox-precaching/src/PrecacheController.ts
```diff
@@ -310,58 +311,94 @@ class PrecacheController {
...
-        const response = await cache.match(cacheKey);
-
+        const response = await this.matchPrecache(request);
         if (response) {
           return response;
         }
 
         // This shouldn't normally happen, but there are edge cases:
         // https://github.com/GoogleChrome/workbox/issues/1441
         throw new Error(`The cache ${this._cacheName} did not have an entry ` +
```
philipwalton: Should this be `WorkboxError`—so we can avoid long strings in production builds?

packages/workbox-precaching/src/PrecacheController.ts
```diff
@@ -310,58 +311,94 @@ class PrecacheController {
...
+   * If for an unexpected reason there is a cache miss for the request,
+   * this will fall back to retrieving the `Response` via `fetch()` when
+   * `fallbackToNetwork` is `true`.
+   *
+   * @param {boolean} [fallbackToNetwork=true] Whether to attempt to get the
+   * response from the network if there's a precache miss.
+   * @return {workbox.routing.Route~handlerCallback}
+   */
+  createHandler(fallbackToNetwork = true): RouteHandlerCallback {
+    return async ({request}: RouteHandlerCallbackOptions) => {
```
philipwalton: Doesn't this also need to take an `event` and call `event.waitUntil()`?

packages/workbox-precaching/src/PrecacheController.ts
```diff
@@ -310,58 +311,94 @@ class PrecacheController {
...
+   */
+  createHandlerBoundToURL(url: string, fallbackToNetwork = true): RouteHandlerCallback {
+    const cacheKey = this.getCacheKeyForURL(url);
+    if (!cacheKey) {
+      throw new WorkboxError('non-precached-url', {url});
+    }
+
+    const handler = this.createHandler(fallbackToNetwork);
+    const request = new Request(url);
+    return () => handler({request});
```
philipwalton: Similar to above, I think this needs to merge `{request}` with whatever options get passed to the handler so that the `event` object can be passed to the handler.

test/workbox-precaching/sw/test-matchPrecache.mjs
```diff
@@ -0,0 +1,56 @@
...
+
+  Use of this source code is governed by an MIT-style
+  license that can be found in the LICENSE file or at
+  https://opensource.org/licenses/MIT.
+*/
+
+import {matchPrecache} from 'workbox-precaching/matchPrecache.mjs';
+import {precache} from 'workbox-precaching/precache.mjs';
+
+describe(`matchPrecache()`, function() {
```
philipwalton: Should this, instead, be testing the `matchPrecache()` method on `PrecacheController`, and then the test for the plain export can just ensure a default controller is created and its `matchPrecache()` method called?

packages/workbox-precaching/src/PrecacheController.ts
jeffposnick: Okay, switching.

test/workbox-precaching/sw/test-matchPrecache.mjs
jeffposnick: This matches what we're doing with other tests modules that follow a similar pattern (c.f. [`test-getCacheKeyForURL.mjs`](https://github.com/GoogleChrome/workbox/blob/f23924a988a3e321dbcd9e3781f07b1de5dce916/test/workbox-precaching/sw/test-getCacheKeyForURL.mjs)).

So... maybe we can keep this test implemented as-is and if we want a different approach, update all of the relevant tests to follow that model?

packages/workbox-precaching/src/PrecacheController.ts
jeffposnick: When used by our `Router` class, the promise that's returned from a handler will be passed to `event.respondWith()`, and by virtue of that, the service worker will be kept alive until the promise is fulfilled:
 https://github.com/GoogleChrome/workbox/blob/f23924a988a3e321dbcd9e3781f07b1de5dce916/packages/workbox-routing/src/Router.ts#L76-L79

If you're using this handler outside of the `Router` class, then I think it's up to you to figure out whether or not you need to explicitly keep the service worker alive.

Also note that `RouteHandlerCallbackOptions` marks `event` as being optional: https://github.com/GoogleChrome/workbox/blob/f23924a988a3e321dbcd9e3781f07b1de5dce916/packages/workbox-core/src/types.ts#L39-L44

So if it *had* to take an event, we would have to use a type other than `RouteHandlerCallbackOptions` for the parameter.

packages/workbox-precaching/src/index.ts
```diff
@@ -1,22 +1,25 @@
 /*
-  Copyright 2018 Google LLC
+  Copyright 2019 Google LLC
```
philipwalton: Nit: I don't think this needs to change for existing files. I'm only mentioned it in the past when a new file was created with a copyright date prior to its existence (due to copy/pasta).

packages/workbox-precaching/src/PrecacheController.ts
jeffposnick: Oops, forgot to push out my local commit.

PR #2262, opened 2019-10-23
Unrevisioned precaching warning

### LINKED ISSUE #2259: Warn when calling precache with a string or without revision property
I'm sure that this was discussed during v5 development, but I'm unable to find the historical issues/PRs with comments about it.

A fairly common issue that we see is developers who pass in string URLs, like `'/offline.html'`, that lack versioning info to `workbox-precaching`. This is a problem because those URLs will never be updated after being cached for the first time. (This is fine if those URLs *do* contain versioning info, like `/app.1234abcd.js`, since it's assumed that the contents will never change, so updates aren't necessary.)

This came up most recently when I took a look at https://github.com/GoogleChrome/web.dev/blob/93694df6ec4874a85975127ce4a1856066e76bf5/src/lib/sw.js#L9

I *think* what we ended up deciding on is that all of the build tools will produce `{url: '/app.1234abcd.js'}` manifest entries, and that we'd very explicitly warn folks who passed in strings in v5 that they need to make sure those URL strings included versioning information, or else dire things would happen. (I'd actually be in favor of doing this in production builds, unlike most of our other warnings that only trigger in dev builds.) And then potentially deprecate passing in strings in v6.

But... I can't find any code in [`PrecacheController`](https://github.com/GoogleChrome/workbox/blob/ac28527b25e301e1330c044a59770af80e8c909a/packages/workbox-precaching/src/PrecacheController.ts) that actually does that.

@philipwalton, is my recollection on that decision accurate? If so, I think we just forgot to add in that extra warning message, and we still have time to do that before the final v5 release.

jeffposnick:
R: @philipwalton

Fixes #2259

Let me know if it's too verbose.

The https://bit.ly/wb-precache link goes to https://developers.google.com/web/tools/workbox/modules/workbox-precaching#explanation_of_the_precache_list right now, which needs to be rewritten a to provide full context about what the problem is and what we want developers to do.

But I think that's the most reasonable place to link to that exists in our docs,  unless we wanted to create a dedicated landing page that elaborated on the topic.

### DIFF
--- a/packages/workbox-precaching/src/PrecacheController.ts
+++ b/packages/workbox-precaching/src/PrecacheController.ts
@@ -100,21 +100,23 @@ class PrecacheController {
       this._urlsToCacheKeys.set(url, cacheKey);
       this._urlsToCacheModes.set(url, cacheMode);
 
-      const warningMessage = `Workbox is precaching URLs without revision ` +
-        `info: ${urlsToWarnAbout.join(', ')}\nOnce precached, the responses ` +
-        `will never be updated.\nThis is only safe for URLs which ` +
-        `contain versioning info (e.g. /v1.0/app.js or /app.1234.js, not ` +
-        `/index.html).\nIf your URLs do contain revision info, pass in ` +
-        `a object with a 'url' property and a 'revision' property set to ` +
-        `null (e.g. {url: '/v1.0/index.js', revision: null}).\nIn a future ` +
-        `release, using revision: null will be required.\nLearn more at ` +
-        `https://bit.ly/wb-precache`;
-      if (process.env.NODE_ENV === 'production') {
-        // Use console directly to display this warning without bloating bundle
-        // sizes by pulling in all of the logger codebase in production.
-        console.warn(warningMessage);
-      } else {
-        logger.warn(warningMessage);
+      if (urlsToWarnAbout.length > 0) {
+        const warningMessage = `Workbox is precaching URLs without revision ` +
+          `info: ${urlsToWarnAbout.join(', ')}\nOnce precached, they ` +
+          `will never be updated.\nThis is only safe for URLs which ` +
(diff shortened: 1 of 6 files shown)


packages/workbox-precaching/src/PrecacheController.ts
```diff
@@ -91,6 +99,25 @@ class PrecacheController {
...
+      if (urlsToWarnAbout.length > 0) {
+        const warningMessage = `Workbox is precaching URLs without revision ` +
+          `info: ${urlsToWarnAbout.join(', ')}\nOnce precached, they ` +
+          `will never be updated.\nThis is only safe for URLs which ` +
+          `contain versioning info (e.g. /v1.0/app.js or /app.1234.js, not ` +
+          `/index.html).\nIf your URLs do contain revision info, pass in ` +
+          `a object with a 'url' property and a 'revision' property set to ` +
+          `null (e.g. {url: '/v1.0/index.js', revision: null}).\nIn a future ` +
+          `release, using revision: null will be required.\nLearn more at ` +
+          `https://bit.ly/wb-precache`;
```
philipwalton: I do think this is pretty long. And since you're already linking to a URL, I think we can make it shorter and put this information on the page we link to. What about:

"Workbox is precaching URLs without revision info. This is generally NOT safe. Learn more at https://bit.ly/wb-precache." (and pass the list of URLs as the second argument to `console.warn()`).

jeffposnick: Yeah, I think you're right. Given that we don't have the right verbiage on https://bit.ly/wb-precache yet, I went long, but we can just get that written up in time for the v5 launch.

I've seen issues with logging "live" variables in the JS console from a service worker, where they show up something generic like `[Array]` after the service worker's process has been killed. I'd rather stringify them first to make sure that the log message is still meaningful after the service worker has stopped.

PR #2274, opened 2019-11-07
Adds in explicit self/window in a few places

jeffposnick:

R: @philipwalton

There are a few instances where using the implicit global (`window` or `self`) was making JS Compiler with a specific set of `externs` unhappy.

### DIFF
--- a/packages/workbox-window/src/Workbox.ts
+++ b/packages/workbox-window/src/Workbox.ts
@@ -104,7 +104,7 @@ class Workbox extends WorkboxEventTarget {
     }
 
     if (!immediate && document.readyState !== 'complete') {
-      await new Promise((res) => addEventListener('load', res));
+      await new Promise((res) => window.addEventListener('load', res));
     }
 
     // Set this flag to true if any service worker was controlling the page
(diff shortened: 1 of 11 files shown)


packages/workbox-window/src/Workbox.ts
```diff
@@ -104,7 +104,7 @@ class Workbox extends WorkboxEventTarget {
     }
 
     if (!immediate && document.readyState !== 'complete') {
-      await new Promise((res) => addEventListener('load', res));
+      await new Promise((res) => window.addEventListener('load', res));
```
philipwalton: I hope it's not a problem here, but I know in the past using `window` has been an issue for people including this code in an SSR context. I don't think switching to `self` changes that though, so I'd rather wait and see if it's a problem before doing something like [I had to do here](https://github.com/w3c/IntersectionObserver/blob/d3eece49746574f43798d411d9fea2968251b075/polyfill/intersection-observer.js#L12-L15).

jeffposnick: This particular code is preceded by `document.readyState !== 'complete'` in the `if()` clause, which seems like it would already be problematic if we're talking about SSR.

PR #2282, opened 2019-11-21
Debug logging toggle via global boolean

### LINKED ISSUE #1619: More control over enabling/disabling debug logging
Currently, it's not possible to disable debugging logs in localhost, and it's a very annoying experience. It would be nice to allow adding a `debug` option to `GenerateSW` configuration and, since we're here, customising log level should also be possible.

jeffposnick:
R: @philipwalton

Fixes #1619

I'm going with `self.__WB_DISABLE_DEBUG_LOGS` for the global runtime toggle, and `disableDebugLogs` as the `generateSW` option name to set the initial value in a generated service worker.


jeffposnick:
I'd be more inclined to go with `__WB_DISABLE_DEV_LOGS`, since `__WB_DISABLE_LOGS` makes it sound like we also wouldn't log on errors (which we still do).

PR #2287, opened 2019-11-22
Await an async cacheWillUpdate in PrecacheController.install()

### LINKED ISSUE #2285: PrecacheController should await potentially async plugin cacheWillUpdate
According to [the docs](https://developers.google.com/web/tools/workbox/guides/using-plugins#custom_plugins) a plugin's `cacheWillUpdate` can be asynchronous but this function is not `await`ed in the PrecacheController ([src](https://github.com/GoogleChrome/workbox/blob/8b22c152c6c30e6b50097bf76503fdd8c38e8378/packages/workbox-precaching/src/PrecacheController.ts#L265)). I believe this is leading to non-deterministic behavior in the installation process.

Here's a reconstruction of my logs, if it's helpful:
```
Installing the PrecacheController...
Network request for '/foo.js' returned a response with status '200'.
customPlugin.cacheWillUpdate called for '/foo.js'.
customPlugin.cacheWillUpdate called for '/foo.js'.
Updating the 'mock' cache with a new Response for /foo.js.
Successfully installed the PrecacheController
```

jeffposnick:
R: @philipwalton

Fixes #2285 

When we made all of the standard  plugins `async`, we forgot add in an `await` inside of the `PrecacheController.install()` method.

### DIFF
--- a/test/workbox-precaching/sw/test-PrecacheController.mjs
+++ b/test/workbox-precaching/sw/test-PrecacheController.mjs
@@ -656,6 +656,36 @@ describe(`PrecacheController`, function() {
       // This should succeed.
       await precacheController.install({plugins});
     });
+
+    it(`should properly await an async cacheWillUpdate plugin`, async function() {
+      sandbox.stub(fetchWrapper, 'fetch').resolves(new Response('', {
+        status: 203,
+      }));
+
+      const precacheController = new PrecacheController();
+      const cacheList = [
+        '/will-be-error.html',
+      ];
+      precacheController.addToCacheList(cacheList);
+
+      const plugins = [{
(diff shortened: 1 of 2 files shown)


test/workbox-precaching/sw/test-PrecacheController.mjs
```diff
@@ -656,6 +656,36 @@ describe(`PrecacheController`, function() {
...
+          expect(request).to.exist;
+
+          if (response.status === 203) {
+            return null;
+          }
+          return response;
+        },
+      }];
+
+      // Assuming the async plugin function is properly await-ed, an error
```
Mohammed7390, line 682: 222222222

PR #2296, opened 2019-12-02
Don't overwrite __WB_DISABLE_DEV_LOGS

jeffposnick:
R: @philipwalton
CC: @westonruter

See https://github.com/GoogleChrome/workbox/pull/2284#issuecomment-560470923

Unfortunately, I can't think of a great way to test this, given that by the time our test suite executes, the `logger` IIFE will have already executed.


jeffposnick:
@philipwalton Just wanted to double-check that you're cool with https://github.com/GoogleChrome/workbox/pull/2296/commits/2f2a98882a8a58865c04cea7f5987bce9b4e85b7, which was necessary to make TypeScript happy. (I think it's safe.)

philipwalton:
I think that's fine, though it does look a bit weird. I haven't tested, but I assume you could also do something like this (though either is fine with me):

```
self.__WB_DISABLE_DEV_LOGS = self.__WB_DISABLE_DEV_LOGS || false;
```

PR #2298, opened 2019-12-04
Update NodeJS to 10

codler:
Update to a supported maintained NodeJS version
https://nodejs.org/en/about/releases/
v8 have end of life this year, 2019-12-31.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ cache:
     - ~/.selenium-assistant
 
 node_js:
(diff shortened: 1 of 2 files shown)


jeffposnick:
Thanks!

There are a few other changes that need to be made to the codebase, so I'm going to close this PR in favor of something more comprehensive that targets the new `v6` branch.

(The issue is tracked in #2445 so it won't be overlooked.)

PR #2312, opened 2019-12-20
Update jsdoc comments to use package names

philipwalton:
R: @jeffposnick

I started exploring what it would take to update all our jsdoc comments to reference package names instead of the global namespace, and with a bit of find-and-replace it ended up not being so bad.

And it was good I did because along the way I realized that all our jsdoc comments in the `_types.ts` files were being dropped (since they were ending up in `.d.ts` files instead of `.js` files), so had we just published as is, a lot of links would have broken.

I also removed the deprecation warnings for the strategy functions (since we said we'd remove those in v5).

To see what the docs look like, you can run: `gulp docs --pretty`. I'm still trying to work out if it's possible to customize how the modules are displayed (e.g. I don't like the `module:` prefix in links and params references, nor do I like the `module-` prefix in the filenames, so hopefully we can get rid of them, but I don't think it's a huge deal if we can't.

### DIFF
--- a/packages/workbox-background-sync/src/BackgroundSyncPlugin.ts
+++ b/packages/workbox-background-sync/src/BackgroundSyncPlugin.ts
@@ -14,16 +14,16 @@ import './_version.js';
  * A class implementing the `fetchDidFail` lifecycle callback. This makes it
  * easier to add failed requests to a background sync Queue.
  *
- * @memberof workbox.backgroundSync
+ * @memberof module:workbox-background-sync
  */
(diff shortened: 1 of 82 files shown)


jeffposnick:
Do you want this reviewed/merged, Phil, or are you still working on the PR?

philipwalton:
Sorry, thought this was already merged. I'll see about the failing test and merge once I resolve the build issues.

philipwalton:
OK, tests are passing. @jeffposnick PTAL and feel free to merge if everything looks good.

PR #2315, opened 2020-01-06
New glitch demos

philkrie:
R: @jeffposnick @philipwalton

Moving the previously discussed new Workbox v5 demos that have already been deployed on Glitch and linked to in the [Workbox Modules page](https://developers.google.com/web/tools/workbox/modules).

Next steps would involve github -> glitch pipeline (to remove manual effort) and a minor build process to reduce steps needed to update demos when a new workbox version comes out.


jeffposnick:
Thanks, @philkrie!

There are currently a number of `gulp` tasks related to updating and publishing the demos, but I don't think there's going to carry over to the new demo setup.

If you're cool with it, I'm going to push a commit to this PR that will remove those `gulp` tasks and the related dependencies from `package.json` prior to merging.

philkrie:
Sure thing, makes sense. I'll do pipeline work within this directory and we can then migrate it to the top level gulp pipeline later.

jeffposnick:
(Sorry, we've been focused on getting v5 finished up. It will probably be cleaner if we wait until after the final release.)

philipwalton:
Hey @philkrie, we published v5 today, so now's probably a good time to address this PR. I'm not sure what needs to happen to update the demos on Glitch, but I noticed they're currently pointing to a prerelease version. Is updating them still a manual process?

@jeffposnick were you still planning to update the PR to remove the gulp tasks referencing the build?

philkrie:
@philipwalton updating glitch is still a manual process though there are [discussions/ideas](https://support.glitch.com/t/tutorial-how-to-auto-update-your-project-with-github/8124) in the glitch forums that I was planning on testing out at the end of this week that might allow us to sync between github and glitch automatically.

Updating the CDN URL in the repo is easy enough, but if we want we could use gulp to create a step that utilizes cdn-details.json

philipwalton:
I think that, long term, finding an automated solution is best. But at the moment it's probably easiest to just update them manually, so at least they're current.

Presumably they won't need to change that often (not until v6, unless we add a feature we really want to show off), so I don't think it'll be that much work to keep them up-to-date—as long as everyone on the team has access.

philkrie:
Updated glitch and the repo to 5.0.0. Still no automated process, haven't had the time, but as mentioned, should be enough for a while

jeffposnick:
Thanks! I've gone ahead and merged.

PR #2320, opened 2020-01-09
Update to the build libs JSDoc

jeffposnick:
R: @philipwalton

This updates the JSDocs for the `workbox-build` and `workbox-webpack-plugin` public interfaces to match what's supported in v5.

I am not 100% sure how JSDoc with that many options will look when rendered in our HTML, so it might need to be tweaked further prior to official publication. But... it should be accurate, at least!


jeffposnick:
(This will need to be updated to use `allow`/`deny` following https://github.com/GoogleChrome/workbox/pull/2325)

philipwalton:
Have you run `gulp docs` or `gulp docs --pretty` locally to see the output? I've found that to be a pretty good representations of what the final output will be most of the time.

packages/workbox-build/src/_types.js
```diff
@@ -26,3 +26,81 @@ import './_version.mjs';
  *
  * @memberof module:workbox-build
  */
+
+/**
+ * @typedef {Object} RuntimeCachingEntry
+ *
+ * @property {string|workbox.routing.Route~handlerCallback} handler
```
philipwalton: I'm going to merge https://github.com/GoogleChrome/workbox/pull/2312 in a bit, and once that's done these should reference the new `module:workbox-routing`-style identifiers (here and throughout).

jeffposnick:
Yeah, I have. It's a big table, but I'm not exactly sure how much horizontal width it will end up needing until seeing how it looks staged on d.g.c.

packages/workbox-webpack-plugin/src/generate-sw.js
```diff
@@ -34,12 +34,156 @@ const _generatedAssetNames = new Set();
...
+  // eslint-disable-next-line jsdoc/newline-after-description
   /**
    * Creates an instance of GenerateSW.
    *
-   * @param {Object} [config] See the
-   * [configuration guide](https://developers.google.com/web/tools/workbox/modules/workbox-webpack-plugin#configuration)
-   * for all supported options and defaults.
+   * @param {Object} config The configuration to use.
+   *
+   * @param {Array<workbox-build~ManifestEntry>} [config.additionalManifestEntries]
```
philipwalton: Does this need to be `module:workbox-build` in order for the link/reference to resolve?

packages/workbox-webpack-plugin/src/inject-manifest.js
```diff
@@ -37,12 +37,81 @@ const _generatedAssetNames = new Set();
  * @module workbox-webpack-plugin
  */
 class InjectManifest {
+  // eslint-disable-next-line jsdoc/newline-after-description
   /**
-   * Creates an instance of InjectManifest.
+   * Creates an instance of GenerateSW.
```
philipwalton: `InjectManifest`?

packages/workbox-webpack-plugin/src/inject-manifest.js
```diff
@@ -37,12 +37,81 @@ const _generatedAssetNames = new Set();
...
    *
-   * @param {Object} [config] See the
-   * [configuration guide](https://developers.google.com/web/tools/workbox/modules/workbox-webpack-plugin#configuration)
-   * for all supported options and defaults.
+   * @param {Object} config The configuration to use.
+   *
+   * @param {string} config.swSrc An existing service worker file that will be
+   * compiled and have a precache manifest injected into it.
+   *
+   * @param {Array<workbox-build~ManifestEntry>} [config.additionalManifestEntries]
```
philipwalton: Same comments as above w.r.t `module:workbox-build` (here and throughout this file).

philipwalton:
Right. Now I'm remembering that when I made my jsdoc changes, I modified the webfu gulp tasks to pull in my local Workbox changes rather than from the remote, so I could see what it would actually look like staged. But now I've reverted those changes, so I can't really share them.

Probably worth updating the `gulp workbox` task to support a local option to make this easier in the future.

packages/workbox-build/src/_types.js
jeffposnick: Okay, I can update this PR once that's submitted.

packages/workbox-webpack-plugin/src/generate-sw.js
jeffposnick: Yup.

packages/workbox-streams/README.md
```diff
@@ -1 +1 @@
-This module's documentation can be found at https://developers.google.com/web/tools/workbox/reference-docs/latest/workbox.streams
+This module's documentation can be found at https://developers.google.com/web/tools/workbox/reference-docs/latest/workbox-streams
```
philipwalton: I think this needs to be `module-workbox-streams`.

jeffposnick: I will take your word for that!

philipwalton: It should be the same as how it currently works for `workbox-build`: https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-build

PR #2325, opened 2020-01-13
Switch to allow/denylist

jeffposnick:
R: @philipwalton

Use `allow`/`deny` terminology, with a custom error message for `navigateFallbackWhitelist`/`navigateFallbackBlacklist` build options letting folks know that they should switch.


tenjojeremy:
Please don't forget to update the docs as well 
https://developers.google.com/web/tools/workbox/modules/workbox-webpack-plugin

philipwalton:
The docs changes are complete. They've been happening in this branch https://github.com/google/WebFundamentals/commits/workbox-v5, and are also tracked in this issue: https://github.com/GoogleChrome/workbox/issues/2122

Issue #2321, opened 2020-01-10
workbox-window integration test failures


jeffposnick:
I previously remember seeing these with Safari, but they're popping up with Firefox Stable as well.

See https://travis-ci.org/GoogleChrome/workbox/builds/633899787?utm_source=github_status&utm_medium=notification

```
[Workbox]: Running Integration test on test/workbox-window/integration with NODE_ENV 'dev' and browser 'Firefox Stable'
  [workbox-window]
    ✓ passes all window unit tests (14421ms)
  [workbox-window] Workbox
    register
      ✓ registers a new service worker (49ms)
      ✓ reports all events for a new SW registration (45ms)
      ✓ reports all events for an updated SW registration (70ms)
@http://localhost:3004/test/workbox-broadcast-update/static/integration.html.njk:4:22
@http://localhost:3004/test/workbox-broadcast-update/static/integration.html.njk:30:10
@http://localhost:3004/test/workbox-broadcast-update/static/integration.html.njk:31:8
Unexpected window opened: about:blank
      1) "before each" hook for "reports all events for an external SW registration"
  4 passing (18s)
  1 failing
  1) [workbox-window] Workbox
       "before each" hook for "reports all events for an external SW registration":
     NoSuchWindowError: Browsing context has been discarded
      at Object.throwDecodedError (node_modules/selenium-webdriver/lib/error.js:550:15)
      at parseHttpResponse (node_modules/selenium-webdriver/lib/http.js:560:13)
      at Executor.execute (node_modules/selenium-webdriver/lib/http.js:486:26)
      at <anonymous>
      at process._tickDomainCallback (internal/process/next_tick.js:229:7)
```

PR #2326, opened 2020-01-13
Set revision: null instead of deleting revision

jeffposnick:
R: @philipwalton

This is needed to avoid a warning message due to a missing `revision` field, introduced in one of the earlier RCs.

See https://github.com/GoogleChrome/workbox/issues/2259#issuecomment-573281342


philipwalton:
Looks like there are some legitimate test failures though.

jeffposnick:
While re-running following the latest commit, the failure is now https://github.com/GoogleChrome/workbox/issues/2321

I'm going to merge since the relevant tests work now.

PR #2357, opened 2020-02-07
Updates various devDependencies

jeffposnick:
R: @philipwalton

This is hopefully a no-op from a code perspective; it just updates a bunch of our `devDependencies` up to their latest releases, including a number of major version bumps.

The `eslint` updates necessitated cleaning up some style violations, which was mostly done automatically via `--fix`.


infra/testing/generate-variant-tests.js
```diff
@@ -8,7 +8,7 @@
 
 /**
  * @callback VariantCallback
- * @param {Object} variant
+ * @param {object} variant
```
philipwalton: Hmmm, I'm pretty sure `Object` is the preferred variant here for Google JS style. Presumably upgrading `eslint-plugin-jsdoc` changed the default. Looks like it can be customized via something like:

```js
{"jsdoc":{"preferredTypes":{"object":"Object"}}}
```

https://github.com/gajus/eslint-plugin-jsdoc

I don't know if this will have any real implications, but I guess my preference would be to not have all of the type definitions change.

jeffposnick: Okay, fair enough. Switching back.

PR #2359, opened 2020-02-07
Initial TS linting

jeffposnick:
R: @philipwalton

This adds in TypeScript linting, and sets up a baseline of rules—mainly turning things off that are prevalent in our codebase. I think we'll have to iterate on the exact ruleset.

In terms of the fixes that are implemented in this PR, they fall into a few categories:
- Cosmetic/single character changes.
- Switching from `<Type>` to `as Type` casting.
- Re-arranging some `function` constant definitions so that they're declared before they're being used.

I've added some inline comments to this PR to call a few things out prior to merging.


packages/workbox-background-sync/src/Queue.ts
```diff
@@ -88,7 +108,8 @@ class Queue {
     }
 
     this._name = name;
-    this._onSync = onSync || this.replayRequests;
+    // TODO (philipwalton): Should this be bound?
```
jeffposnick: @philipwalton

packages/workbox-background-sync/src/Queue.ts
```diff
@@ -361,7 +382,8 @@ class Queue {
 
             let syncError;
             try {
-              await this._onSync({queue: this});
+              // TODO (philipwalton): Should the await be here?
```
jeffposnick: @philipwalton

packages/workbox-background-sync/src/lib/QueueStore.ts
```diff
@@ -49,7 +49,8 @@ export class QueueStore {
   constructor(queueName: string) {
     this._queueName = queueName;
     this._db = new DBWrapper(DB_NAME, DB_VERSION, {
-      onupgradeneeded: this._upgradeDb,
+      // TODO (philipwalton): Should this be bound?
+      onupgradeneeded: this._upgradeDb, // eslint-disable-line @typescript-eslint/unbound-method
```
jeffposnick: @philipwalton

packages/workbox-broadcast-update/src/BroadcastCacheUpdate.ts
```diff
@@ -24,11 +24,11 @@ const isSafari = /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
 
 
 // Give TypeScript the correct global.
-declare var self: ServiceWorkerGlobalScope;
+declare let self: ServiceWorkerGlobalScope;
 
 export interface BroadcastCacheUpdateOptions {
   headersToCheck?: string[];
-  generatePayload?: (options: CacheDidUpdateCallbackParam) => Object;
+  generatePayload?: (options: CacheDidUpdateCallbackParam) => Record<string, any>;
```
jeffposnick: I _think_ this is a safe change (it was done automatically via `--fix`), but I guess there's a chance it might be considered a breaking change?

packages/workbox-background-sync/src/Queue.ts
philipwalton: Yes, we have to await here. If we don't—and if the `onSync` function is an async function (which is should be)—then any errors thrown during the execution of that function won't be caught.

We could type that `onSync` returns a promise, but technically it might not since it's defined by the user.

packages/workbox-background-sync/src/lib/QueueStore.ts
philipwalton: Similar comment to above. It feels strange that we need to explicitly bind all these functions. Is this one of the rules enforced by Google TS style?

packages/workbox-broadcast-update/src/BroadcastCacheUpdate.ts
philipwalton: Seems safe to me? I guess the worse case scenario is someone has to update their typings. I'm not sure if that's considered breaking or not.

packages/workbox-background-sync/src/Queue.ts
jeffposnick: Okay, I'll change

```
interface OnSyncCallback {
  (options: OnSyncCallbackOptions): void;
}
```

to

```
interface OnSyncCallback {
  (options: OnSyncCallbackOptions): Promise<void>;
}
```

in that case. 

(Again, really not sure if this needs to be marked as a breaking change?)

packages/workbox-background-sync/src/lib/QueueStore.ts
jeffposnick: I don't think we need to enforce this one. I can disable it globally in our repo if you'd like.

I just wanted to confirm before doing so.

PR #2369, opened 2020-02-14
Various npm dependency bumps

jeffposnick:
R: @philipwalton

No breaking changes in the dependencies that I could see.

We had gotten pretty far behind on the various Rollup plugins in particular, as the "official" ones had all been renamed to `@rollup/plugin-...`.


jeffposnick:
(I had to rename a few files that were still using `require()` instead of `import` from `.mjs` to `.js`, because as of https://github.com/mochajs/mocha/pull/4038, `mocha` requires that any file with the `.mjs` extension to only use `import`.)

philipwalton:
Ahh, yeah and that fits with Node ESM requirements as well.

PR #2371, opened 2020-02-14
Include the hash in our "friendly" URL logging

### LINKED ISSUE #2323: Misleading log messages when resource URL has fragment
**Libraries Affected**:
workbox-routing / workbox-precaching

**Browser & Platform**:
Google Chrome v81.0.4023.0

**Issue**:
I just spent a while debugging some seemingly strange behavior in my project, which uses the `GenerateSW` workbox webpack plugin. I had set up a `runtimeCaching` rule with the URL pattern 
  `/\.(png|jpg|jpeg|svg)$/`; yet many SVG files were not being cached as expected. I turned to the verbose-level log & saw several instances of `No route found for: /file.svg`, which was obviously super jarring - it even happened when I set the URL pattern to the exact file name!
I figured out what was going on only when I set a debugger breakpoint at the line that logs that message - it turns out the real URL that was being matched against had a fragment (i.e `/file.svg#content`), & the log was obscuring this, I believe due to its use of the `getFriendlyURL` function: 
https://github.com/GoogleChrome/workbox/blob/ef5c2ca8fcbad586eccbaf8ab1df6c90afec7541/packages/workbox-routing/src/Router.ts#L202-L209
_(note: I need the fragment because I'm using an svg \<use\> tag.)_

I think URLs are being similarly obscured in workbox-precaching debug logs as well. The solution should be pretty simple - keep URL fragments when logging so the user knows exactly what URL patterns are being tested against.

jeffposnick:
R: @philipwalton

Fixes #2323 by including the `hash` portion of a URL in our log messages.

### DIFF
--- a/packages/workbox-core/src/_private/getFriendlyURL.ts
+++ b/packages/workbox-core/src/_private/getFriendlyURL.ts
@@ -10,10 +10,9 @@ import '../_version.js';
 
 const getFriendlyURL = (url: URL | string) => {
   const urlObj = new URL(String(url), location.href);
-  if (urlObj.origin === location.origin) {
-    return urlObj.pathname + urlObj.hash;
-  }
-  return urlObj.href;
+  // See https://github.com/GoogleChrome/workbox/issues/2323
+  // We want to include everything, except for the origin if it's same-origin.
+  return urlObj.href.replace(new RegExp(`^${location.origin}`), '');
 };
 
 export {getFriendlyURL};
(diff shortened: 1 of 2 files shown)


packages/workbox-core/src/_private/getFriendlyURL.ts
```diff
@@ -11,7 +11,7 @@ import '../_version.js';
 const getFriendlyURL = (url: URL | string) => {
   const urlObj = new URL(String(url), location.href);
   if (urlObj.origin === location.origin) {
-    return urlObj.pathname;
+    return urlObj.pathname + urlObj.hash;
```
philipwalton: I wonder if it makes sense to keep the `search` portion as well? Based on that issue I think someone could run into a similar problem with a stray query param.

jeffposnick: Yeah, good point.

jeffposnick: I'm just stripping the leading origin now, if it's a same-origin URL.

PR #2381, opened 2020-02-21
Fix Deferred type and run lint fix

philipwalton:
R: @jeffposnick

This PR updates the `Deferred` types and then any consumers of that class via `eslint --fix`.

### DIFF
--- a/packages/workbox-core/src/_private/Deferred.ts
+++ b/packages/workbox-core/src/_private/Deferred.ts
@@ -19,9 +19,9 @@ import '../_version.js';
  */
 class Deferred<T> {
   promise: Promise<T>;
-  resolve?: (value?: T) => void;
-  reject?: (reason?: any) => void;
-  
+  resolve!: (value?: T) => void;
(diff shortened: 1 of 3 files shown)


packages/workbox-core/src/_private/Deferred.ts
```diff
@@ -19,9 +19,9 @@ import '../_version.js';
  */
 class Deferred<T> {
   promise: Promise<T>;
-  resolve?: (value?: T) => void;
-  reject?: (reason?: any) => void;
-  
+  resolve!: (value?: T) => void;
```
jeffposnick: Could you explain the use of `!` here?

I don't think I've seen it in a type definition before, only when used as a non-`null`/`undefined` assertion as per https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator

Isn't the default assumption, when you leave out the `?`, that these properties have to be a function that returns `void`?

PR #2412, opened 2020-03-18
Add compileSrc option to InjectManifet webpack plugin

### LINKED ISSUE #2383: Use workbox-webpack-plugin to generate a standalone precache manifest
**Library Affected**:
workbox-webpack-plugin v5.0.0

My goal: I use the micro-front-end architecture in my project, and use navigator.serviceWorker in the main application. Now I need to generate the preaceche manifest file in the sub-application, and then read the main application's InjectManifest manifestTransforms parameter to finally generate the preaceche manifest.
But the question comes, how can I use workbox-webpack-plugin to generate a preaceche manifest JSON file in the sub-app project

jeffposnick:
R: @philipwalton
CC: @bailnl

Fixes #2383

When `compileSrc` is set to `false`, the `swSrc` file won't be run through the `webpack` compilation process. That allows for the use case described in #2383, where the manifest ends up injected into an arbitrary file that might not be something that should be processed by `webpack`.

The default behavior has not changed.

### DIFF
--- a/packages/workbox-build/src/options/defaults.js
+++ b/packages/workbox-build/src/options/defaults.js
@@ -10,6 +10,7 @@ module.exports = {
   babelPresetEnvTargets: ['chrome >= 56'],
   cleanupOutdatedCaches: false,
   clientsClaim: false,
+  compileSrc: true,
   disableDevLogs: false,
   exclude: [
     /\.map$/,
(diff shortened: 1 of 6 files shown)


bailnl:
thanks ~  @jeffposnick

PR #2433, opened 2020-03-31
gulp-glitch first pass

philkrie:
Hey Jeff,

I've edited the code to use await instead of on but wasn't sure how to move away from the usual pipe syntax w/in gulp. This currently works but would like your thoughts on how to revise. I'll also start recording some documentation on how to scale this to other glitch projects.

Phillip


gulp-tasks/publish-glitch.js
```diff
@@ -0,0 +1,62 @@
...
+  Copyright 2018 Google LLC
+
+  Use of this source code is governed by an MIT-style
+  license that can be found in the LICENSE file or at
+  https://opensource.org/licenses/MIT.
+*/
+
+const gulp = require('gulp');
+// const del = require('del');
+const git = require('gulp-git');
```
jeffposnick, line 11: I don't think you should use `gulp-git` here. It is probably making you more tied to the `gulp` way of doing things than you want to be.

There are also some helper libraries like `tempy` and `fs-extra` that we use elsewhere and could help here.

There's an example at https://github.com/GoogleChrome/workbox/commit/89840b22da92a76b09da0d69387160e4f2888255 of just spawning a local process using `async`/`await` (in that case, `yarn`, but you should be able to do it with `git` as well.)

```
const childProcess = require('child_process');
const fse = require('fs-extra');
const tempy = require('tempy');
const util = require('util');

const exec = util.promisify(childProcess.exec);
```

gulp-tasks/publish-glitch.js
```diff
@@ -0,0 +1,62 @@
...
+
+gulp.task('publish-glitch', async () => {
+  for (const project of glitchProjects) {
+    // repo URLs can be associated w/ the demo folder through a predefined object OR th
+    // for each repo...
+    const projectURL = 'https://' + key + '@api.glitch.com/git/' + project;
+    const projectPath = './temp_' + project;
+    const date = new Date();
+
+    await git.clone(projectURL, {args: projectPath}, async () => {
```
jeffposnick, line 41: Assuming you have the promisified `exec` set up based on the earlier comment, you could move from `gulp-git` to

```
const tempDir = await tempy.directory();
await exec(`git clone ${projectURL}`, {cwd: tempDir});
// See https://github.com/jprichardson/node-fs-extra/blob/master/docs/copy.md
await fse.copy(..., ...);
// Etc.
```

philkrie: got it, makes sense. Given that I added axios as well, do you recommend using exec to invoke cURL for the hook request?

philkrie:
Updated w/o gulp piping, um yeah you are totally correct, way easier and clearer.

gulp-tasks/publish-glitch.js
jeffposnick: I'm cool with using `axios` instead of shelling out to `cURL`. I view using `exec` as more of a second-choice when the `node`-native choice (like `gulp-git`) is way too much effort.

jeffposnick: (Or if you already have the code working with `cURL`, as you seem to, that's fine with me.)

philkrie: it should be good to go now. we can review on Tuesday if you'd like. I also have a write up on the full funnel.

jeffposnick:
Great! Glad that approached worked for you.

gulp-tasks/publish-glitch.js
```diff
@@ -0,0 +1,58 @@
...
+const util = require('util');
+const del = require('del');
+
+const exec = util.promisify(childProcess.exec);
+
+const key = '9c5af840-16dd-42d8-b5d1-0fa4a52939b8';
+
+const glitchProjects = [
+  'workbox-core',
+  // 'workbox-sw',
```
jeffposnick, line 22: FYI, you still have these commented out.

gulp-tasks/publish-glitch.js
```diff
@@ -0,0 +1,58 @@
...
+const gulp = require('gulp');
+const childProcess = require('child_process');
+const fse = require('fs-extra');
+const tempy = require('tempy');
+const util = require('util');
+const del = require('del');
+
+const exec = util.promisify(childProcess.exec);
+
+const key = '9c5af840-16dd-42d8-b5d1-0fa4a52939b8';
```
jeffposnick, line 18: This key was going to be moved elsewhere, correct?

gulp-tasks/publish-glitch.js
philkrie: yeah still need to update the server in each glitch project but can finish by end of this week

gulp-tasks/publish-glitch.js
philkrie: whoops yeah that's technically sensitive, we'd want it to be in an env variable. I can roll back the commit to remove the history but how do we remove this snippet from the comment chain?

gulp-tasks/publish-glitch.js
```diff
@@ -0,0 +1,67 @@
...
+  // 'workbox-streams',
+  // 'workbox-range-requests',
+  // 'workbox-broadcast-update-demo',
+  // 'workbox-window',
+  // 'workbox-navigation-preload'
+];
+
+
+gulp.task('publish-glitch', async () => {
+  if (!process.env.GLITCH_PERSONAL_TOKEN) {
```
philkrie, line 37: @jeffposnick I went for the scorched earth strategy and just rolled back to clean up all problematic history, and recommited everything at once. Instead of using fs-extra, I thought it'd make sense to follow the convention used in publish-github.js and use node environment vars. I noticed you don't have dotenv: do you just pass the env vars with the command?

jeffposnick: Okay, that's fine.

I've got some environment vars set in my `~/.bashrc` config.

philkrie: Sounds good. I'll push demo server changes later this week (since I'll be adding or editing server.js code for each demo), make sure this task works when deploying to every demo, and then we can merge

philkrie:
@jeffposnick this latest commit demonstrates for two demos what a top level express server might look like. Since we are launching an express server, we need package.json and updateServer.js at the top level. In addition, the express server has to manage all requests, so by doing this I am forced (unless I'm mistaken) to split updateServer.js into two parts: the top serves index.html, sw.js, and any static files. I then use a comment to indicate that the bottom half is unrelated to the demo. That's where we place the gulp hook listener. Let me know if this seems reasonable.

demos/src/workbox-core/updateServer.js
```diff
@@ -0,0 +1,39 @@
+const express = require('express');
```
jeffposnick, line 1: Is it possible to put the logic in this file into a reusable `demos/src/lib/server.js` file that just gets `require()`ed and run in each individual project's `updateServer.js`?

If it's not possible to refactor things that way (I'm not sure if each Glitch project would have access to code at the `demos/src/lib/` path level), that's okay, but it would help reduce the repetition if you could.

philkrie: there's no common structure that separate glitch projects can access (as far as I'm aware). The only way would be to place every demo into the same project, each in a subdirectory. However, that might reduce clarity for users inexperienced w/ service workers.

To reduce duplication in this git repo, we could use gulp to concat the final file together.

philkrie: @jeffposnick I raised a [thread](https://support.glitch.com/t/is-there-a-way-to-access-code-in-another-glitch-project-from-another-project/22601) about this on the glitch forums. Looks like we could import this boilerplate using package.json, such as:

{
    "name": "external-package-example",
    "dependencies": {
        "backpack": "git://github.com/philkrie/demo_server.git"
    }
}

Sadly we can't target a specific directory w/in a repo, so instead of hosting this within the demos folder in the workbox repo, we might need to create a separate git repo to do this. On the plus side, if we want to expand this to all web.dev demos, we can essentially make this a package that a demo creator can import, making the git listener easier to use. What do you think?

jeffposnick: Eh, I'd rather not spin up a new GitHub repo if we could avoid it. I wouldn't worry about the code duplication in this case, in the interest of getting something finalized. We could always revisit after the initial setup is in place.

Issue #2439, opened 2020-04-05
What is the signature for a custom route handler?


daffinm:
**Library Affected**:
https://developers.google.com/web/tools/workbox/modules/workbox-routing#matching_and_handling_in_routes

**Issue or Feature Request Description**:
The handler callback method in the docs has the following signature:
```javascript
const handlerCb = ({url, event, params})
```
But the handle method in all of the workbox strategies seems to have the following signature;
```javascript
async handle({ event, request }) 
```
I see from the debugger that the object passed to the handle method has the following properties:
```javascript
{
   url: URL, 
   request: Request, 
   event: FetchEvent, 
   params: undefined
}
````
Could these docs be updated? Or could they point to the API docs for Workbox? (Sorry. I am a java programmer struggling to adjust to this crazy javascript stuff.)

jeffposnick:
The relevant TypeScript definitions are at https://github.com/GoogleChrome/workbox/blob/3ac095656354157eb4b58c93d3b1e390a5ba34dd/packages/workbox-core/src/types.ts#L36-L56

Ultimately, that's the intended "source of truth".

I'm not sure which docs you were looking at, but if anything doesn't match that (which is unfortunately possible; our docs are generated from JSDocs, not the TypeScript signature) then it's an issue with the docs that we'll resolve.

https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-routing#~handlerCallback is the "official" docs—is that what you were looking at? If not, what's the URL of the page?

philipwalton:
> But the handle method in all of the workbox strategies seems to have the following signature;
> 
> ```js
> async handle({ event, request }) 
> ```

@daffinm note that, in JavaScript, you generally only list the parameters a given method is using—not all the parameters it's invoked with.

In the case of Workbox strategies, they all only use the `request` and `event` params, but since Workbox's routing interface accepts handlers that may also use `url` and/or `params`, every invocation of a handler function (which all strategy classes `handle()` method is) from a `Router` object needs to be invoked with all four object params, even if the function doesn't end up using them.

philipwalton:
@jeffposnick looking at the docs and TypeScript typings, I think everything is correct (though it is the case that strategy classes don't use the `url` or `params` options, which is OK since they're marked as optional).

daffinm:
[...]

PR #2440, opened 2020-04-06
handlerCallback JSDocs update

jeffposnick:
Prompted by https://github.com/GoogleChrome/workbox/issues/2439, I found a discrepancy in our `handlerCallback` JSDocs and TypeScript definition: https://github.com/GoogleChrome/workbox/blob/3ac095656354157eb4b58c93d3b1e390a5ba34dd/packages/workbox-core/src/types.ts#L36-L56

### DIFF
--- a/packages/workbox-routing/src/_types.ts
+++ b/packages/workbox-routing/src/_types.ts
@@ -68,9 +68,8 @@ import './_version.js';
  *
  * @callback ~handlerCallback
  * @param {Object} context
- * @param {URL} context.url The URL that matched.
- * @param {Request} [context.request] The corresponding request,
- *     if available.
+ * @param {Request|string} context.request The corresponding request.
+ * @param {URL} [context.url] The URL that matched, if available.
  * @param {FetchEvent} [context.event] The corresponding event that triggered
  *     the request, if available.
  * @param {Object} [context.params] Array or Object parameters returned by the


philipwalton:
FWIW, I have a *slight* concern that adding this docs change will signal to developers that they need to handle `request` as a string themselves in their own handler code.

I'd like a rough policy where we _always_ invoke user code with a real `Request` object, but we also make our first-party handlers callable with a string (for developer convenience).

Unfortunately, I don't think we can easily convey this through jsdocs though...

jeffposnick:
We don't have control over whether handler code gets invoked via the `Router` or via some other mechanism, so having the JSDocs accurately reflect all of the possibilities seems like the way to go.

philipwalton:
> We don't have control over whether handler code gets invoked via the Router or via some other mechanism.

If they're invoking their handlers themselves, then they can pass in whatever args they way, but we're currently never calling them with a string.

The only reason we switched this to `Request|string` is because we wanted developers to be able to call _our_ strategy classes from _their_ code in a convenient way.

jeffposnick:
I think we should do what we'd discussed previously and create a sub-type in which `request` is always a `Request`, and document `Route` as taking that sub-type.

Until we do that, we should just document the actual behavior.

PR #2444, opened 2020-04-07
Swap node versions in our CI tests

### LINKED ISSUE #2443: Fix chromedriver issue with our CI tests
```
> chromedriver@80.0.2 install /Users/runner/runners/2.168.0/work/workbox/workbox/node_modules/chromedriver
> node install.js

internal/util.js:214
    throw new errors.TypeError('ERR_INVALID_ARG_TYPE', 'original', 'function');
    ^

TypeError [ERR_INVALID_ARG_TYPE]: The "original" argument must be of type function
    at promisify (internal/util.js:214:11)
    at Object.<anonymous> (/Users/runner/runners/2.168.0/work/workbox/workbox/node_modules/extract-zip/index.js:11:18)
    at Module._compile (module.js:653:30)
    at Object.Module._extensions..js (module.js:664:10)
    at Module.load (module.js:566:32)
    at tryModuleLoad (module.js:506:12)
    at Function.Module._load (module.js:498:3)
    at Module.require (module.js:597:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/Users/runner/runners/2.168.0/work/workbox/workbox/node_modules/chromedriver/install.js:14:20)
```

We currently pull in the latest version of https://www.npmjs.com/package/chromedriver each time we run the macOS CI, instead of pinning to a specific revision. (This is because the version of Chrome we test against changes over time, and the `chromedriver` version is supposed to match the Chrome version.)

It looks like the `chromedriver@80.0.2` install flow is broken. This might be because we're still on node v8 on the macOS CI, and I think downstream dependencies are starting to remove node v8 support.

Workbox v5 still supports node v8, so I want some test coverage still, but I think I can swap our CI config so that our Windows CI uses node v8 and our macOS CI runs a newer node version. Since our Windows CI doesn't install `chromedriver` (it just runs the build tool tests), that might be sufficient.

jeffposnick:
R: @philipwalton

Fixes #2443

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -18,7 +18,7 @@ jobs:
           
     - uses: actions/setup-node@v1
       with:
-        node-version: 13.x
+        node-version: 8.x
 
     - name: Setup
       run: |
@@ -47,7 +47,7 @@ jobs:
           
     - uses: actions/setup-node@v1
       with:
-        node-version: 8
+        node-version: 13.x
 
     - name: Setup
       run: |
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -22,7 +22,8 @@ jobs:
 
     - name: Setup
       run: |
-        npm ci
+        # Use `npm i` to work around https://github.com/npm/cli/issues/558
+        npm i
         gulp build
 
     - run: npx gulp test-node
--- a/package-lock.json
+++ b/package-lock.json
@@ -11273,9 +11273,9 @@
       }
     },
     "jszip": {
-      "version": "3.2.2",
-      "resolved": "https://registry.npmjs.org/jszip/-/jszip-3.2.2.tgz",
-      "integrity": "sha512-NmKajvAFQpbg3taXQXr/ccS2wcucR1AZ+NtyWp2Nq7HHVsXhcJFR8p0Baf32C2yVvBylFWVeKf+WI2AnvlPhpA==",
+      "version": "3.3.0",
+      "resolved": "https://registry.npmjs.org/jszip/-/jszip-3.3.0.tgz",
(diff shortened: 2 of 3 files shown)


jeffposnick:
So in this PR, our macOS CI tests are now failing with

```
[14:41:44] 'test-integration' errored after 11 min
[14:41:44] Error: Server terminated early with status 1
    at /Users/runner/runners/2.168.0/work/workbox/workbox/node_modules/selenium-webdriver/remote/index.js:251:52
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
[14:41:44] 'test' errored after 17 min
```

right at the point where the first Safari test should start, after the Firefox tests complete. Investigating further...

jeffposnick:
I can reproduce the `Error: Server terminated early with status 1` error when testing against Safari while running the test suite for the underlying `selenium-assistant` library. Upgrading to `selenium-webdriver ^4.0.0-alpha.7` in that underlying library didn't help, either.

Still investigating, but it does appear to be more of an issue with the testing infrastructure in general and less with anything we're doing in our tests.

jeffposnick:
Okay, it's due to https://github.com/GoogleChromeLabs/selenium-assistant/blob/92a799db2025b08bc698a9bcf84f61e5d4142638/src/local-browsers/safari.js#L64-L69, which was a workaround for Safari >= 12 that doesn't appear to be necessary anymore in Safari 13.1.

I am going to push out a new release of `selenium-assistant` that changes that logic so that it's not in effect for Safari 13.1, and then hopefully the tests will work moving forward.

jeffposnick:
...now Windows is failing due to https://github.com/npm/cli/issues/558

Fun. There are some workarounds in that thread.

jeffposnick:
Okay! That's better.

PR #2451, opened 2020-04-13
Don't check swSrc for hardcoded injection point in wizard flow

### LINKED ISSUE #2447: The wizard validation procedure not corresponds to the error message
https://github.com/GoogleChrome/workbox/blob/43406c78aa541b4a713181f16dc738c543a8ef22/packages/workbox-cli/src/lib/assert-valid-sw-src.js#L14
In this `assert-valid-sw-src.js`, `/\.precacheAndRoute\(\s*\[\s*\]\s*\)/` is used to check sw file content, while the errors show `self.__WB_MANIFEST` should be contained.

https://github.com/GoogleChrome/workbox/blob/43406c78aa541b4a713181f16dc738c543a8ef22/packages/workbox-cli/src/lib/errors.js#L30

jeffposnick:
R: @philipwalton

Fixes #2447 

Using the `wizard` mode in the `workbox-cli`, there was previously a validation step that would check to see if your `swSrc` file contained the manifest injection point. This check wasn't updated for Workbox v5, so it was still looking for `precacheAndRoute([])`.

There's already runtime validation that will check for the proper injection point when you run `workbox injectManifest` after you complete the `wizard` step. That runtime validation will work properly if you happen to override the `injectionPoint` configuration option, which this previous  hardcode assertion didn't support.

So I'm just dropping the assertion from the `wizard` flow entirely.


linonetwo:
so we can not use workbox now...Please release a verson!

jeffposnick:
In the meantime, you can also work around this issue by just adding in

```
// precacheAndRoute([])
```

commented out in your `swSrc` file.

PR #2452, opened 2020-04-13
Fixes unhandled rejections in Windows tests

jeffposnick:
R: @philipwalton

I noticed while looking at our CI output that there are some test cases in the Windows suite that should have been failed, but passed because of a missing `await`:

https://github.com/GoogleChrome/workbox/pull/2451/checks?check_run_id=583425178#step:6:60

This PR adds that `await` to ensure that an exception in the `async` sub-function properly fails our main test case. It also fixes the logic so that the Windows tests should pass.

(They were failing because a Windows path separator isn't meant to be used in a `globby` pattern.)

### DIFF
--- a/infra/testing/confirm-directory-contains.js
+++ b/infra/testing/confirm-directory-contains.js
@@ -8,11 +8,11 @@
 
 const expect = require('chai').expect;
 const globby = require('globby');
-const path = require('path');
+const upath = require('upath');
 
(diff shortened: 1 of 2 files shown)


jeffposnick:
That test looks good now: https://github.com/GoogleChrome/workbox/pull/2452/checks?check_run_id=583544083#step:6:60

Issue #2415, opened 2020-03-22
workbox-build is droping V8 support in 5.1.1


Mister-Hope:
workbox-build@5.1.1 requires `strip-comments^2.0.1`

while it requires Node >= 10

https://www.npmjs.com/package/strip-comments

> workbox-build
> The generateSWString mode has been removed. We expect the impact of this to be minimal, as it was primarily used internally by workbox-webpack-plugin.
> 
> The minimum required version of node has been increased to v8.0.0. This also applies to build tools that use workbox-build, like workbox-cli and workbox-webpack-plugin.

This issue means all the CI process which requires `workbox-build` V5 fails on node 8. 😢 I hope you can fix it.

jeffposnick:
Sorry, that was definitely inadvertent, and is kind of embarrassing since I think this is the second time this has happened with `strip-comments`.

We do have CI tests set up that use Node v8, but I guess it's not exercising that code path.

We'll cut a new release that rolls back `strip-comments` and ensure that the code coverage for our CI tests will catch this moving forward.

jeffposnick:
So actually, could you tell me how this failure manifested itself? I just tried running the `workbox-build` code that should use `strip-comments` in Node v8, and it ran successfully.

I know that `strip-comments` has `node >= 10` listed in its `engines` metadata, but how were you able to trigger a failure in Node v8?

Again, we'll bump things back in the next release anyway, but I would like to come up with a "failing" test to add to our test suite that will hopefully catch this in the future.

Mister-Hope:
> 11:03:41 AM: error strip-comments@2.0.1: The engine "node" is incompatible with this module. Expected version ">=10".

https://app.netlify.com/sites/vuepress/deploys/5e77172e94c6e90008aee2c2

Related to https://github.com/vuejs/vuepress/pull/2230

jeffposnick:
Okay, so it looks like `yarn` is stricter about the `engines` field than `npm` is.

(It looks like [`yarn install --ignore-engines`](https://classic.yarnpkg.com/en/docs/cli/install/#toc-yarn-install-ignore-engines) would get this to work, but again, we're going to roll this back to an earlier version of `strip-comments`.)

jeffposnick:
It looks like adding in https://docs.npmjs.com/misc/config#engine-strict to our CI's `npm` installation might give us the same strict checking behavior.

jeffposnick:
[...]

PR #2459, opened 2020-04-16
Update precaching to use routing and strategies

### LINKED ISSUE #1760: Support for setting fetchOptions in workbox-precaching
**Library Affected**:
*workbox-precaching.dev.js, workbox-precaching.prod.js.*

**Browser & Platform**:
*All browser*

*Provide an option to explicitly pass fetchOptions while adding precache route .*

Provide an option to pass fetchOptions while adding precache route either via workbox.precaching.PrecacheController, workbox.precaching.precacheAndRoute or workbox.precaching.addRoute

This is helpful in applications where custom response is served to user based on browser compatibility or user state. This could be easily achieved in custom request handler but for precaching no option is there.

By looking at source, it is observed that workbox precaching in background uses fetch calls for each asset or page listed to be precached.

_cacheEntryInTemp in class PrecacheController (file PrecacheController.mjs) Line is referenced
https://github.com/GoogleChrome/workbox/blob/b146d8d3c2db89fa51afc322049b9795894bb4f2/packages/workbox-precaching/controllers/PrecacheController.mjs#L268 fetchOptions is hardcoded as null.

I couldn't find any way to override method _cacheEntryInTemp in PrecacheController class, so to append custom header on precache fetch calls, As the temporary workaround, I've copied entire workbox and started serving it from my server by modifying _cacheEntryInTemp in PrecacheController as below:

    _cacheEntryInTemp({ precacheEntry, event, plugins }) {
      var _this3 = this;

      return babelHelpers.asyncToGenerator(function* () {
		var myHeader = new Headers(precacheEntry._networkRequest.headers);
		myHeader.append(myHeaderK, myHeaderV);
        let response = yield fetchWrapper_mjs.fetchWrapper.fetch({
          request: precacheEntry._networkRequest,
          event,
          fetchOptions: {
				  credentials: 'include',
				  headers: myHeader,
				},
          plugins
        });

        if (response.redirected) {
          response = yield cleanRedirect(response);
        }

        yield cacheWrapper_mjs.cacheWrapper.put({
          cacheName: _this3._getTempCacheName(),
          request: precacheEntry._cacheRequest,
          response,
          event,
          plugins
        });

        yield _this3._precacheDetailsModel._addEntry(precacheEntry);

        return true;
      })();
    }

before initializing workbox i set myHeaderK (key), myHeaderV (value) for the application based on user browser and other parameters. This is just an example on how this could be beneficial for applications requiring explicit control over dynamic serving of precache pages based on custom parameters.

Therefore, please check if fetchOptions could be added as args in precache in next release.

### LINKED ISSUE #1767: Allow specifying specific strategies for precaching
**Library Affected**:
workbox-precaching

**Browser & Platform**:
All

By default precaching uses a bespoke cache-first strategy. This is the right strategy most of the time, but not all the time. See discussion here: https://twitter.com/jeffposnick/status/1067081651803619328.

It is possible to work around this right now, but only by creating routes from scratch and configuring them to use the precache cache directly, and doing so can silently cause problems and break things with some strategies.

It would be useful to be able to pass a strategy into precache route setup directly. With that in place, workbox could:
* Allow new combinations of existing strategies and precaching options
* Allow custom strategies and other advanced fine tuning of precaching
* Catch & warn users directly when trying to set up known bad combinations of strategies (e.g. stale-while-revalidate of precached revisioned resources)

### LINKED ISSUE #1857: workbox-precaching routes are not compatible with other packages
**Library Affected**:
*workbox-precaching*

The `addRoute()` (and by extension `precacheAndRoute()`) method in `workbox-precaching` adds its own fetch event listener + cache-first response logic rather than depending on the `Router` and `Route` modules from `workbox-routing` and the `CacheFirst` module from `workbox-strategies`.

This is not ideal for a few reasons:

- Code is duplicated and needs to be tested twice
- Logging for precached routes is different from logging for all other routes
- Precache route logic (the [configuration options](https://developers.google.com/web/tools/workbox/reference-docs/latest/workbox.precaching#.addRoute) that `addRoute()` and `precacheAndRoute()` offer) is not available to runtime routes.
- Precache route handlers are not exposed to the developer, so they can't be reused with `workbox-streams` code (instead, developers have to create their own `CacheFirst` strategy instances anyway).

I'm sure there's historical reasons for (from the sw-precache/sw-toolbox merge), but I think we should explore whether it would be possible to consolidate these features in the future.

IMO, ideally, `workbox-precaching` would depend on `workbox-routing` and `workbox-strategies`, and the configuration options for customizing preaching routes could be added to the `workbox-routing` package and thus available to all routes.

philipwalton:
R: @jeffposnick

This PR fixes #1857 and fixes #1760 and fixes #1767 and is a slight rewrite of the main `workbox-precaching` logic to leverage the changes added in #2415.

These changes don't yet address the use case of passing your own strategy or fetch/match options because I wanted to add the initial set of changes without modifying the API (and changing the tests as little as possible) in order to ensure we're not losing any functionality as part of this change.

Once this is merged I can put together a proposal for how developers could use their own custom strategy for precaching.


test/workbox-precaching/sw/utils/test-addFetchListener.mjs
```diff
@@ -7,16 +7,25 @@
...
-import {addFetchListener} from 'workbox-precaching/utils/addFetchListener.mjs';
+import {addRoute} from 'workbox-precaching/addRoute.mjs';
 import {precache} from 'workbox-precaching/precache.mjs';
-import {getOrCreatePrecacheController} from 'workbox-precaching/utils/getOrCreatePrecacheController.mjs';
+import {resetDefaultPrecacheController} from '../resetDefaultPrecacheController.mjs';
 import {dispatchAndWaitForResponse} from '../../../../infra/testing/helpers/extendable-event-utils.mjs';
 
-describe(`addFetchListener()`, function() {
+
+// TODO(philipwalton): move these tests into the `../test-addRoute.mjs` after
```
philipwalton, line 16: cc: @jeffposnick FYI, to make the diff easier to view for this PR, I kept these tests in this file. Once you're comfortable with the changes, I can file a separate PR to move them into the `test-addRoute` file.

packages/workbox-precaching/src/PrecacheController.ts
```diff
@@ -356,6 +426,36 @@ class PrecacheController {
...
+   * @param {Object} [options] See
+   * [addRoute() options]{@link module:workbox-precaching.PrecacheController#addRoute}.
+   */
+  createMatchCallback(options?: FetchListenerOptions) {
+    return ({request}: RouteMatchCallbackOptions) => {
+      const urlsToCacheKeys = this.getURLsToCacheKeys();
+      for (const possibleURL of generateURLVariations(request.url, options)) {
+        const cacheKey = urlsToCacheKeys.get(possibleURL);
+        if (cacheKey) {
+          return {cacheKey};
```
philipwalton, line 448: cc: @jeffposnick not sure how you feel about passing the `cacheKey` from the match callback to the handler callback. Originally I assume this logic would live in a plugin, but since it needs to be run in the match callback, I figure it's better to not have to run it twice. Let me know what you think.

jeffposnick: That sounds like it's in keeping with the intended purpose of `matchCallback` return values, so I'm 👍

packages/workbox-precaching/src/utils/PrecacheStrategy.ts
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  async _handleFetch(request: Request, handler: StrategyHandler) {
+    let response = await handler.cacheMatch(request);
+
+    if (!response) {
+      // Fall back to the network if we don't have a cached response
+      // (perhaps due to manual cache cleanup).
+      if (handler.params &&
+          handler.params.fallbackToNetwork === false) {
```
jeffposnick, line 52: When would this code get called with `handler.params.fallbackToNetwork === false`?

Also, we can start introducing the optional chaining operator into the codebase, since we run everything though `preset-env` anyway.

packages/workbox-precaching/src/utils/PrecacheStrategy.ts
```diff
@@ -0,0 +1,138 @@
...
+   *
+   * This method indicates whether the default cacheable response logic (i.e.
+   * <400, including opaque responses) should be used. If a custom plugin
+   * with a `cacheWillUpdate` callback is passed, then the strategy should
+   * defer to that plugin's logic.
+   *
+   * @private
+   */
+  _usesCustomCacheableResponseLogic(): boolean {
+    for (const plugin of this.plugins) {
```
jeffposnick, line 128: Optional, but I prefer this logic with `some()`:

```
return this.plugins.some((plugin) => plugin.cacheWillUpdate &&
    plugin.cacheWillUpdate !== copyRedirectedResponses);
```

packages/workbox-precaching/src/utils/PrecacheStrategy.ts
```diff
@@ -0,0 +1,138 @@
...
+import {WorkboxError} from 'workbox-core/_private/WorkboxError.js';
+import {WorkboxPluginCallbackParam} from 'workbox-core/types.js';
+import {Strategy, StrategyOptions} from 'workbox-strategies/Strategy.js';
+import {StrategyHandler} from 'workbox-strategies/StrategyHandler.js';
+
+import '../_version.js';
+
+
+async function copyRedirectedResponses({response}: WorkboxPluginCallbackParam['cacheWillUpdate']) {
+  return response.redirected ? await copyResponse(response) : response;
```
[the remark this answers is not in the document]
philipwalton: Let's do it in a separate PR. I filed https://github.com/GoogleChrome/workbox/issues/2468.

packages/workbox-precaching/src/utils/PrecacheStrategy.ts
philipwalton: > When would this code get called with `handler.params.fallbackToNetwork === false`?

That happens here: https://github.com/GoogleChrome/workbox/blob/e4b8c673ec172767592aca46ac1683ab38184666/packages/workbox-precaching/src/PrecacheController.ts#L479.

> Also, we can start introducing the optional chaining operator into the codebase, since we run everything though `preset-env` anyway.

I'm not sure about that. We do use babel for our production builds, but for anyone who uses their own bundler, they'd just get our source code. We could change that, but unless we do, I don't want us to use JS features that require specific transpilation targets that would be easy for developers to miss.

packages/workbox-precaching/src/utils/PrecacheStrategy.ts
philipwalton: Hmmm, this comment actually got me wondering if it'd make more sense to be checking for the presence of a plugin rather than checking for the presence of a callback. I'll update to do that.

philipwalton: Updated. I change the logic slightly to be comparing against a plugin instead of the callback.

packages/workbox-precaching/src/utils/PrecacheStrategy.ts
jeffposnick: Okay, no optional chaining then.

PR #2463, opened 2020-04-21
Per-method setDefaultHandler()

### LINKED ISSUE #2434: Associate setDefaultHandler() calls with a HTTP method
**Library Affected**:
*workbox-routing*

`registerRoute()` takes a third parameter for HTTP method, defaulting to `'GET'`.

`setDefaultHandler()` doesn't take an HTTP method as a parameter, meaning that it ends up being the default for _all_ HTTP requests that are not explicitly matched by another route, regardless of method.

I think we should change the behavior of `setDefaultRoute()` so that it also takes an HTTP method parameter, and change the default so that it only applies to `'GET'`. It's fairly unlikely that there's a `handleCallback` that you actually want to use for all HTTP methods.

This comes up from time to time as a source of confusion, and might be a good thing to plan for v6, when we have more of an opportunity to make some breaking changes.

jeffposnick:
R: @philipwalton

Fixes #2434

The default method is `GET` and that's likely to be the most widely used method, but this is a breaking change since moving forward, a call to `setDefaultHandler()` won't cause `POST`, `PUT`, etc. requests to also match.

### DIFF
--- a/packages/workbox-routing/src/Router.ts
+++ b/packages/workbox-routing/src/Router.ts
@@ -7,18 +7,17 @@
 */
 
 import {assert} from 'workbox-core/_private/assert.js';
-import {logger} from 'workbox-core/_private/logger.js';
-import {WorkboxError} from 'workbox-core/_private/WorkboxError.js';
 import {getFriendlyURL} from 'workbox-core/_private/getFriendlyURL.js';
-import {Route} from './Route.js';
-import {HTTPMethod} from './utils/constants.js';
-import {normalizeHandler} from './utils/normalizeHandler.js';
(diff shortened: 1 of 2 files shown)


packages/workbox-routing/src/Router.ts
```diff
@@ -47,14 +46,15 @@ interface CacheURLsMessageData {
  */
 class Router {
   private readonly _routes: Map<HTTPMethod, Route[]>;
-  private _defaultHandler?: HandlerObject;
+  private readonly _defaultHandler: Map<HTTPMethod, HandlerObject>;
```
philipwalton: Naming nit: can we call this `_defaultHandlerMap` (or something like that) since it's no longer a handler type.

Issue #2479, opened 2020-04-29
Upgrade "stuck" devDependencies


jeffposnick:
This is a meta-issue to track some `devDependecies` in our top-level `package.json` that I can't find a way to upgrade cleanly.

- [x] `@octokit/rest`: The API changed fairly significantly, and the last time I tried updating it, our GitHub publish process broke.
- [ ] `service-worker-mock`: There are additional mocks that ended up breaking our unit tests.
- [x] `rollup`: When I tried updating it, our "hack" to include the versioning info in the final bundles broke.

philipwalton:
> * `rollup`: When I tried updating it, our "hack" to include the versioning info in the final bundles broke.

My guess is this is because we mark our modules as "pure", which technically the version modules are not.

tomayac:
Hi there,

Workbox is [moving to a new engineering team within Google](https://github.com/GoogleChrome/workbox?tab=readme-ov-file#maintenance-update:~:text=Maintenance%20update,owners%20of%20Workbox.). As part of this move, we're declaring a partial bug bankruptcy to allow the new team to start fresh. We realize this isn't optimal, but realistically, this is the only way we see it working. For transparency, here're the criteria we applied:

- For PRs, we closed [everything that doesn't target the current `v7` branch](https://github.com/GoogleChrome/workbox/pulls?q=is%3Apr+is%3Aclosed+-base%3Av7).
- For Issues, we closed [everything that was created before 2024](https://github.com/GoogleChrome/workbox/issues?q=is%3Aissue+is%3Aclosed+created%3A%3C2024-01-01+). 

Thanks, and we hope for your understanding!
The Workbox team

PR #2480, opened 2020-04-29
Update various dependencies and devDependencies

jeffposnick:
This updates a number of our top-level project `devDependencies`, and removes some build dependencies that are no longer needed now that we've bumped our minimum supported version of Node.

There are a few `devDependencies` that I can't cleanly update, and I opened https://github.com/GoogleChrome/workbox/issues/2479 to track those.

NOTE: I accidentally pushed a few build-tool `dependency` updates to this branch as well, but that should be fine. It's all targeted for `v6` anyway.


gulp-tasks/utils/build-node-package.js
```diff
@@ -35,12 +35,5 @@ module.exports = (packagePath) => {
         },
       }],
     ],
-    plugins: [
-      // This ensures that the helpers used by Babel to accompany transpilation
-      // are only included in our Rollup bundles once, even if they're used
-      // in multiple source files.
-      // See https://github.com/rollup/rollup-plugin-babel#helpers
-      '@babel/transform-runtime',
```
philipwalton, line 43: Not needed anymore?

philipwalton (approved):

gulp-tasks/utils/build-node-package.js
jeffposnick: It shouldn't be. We target Node 10 now, and nothing in our code base ends up being transpiled into something that requires the Babel runtime.

(As far as I can tell!)

PR #2484, opened 2020-04-30
Don't allow copyResponse() on cross-origin redirections

### LINKED ISSUE #2468: The copyResponse() method should be limited to same-origin requests
The `Response` object in JavaScript has several [built-in security protections](https://bugs.chromium.org/p/chromium/issues/detail?id=669363&desc=2#c1) that we're kinda throwing out the window by exposing our [`copyResponse()`](https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-core#.copyResponse) helper object.

This method was originally intended to allow Workbox precaching to be able to work with redirected responses, but it probably makes sense to limit this ability to same-origin responses only.

jeffposnick:
R: @philipwalton

Fixes #2468

The most immediate impact of this PR is that if `workbox-precaching` is used with a URL that results in a redirected response, and the URL of that redirected response isn't for the same-origin as the current web app, then precaching will fail.

(Note that cross-origin URLs are still allowed when precaching; this only impacts _redirected_ cross-origin URLs.)

### DIFF
--- a/packages/workbox-core/src/copyResponse.ts
+++ b/packages/workbox-core/src/copyResponse.ts
@@ -7,8 +7,9 @@
 */
 
 import {canConstructResponseFromBodyStream} from './_private/canConstructResponseFromBodyStream.js';
-import './_version.js';
+import {WorkboxError} from './_private/WorkboxError.js';
 
+import './_version.js';
 
 /**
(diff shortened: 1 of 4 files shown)


philipwalton (approved):
Thanks!

PR #2489, opened 2020-05-01
messageSkipWaiting() method for workbox-window

### LINKED ISSUE #2394: [workbox-window] New messageSkipWaiting() method
There's a subtle way that the "[Offer a page reload for users](https://developers.google.com/web/tools/workbox/guides/advanced-recipes#offer_a_page_reload_for_users)" recipe can go wrong, in that the

```
wb.messageSW({type: 'SKIP_WAITING'});
```

could, under some circumstances, end up sending that message to the already-`active` service worker, not the service worker that's currently in the `waiting` state.

After discussing with @philipwalton, we're going to keep the `messageSW()` behavior the same (it will always call `postMessage()` on the service worker instance associated with the `Workbox` object), and instead come up with a new method—my proposed name is `messageSkipWaiting()`—that will be equivalent to using [`messageSW(sw, {type: 'SKIP_WAITING'})`](https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-window#.messageSW) with the `sw` parameter set to whatever the currently `waiting` service worker is.

We can then update the recipe to read

```
wb.messageSkipWaiting();
```

and it should behave as intended.

### LINKED ISSUE #2470: Broken skip waiting process when service worker gets updated multiple times
**Library Affected**:
*workbox-window*

**Browser & Platform**:
*all browsers*

I'm basically following this recipe: https://developers.google.com/web/tools/workbox/guides/advanced-recipes#offer_a_page_reload_for_users
The prompt that is offered to the user is not blocking, so the user can keep working without updating immediately.

The problem I encounter is that after the service worker has been updated 2 times without the user reloading my app, I'm not receiving an event when the service worker is updated for the 3rd time, and hence I don't know where to send the skip waiting event to.

Scenario:
1. App is running, only open in 1 tab
2. Service worker is updated
App receives the waiting event and handle to new SW
App displays non blocking reload prompt
User does not reload
3. Service worker is updated again
App receives the externalwaiting event and handle to new SW
App is still displaying non blocking reload prompt
User does not reload
4. Service worker is updated again
App does **not** receive waiting or externalwaiting event, and of course **not** the handle to the new SW
App is still displaying non blocking reload prompt
5. User tries to reload through prompt
Since the app does not have the handle for the most recent SW, nothing happens

The change from waiting to externalwaiting event is caused by _updateFoundCount.
Not a problem, but it doesn't make sense to me, looks like the code only ever expects 1 waiting update.
https://github.com/GoogleChrome/workbox/blob/master/packages/workbox-window/src/Workbox.ts#L347

The reason for the missing externalwaiting event on the 3rd update is that the updatefound event listener is removed when the first "external" event is encountered:
https://github.com/GoogleChrome/workbox/blob/master/packages/workbox-window/src/Workbox.ts#L363
Commenting out that line fixed my problem.

jeffposnick:
R: @philipwalton

Fixes #2394 and fixes #2470.

I decided to target this for the v6 branch.

### DIFF
--- a/infra/testing/server/templates/sw-skip-waiting-on-message.js.njk
+++ b/infra/testing/server/templates/sw-skip-waiting-on-message.js.njk
@@ -8,7 +8,7 @@
 
 // {{ version }}
 
-addEventListener('message', async (event) => {
+addEventListener('message', (event) => {
   if (event.data.type === 'SKIP_WAITING') {
     skipWaiting();
   }
--- a/packages/workbox-window/src/Workbox.ts
+++ b/packages/workbox-window/src/Workbox.ts
@@ -27,6 +27,10 @@ const WAITING_TIMEOUT_DURATION = 200;
 // that the registration didn't trigger an update.
 const REGISTRATION_TIMEOUT_DURATION = 60000;
 
+// The de facto standard message that a service worker should be listening for
+// to trigger a call to skipWaiting().
+const SKIP_WAITING_MESSAGE = {type: 'SKIP_WAITING'};
+
 /**
  * A class to aid in handling service worker registration, updates, and
  * reacting to service worker lifecycle events.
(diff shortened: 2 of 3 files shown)


jeffposnick:
Ping, @philipwalton

philipwalton (approved):

Issue #2487, opened 2020-05-01
Pass a new sameOrigin param to matchCallback


jeffposnick:
In the course of #2486 I realized there are a lot of places where I end up writing `url.origin === self.location.origin && ...`.

We could simplify this use case by passing in an additional parameter to the `matchCallback` function, `sameOrigin`, that's set to the result of evaluating `url.origin === self.location.origin` inside of the `Router`.

dman777:
I am having a similar issue. 

The front end web app is `https://bar-1.dev.foo.io/` but api calls(which I am trying to cache) are made to `https://bar-api-1.dev.foo.io/v1/bedlist_providers?size=16`. 

The service worker runs with no complaints. It's just not caching the api as it should:

```
registerRoute(
  /v1\/bedlist_providers\?size=/,
  new StaleWhileRevalidate({
    cacheName: 'providers-list'
  }),
);

registerRoute(
  /v1\/providers\/\d*\/bedlist\?size=/,
  new StaleWhileRevalidate({
    cacheName: 'bedlists'
  }),
```

I believe it is because the api route origin is different than what it is registered to. Any ideas what I can do?

jeffposnick:
You can use a [`matchCallback`](https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-routing#~matchCallback) as the first parameter to `registerRoute()` instead of a RegExp. That would probably end up being a lot cleaner for your use case.

```
registerRoute(
  ({url}) => url.origin === 'https://bar-api-1.dev.foo.io' &&
             url.pathname === '/v1/bedlist_providers',
  new StaleWhileRevalidate({
    cacheName: 'providers-list'
  }),
);
```

PR #2505, opened 2020-05-08
sameOrigin param for matchCallback

jeffposnick:
R: @philipwalton 

Fixes https://github.com/GoogleChrome/workbox/issues/2487

### DIFF
--- a/packages/workbox-routing/src/Router.ts
+++ b/packages/workbox-routing/src/Router.ts
@@ -13,7 +13,8 @@ import {getFriendlyURL} from 'workbox-core/_private/getFriendlyURL.js';
 import {Route} from './Route.js';
 import {HTTPMethod} from './utils/constants.js';
 import {normalizeHandler} from './utils/normalizeHandler.js';
-import {Handler, HandlerObject, HandlerCallbackOptions} from './_types.js';
+import {Handler, HandlerObject, HandlerCallbackOptions,
+  MatchCallbackOptions} from './_types.js';
 import './_version.js';
 
 
@@ -171,7 +172,13 @@ class Router {
       return;
     }
 
-    const {params, route} = this.findMatchingRoute({url, request, event});
+    const sameOrigin = url.origin === location.origin;
+    const {params, route} = this.findMatchingRoute({
+      event,
+      request,
+      sameOrigin,
+      url,
+    });
     let handler = route && route.handler;
 
     const debugMessages = [];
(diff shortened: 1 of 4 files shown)


packages/workbox-routing/src/Router.ts
```diff
@@ -264,30 +271,12 @@ class Router {
...
-      assert!.isInstance(request, Request, {
-        moduleName: 'workbox-routing',
-        className: 'Router',
-        funcName: 'findMatchingRoute',
-        paramName: 'options.request',
-      });
-    }
-
+  findMatchingRoute({url, sameOrigin, request, event}: MatchCallbackOptions):
+      {route?: Route; params?: HandlerCallbackOptions['params']} {
```
jeffposnick, line 275: I think we agreed that the dev-only `assert!.isInstance()`-style checks for parameters was less useful now that folks are migrating towards TypeScript, so instead of adding a check here for `sameOrigin`, I remove the existing checks.

packages/workbox-routing/src/_types.ts
```diff
@@ -47,10 +48,11 @@ import './_version.js';
  * @callback ~matchCallback
  * @param {Object} context
  * @param {URL} context.url The request's URL.
- * @param {Request} [context.request] The corresponding request,
- *     if available.
+ * @param {boolean} context.sameOrigin The result of comparing `url.origin`
+ * against the current origin.
+ * @param {Request} context.request The corresponding request.
```
jeffposnick, line 53: As per the TypeScript typing (and the code), `request` is always going to be set. I figured I'd fix that while I was here.

packages/workbox-routing/src/Router.ts
philipwalton: SGTM

packages/workbox-routing/src/_types.ts
```diff
@@ -47,10 +48,11 @@ import './_version.js';
...
  * @param {FetchEvent} [context.event] The corresponding event that triggered
- *     the request, if available.
+ * the request, if available.
```
philipwalton, line 55: BTW, I indent these 4 spaces because that's what's recommended by the Google JS style (go/js then search for `@param`).

packages/workbox-routing/src/Router.ts
philipwalton: SGTM

packages/workbox-routing/src/_types.ts
philipwalton: I don't have a preference for which we do, just explaining why they're there (ESLint doesn't have an option to enforce).

philipwalton (approved):

PR #2522, opened 2020-05-24
fix(build): migrate to the main lodash package

merceyz:
**What's the problem this PR addresses?**

The per method packages for lodash has been deprecated and the use of them is discouraged.
See https://lodash.com/per-method-packages for more

**How did you fix it?**

Replaced the per method packages with lodash and required the method directly


jeffposnick:
Thanks! I'm going change this to target the `v6` branch instead, as that's likely to be our next release.

jeffposnick:
Actually, that ends up being pretty messy when targeted against `v6`. If you don't mind, I can just close this PR and create a similar one myself using `v6` as the base—or alternatively, if you'd prefer, you could do the same.

Just let me know.

merceyz:
I just rebased the PR instead :)

jeffposnick:
Thank you!

PR #2529, opened 2020-06-04
canConstructResponseFromBodyStream returns false

aliai:
Looks like `supportStatus` is always false no matter what.

Do i need to open an issue for this small fix? R: @jeffposnick @philipwalton

### DIFF
--- a/packages/workbox-core/src/_private/canConstructResponseFromBodyStream.ts
+++ b/packages/workbox-core/src/_private/canConstructResponseFromBodyStream.ts
@@ -31,8 +31,9 @@ function canConstructResponseFromBodyStream(): boolean {
       } catch (error) {
         supportStatus = false;
       }
+    } else {
+      supportStatus = false;
     }
-    supportStatus = false;
   }
 
   return supportStatus;


aliai:
@googlebot I signed it!

philipwalton:
@aliai can you rebase your changes against the v6 branch? It looks like your base branch is an older commit and that's messing up the diff.

jeffposnick:
Sorry, that was my fault for changing the base branch. I can rebase if the code change looks good to you, @philipwalton. (I think it's code you touched last.)

jeffposnick:
(The change looks fine to me, but I wanted to double-check.)

philipwalton:
The actual change commit (https://github.com/GoogleChrome/workbox/pull/2529/commits/8fd36eb7503f8c41d978c9e3dd2e80f5742141f1) looks fine to me, but we definitely need to make sure that's the only commit we merge.

jeffposnick:
Okay, so this is interesting.

It turns out that when you construct a new `Response` using a `Blob` as the `BodyInit`, the `Content-Type` header associated with the `Blob` is carried over to the new `Response`. If you construct a new `Response` using a `ReadableStream` as the `BodyInit`, then the `Content-Type` header isn't carried over.

This is alluded to in our tests, which now fail:

https://github.com/GoogleChrome/workbox/blob/8ffe56090f712e9b8bb6cff14e25f7ec090a1ce4/test/workbox-core/sw/test-copyResponse.mjs#L75-L76

(And this wasn't an issue with the tests previously, since the logic changed in this PR which always returned `false` meant that a `Blob` would always be used as the `BodyInit`.)

Given this inconsistency, and given that using a `Blob` a) works in all browsers and b) is the behavior that everyone has had up until now, I'm now thinking that we should just ditch the `canConstructResponseFromBodyStream` logic entirely and always use a `Blob` as the `BodyInit`. That seems preferable than potentially having different `Content-Type` logic in Chrome vs. other browsers.

CC: @philipwalton

aliai:
Hey @philipwalton, sorry for the delay. I rebased it on v6 now :)

aliai:
@jeffposnick looks like your solution is better. If decided, feel free to close my PR :)

jeffposnick:
I'm going to close this and open a new PR as discussed in https://github.com/GoogleChrome/workbox/pull/2529#issuecomment-646075682

Thanks for bringing this issue to our attention, @aliai!

PR #2532, opened 2020-06-09
Ignore fbclid by default when matching precached URLs

### LINKED ISSUE #2417: Add `fbclid` to default-ignored URL caching parameters?
Per https://developers.google.com/web/tools/workbox/modules/workbox-precaching :

> By default, search parameters starting with utm_ are removed, meaning that a request for /about.html?utm_campaign=abcd will be fulfilled with a precached entry for /about.html.

I previously had issues configuring `ignoreURLParametersMatching` with https://github.com/mischnic/parcel-plugin-sw-cache (the main issue being "I couldn't get it to work at all"), so I haven't been using it. That's mostly fine because the app I'm using it for can work without adding params yet.

However, Facebook forcibly appends a `fbclid` param, which prevents the app from loading without a hard-refresh. :-/

jeffposnick:
R: @philipwalton

Fixes #2417

This adds `fbclid` to the default list of URL query parameters that should be ignored when matching against precached URLs.

### DIFF
--- a/test/workbox-precaching/integration/test-precache-and-update.js
+++ b/test/workbox-precaching/integration/test-precache-and-update.js
@@ -55,12 +55,12 @@ describe(`[workbox-precaching] Precache and Update`, function() {
     // Include the default ignoreURLParametersMatching query parameters.
     await global.__workbox.webdriver.get(`${baseURL}index.html`);
     await global.__workbox.webdriver.get(`${baseURL}index.html?utm_source=test`);
-    await global.__workbox.webdriver.get(`${baseURL}index.html?^fbclid$=test`);
+    await global.__workbox.webdriver.get(`${baseURL}index.html?fbclid=test`);
 
     expect(requestCounter.getURLCount('/test/workbox-precaching/static/precache-and-update/')).to.eql(0);
     expect(requestCounter.getURLCount('/test/workbox-precaching/static/precache-and-update/index.html')).to.eql(0);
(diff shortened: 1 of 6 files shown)


test/workbox-precaching/integration/test-precache-and-update.js
```diff
@@ -52,10 +52,15 @@ describe(`[workbox-precaching] Precache and Update`, function() {
     requestCounter = global.__workbox.server.startCountingRequests();
 
     // Request the page and check that the precached assets weren't requested from the network.
+    // Include the default ignoreURLParametersMatching query parameters.
     await global.__workbox.webdriver.get(`${baseURL}index.html`);
+    await global.__workbox.webdriver.get(`${baseURL}index.html?utm_source=test`);
+    await global.__workbox.webdriver.get(`${baseURL}index.html?^fbclid$=test`);
```
philipwalton: Is this URL supposed to have the `^` and `$` RegExp characters in it? (Same question for line 63.)

philipwalton: BTW, if this was a mistake and the tests are still passing, maybe update the tests to also check the inverse?

philipwalton (approved):
LGTM once the issue in the test has been resolved.

test/workbox-precaching/integration/test-precache-and-update.js
jeffposnick: Sorry, bad find-and-replace when I wanted to make the RegExp more restrictive.

PR #2537, opened 2020-06-11
Initial pass of workbox-cli conversion

### LINKED ISSUE #2476: Migrate workbox-cli to TypeScript
Placeholder to track the migration.

CC: @philkrie

philkrie:
R: @jeffposnick @philipwalton

*begins* to fix #2476

Initial pass. I'm able to get it to run, but tests aren't passing. Please take a look when you have a chance and let me know if I'm on the right track, i'm not very experienced w/ Typescript and there are some a few hacky solutions for now (search for //FIX).

package.json contains devDependencies for the types of some npm packages, which if I recall from the tests shouldn't be used. I also was messing around w/ module/target configs in tsconfig: since this is not imported as a module like most of the workbox packages, I wasn't sure if the default tsconfig should be changed.

In terms of tests, when running `gulp test --package="workbox-cli" --fix` I get almost entirely errors like this: 

`48) [workbox-cli] lib/run-wizard.js should write the configuration to the expected location based on the answers provided:
TypeError: runWizard is not a function
      at Context.<anonymous> (test/workbox-cli/node/lib/run-wizard.js:35:11)
      at processImmediate (internal/timers.js:439:21)
      at process.topLevelDomainCallback (domain.js:130:23)`

It doesn't seem to like how I exported functions in my typescript, but I'm unsure what the solution is.


jeffposnick:
Hey @philkrie! I'll leave some inline feedback, but in the meantime, could you rebase this so that the PR goes against the `v6` branch instead of `master`?

packages/workbox-cli/src/app.ts
```diff
@@ -6,28 +6,44 @@
...
  * Runs the specified build command with the provided configuration.
  *
  * @param {Object} options
  */
-async function runBuildCommand({command, config, watch}) {
+async function runBuildCommand({command, config, watch}: BuildCommand) {
   try {
     const {count, filePaths, size, warnings} =
-        await workboxBuild[command](config);
+        await (workboxBuild as any)[command](config); //FIX this is dubious
```
jeffposnick: If you limit `command` to 'generateSW' or `'injectManifest'` as per the earlier comment, you shouldn't have to cast `workboxBuild` to `any`.

You might still need to massage `config` a bit since the two take a different set of options, though.

packages/workbox-cli/src/app.ts
```diff
@@ -38,7 +54,7 @@ async function runBuildCommand({command, config, watch}) {
     } else {
       const message = filePaths
           .sort()
-          .map((filePath) => `  • ${filePath}`)
+          .map((filePath: string) => `  • ${filePath}`)
```
jeffposnick: `: string` shouldn't be necessary; TypeScript can infer it based on the type of `filePaths`.

packages/workbox-cli/src/app.ts
```diff
@@ -6,28 +6,44 @@
...
+import {constants} from './lib/constants.js';
+import {errors} from './lib/errors.js';
+import {logger} from './lib/logger.js';
+import {readConfig} from './lib/read-config.js';
+import {runWizard} from './lib/run-wizard.js';
+import * as meow from 'meow';
+
+
+interface BuildCommand {
+  command: string;
```
jeffposnick: You can limit the possible values via:

```
command: 'generateSW'|'injectManifest';
```

packages/workbox-cli/package.json
```diff
@@ -18,32 +18,37 @@
...
   },
   "workbox": {
     "packageType": "node"
+  },
+  "devDependencies": {
+    "@types/common-tags": "^1.8.0",
+    "@types/fs-extra": "^9.0.1",
+    "@types/glob-watcher": "^5.0.0",
+    "@types/inquirer": "^6.5.0",
+    "@types/meow": "5.0.0",
```
jeffposnick: I haven't checked all of these, but `meow`, at least, publishes its own types alongside the official module. You shouldn't have to import unofficial types from DefinitelyTyped.

https://github.com/sindresorhus/meow/blob/master/index.d.ts

packages/workbox-cli/src/bin.ts
```diff
@@ -8,17 +8,18 @@
...
+
+import {app} from "./app";
+import {cleanupStackTrace} from './lib/cleanup-stack-trace.js';
+import {helpText} from'./lib/help-text';
+import {logger} from './lib/logger';
 
 (async () => {
-  const params = meow(helpText);
-  updateNotifier({pkg: params.pkg}).notify();
+  const params: meow.Result<any> = meow(helpText);
```
jeffposnick, line 27: You can do

```
export interface SupportedFlags extends meow.AnyFlags {
  debug: meow.BooleanFlag;
  injectManifest: meow.BooleanFlag;
  watch: meow.BooleanFlag;
}
```

(those are the three supported flags) and then write `meow.Result<SupportedFlags>` instead of `meow.Result<any>`.

If you export the interface here, you can then import it in `app.ts` and reuse it.

packages/workbox-cli/src/app.ts
```diff
@@ -60,7 +76,7 @@ async function runBuildCommand({command, config, watch}) {
   }
 }
 
-module.exports = async (params = {}) => {
+export const app = async (params: meow.Result<any>) => {
```
jeffposnick: See later comment—this can be `meow.Result<SupportedFlags>`.

packages/workbox-cli/src/lib/cleanup-stack-trace.ts
```diff
@@ -10,18 +10,18 @@
 // Removes the initial portion, since that's obtained from error.message.
 // Removes every stack frame earlier than the last instance of moduleName,
 // since that's just frames related to the Node runtime/loader.
-module.exports = (error, moduleName) => {
-  const frames = error.stack.split(`\n`);
-  let startFrame = null;
+export function cleanupStackTrace(error: Error, moduleName: string) {
+  const frames = error?.stack?.split(`\n`);
```
jeffposnick: In our use case, if `error.stack` isn't set, it's probably cleanest to just return early.

Adding this to the top of the function will allow you to drop all of the `?` characters, since if you get that far, `error.stack` was set.

```
if (!error.stack) {
  return '';
}
```

packages/workbox-cli/src/lib/cleanup-stack-trace.ts
```diff
@@ -10,18 +10,18 @@
...
+  let startFrame = -1;
```
jeffposnick: You should be able to say:

```
let startFrame: number;
```

packages/workbox-cli/src/lib/cleanup-stack-trace.ts
```diff
@@ -10,18 +10,18 @@
...
   let lastFrame = 0;
-  frames.forEach((frame, index) => {
-    if (startFrame === null && frame.includes(`    at `)) {
+  frames?.forEach((frame, index) => {
+    if (startFrame === -1 && frame.includes(`    at `)) {
```
jeffposnick: Assuming you've set the type of `startFrame` to `number` as per the previous suggestion, this can be

```
if (startFrame === undefined && ...)
```

(I would rather keep the logic as-is instead of using `-1` as a "magic" value, although it's probably equivalent.)

packages/workbox-cli/src/lib/questions/ask-config-location.ts
```diff
@@ -6,31 +6,31 @@
...
 
 // The key used for the question/answer.
 const name = 'configLocation';
 
 /**
  * @return {Promise<Object>} The answers from inquirer.
  */
-function askQuestion() {
-  return inquirer.prompt([{
+function askQuestion(): Promise<Answers> {
```
jeffposnick: I think you can leave out the `: Promise<Answers>` here, as the type is inferred by the only return value.

packages/workbox-cli/src/lib/questions/ask-extensions-to-cache.ts
```diff
@@ -23,16 +23,16 @@ const name = 'globPatterns';
  * @return {Promise<Array<string>>} The unique file extensions corresponding
  * to all of the files under globDirectory.
  */
-async function getAllFileExtensions(globDirectory) {
-  const files = await new Promise((resolve, reject) => {
+async function getAllFileExtensions(globDirectory: string): Promise<Array<string>> {
```
jeffposnick: You can leave out the `: Promise<Array<string>>` here, as the type is inferred by the `return` value.

packages/workbox-cli/src/lib/questions/ask-extensions-to-cache.ts
```diff
@@ -59,7 +59,7 @@ async function getAllFileExtensions(globDirectory) {
  * @param {string} globDirectory The directory used for the root of globbing.
  * @return {Promise<Object>} The answers from inquirer.
  */
-async function askQuestion(globDirectory) {
+async function askQuestion(globDirectory: string): Promise<Answers> {
```
jeffposnick: Again, you should be able to leave out this `return` type.

packages/workbox-cli/src/lib/questions/ask-root-of-web-app.ts
```diff
@@ -24,10 +24,10 @@ const name = 'globDirectory';
  * @return {Promise<Array<string>>} The subdirectories of the current
  * working directory, with hidden and ignored ones filtered out.
  */
-async function getSubdirectories() {
+async function getSubdirectories(): Promise<Array<string>> {
```
jeffposnick, line 27: You can remove the `return` type definition.

packages/workbox-cli/src/lib/questions/ask-root-of-web-app.ts
```diff
@@ -41,18 +41,18 @@ async function getSubdirectories() {
 /**
  * @return {Promise<Object>} The answers from inquirer.
  */
-async function askQuestion() {
+async function askQuestion(): Promise<Answers> {
```
jeffposnick: Ditto on removing the `return` type.

packages/workbox-cli/src/lib/questions/ask-root-of-web-app.ts
```diff
@@ -41,18 +41,18 @@ async function getSubdirectories() {
...
     const manualEntryChoice = 'Manually enter path';
-    return inquirer.prompt([{
+    return prompt([{
       name,
       type: 'list',
       message: ol`What is the root of your web app (i.e. which directory do
         you deploy)?`,
       choices: subdirectories.concat([
-        new inquirer.Separator(),
+        new Separator() as unknown as string, //FIX This is kinda funky, but was triggering type error
```
jeffposnick: Try `new Separator().toString()`, as per https://github.com/SBoudrias/Inquirer.js/blob/039a55c7ec04d95887bdeb9b164cc634f26beb57/packages/inquirer/lib/objects/separator.js#L22-L24

The string conversion was previously implicit.

packages/workbox-cli/src/lib/questions/ask-sw-dest.ts
```diff
@@ -19,16 +19,16 @@ const name = 'swDest';
  * @param {string} defaultDir
  * @return {Promise<Object>} The answers from inquirer.
  */
-function askQuestion(defaultDir) {
-  return inquirer.prompt([{
+function askQuestion(defaultDir: string): Promise<Answers> {
```
jeffposnick: You can drop the `return` type.

packages/workbox-cli/src/lib/questions/ask-sw-src.ts
```diff
@@ -6,17 +6,17 @@
...
 
 // The key used for the question/answer.
 const name = 'swSrc';
 
 /**
  * @return {Promise<Object>} The answers from inquirer.
  */
-function askQuestion() {
-  return inquirer.prompt([{
+function askQuestion(): Promise<Answers> {
```
jeffposnick: You can drop the `return` type.

packages/workbox-cli/src/lib/read-config.ts
```diff
@@ -8,4 +8,4 @@
 
 // A really light wrapper on top of Node's require() to make it easier to stub
 // out reading the configuration during tests.
-module.exports = (configFile) => require(configFile);
+export function readConfig(configFile: string):any{require(configFile)};
```
jeffposnick: This should be

```
export function readConfig(configFile: string) {
  return require(configFile);
};
```

packages/workbox-cli/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
```
jeffposnick, line 1: I am not 100% sure about the values to use here. If you get caught up, @philipwalton might be able to provide some guidance.

jeffposnick (changes requested):

packages/workbox-cli/src/lib/cleanup-stack-trace.ts
philkrie: This throws an error on line 29 that the variable is used before assignment. How about:

`let startFrame: number | undefined;`

packages/workbox-cli/src/lib/questions/ask-extensions-to-cache.ts
philkrie: It doesn't appear to be: leaving this out results in `const subdirectories` being unknown on line 45.

packages/workbox-cli/src/lib/questions/ask-root-of-web-app.ts
philkrie: the inferred type is Promise<unknown>, which causes line 45 `const subdirectories`  to be typed as unknown. Is there a better way to resolve this?

packages/workbox-cli/src/app.ts
philkrie: I was able to use `GenerateSWConfig & InjectManifestConfig;`

philkrie:
Tests are still failing, but I think those need to be rewritten. For example, currently I'm getting errors due to `const app = require(MODULE_PATH);` not being a function. I have to unpack it a la `const {app} = require(MODULE_PATH);` Should I rewrite the tests, or is the issue in how I'm using export?

jeffposnick:
I can take a look at the tests for you. I think it's going to require (hah!) a bit of changing the underlying module loading plumbing.

jeffposnick:
I changed the build to run anything with a `packageType: 'node_ts'` field in its `package.json` through `tsc`. That, plus massaging some of the tests to use named exports instead of `default` and I think we're ready to go with this PR!

The one messy thing is that your original PR differs a bit from the `v6` branch's `demos/` directory, so this PR includes some of those changes as well. At this point, I think the easiest thing to do will be to merge and then sanity check the `demos/` later. (When we switched the base from `master` to `v6` that led to the confusion.)

jeffposnick (approved):

PR #2546, opened 2020-06-19
Switch from external<event> to isExternal property

### LINKED ISSUE #2031: Remove "external" flavor of workbox-window events
**Library Affected**:
workbox-window

**Browser & Platform**:
Chrome 72

If the service worker update is triggered from within the app using [ServiceWorkerRegistration.update()](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/update) instead of automatically at page load then the `workbox-window` events don't work as they normally do, breaking the update mechanism which otherwise works as expected when used with normal app re-launch.

This would be useful for long-running devices where the trigger for a version check can be pushed out (e.g. by the device subscribing to firebase, having a notification sent etc...)

Will try to create a simplified reproduction.

jeffposnick:
R: @philipwalton

Fixes #2031 

This keeps most of the logic the same, but swaps out the `external<something>` events for a `<something>` event that has a new `isExternal` property set.

This is a breaking change, obviously, and means that folks who did care about the distinction between non-external and external events need to explicitly check `isExternal`. Most folks didn't seem to care about the distinction, though, so the net effect is that you only need to listen to one type of event from now on.

The test suite hopefully covers the same ground as what we previously tested, but it was a bit of a struggle...

### DIFF
--- a/packages/workbox-window/src/Workbox.ts
+++ b/packages/workbox-window/src/Workbox.ts
@@ -41,9 +41,6 @@ const SKIP_WAITING_MESSAGE = {type: 'SKIP_WAITING'};
  * @fires [controlling]{@link module:workbox-window.Workbox#controlling}
  * @fires [activated]{@link module:workbox-window.Workbox#activated}
  * @fires [redundant]{@link module:workbox-window.Workbox#redundant}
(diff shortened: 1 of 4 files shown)


philipwalton (approved):
LGTM, and thanks for addressing the flaky test!

PR #2548, opened 2020-06-19
Route/Handler TypeScript normalization

### LINKED ISSUE #2511: RouteHandlerCallbackOptions type is not conform to the doc
**Library Affected**:
workbox-core (types exported)

**Browser & Platform**:
N/A

According to the doc (https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-routing#~handlerCallback), options should be:

```js
export interface RouteHandlerCallbackOptions {
  request: Request;
  url?: URL;
  event?: FetchEvent;
  params?: string[] | {[paramName: string]: string};
}
```

but the type exported by workbox-core are https://github.com/GoogleChrome/workbox/blob/04ba6442c466d2e8197fe586672143d201af3a61/packages/workbox-core/src/types.ts#L36-L44

This type is too loose (`ExtendableEvent` instead of `FetchEvent` and `Request | string` instead of `Request`).

This makes it hard to use for example :

```js
const saveHandler: RouteHandler = ({ url, event, params }) => {
    let initid = params[1];
    let viewId = params[2];
    const workingRequest = event.request.clone();
```

which triggers `TS2339: Property 'request' does not exist on type 'ExtendableEvent'.`

neither

```js
  const saveHandler: RouteHandler = ({ url, event, params , request}) => {
    let initid = params[1];
    let viewId = params[2];
    const workingRequest = request.clone();
```

which triggers `TS2339: Property 'clone' does not exist on type 'string | Request'.   Property 'clone' does not exist on type 'string'.`

jeffposnick:
R: @philipwalton

Fixes #2511

I am not 100% confident about the changes in this PR, especially as it relates to the new `workbox-strategy` code in v6. So your close attention and feedback is appreciated, @philipwalton. It will be good to shake this out during the v6 pre-releases to make sure nothing unexpected was done.

There are a number of JSDoc changes that need to be made along with these TypeScript changes, and I'll do them in a follow-up commit once we're happy with all the TypeScript types.

As part of this PR, I dropped the "shadow" types that were re-exported in `workbox-routing/src/_types.ts` and just refer to everything from a single source of truth, `workbox-core/src/types.ts`.

I tested this manually with some sample usage scenarios to confirm that `tsc` didn't flag any problems. All of the following worked fine:

```ts
registerRoute(
  /\.css$/,
  ({url}) => fetch(url.href),
);

const myHandler = ({request} : {request: Request}) => {
  return fetch(request);
};
registerRoute(
  ({url}) => url.pathname.endsWith('.js'),
  myHandler,
);

const networkFirst = new NetworkFirst();
registerRoute(
  '/abcd',
  networkFirst,
);

async function testHandler(request: Request) {
  const response = await networkFirst.handle({request});
  return response;
}
registerRoute(
  '/abcd',
  ({request}) => testHandler(request),
);

async function anotherTestHandler() {
  const response = await networkFirst.handle({request: '/blah'});
  return response;
}
registerRoute(
  '/abcd',
  anotherTestHandler,
);

const saveHandler = async ({url, event, params} : {
  url: URL, event: FetchEvent, params?: string[] | {},
}) => {
  console.log(url, event, params);
  return new Response('blah');
};
registerRoute(
  '/abcd',
  saveHandler,
);
```

### DIFF
--- a/packages/workbox-core/src/types.ts
+++ b/packages/workbox-core/src/types.ts
@@ -108,7 +108,7 @@ export type RouteHandler = RouteHandlerCallback | RouteHandlerObject;
 
 export interface HandlerWillStartCallbackParam {
   request: Request;
-  event?: ExtendableEvent;
+  event: ExtendableEvent;
   state?: PluginState;
 }
 
@@ -120,7 +120,7 @@ export interface CacheDidUpdateCallbackParam {
   cacheName: string;
   newResponse: Response;
   request: Request;
-  event?: ExtendableEvent;
+  event: ExtendableEvent;
   oldResponse?: Response | null;
   state?: PluginState;
 }
@@ -132,7 +132,7 @@ export interface CacheDidUpdateCallback {
 export interface CacheKeyWillBeUsedCallbackParam {
   mode: string;
   request: Request;
-  event?: ExtendableEvent;
+  event: ExtendableEvent;
   params?: any;
   state?: PluginState;
 }
@@ -144,7 +144,7 @@ export interface CacheKeyWillBeUsedCallback {
 export interface CacheWillUpdateCallbackParam {
   request: Request;
   response: Response;
-  event?: ExtendableEvent;
+  event: ExtendableEvent;
   state?: PluginState;
 }
 
@@ -156,7 +156,7 @@ export interface CachedResponseWillBeUsedCallbackParam {
   cacheName: string;
   request: Request;
   cachedResponse?: Response;
-  event?: ExtendableEvent;
+  event: ExtendableEvent;
   matchOptions?: CacheQueryOptions;
   state?: PluginState;
 }
@@ -169,7 +169,7 @@ export interface FetchDidFailCallbackParam {
   error: Error;
   originalRequest: Request;
   request: Request;
-  event?: ExtendableEvent;
+  event: ExtendableEvent;
   state?: PluginState;
 }
 
@@ -180,7 +180,7 @@ export interface FetchDidFailCallback {
 export interface FetchDidSucceedCallbackParam {
   request: Request;
   response: Response;
-  event?: ExtendableEvent;
+  event: ExtendableEvent;
   state?: PluginState;
 }
 
@@ -190,7 +190,7 @@ export interface FetchDidSucceedCallback {
 
 export interface RequestWillFetchCallbackParam {
   request: Request;
-  event?: ExtendableEvent;
+  event: ExtendableEvent;
   state?: PluginState;
 }
 
@@ -201,7 +201,7 @@ export interface RequestWillFetchCallback {
 export interface HandlerWillRespondCallbackParam {
   request: Request;
   response: Response;
-  event?: ExtendableEvent;
+  event: ExtendableEvent;
   state?: PluginState;
 }
 
@@ -211,7 +211,7 @@ export interface HandlerWillRespondCallback {
 
 export interface HandlerDidRespondCallbackParam {
   request: Request;
-  event?: ExtendableEvent;
+  event: ExtendableEvent;
   response?: Response;
   state?: PluginState;
 }
@@ -223,7 +223,7 @@ export interface HandlerDidRespondCallback {
 export interface HandlerDidCompleteCallbackParam {
   request: Request;
   error?: Error;
-  event?: ExtendableEvent;
+  event: ExtendableEvent;
   response?: Response;
   state?: PluginState;
 }
--- a/packages/workbox-core/src/types.ts
+++ b/packages/workbox-core/src/types.ts
@@ -23,10 +23,10 @@ export type PluginState = MapLikeObject;
  * Options passed to a `RouteMatchCallback` function.
  */
 export interface RouteMatchCallbackOptions {
-  url: URL;
-  sameOrigin: boolean;
+  event: ExtendableEvent;
   request: Request;
-  event?: ExtendableEvent;
+  sameOrigin: boolean;
+  url: URL;
 }
 
 /**
@@ -48,11 +48,22 @@ export interface RouteMatchCallback {
  * Options passed to a `RouteHandlerCallback` function.
  */
 export interface RouteHandlerCallbackOptions {
-  request: Request | string;
-  url?: URL;
+  event: ExtendableEvent;
+  request: Request;
+  url: URL;
   params?: string[] | MapLikeObject;
+}
+
+/**
+ * Options passed to a `ManualHandlerCallback` function.
+ */
+export interface ManualHandlerCallbackOptions {
+  request: Request | string;
   event?: ExtendableEvent;
 }
+
+export type HandlerCallbackOptions =
+    RouteHandlerCallbackOptions | ManualHandlerCallbackOptions;
 
 /**
  * The "handler" callback is invoked whenever a `Router` matches a URL/Request
@@ -64,6 +75,18 @@ export interface RouteHandlerCallbackOptions {
  */
 export interface RouteHandlerCallback {
   (options: RouteHandlerCallbackOptions): Promise<Response>;
+}
+
+/**
+ * The "handler" callback is invoked whenever a `Router` matches a URL/Request
+ * to a `Route` via its `RouteMatchCallback`. This handler callback should
+ * return a `Promise` that resolves with a `Response`.
+ *
+ * If a non-empty array or object is returned by the `RouteMatchCallback` it
+ * will be passed in as this handler's `options.params` argument.
+ */
+export interface HandlerCallback {
+  (options: HandlerCallbackOptions): Promise<Response>;
 }
 
 /**
@@ -95,10 +118,10 @@ export interface HandlerWillStartCallback {
 
 export interface CacheDidUpdateCallbackParam {
   cacheName: string;
-  oldResponse?: Response | null;
   newResponse: Response;
   request: Request;
   event?: ExtendableEvent;
+  oldResponse?: Response | null;
   state?: PluginState;
 }
 
@@ -107,10 +130,10 @@ export interface CacheDidUpdateCallback {
 }
 
 export interface CacheKeyWillBeUsedCallbackParam {
-  request: Request;
   mode: string;
-  params?: any;
+  request: Request;
   event?: ExtendableEvent;
+  params?: any;
   state?: PluginState;
 }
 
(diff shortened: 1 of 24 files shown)


packages/workbox-core/src/types.ts
```diff
@@ -48,12 +48,23 @@ export interface RouteMatchCallback {
...
+  url: URL;
   params?: string[] | MapLikeObject;
+}
+
+/**
+ * Options passed to a `ManualHandlerCallback` function.
+ */
+export interface ManualHandlerCallbackOptions {
+  request: Request | string;
   event?: ExtendableEvent;
```
philipwalton: I believe we want to make this required in v6.

packages/workbox-core/src/types.ts
```diff
@@ -48,12 +48,23 @@ export interface RouteMatchCallback {
...
+  event: ExtendableEvent;
+  request: Request;
+  url: URL;
   params?: string[] | MapLikeObject;
+}
+
+/**
+ * Options passed to a `ManualHandlerCallback` function.
+ */
+export interface ManualHandlerCallbackOptions {
```
philipwalton, line 60: Could `RouteHandlerCallbackOptions` extend `ManualHandlerCallbackOptions` (especially if we make `event` required)?

packages/workbox-core/src/types.ts
```diff
@@ -66,6 +77,18 @@ export interface RouteHandlerCallback {
...
 
+/**
+ * The "handler" callback is invoked whenever a `Router` matches a URL/Request
+ * to a `Route` via its `RouteMatchCallback`. This handler callback should
+ * return a `Promise` that resolves with a `Response`.
+ *
+ * If a non-empty array or object is returned by the `RouteMatchCallback` it
+ * will be passed in as this handler's `options.params` argument.
+ */
+export interface HandlerCallback {
```
philipwalton: This looks identical to the `RouteHandlerCallback` type above, and it doesn't seem to be used anywhere in the code base. Is it needed?

packages/workbox-routing/src/_types.ts
```diff
@@ -6,25 +6,8 @@
...
-  RouteHandlerObject as HandlerObject,
-  RouteHandlerCallback as HandlerCallback,
-  RouteHandlerCallbackOptions as HandlerCallbackOptions,
-  RouteMatchCallback as MatchCallback,
-  RouteMatchCallbackOptions as MatchCallbackOptions,
-}
-
 // * * * IMPORTANT! * * *
 // ------------------------------------------------------------------------- //
 // jdsoc type definitions cannot be declared above TypeScript definitions or
```
philipwalton, line 13: This was a problem before, but it looks like the jsdoc comments in this file are no longer up-to-date. Can you fix that as part of this PR?

packages/workbox-strategies/src/Strategy.ts
```diff
@@ -115,7 +110,7 @@ abstract class Strategy implements RouteHandlerObject {
    * @param {URL} [options.url]
    * @param {*} [options.params]
    */
-  handle(options: FetchEvent | RouteHandlerCallbackOptions): Promise<Response> {
+  handle(options: HandlerCallbackOptions): Promise<Response> {
```
philipwalton: I think I'd still like to have this to accept a `FetchEvent` object, since that's how it's most often going to be used:

```js
addEventListener('fetch', (event) => {
   event.respondWith(someStrategy.handle(event));
});
```

I remember in the past you expressed a preference for `handle({event})` (using object shorthand), but that doesn't work if we're going to require both the `request` and the `event` param (which I think we want to?).

WDYT?

packages/workbox-strategies/src/Strategy.ts
```diff
@@ -150,12 +145,14 @@ abstract class Strategy implements RouteHandlerObject {
...
-      // `options.request` can be a string, similar to what `fetch()` accepts.
-      options.request = new Request(options.request);
     }
 
-    const {event, request, params} = options as StrategyHandlerOptions;
+    const event = options.event;
+    const request = typeof options.request === 'string' ?
+        new Request(options.request) :
+        options.request;
+    const params = 'params' in options ? options.params : undefined;
```
philipwalton, line 158: Why the change away from destructuring here?

packages/workbox-core/src/types.ts
```diff
@@ -48,12 +48,23 @@ export interface RouteMatchCallback {
...
+
+/**
+ * Options passed to a `ManualHandlerCallback` function.
+ */
+export interface ManualHandlerCallbackOptions {
+  request: Request | string;
   event?: ExtendableEvent;
 }
 
+export type HandlerCallbackOptions =
```
philipwalton, line 65: Similar question to above. Would it work to make something like `HandlerCallbackOptions` a base type, and then any modules that need to accept `RouteHandlerCallbackOptions | ManualHandlerCallbackOptions` could accept any subtype of `HandlerCallbackOptions`?

philipwalton (changes requested):

packages/workbox-core/src/types.ts
jeffposnick: Oh, right. We had talked about how using strategies manually outside of an event handler wasn't a great idea.

packages/workbox-core/src/types.ts
jeffposnick: Yup.

packages/workbox-core/src/types.ts
jeffposnick: One follow-up from that is that I need to change a lot of the parameter signatures in the `PrecacheController` class to mark the `ExtendableEvent` as being required. (That's probably a good thing, since using `PrecacheController` outside of an event sounds like a recipe for disaster.)

jeffposnick: This change also filters through to the `MatchWrapperOptions` and `PutWrapperOptions` interfaces, which now need to mark the `event` as required.

packages/workbox-core/src/types.ts
jeffposnick: Actually, no, because `ManualHandlerCallbackOptions` defines `request: Request | string;` and we want `RouteHandlerCallbackOptions` to be `request: Request;`.

There are enough different about the two that it would be easier to keep them separate rather than extend and override.

packages/workbox-core/src/types.ts
jeffposnick: So my intention here was to export an interface that developers could implement when they wrote their own "manual" handler functions. But `HandlerCallback `, as implemented, is probably too generic, since it could take either `RouteHandlerCallbackOptions` or `ManualHandlerCallbackOptions` as its parameter.

I've gotten more specific by removing `HandlerCallback` and exporting an explicit `ManualHandlerCallback` instead, that takes `ManualHandlerCallbackOptions`.

So even though `ManualHandlerCallbackOptions` isn't used by the codebase, developers could take advantage of it via

```
import {ManualHandlerCallback} from 'workbox-core/types';

const myHandler: ManualHandlerCallback = async({request, event}) => {
  const networkFirst = new NetworkFirst();
  const networkResponse = networkFirst.handle({event, request});
  return networkResponse || new Response('...');
};
```

packages/workbox-routing/src/_types.ts
jeffposnick: Once the actual types are 👍 , I'll go back and adjust the JSDocs. It's going to be too much wasted effort otherwise.

packages/workbox-strategies/src/Strategy.ts
jeffposnick: Yeah, I'm adding that `FetchEvent` back as one of the supported types, especially since the code supports it.

packages/workbox-strategies/src/Strategy.ts
jeffposnick: `HandlerCallbackOptions` may not contain `params` (`RouteHandlerCallbackOptions` will, but `ManualHandlerCallbackOptions` won't), and explicitly checking instead of just using destructing was necessary to get TypeScript happy.

packages/workbox-core/src/types.ts
jeffposnick: See my earlier comment about wanting to keep `RouteHandlerCallbackOptions` and `ManualHandlerCallbackOptions` independent. Making `HandlerCallbackOptions` a composite of both of them seems to work for our use case.

jeffposnick:
PTAL.

packages/workbox-core/src/types.ts
philipwalton: Gotcha.

packages/workbox-core/src/types.ts
philipwalton: SGTM

packages/workbox-routing/src/_types.ts
philipwalton: SGTM

packages/workbox-strategies/src/Strategy.ts
philipwalton: Cool.

packages/workbox-core/src/types.ts
philipwalton: SGTM.

packages/workbox-strategies/src/Strategy.ts
philipwalton: Kk

packages/workbox-core/src/types.ts
philipwalton: Not sure if you want to do this as part of this PR or a separate on, but there are many cases in our code where we've had to make an event optional in a type definition due to us not making it required in handlers (e.g. [here](https://github.com/GoogleChrome/workbox/blob/f6cecf266436b4b199d6a80ece4f09a8515609d0/packages/workbox-strategies/src/StrategyHandler.ts#L41)). I did a in VSCode for `event?:`, and lots of cases come up (and there are probably others where the name "event" isn't used).

Also, there are cases where we're conditionally checking whether an event was passed (e.g. [here](https://github.com/GoogleChrome/workbox/blob/f6cecf266436b4b199d6a80ece4f09a8515609d0/packages/workbox-strategies/src/StrategyHandler.ts#L120-L122)), and it probably makes sense to remove those now.

How do you want to handle these?

jeffposnick: Okay, I can get those updated in this PR as well.

philipwalton (approved):

packages/workbox-core/src/types.ts
philipwalton: Thanks, I went through and updated the corresponding jsdoc comments for a few that where missed (it was easier than going back and forth).

jeffposnick:
Okay! I think with your JSDoc updates, @philipwalton, and all the tests now passing, that's it for this PR.

If there's anything I left out, we can pick it up prior to the v6 release, but I'm merging for now.

PR #2560, opened 2020-06-30
Ensure getSW() always returns a promise

### LINKED ISSUE #2545: UMD version workbox-window getSW() does not always return a Promise
**Library Affected**:
*workbox-window*

**Browser & Platform**:
*all browsers*

In Workbox.ts, getSW() is defined as async:
```javascript
 async getSW(): Promise<ServiceWorker> {
    // If `this._sw` is set, resolve with that as we want `getSW()` to
    // return the correct (new) service worker if an update is found.
    return this._sw !== undefined ? this._sw : this._swDeferred.promise;
  }
}
```
UMD version is as follows:
```javascript
  _proto.getSW = function getSW() {
        try {
          var _this7 = this;

          // If `this._sw` is set, resolve with that as we want `getSW()` to
          // return the correct (new) service worker if an update is found.
          return _this7._sw !== undefined ? _this7._sw : _this7._swDeferred.promise;
        } catch (e) {
          return Promise.reject(e);
        }
      }
```
(from https://storage.googleapis.com/workbox-cdn/releases/5.1.2/workbox-window.dev.umd.js)
So when __sw_ is set to a ServiceWorker object, it is just returned and not wrapped in a Promise.
This causes calling code to fail with:
 _Uncaught TypeError: wb.getSW(...).then is not a function_

jeffposnick:
R: @philipwalton
CC: @housseindjirdeh 

Fixes #2545

I think the `async` without any usage of `await` might have confused the transpilation? Anyway, being explicit seems like the best option.

### DIFF
--- a/packages/workbox-window/src/Workbox.ts
+++ b/packages/workbox-window/src/Workbox.ts
@@ -254,10 +254,12 @@ class Workbox extends WorkboxEventTarget {
    *
    * @return {Promise<ServiceWorker>}
    */
-  async getSW(): Promise<ServiceWorker> {
+  getSW(): Promise<ServiceWorker> {
     // If `this._sw` is set, resolve with that as we want `getSW()` to
     // return the correct (new) service worker if an update is found.
-    return this._sw !== undefined ? this._sw : this._swDeferred.promise;
+    return this._sw !== undefined ?
+      Promise.resolve(this._sw) :
+      this._swDeferred.promise;
   }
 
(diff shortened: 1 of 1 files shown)


philipwalton (approved):
Yeah, I think (from a style perspective) it's best to always use await if we're marking a function as async. Otherwise it's not clear if the promise-returning behavior was intended or it was just a mistake.

PR #2562, opened 2020-06-30
Precache one request at a time

### LINKED ISSUE #2528: Batching of precache requests to prevent net::ERR_INSUFFICIENT_RESOURCES in Chrome
**Library Affected**:
This likely is only related to *workbox-precache*

**Browser & Platform**:
*Google Chrome*

I originally wrote this as a comment on #570, but I since that issue is closed, I thought I would post it as a new issue to get more visibility

It seems that workbox uses [`Promise.all`](https://github.com/GoogleChrome/workbox/blob/master/packages/workbox-precaching/src/PrecacheController.ts#L171) when making precache requests instead of explicitly batching or rate-limiting the requests, and it seems that Chrome can't handle this under certain circumstances. 

Specifically, I'm getting sporadic `net::ERR_INSUFFICIENT_RESOURCES` failures.

Screen Shot 2020-06-02 at 1 35 10 AM

After looking up this error, it seems that this represents some sort of resource exhaustion within chrome. A few other people have come across this [here](https://github.com/ryanmr/err-insufficient-resources-demo), and [here](https://bugs.chromium.org/p/chromium/issues/detail?id=958108), and it seems that the answer is to simply make fewer concurrent requests. 

I noticed that @nachoab came up against this [here](https://github.com/GoogleChromeLabs/sw-precache/issues/118#issuecomment-338593470) and solved it by batching requests in chunks of 20, effectively limiting concurrency to 20 inflight requests at most... but that's against the old `sw-precache` repo and not directly applicable here.

The easiest and most flexible solution as I see it would be to make `precacheAndRoute` (or just `precache`) return a promise that resolves when all of the specified routes have been precached. This way I could do the rate-limiting in my serviceworker, instead of adding the burden of rate-limiting onto workbox. 

Thoughts?

jeffposnick:
R: @philipwalton

Fixes #2528

This is basically batched precaching with a forced batch size of 1, i.e. only have one preaching request in flight at a given time.

I think this is a win from the perspective of reducing possible bandwidth contention with the main thread during precaching as well.

### DIFF
--- a/packages/workbox-precaching/src/PrecacheController.ts
+++ b/packages/workbox-precaching/src/PrecacheController.ts
@@ -275,18 +275,19 @@ class PrecacheController {
       }
     }
 
-    const precacheRequests = toBePrecached.map(({cacheKey, url}) => {
+    // Cache entries one at a time.
+    // See https://github.com/GoogleChrome/workbox/issues/2528
+    for (const {cacheKey, url} of toBePrecached) {
       const integrity = this._cacheKeysToIntegrities.get(cacheKey);
       const cacheMode = this._urlsToCacheModes.get(url);
-      return this._addURLToCache({
+      await this._addURLToCache({
         cacheKey,
         cacheMode,
         event,
         integrity,
         url,
       });
-    });
-    await Promise.all(precacheRequests);
+    }
 
     const updatedURLs = toBePrecached.map((item) => item.url);
 


philipwalton (approved):
👍

jobvs:
@jeffposnick Would it be possible to get this in a 5.1.4? The changes are not conflicting with the 5.x codebase. Can make the MR if you want.

PR #2564, opened 2020-07-01
Document skipWaiting message in GenerateSW

jwagner:
R: @jeffposnick @philipwalton

It wasn't very clear from the documentation how `skipWaiting` could be triggered via `postMessage`.


jeffposnick:
Thanks!

jeffposnick (changes requested):
Would you mind making this change to

https://github.com/GoogleChrome/workbox/blob/9498e9192fd3ed6d60fc54507776d612d3cbde3c/packages/workbox-build/src/generate-sw.js#L171-L174

as well?

jwagner:
Hi @jeffposnick first off, sorry for taking so incredibly long to reply. I added the change to the second location and removed a stray period.

jeffposnick (approved):
Thank you!

PR #2567, opened 2020-07-06
Add fetch fallback option for generate SW

### LINKED ISSUE #2566: Fallback on fetch when both network and cache not available
# Summary

I'm using `workbox-webpack-plugin` to generate `service-worker.js` and `workbox-xxxxxx.js` runtime code. Using current implementation, there is no way to fallback when fetch in service worker throws an error due to network offline. What's worse is that when there is no pre-caches or runtime caches hit, the web app would show the browser's network offline error page. Which is not ideal from the user experience perspective. Besides, fallback image could help to workout a placeholder image scenario.

**NOTE**
This is totally different requirement from `navigationFallback` , which assuming all the web app resources are pre-cached. And when a navigate to a route that is not pre-cached, the navigation fallback url will be responded, even the network is online.

# Propose Changes

There is solution to implement comprehensive fallback:

> All of the built-in caching strategies reject in a consistent manner when there's a network failure and/or a cache miss. This promotes the pattern of setting a global "catch" handler to deal with any failures in a single handler function.
>
> Reference: https://developers.google.com/web/tools/workbox/guides/advanced-recipes#comprehensive_fallbacks

However the global [setCatchHandler](https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-routing#.setCatchHandler) API is not exposed in the generated workbox runtime. Hence there is no way to utilized this so far.

The ideas here are:

1. Add a configuration option for fetch fallback
2. Add comprehensive fallback solution into the service worker template
3. Generate related code

Generate service worker with extra following code:

image

**Library Affected**:
*workbox-build, workbox-webpack-plugin*

**Browser & Platform**:
*all browsers*

shadowwalker:
This provides a fallback mechanism when both network and cache is not available

All of the built-in caching strategies reject in a consistent manner when there's a network failure and/or a cache miss. This promotes the pattern of setting a global "catch" handler to deal with any failures in a single handler function.

Reference:
https://developers.google.com/web/tools/workbox/guides/advanced-recipes#comprehensive_fallbacks

R: @jeffposnick @philipwalton

Fixes: https://github.com/GoogleChrome/workbox/issues/2566 (see the issue for more detailed description)

Following is generated in the `service-worker.js` when generate SW
image

### DIFF
--- a/packages/workbox-build/src/lib/populate-sw-template.js
+++ b/packages/workbox-build/src/lib/populate-sw-template.js
@@ -30,6 +30,7 @@ module.exports = ({
   offlineGoogleAnalytics,
   runtimeCaching = [],
   skipWaiting,
+  fetchFallback,
 }) => {
   // There needs to be at least something to precache, or else runtime caching.
   if (!(manifestEntries.length > 0 || runtimeCaching.length > 0)) {
@@ -85,6 +86,7 @@ module.exports = ({
       precacheOptionsString,
       runtimeCaching: runtimeCachingConverter(moduleRegistry, runtimeCaching),
       skipWaiting,
+      fetchFallback,
       use: moduleRegistry.use.bind(moduleRegistry),
     });
 
--- a/packages/workbox-build/src/lib/write-sw-using-default-template.js
+++ b/packages/workbox-build/src/lib/write-sw-using-default-template.js
@@ -34,6 +34,7 @@ module.exports = async ({
   skipWaiting,
   sourcemap,
   swDest,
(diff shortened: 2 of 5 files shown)


jeffposnick:
We're going to go with the approach outlined in https://github.com/GoogleChrome/workbox/issues/2569#issuecomment-656867032, with a `runtimeCaching` configuration option implemented after we get the `handlerDidError` callback in place.

Thanks for the contribution, though, as it forced us to think about a cleaner design that we've been putting off for a while 😄

PR #2517, opened 2020-05-19
gulp modernization

jeffposnick:
R: @philipwalton 

Fixes #2508

This is a substantial rewrite of the `gulp` build tasks, following recommended `gulp` v4 best practices.

It should hopefully result in cleaner, more maintainable tasks, with fewer wrapper/throwaway dependencies.

In general, the current functionality should be equivalent to what we previously had, with one major change being that task names all have `_` separators instead of `:`.

This also removes the `npm scripts` that we previously using in favor for `gulp` equivalents, (except for the `version` script, because that's used by `lerna` directly).

PR #2570, opened 2020-07-08
publish_glitch gulp task update

jeffposnick:
R: @philkrie 

The `publish-glitch.js` task wasn't properly converted to the new `gulp` task setup as part of #2517. The old task got imported into the v6 branch via #2537.

This PR updates the `publish-glitch.js` task to match what we're doing in the other v6 tasks. I don't have things set up to run it myself, so it would be appreciated if you could try it out for me via `gulp publish_glitch`, @philkrie  (note the underscore) and confirm that it works as expected.

### DIFF
--- a/gulp-tasks/publish-glitch.js
+++ b/gulp-tasks/publish-glitch.js
@@ -6,68 +6,57 @@
   https://opensource.org/licenses/MIT.
 */
 
-const gulp = require('gulp');
-const childProcess = require('child_process');
-const fse = require('fs-extra');
-const tempy = require('tempy');
-const util = require('util');
 const del = require('del');
+const execa = require('execa');
+const fse = require('fs-extra');
+const globby = require('globby');
+const ol = require('common-tags').oneLine;
+const olt = require('common-tags').oneLineTrim;
+const tempy = require('tempy');
+const upath = require('upath');
 
-const exec = util.promisify(childProcess.exec);
+const DEMOS_DIR = 'demos/src';
 
-const glitchProjects = [
-  'workbox-core',
-  'workbox-sw',
-  'workbox-precaching',
-  'workbox-strategies',
-  'workbox-background-sync-demo',
-  'workbox-routing',
-  'workbox-expiration',
-  'workbox-cacheable-response',
-  'workbox-google-analytics',
-  'workbox-streams',
-  'workbox-range-requests',
-  'workbox-broadcast-update-demo',
-  'workbox-window',
-  'workbox-navigation-preload',
-];
-
-
-gulp.task('publish-glitch', async () => {
+async function publish_glitch() {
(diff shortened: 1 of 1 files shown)


gulp-tasks/publish-glitch.js
```diff
@@ -6,68 +6,57 @@
...
+    await execa('git', ['commit', `-m'Autocommit on ${new Date()}'`], {
+      cwd: projectPath,
+    });
+    await execa('git', ['push', 'origin', 'glitch', '-f', '--set-upstream',
+      '--no-verify'], {cwd: projectPath});
+
+    const deployURL = new URL(`https://${project}.glitch.me/deploy`);
+    deployURL.searchParams.set('secret', process.env.GLITCH_WORKBOX_SECRET);
+    deployURL.searchParams.set('repo', `https://api.glitch.com/git/${project}`);
+    await execa('curl', ['-X', 'POST', `"${deployURL}"`]);
```
philkrie, line 54: Double quotes around deployURL cause curl to fail, took me longer than I care to admit to see that...

gulp-tasks/publish-glitch.js
```diff
@@ -6,68 +6,57 @@
...
-      await exec(`git push origin glitch -f --set-upstream --no-verify`,
-          {cwd: projectPath});
-      await exec(`curl -X POST "${'https://' + project + '.glitch.me/deploy?secret=' + process.env.GLITCH_WORKBOX_SECRET + '&repo=https://api.glitch.com/git/' + project}"`);
-      await del(projectPath, {force: true});
-      console.log('Push attempted to ' + project);
-    } catch (e) {
-      console.log(project);
-      console.log(e.stdout);
-    }
+    await execa('git', ['clone', projectURL, projectPath]);
```
philkrie, line 39: Everything from 39-54 should be in a try catch, otherwise the loop stops in certain cases, such as no changes in the working tree of one of the folders.

philkrie (changes requested):

gulp-tasks/publish-glitch.js
jeffposnick: Your original code had double-quotes as well. But I'll remove all the quotes as perhaps the was `execa` interprets them is different than in your code.

gulp-tasks/publish-glitch.js
jeffposnick: Okay.

jeffposnick:
(I just pushed the https://github.com/GoogleChrome/workbox/commit/d26e7104ebe9517dfd01fa2b9917d1e8cbf1f282 commit to this branch, but for some reason, the updates aren't showing up in this PR. If that doesn't resolve itself soon, I'm going to close the PR and re-open one with the current contents of the `publish-glitch-update` branch.)

PR #2571, opened 2020-07-08
Remove canConstructResponseFromBodyStream

jeffposnick:
R: @philipwalton
CC: @aliai 

This is an alternative to #2529

See https://github.com/GoogleChrome/workbox/pull/2529#issuecomment-646075682 for context.


philipwalton:
Can we confirm that this still supports streaming in some form? I wouldn't want this utility method to always read the entire response into memory/disk prior to creating the new response.

jeffposnick:
"_still_ supports" may not be the right question, because v5 unconditionally invoked the `blob` codepath.

So if you were fine with the v5 behavior—as in, you've tested scenarios you care about and haven't noticed any problems—this PR should not result in any changes.

There might still be some benefits from using a `ReadableStream` instead of a `blob` as the constructor, but given the differing `Content-Type` header behavior, they can't just be used interchangeably.

philipwalton (changes requested):
Sorry for getting back to this late...

My recollection from our discussion with the SW team is that using a blob when copying a response could definitely slow it down. I've been using this method to add metadata to headers on my site, and I'd prefer to use the stream when available, to ensure responses are as fast as possible.

We can just get the `Content-Type` header from the original response, right?

tropicadri:
Closing this PR due to some maintenance clean up.

PR #2572, opened 2020-07-09
Fix workbox-build types in workbox-cli

jeffposnick:
R: @philipwalton

This changes the dependency on `workbox-cli`'s TypeScript typings from `@types/workbox-build` to a local override file for the time being.

Once `workbox-build` is migrated to TypeScript, we can remove it.

This is necessary because `@types/workbox-build` contains references to some types exported from `workbox-routing` that are no longer valid in v6. When trying to build `workbox-cli` in a local project that has the new v6 version of `workbox-routing` linked in via `lerna bootstrap`, the `tsc -b` will fail. (This took me forever to debug.)

### DIFF
--- a/packages/workbox-cli/src/index.d.ts
+++ b/packages/workbox-cli/src/index.d.ts
@@ -1,0 +1,12 @@
+declare module 'workbox-build' {
+  type BuildReturnValue = {
+    count: number;
+    filePaths: Array<string>;
+    size: number;
+    warnings: Array<string>;
+  };
+
+  export function generateSW(args: any): Promise<BuildReturnValue>;
+  export function injectManifest(args: any): Promise<BuildReturnValue>;
+  export function copyWorkboxLibraries(dir: string): Promise<string>;
+}
--- a/.gitignore
+++ b/.gitignore
@@ -21,11 +21,10 @@ firebase-debug.log
 # Generated TypeScript files and build data
 tsconfig.tsbuildinfo
 packages/workbox-*/**/*.d.ts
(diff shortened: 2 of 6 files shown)


packages/workbox-cli/src/index.d.ts
```diff
@@ -0,0 +1,12 @@
+declare module 'workbox-build' {
```
philipwalton, line 1: Hmmm, hopefully having a `d.ts` file in the `src` directory won't cause any issues. Could this also be at the top level?

philipwalton (approved):
LGTM with one question...

packages/workbox-cli/src/index.d.ts
jeffposnick: Could you be more specific about what issues it might cause?

FWIW, the output in `build/` looks like:

```
$ ls packages/workbox-cli/build/
app.d.ts        app.js          bin.d.ts        bin.js          lib
```

without the `index.d.ts` carried over. We only publish the `build/` directory to `npm`, so I don't think anything would "leak".

jeffposnick: (Moving it to the top-level meant that it wasn't automatically picked up by `tsc`. I guess I could change the compiler settings to override that, but unless there's an actual problem, I'd rather keep it simple.)

jeffposnick: ...I'm just going to merge (as this is causing problems for me whenever I run `gulp build` locally) and if there's some reason that it needs to be moved, we can do that in a follow-up PR.

PR #2576, opened 2020-07-11
Adds handlerDidError plugin callback

### LINKED ISSUE #2569: Add handlerDidError plugin callback
**Library Affected**:
*workbox-strategies*

`workbox-routing` currently provides "fallback" support via the `setCatchHandler()` method, which is [usually written](https://developers.google.com/web/tools/workbox/guides/advanced-recipes#provide_a_fallback_response_to_a_route) as a big `switch` statement with a bunch of different response strategies for different types of requests all bunched together.

In addition to this approach, which is implemented on the global-router level, it would be a good developer experience to expose `Strategy`-level fallback support, with a handler that gets triggered whenever the "main" handler for a `Strategy` results in a `WorkboxError`.

This new `setCatchHandler` method on the `Strategy` base class would also be associated with a new TypeScript callback signature, that matched the normal `handlerCallback` signature, but included an additional `error` property, in case the developer wants access to the underlying error. (The `workbox-routing` `setCatchHandler()` should also use this new callback signature, and pass the underlying error through there as well.)

You can approximate this now by creating your own `handlerCallback` with a `try`/`catch` block and a call to a `Strategy` inside of it, but that's more awkward than it needs to be. Code that's currently:

```js
const networkFirst = new NetworkFirst({...});

const handlerWithFallback = async (params) => {
  try {
    // Hopefully you remember the await here, or catch will never be triggered!
    return await networkFirst.handle(params);
  } catch (error) {
    return caches.match('/offline');
  }
};
```

Could be rewritten as:

```js
const networkFirst = new NetworkFirst({...});

networkFirst.setCatchHandler((params) => {
  // params.error has the underlying error, if needed.
  return caches.match('/offline');
});
```

I think it should be pretty clean to support this with the new `Strategy` base class that @philipwalton put in place for v6.

jeffposnick:
R: @philipwalton

Fixes #2569

The one thing I wasn't 100% sure about was what should happen when there are multiple `handlerDidError` plugins defined for a given `Strategy`. The implementation in this PR will run them one by one until a truthy response is returned, at which point that becomes the effective response and any additional callbacks are not run.

### DIFF
--- a/packages/workbox-strategies/src/Strategy.ts
+++ b/packages/workbox-strategies/src/Strategy.ts
@@ -177,8 +177,8 @@ abstract class Strategy implements RouteHandlerObject {
       response = await this._handle(request, handler);
       // The "official" Strategy subclasses all throw this error automatically,
       // but in case a third-party Strategy doesn't, ensure that we have a
-      // consistent failure when there's no response.
-      if (!response) {
+      // consistent failure when there's no response or an error response.
+      if (!response || response.type === 'error') {
         throw new WorkboxError('no-response', {url: request.url});
       }
     } catch (error) {
--- a/packages/workbox-strategies/src/Strategy.ts
+++ b/packages/workbox-strategies/src/Strategy.ts
@@ -7,6 +7,9 @@
 */
 
 import {cacheNames} from 'workbox-core/_private/cacheNames.js';
+import {WorkboxError} from 'workbox-core/_private/WorkboxError.js';
+import {logger} from 'workbox-core/_private/logger.js';
+import {getFriendlyURL} from 'workbox-core/_private/getFriendlyURL.js';
 import {HandlerCallbackOptions, RouteHandlerObject, WorkboxPlugin}
     from 'workbox-core/types.js';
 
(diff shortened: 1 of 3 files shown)


jeffposnick:
(I'll do a follow-up PR for the proposed `PrecacheFallbackPlugin` bit, and the `runtimeCaching` changes for `generateSW` to use it.)

packages/workbox-strategies/src/Strategy.ts
```diff
@@ -168,11 +171,37 @@ abstract class Strategy implements RouteHandlerObject {
 
   async _getResponse(handler: StrategyHandler, request: Request, event: ExtendableEvent) {
     await handler.runCallbacks('handlerWillStart', {event, request});
-    let response = await this._handle(request, handler);
+
+    let response: Response | undefined = undefined;
+    try {
+      response = await this._handle(request, handler);
```
philipwalton, line 177: What do you think about the case where `_handle()` responds with a `Response.error()` object?

jeffposnick: I hadn't thought about that

 Looks like it can be [detected](https://fetch.spec.whatwg.org/#responsetype) via `response.type === 'error'`. I'll add in support for triggering `handlerDidError` when that's encountered.

jeffposnick: Updated.

philipwalton (approved):
LGTM, thanks for adding this! I think it nicely rounds out the plugin lifecycle.

PR #2579, opened 2020-07-14
Adds the PrecacheFallbackPlugin

jeffposnick:
R: @philipwalton

This is a follow-up to https://github.com/GoogleChrome/workbox/pull/2576

Once this is merged, I'm going to do one more PR to add in support for using this plugin from within `runtimeCaching` as using `generateSW`.

### DIFF
--- a/packages/workbox-precaching/src/PrecacheFallbackPlugin.ts
+++ b/packages/workbox-precaching/src/PrecacheFallbackPlugin.ts
@@ -8,7 +8,9 @@
 
 import {WorkboxPlugin} from 'workbox-core/types.js';
 
-import {matchPrecache} from './matchPrecache.js';
+import {getOrCreatePrecacheController} from
+    './utils/getOrCreatePrecacheController.js';
+import {PrecacheController} from './PrecacheController.js';
 
 import './_version.js';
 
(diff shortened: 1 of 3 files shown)


packages/workbox-precaching/src/PrecacheFallbackPlugin.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+   * Constructs a new PrecacheFallbackPlugin with the associated fallbackURL.
+   *
+   * @param {Object} config
+   * @param {string} config.fallbackURL A precached URL to use as the fallback
+   *     if the associated strategy can't generate a response.
+   */
+  constructor({fallbackURL}: {fallbackURL: string}) {
+    this._fallbackURL = fallbackURL;
```
philipwalton, line 50: I think you should add the ability to pass in an optional `PrecacheController` instance, which would default to calling `getOrCreatePrecacheController()` if not passed.

If we don't do that then the use of `matchPrecache()` below would only work if using the default controller.

packages/workbox-precaching/src/PrecacheFallbackPlugin.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  /**
+   * @return {Promise<Response>} The precache response for the fallback URL.
+   *
+   * @private
+   */
+  handlerDidError: WorkboxPlugin['handlerDidError'] =
+    () => matchPrecache(this._fallbackURL);
```
philipwalton: This should probably call `matchPrecache()` on the instance's internal `PrecacheController` instance (see above).

philipwalton (changes requested):

packages/workbox-precaching/src/PrecacheFallbackPlugin.ts
jeffposnick: Done.

packages/workbox-precaching/src/PrecacheFallbackPlugin.ts
jeffposnick: Done.

jeffposnick:
PTAL.

hpierre74 (approved):
👍 Looking forward for the next PR :)

philipwalton (approved):

PR #2581, opened 2020-07-15
publish_glitch gulp task update

jeffposnick:
(Moving this to a fresh PR as https://github.com/GoogleChrome/workbox/pull/2570 wasn't picking up the latest commits for reasons unknown.)

R: @philkrie 

The `publish-glitch.js` task wasn't properly converted to the new `gulp` task setup as part of #2517. The old task got imported into the v6 branch via #2537.

This PR updates the `publish-glitch.js` task to match what we're doing in the other v6 tasks. I don't have things set up to run it myself, so it would be appreciated if you could try it out for me via `gulp publish_glitch`, @philkrie  (note the underscore) and confirm that it works as expected.

### DIFF
--- a/gulp-tasks/publish-glitch.js
+++ b/gulp-tasks/publish-glitch.js
@@ -15,6 +15,8 @@ const olt = require('common-tags').oneLineTrim;
 const tempy = require('tempy');
 const upath = require('upath');
 
+const logHelper = require('../infra/utils/log-helper');
+
 const DEMOS_DIR = 'demos/src';
 
 async function publish_glitch() {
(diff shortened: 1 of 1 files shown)


philkrie:
thanks for the changes, just confirmed this works. might be good to add some more logging when a push completes successfully, but let's merge this and I can add that in a different PR.

jeffposnick:
Okay! In general, the logs in v6 are much less verbose, with the philosophy "if nothing was logged, that means it worked".

That might have been a bit too extreme in this case, so yeah, feel free to add in any logs after.

PR #2591, opened 2020-07-29
Warn when an async matchCallback is used

jeffposnick:
R: @philipwalton

Some dev-only logging to help developers who are writing their own `matchCallback` functions, and mistakenly write them as `async` functions.


jeffposnick:
FWIW, this isn't something that TypeScript could catch. Right now we have:

```
export interface RouteMatchCallback {
  (options: RouteMatchCallbackOptions): any;
}
```

Ideally that `any` would be "`any`, excluding `Promise`", but that's not possible until https://github.com/Microsoft/TypeScript/pull/29317 is available.

philipwalton (approved):

PR #2605, opened 2020-08-14
Tweaks to the publishing process

jeffposnick:
R: @philipwalton 

This tweaks the release process in two ways:

- We no longer generate .tar.gz and .zip archives of each release and attach them to the corresponding GitHub release, since there are many other ways of consuming the Workbox code nowadays, and it's not clear that anyone uses those archives.

- It explicitly requires passing in `--distTag=...` when running `gulp publish`, removing the previous logic that checked for the `master` branch name and switched between `latest` and `next` based on that. Being explicit is a bit more work, but since we may not go back to using `master`, it seemed like the thing to do.

### DIFF
--- a/gulp-tasks/publish.js
+++ b/gulp-tasks/publish.js
@@ -9,6 +9,7 @@
 const {series} = require('gulp');
 const execa = require('execa');
 const fse = require('fs-extra');
+const ol = require('common-tags').oneLine;
 
 const {publish_cdn} = require('./publish-cdn');
 const {publish_github} = require('./publish-github');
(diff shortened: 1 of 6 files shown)


gulp-tasks/publish.js
```diff
@@ -24,7 +25,14 @@ async function publish_sign_in_check() {
   await execa('npm', ['whoami']);
 }
 
+function dist_tag_check() {
```
jeffposnick, line 28: This is also checked inside of `publish_lerna`, but failing early (before the `test` task runs) seemed ideal.

philipwalton (approved):

Issue #2593, opened 2020-08-02
Deduplicate requests


ianldgs:
**Library Affected**:
*workbox-core, workbox-strategies*

**Browser & Platform**:
*all browsers*

**Issue or Feature Request Description**:
*Support for deduplicating requests.*
Or
*Support for creating custom strategies*
Or
*Extend plugin API*

**Use case**
*2 API calls to the same endpoint, with the same query, being done at the same time*

**What I've got so far**
*I have cloned the repo (V6) and created a custom strategy. However, the structure is not being very helpful for this case. I cannot base my strategy on another strategy, or even use the strategy as secondary and it depends on a lot of internals.**

**Idea**
*I do understand that, apart from the problems, this strategy wouldn't be as useful as the others currently provided and that creating custom strategies isn't something that people do every day. So the most elegant solution would probably be to provide a `fetch` method on the plugin API. The `requestWillFetch` doesn't work for this case.*

PS: If you like the idea, I can create a pull request. I've got a bit familiar with the project and internals.

jeffposnick:
Hmm, that's interesting. I'd like to hear more about exactly how you envision this working, and whether caching comes into play at all, or if it's network-only.

Here's what I _think_ you're asking for: a custom `Strategy` that will detect whether a request is for a URL that is already in the middle of being requested. While there's already an in-flight request, the second request will pause; once there's a response for the first request, that response will be cloned and used to satisfy all other paused requests.

In that model, there's no caches that come into play, and the pausing only happens if there is an actual overlap between two inflight requests; if two requests are made sequentially, without and overlap, then the second one would behave independent of the first.

Does that sound like what you had in mind?

jeffposnick:
I also wanted to note that, especially with the upcoming support for custom `Strategy` subclasses in Workbox v6, we're really hoping that developers will address creative use cases like this not with pull requests against Workbox, but by publishing their own `Strategy` classes and plugins on `npm`.

An ecosystem of third-party modules that worked with Workbox would be 💯 .

ianldgs:
> Hmm, that's interesting. I'd like to hear more about exactly how you envision this working, and whether caching comes into play at all, or if it's network-only.
> 
> Here's what I _think_ you're asking for: a custom `Strategy` that will detect whether a request is for a URL that is already in the middle of being requested. While there's already an in-flight request, the second request will pause; once there's a response for the first request, that response will be cloned and used to satisfy all other paused requests.
> 
[...]

PR #2608, opened 2020-08-16
DX improvements for extending Strategy classes

jeffposnick:
R: @philipwalton 

This fixes a couple of awkward things I found when extending `NetworkFirst` in https://github.com/GoogleChrome/workbox/issues/2593#issuecomment-674455562:

- The log messages should use `this.constructor.name` instead of a hardcoded string.
- `NetworkFirstOptions` wasn't exported, making it awkward to call the `NetworkFirst` constructor via `super()` without using `any`.

### DIFF
--- a/packages/workbox-core/src/_private/dontWaitFor.ts
+++ b/packages/workbox-core/src/_private/dontWaitFor.ts
@@ -12,7 +12,7 @@ import '../_version.js';
  *
  * @private
  **/
-export function dontWaitFor(promise: Promise<any>) {
+export function dontWaitFor(promise: Promise<any>): void {
   // Effective no-op.
   promise.then(() => {});
 }
--- a/packages/workbox-strategies/src/NetworkFirst.ts
+++ b/packages/workbox-strategies/src/NetworkFirst.ts
@@ -17,7 +17,7 @@ import {messages} from './utils/messages.js';
 import './_version.js';
 
 
-interface NetworkFirstOptions extends StrategyOptions {
+export interface NetworkFirstOptions extends StrategyOptions {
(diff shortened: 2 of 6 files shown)


packages/workbox-core/src/_private/dontWaitFor.ts
```diff
@@ -12,7 +12,7 @@ import '../_version.js';
  *
  * @private
  **/
-export function dontWaitFor(promise: Promise<any>) {
+export function dontWaitFor(promise: Promise<any>): void {
```
jeffposnick, line 15: I think the `tsc` dependency bump did something to make this required.

https://github.com/GoogleChrome/workbox/pull/2608/checks?check_run_id=991709793#step:6:131

packages/workbox-strategies/src/NetworkFirst.ts
```diff
@@ -17,7 +17,7 @@ import {messages} from './utils/messages.js';
 import './_version.js';
 
 
-interface NetworkFirstOptions extends StrategyOptions {
+export interface NetworkFirstOptions extends StrategyOptions {
```
jeffposnick, line 20: I believe we agreed that these should be sub-module exports rather than top-level exports, right? This matches what we're doing for the `StrategyOptions` interface as well, and means you have to do `import {NetworkFirstOptions} from 'workbox-strategies/NetworkFirst'`.

https://github.com/GoogleChrome/workbox/blob/0f52e63616888bb6b568ed8dba0ec049654529c5/packages/workbox-strategies/src/Strategy.ts#L20-L25

philipwalton (approved):

PR #2616, opened 2020-08-24
Allow cacheKeyWillBeUsed to influence the request method check

### LINKED ISSUE #2615: Allow post caching
**Library Affected**:
workbox-strategies

**Browser & Platform**:
all browsers

Allow post caching

Developers should be able to cache `post` requests as `get` requests using `cacheKeyWillBeUsed`.  Currently there is a check that prevents post caching, which checks the initial request instead of the request returned by `cacheKeyWillBeUsed`, which should be respected as a custom cache key.

A related issue #2574 was filed but closed by the author.

markbrocato:
Fixes #2615

Moves the request method check after the call to `cacheKeyWillBeUsed` so that developers can potentially cache `post` requests as `get` requests.

### DIFF
--- a/packages/workbox-strategies/src/StrategyHandler.ts
+++ b/packages/workbox-strategies/src/StrategyHandler.ts
@@ -197,7 +197,13 @@ class StrategyHandler {
    */
   async cachePut(key: RequestInfo, response: Response): Promise<void> {
     const request: Request = toRequest(key);
-
+    
+    // Run in the next task to avoid blocking other cache reads.
+    // https://github.com/w3c/ServiceWorker/issues/1397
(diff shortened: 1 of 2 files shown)


markbrocato:
@googlebot I signed it!

jeffposnick (approved):
Thanks for the contribution! As mentioned in https://github.com/GoogleChrome/workbox/issues/2615, I'm not sure that I want to advertise this as a way to cache arbitrary `POST` requests, but it does makes sense to perform the method check on the `cacheKeyWillBeUsed` return value.

If folks who know what they're doing use that as a way to cache `POST` requests, then so be it...

PR #2620, opened 2020-08-26
networkTimeoutSeconds in NetworkOnly

### LINKED ISSUE #2030: Add networkTimeoutSeconds to NetworkOnly strategy
At present, it is possible to set timeout only for ``NetworkFirst`` strategy:

```
workbox.strategies.networkFirst({networkTimeoutSeconds: 10});
```

When ``NetworkOnly`` strategy is used to retrieve _non-cacheable data_, it keeps waiting indefinitely in case of a network issue.

Suggested implementation: https://github.com/github/fetch/issues/175#issuecomment-216791333

jeffposnick:
R: @philipwalton
CC: @Snugug 

Fixes #2030

This adds support for `networkTimeoutSeconds` to the `NetworkOnly` strategy, for folks who are okay with a network request failing after a period of time.


packages/workbox-strategies/src/NetworkOnly.ts
```diff
@@ -42,15 +68,29 @@ class NetworkOnly extends Strategy {
...
+      const promises: Promise<Response|undefined>[] = [handler.fetch(request)];
+
+      if (this._networkTimeoutSeconds) {
+        const timeoutPromise = new Promise((resolve: (_: undefined) => void) => {
+          setTimeout(resolve, this._networkTimeoutSeconds * 1000);
+        });
+        promises.push(timeoutPromise);
+      }
+
+      response = await Promise.race(promises);
```
jeffposnick, line 84: I know `Promise.race()` is often not the right choice... but here I think it is?

packages/workbox-strategies/src/NetworkOnly.ts
```diff
@@ -42,15 +68,29 @@ class NetworkOnly extends Strategy {
...
-    let response;
+    let error: Error | undefined = undefined;
+    let response: Response | undefined;
+
     try {
-      response = await handler.fetch(request);
+      const promises: Promise<Response|undefined>[] = [handler.fetch(request)];
+
+      if (this._networkTimeoutSeconds) {
+        const timeoutPromise = new Promise((resolve: (_: undefined) => void) => {
```
philipwalton: This can use the `timeout` function in `workbox-core`: https://github.com/GoogleChrome/workbox/blob/v6/packages/workbox-core/src/_private/timeout.ts

packages/workbox-strategies/src/NetworkOnly.ts
```diff
@@ -30,6 +34,28 @@ import './_version.js';
  * @memberof module:workbox-strategies
  */
 class NetworkOnly extends Strategy {
+  private readonly _networkTimeoutSeconds: number;
+
+  /**
+   * @param {Object} options
```
philipwalton: Note, all these options are optional, but they're not indicated as such by the jsdocs.

I've made that change as part of my [`precaching-v6` changes](https://github.com/GoogleChrome/workbox/compare/v6...precaching-v6#diff-0c320ed20894cc8a17c07ff33737889a), but you may as well make them here to avoid merge conflicts.

packages/workbox-strategies/src/NetworkOnly.ts
jeffposnick: Switched.

packages/workbox-strategies/src/NetworkOnly.ts
jeffposnick: Done.

jeffposnick:
@philipwalton, PTAL.

PR #2627, opened 2020-09-10
Add missing spaces in multiline logs

piotr-cz:
R: @jeffposnick @philipwalton 

Add missing spaces in multiline logs

Should I post PR against `v6` or `master` branch?
The `gulp lint` command shows lots of unrelated issues, and `gulp test` breaks with meassage `Error: Cannot find module '@hapi/joi'`

### DIFF
--- a/packages/workbox-background-sync/src/Queue.ts
+++ b/packages/workbox-background-sync/src/Queue.ts
@@ -323,14 +323,14 @@ class Queue {
         await fetch(entry.request.clone());
 
         if (process.env.NODE_ENV !== 'production') {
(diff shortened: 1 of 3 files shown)


jeffposnick (approved):
Thanks for the contribution!

Yes, PR this should target the `v6` branch.

The CI run is happy, so I think you might have just run into dependency trouble when switching between the `v5` and `v6` codebase?

piotr-cz:
It's because simple  `npm install` doesn't install subpackages dependencies like '@hapi/joi'

jeffposnick:
Ah. The dependencies of the various `packages/*` should all end up installed by `lerna` as a side-effect of running `gulp build`.

PR #2639, opened 2020-09-26
Precaching updates for v6

### LINKED ISSUE #1857: workbox-precaching routes are not compatible with other packages

### LINKED ISSUE #2402: workbox-precaching should have precedence over workbox-routing
**Library Affected**:
*workbox-precaching*

We currently delay registering the `fetch` handler for `workbox-precaching` until the point where `precacheAndRoute()` is called:

https://github.com/GoogleChrome/workbox/blob/a284beedb84f7e5d36694fce60bb4267000b1d5a/packages/workbox-precaching/src/precacheAndRoute.ts#L35-L38

We should consider registering this earlier, as a side effect of importing `workbox-precaching`. If the user doesn't end up calling `precacheAndRoute()`, then the `fetch` handler will effectively be a no-op. Alternatively, we should consider finding a way to share a common `fetch` handler between `workbox-precaching` and `workbox-routing`.

This works around an unintuitive issues where including any reference to `workbox-routing`'s `registerRoute()` prior to calling `precacheAndRoute()` would cause *all* runtime caching routes that are registered to take precedence of `workbox-precaching`, even if those runtime caching routes are listed after `precacheAndRoute()`.

The downside is that if a developers *wants* a runtime caching route registered with `workbox-routing` to take precedence over precaching, they won't be able to accomplish that. I'd argue that they shouldn't precache a URL that they plan on serving with runtime caching, but maybe there are some valid use cases I'm not thinking of.

This recently caused some confusion for the https://web.dev/ folks. CC: @samthor @philipwalton

### LINKED ISSUE #2574: Workbox POST request can't be cached.
**Library Affected**:
*workbox-core*

Hi @jeffposnick 

I have an url which is something like this : `https://api.com/data` .. In my app, I use `axios` to make requests in multiple places for this same url, but with different request payloads.

1) One page might create a request with the body: `example:1` and another one `example:2`. Since the url for these two are the same, I can't just use the route and strategy of workbox. So, should I use `cacheKeyWillBeUsed` and figure out in there what the request payload's key is to distinguish between these same urls, but different payloads ? 

2) I tried this:

```js
registerRoute(
        new RegExp(`${config.api.url.replace(/\./g,'\\.')}.*/metric`),
        new NetworkFirst({
            cacheName: config.app.cache.api,
        }),
        'POST'
    )
```

but it tells me: 

cacheWrapper.js:192 Uncaught (in promise) attempt-to-cache-non-get-request: Unable to cache 'https://test.com/metric' because it is a 'POST' request and only 'GET' requests can be cached.

Then I looked through the src code and found this:

```js
    if (request.method && request.method !== 'GET') {
      throw new WorkboxError('attempt-to-cache-non-get-request', {
        url: getFriendlyURL(request.url),
        method: request.method,
      });
    }
  }
```

This code is located in `workbox-core/src/_private/cacheWrapper.ts`.

I tried `CacheFirst` strategy too, but with no luck.  Also it looks like service worker intercepts POST request , but maybe, this error is due to the fact that after getting the response back, workbox can't put it into the Cache API because Cache API doesn't support this. Is that right and the only choice i got is to store it in indexeddb?

### LINKED ISSUE #2638: [RFC] v6 Precaching Updates and Changes
## Objective

Decouple routing and strategy logic from the `PrecacheController` class in order to:

* Reduce bugs/confusion stemming from route order/precedence of precached routes.
* Give developers more control over how assets are stored and served from the precache
* Reduce bundle sizes for developers not using many of the existing built-in precaching options
* Make it possible (in the future) to swap out `PrecacheStrategy` for a custom strategy

## Background

The `workbox-precaching` package was [recently updated](https://github.com/GoogleChrome/workbox/pull/2459) to use modules from `workbox-strategies` and `workbox-routing` under the hood, but it did not change the API—which means the degree of decoupling was limited, and it could not fully address several issues:

* https://github.com/GoogleChrome/workbox/issues/1857
* https://github.com/GoogleChrome/workbox/issues/2402
* https://github.com/GoogleChrome/workbox/issues/2574#issuecomment-672572043

With v6 being a major release and breaking changes being possible, this proposals continues the work started in [#2459](https://github.com/GoogleChrome/workbox/pull/2459) and fully removes any custom route-matching or strategy-handling logic done in `workbox-precaching`. All of that work is now being done by subclasses of `workbox-routing` or `workbox-strategy` classes.

## New modules

This proposal introduces several new modules (both public and private) to the workbox-precaching packages. These modules are largely self-contained and independent, so developers will only pay the cost for the modules they import:

### PrecacheStrategy

`PrecacheStrategy` is a [custom strategy](https://github.com/GoogleChrome/workbox/issues/2446) that can populate the precache using the same logic currently used in v5. This includes:

* Adding a request to the cache on install
* Retrieving a request from the cache on fetch
* Falling back to network on cache miss (without updating the cache)
* Handling same-origin redirects
* Handling opaque responses (if a `cacheWillUpdate` plugin callback isn't used).

```js
interface PrecacheStrategyOptions extends StrategyOptions {
  fallbackToNetwork: boolean;
}

declare class PrecacheStrategy extends Strategy {
  constructor(options?: PrecacheStrategyOptions);
}
```

Note, two items explicitly NOT handled by the `PrecacheStrategy` class are:

* Handling revision cache keys
* Generating a list of updated/not-updated URLs on install

These are handled by the `PrecacheCacheKeyPlugin` and `PrecacheInstallReportPlugin` plugins (respectively), and both of these plugins are added to the strategy by the `PrecacheController` on instantiation.

There are two primary reasons this logic is handled by plugins rather than `PrecacheStrategy` itself:

* To avoid coupling of concerns between the `PrecacheController` (which maintains the precache manifest and cacheKey mapping) and the `PrecacheStrategy` (which handles how requests are fetched/cached).
* To make it possible (in the future, not in v6) for a developer to use their own custom strategy (or a built-in strategy like `CacheFirst`) with precaching.

_**Note:** while this proposal doesn't include any API changes to support custom precaching strategies, it does make it possible through module aliasing with a build tool. While this is not currently recommended, the fact that it does work validates the flexibility of the new design._

### PrecacheRoute

`PrecacheRoute` is a custom route that gets created passing it a `PrecacheController` instance as well as the [configuration options](https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-precaching#.addRoute) currently accepted by the `addRoute()` method in v5.

The route will use the urls-to-cache-keys map of the `PrecacheController` instance to determine whether an incoming request "matches", and it will use the strategy created by the controller as the handler. The matcher will also pass the `cacheKey` value as a custom param to the handler so it doesn't have to re-lookup that value.

Since the `PrecacheRoute` logic in v6 is separate from the `PrecacheController` logic, developers who don't need to use the [configuration options](https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-precaching#.addRoute) can avoid bundling that logic in their service worker file.

```ts
declare class PrecacheRoute extends Route {
  constructor(pc: PrecacheController, options?: FetchListenerOptions);
}
```

### PrecacheCacheKeyPlugin _(private)_

`PrecacheCacheKeyPlugin` (not publicly exposed) is a custom plugin that gets automatically set on the strategy created by the `PrecacheController`.

The plugin contains a `cacheKeyWillBeUsed` callback to map request URLs to revisioned assets in the precache.

As mentioned above, the rationale for putting this logic into its own plugin (rather than as part of `PrecacheStrategy`) is that it can then be used with any other strategy as well.

```ts
declare class PrecacheCacheKeyPlugin implements WorkboxPlugin {
  constructor({ precacheController }: {
      precacheController: PrecacheController;
  });
  cacheKeyWillBeUsed: WorkboxPlugin['cacheKeyWillBeUsed'];
}
```

### PrecacheInstallReportPlugin _(private)_

Similar to `PrecacheCacheKeyPlugin`, the `PrecacheInstallReportPlugin` (not publicly exposed) gets automatically set on the strategy created by the `PrecacheController`, and it's responsible for generating the list of updated (and not updated) URLs during service worker installation.

For each handler instance run during the `install` event, if the URL is found in the cache already, it'll be added to the `notUpdatedURLs` array, otherwise it'll be added to the `updatedURLs` array.

```ts
declare class PrecacheInstallReportPlugin implements WorkboxPlugin {
  updatedURLs: string[];
  notUpdatedURLs: string[];
  handlerWillStart: WorkboxPlugin['handlerWillStart'];
  cachedResponseWillBeUsed: WorkboxPlugin['cachedResponseWillBeUsed'];
}
```

Again, this logic is implemented via a plugin to make it possible for other strategies to be used with `PrecacheController`, and to make it possible for the `PrecacheStrategy` to be used as a runtime strategy (not sure if there are real use cases for this, but it would be technically possible).

## Breaking changes

This proposal intentionally limits the amount of breaking changes to the top level function modules exported by `workbox-precaching`. Most of the breaking changes are scoped to the lesser-used `PrecacheController` class and to the recently-introduced (and also lesser-used) `createHandler` and `createHandlerBoundToURL` functions.

#### Changes to the `constructor()` parameter{#changes-to-the-constructor-parameters}

In v5 the only option you could pass to the `PrecacheController` constructor was a single `cacheName` param. In v6 the `PrecacheController` will take an object in order to make it more extensible in the future.

The current set of accepted object properties is:

* `cacheName`
* `plugins`
* `fallbackToNetwork` (defaults to `true`)

```js
// In v5

const pc = new PrecacheController('precache');

pc.addPlugins([...])

// In v6

const pc = new PrecacheController({
  cacheName: 'precache',
  plugins: [...]
});
```

In v5 the only way to add plugins was via the top-level `addPlugins()` function, or by passing them to the `install()` method of `PrecacheController`.

In v6, plugins are passed when creating the `PrecacheController` instance.

Lastly, in v6 the `fallbackToNetwork` option (previously set when calling [`createHandler()`](https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-precaching.PrecacheController#createHandler) or [`createHandlerBoundToURL()`](https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-precaching.PrecacheController#createHandlerBoundToURL)) is now set when creating the `PrecacheController` instance, which means it will now the same setting will apply to all handled requests.

If a developer needs different fallback logic for different resources, they will need to create multiple `PrecacheController` instances.

#### Changes to the `createHandlerBoundToURL()` method parameter{#changes-to-the-createhandlerboundtourl-method-parameters}

The `createHandlerBoundToURL` method no longer accepts a second, `fallbackToNetwork` parameter.

As mentioned above, the fallback-to-network logic is now set on the `PrecacheController` instance itself (and applied to the `PrecacheStrategy` the controller creates upon instantiation) and thus it applies to all requests handled by the controller.

_**Note:** the same change will be made to the top-level `createHandlerBoundToURL()` function, see below._

#### Changes to the `install()` method parameter{#changes-to-the-install-method-parameters}

The install method previously took an optional `{event, plugins}` config object. In v6 it will only take an `event` object, which will be required.

In addition (and not technically a breaking change), the install event will now automatically call `event.waitUntil()`, since it's easy for developers to forget that this is needed, which can cause installations to fail.

```js
// In v5

const pc = new PrecacheController();

addEventListener('install', (event) => event.waitUntil(pc.install({event})));

// In v6

const pc = new PrecacheController();

addEventListener('install', pc.install);
```

#### Changes to the `activate()` method parameter{#changes-to-the-activate-method-parameters}

Similar to the `install()` method, the `activate()` method will also now require an event (previously it took no parameters), and that event will automatically call `event.waitUntil()`.

```js
// In v5

const pc = new PrecacheController();

addEventListener('activate', (event) => event.waitUntil(pc.activate()));

// In v6

const pc = new PrecacheController();

addEventListener('activate', pc.activate);
```

#### Remove the `addRoute()` method

Since this proposal introduces a new `PrecacheRoute` class, the `addRoute()` method on `PrecacheController` will be removed in favor of developers explicitly adding routes for precaching.

This will help prevent cases where the order of initializing precaching can affect what routes match particular requests:

```js
// In v5

const pc = new PrecacheController();

// Adds a fetch event listener to match precaching requests
pc.addRoute({...});

// In v6

const pc = new PrecacheController();

// Creates and registers a route explicitly on the default router.
registerRoute(new PrecacheRoute(pc, {...}))
```

_**Note:** for better backwards compatibility, the top-level `addRoute()` function will continue to work as it does today, only the `PrecacheController` method is being removed._

#### Remove the `precacheAndRoute()` method

Since the `addRoute()` method is being removed (see above), it also makes sense to remove the `precacheAndRoute()` method from the `PrecacheController` class.

_**Note:** as with the `addRoute()` method, the top-level `precacheAndRoute()` function will continue to be available, it's just the `PrecacheContoller` method that's being removed.

#### Remove the `createMatchCalback()` metho{#remove-the-creatematchcalback-method}

The primary purpose of the `createMatchCallback()` method was to create a route to match precached assets, but with this proposal's addition of the `PrecacheRoute` class, this is no longer needed.

#### Remove the `createHandler()` method

Since the `PrecacheController` object now creates and exposes a strategy instance, and since all strategies have a `handle()` method, the `createHandler()` method of `PrecacheController` will be removed.

Developers who need a handler to respond to requests for precached assets can use the strategy directly:

```js
// In v5

const pc = new PrecacheController();

registerRoute(
  ({request}) => request.destination === 'script',
  pc.createHandler(),
)

// In v6

const pc = new PrecacheController();

registerRoute(
  ({request}) => request.destination === 'script',
  pc.strategy,
);
```

The `createHandler()` method also took a configuration option `fallbackToNetwork`, which defaulted to `true`. Developers who do not want their precaching logic to fall back to network can set this to false when creating the `PrecacheController` instance:

```js
// In v5

const pc = new PrecacheController();

registerRoute(
  ({request}) => request.destination === 'script',
  pc.createHandler({fallbackToNetwork: false}),
)

// In v6

const pc = new PrecacheController({fallbackToNetwork: false});

registerRoute(
  ({request}) => request.destination === 'script',
  pc.strategy,
);
```

### createHandlerBoundToURL()

As with the changes to the `createHandlerBoundToURL()` method (mentioned above), the top-level `createHandlerBoundToURL()` method no longer accepts a second `fallbackToNetwork` parameter.

Developers who do not want the default behavior of falling back to network on precache misses will not be able to use the default precach controller, they'll need to create their own `PrecacheController` instance and set the `fallbackToNetwork` option to `false`.

### createHandler()

As with the removal of the `createHandler()` method (mentioned above), the top-level `createHandler()` will be removed.

Developers who need to use a handler for precached assets will not be able to use the default precache controller, they'll need to create their own `PrecacheController` instance and use its `strategy` property.

### Precaching an app shell

```js
import {PrecacheController} from 'workbox-precaching';
import {NavigationRoute, registerRoute} from 'workbox-routing';

const pc = new PrecacheController();

pc.precache([
  // Manifest goes here...
]);

const handler = pc.createHandlerBoundToURL('/app-shell.html');
const navigationRoute = new NavigationRoute(handler, {
  allowlist: [...],
  denylist: [...],
});
registerRoute(navigationRoute);
```

### Precaching with custom routing

```js
import {PrecacheController} from 'workbox-precaching';
import {registerRoute} from 'workbox-routing';

const pc = new PrecacheController({cacheName: 'precache'});

pc.precache([
  // Manifest goes here...
]);

registerRoute(
  ({request}) => request.destination === 'script',
  pc.strategy
);
```

## WIP Implementation

https://github.com/GoogleChrome/workbox/tree/precaching-v6

philipwalton:
R: @jeffposnick 

Fixes #2638, fixes #1857, fixes #2402, fixes #2574

This PR implements the changes proposed in [[RFC] v6 Precaching Updates and Changes #2638
](https://github.com/GoogleChrome/workbox/issues/2638)


packages/workbox-core/src/_private/cacheMatchIgnoreParams.ts
```diff
@@ -0,0 +1,56 @@
...
+  const strippedURL = new URL(fullURL);
+  for (const param of ignoreParams) {
+    strippedURL.searchParams.delete(param);
+  }
+  return strippedURL.href;
+}
+
+
+/**
+ * Matches an item in the cache, ignoring specific URL params. This is
```
jeffposnick: missing word: "similar"

packages/workbox-core/src/_private/cacheMatchIgnoreParams.ts
```diff
@@ -0,0 +1,56 @@
...
+ * @param {Cache} cache
+ * @param {Request} request
+ * @param {Object} matchOptions
+ * @param {Array<string>} ignoreParams
+ * @return {Promise<Response|undefined>}
+ */
+export async function cacheMatchIgnoreParams(
+  cache: Cache,
+  request: Request,
+  ignoreParams: string[],
```
jeffposnick, line 35: Maybe the intended usage is a bit different, but we've supported ignoring parameters via an array of RegExps, which so that you can, e.g., ignore everything that starts with `utm_`.

Is there a reason that this code takes in strings instead?

packages/workbox-core/src/_private/cacheMatchIgnoreParams.ts
```diff
@@ -0,0 +1,56 @@
...
+  matchOptions?: CacheQueryOptions,
+): Promise<Response | undefined> {
+  const strippedRequestURL = stripParams(request.url, ignoreParams);
+
+  // If the request doesn't include any ignored params, match as normal.
+  if (request.url === strippedRequestURL) {
```
jeffposnick, line 41: I must be missing something, but why isn't this whole method just:

```
const strippedRequestURL = stripParams(request.url, ignoreParams);
return cache.match(strippedRequestURL, matchOptions);
```

and you could omit the whole process of iterating over the cache keys, etc.?

packages/workbox-core/src/_private/waitUntil.ts
```diff
@@ -0,0 +1,26 @@
...
+ * @return {Function}
+ * @private
+ */
+function waitUntil(event: ExtendableEvent, asyncFn: () => Promise<any>) {
+  const returnPromise = asyncFn();
+  event.waitUntil(returnPromise);
+  return returnPromise;
+}
+
+export {waitUntil}
```
jeffposnick: Please add a missing `;`

packages/workbox-precaching/src/PrecacheRoute.ts
```diff
@@ -0,0 +1,67 @@
...
+  license that can be found in the LICENSE file or at
+  https://opensource.org/licenses/MIT.
+*/
+
+import {logger} from 'workbox-core/_private/logger.js';
+import {getFriendlyURL} from 'workbox-core/_private/getFriendlyURL.js';
+import {RouteMatchCallback, RouteMatchCallbackOptions} from 'workbox-core/types.js';
+import {Route} from 'workbox-routing/Route.js';
+
+import {PrecacheRouteOptions, } from './_types.js';
```
jeffposnick: Please remove the `, `.

packages/workbox-core/src/_private/cacheMatchIgnoreParams.ts
```diff
@@ -0,0 +1,56 @@
...
+ * params (while continuing to match on the others).
+ *
+ * @private
+ * @param {Cache} cache
+ * @param {Request} request
+ * @param {Object} matchOptions
+ * @param {Array<string>} ignoreParams
+ * @return {Promise<Response|undefined>}
+ */
+export async function cacheMatchIgnoreParams(
```
jeffposnick: We've been following the pattern of declaring things first, and then saving all the `export {...};` statements for the end instead of declaring and exporting at the same time.

packages/workbox-core/src/_private.ts
```diff
@@ -20,12 +21,14 @@ import {getFriendlyURL} from './_private/getFriendlyURL.js';
...
 import {resultingClientExists} from './_private/resultingClientExists.js';
 import {timeout} from './_private/timeout.js';
+import {waitUntil} from './_private/waitUntil.js';
 import {WorkboxError} from './_private/WorkboxError.js';
 
 import './_version.js';
 
 export {
   assert,
+  cacheMatchIgnoreParams,
```
jeffposnick, line 31: Is `cacheMatchIgnoreParams()` calling anywhere internally, or is it just meant as a new, externally-visible helper method? I don't see it in the RFC.

As mentioned later, it's awkward that it takes in an array of strings and not RegExps, as RegExps are used elsewhere for a similar purpose.

jeffposnick (changes requested):
This PR mentions that it's an "Initial" set of updates. How complete is it? What's missing?

I am a little confused about what ends up invoking `_handle()` on the `PrecacheStrategy`, and I am not sure if that's because this isn't yet the full set of changes.

jeffposnick:
Sorry, after realizing that there were several large files whose diffs were hidden, I need to re-review this PR.

packages/workbox-precaching/src/PrecacheController.ts
```diff
@@ -35,83 +33,55 @@ declare global {
...
-  private readonly _urlsToCacheKeys: Map<string, string>;
-  private readonly _urlsToCacheModes: Map<string, "reload" | "default" | "no-store" | "no-cache" | "force-cache" | "only-if-cached">;
-  private readonly _cacheKeysToIntegrities: Map<string, string>;
-
-  private _router?: Router;
-  private _strategy?: Strategy;
-  private _installed?: boolean = false;
-  private readonly _cacheKeyPlugin: WorkboxPlugin;
-  private readonly _plugins: WorkboxPlugin[] = [];
+  private _installAndActiveListenersAdded?: boolean = false;
```
jeffposnick: Since it's initialized to `false`, I don't think the `?` annotation is appropriate.

packages/workbox-precaching/src/PrecacheController.ts
```diff
@@ -234,135 +174,77 @@ class PrecacheController {
...
-      if (plugins) {
-        assert!.isArray(plugins, {
-          moduleName: 'workbox-precaching',
-          className: 'PrecacheController',
-          funcName: 'install',
-          paramName: 'plugins',
+  install(event: ExtendableEvent) {
+    return waitUntil(event, async () => {
+      const installReportPlugin = new PrecacheInstallReportPlugin();
+      this.strategy.plugins.push(installReportPlugin);
```
jeffposnick, line 187: Is there any use case in which someone might call `install()` manually, instead of having it invoked by the event listener?

If so, if they end up calling it twice, this plugin will be added twice.

(Feel free to ignore if you don't think that's a scenario that's likely to happen.)

jeffposnick (changes requested):
Thanks; this mostly looks good now that I have found the entirety of the code changes.

packages/workbox-core/src/_private/cacheMatchIgnoreParams.ts
philipwalton: Done

packages/workbox-core/src/_private/waitUntil.ts
philipwalton: Done.

packages/workbox-precaching/src/PrecacheRoute.ts
philipwalton: Done.

packages/workbox-core/src/_private/cacheMatchIgnoreParams.ts
philipwalton: Done.

packages/workbox-precaching/src/PrecacheController.ts
philipwalton: I kept the `?` and didn't initialize to false to save a few bytes.

packages/workbox-precaching/src/PrecacheController.ts
philipwalton: Hmmm, there's definitely a use case for calling `install()` manually (e.g. you want to get the list of updated and not updated URLs). However, I don't think it's an issue to add the plugin more than once, since it's only used to determine the `install()` method's return value.

Calling the install method more than once—in general—isn't great, but that was always possible. In that case the second run would discover that everything is cached and report no updated URLs.

We could try to prevent calling it more than once, but since it's not really destructive (and users should see in the logs that it's happening twice), I'm not sure it's worth the additional code to prevent the possibility.

WDYT?

jeffposnick: Okay, we can leave this as-is if it's not a big concern.

jeffposnick (approved):
Thanks for implementing this!

PR #2641, opened 2020-10-01
webpack v5 compatibility

### LINKED ISSUE #2630: webpack v5 compatibility
**Library Affected**:

- workbox-webpack-plugin

**Browser & Platform**:

None

webpack 5 [exposes `webpack-sources`](https://github.com/webpack/webpack/pull/11345) via `webpack.sources`, and it's advised to use it instead of `require('webpack-sources')`, if you agree to do it, I can submit a pull request.

jeffposnick:
R: @philipwalton @housseindjirdeh (who might have some more `webpack` background)

Fixes #2630 

There are unfortunately a lot of changes; they basically follow the guidance I received from the webpack team at https://github.com/webpack/webpack/issues/11425

The codebase should work with webpack v4.40.0 or higher or webpack v5.0.0, so that's a breaking change (we previously worked with webpack v4.0.0 or higher).

There are now two different sets of tests, each going against a different major version of webpack. (Thanks to @developit for help getting the proper node module resolution set up.)

One major outstanding issue is compatibility with the latest `html-webpack-plugin` (see https://github.com/jantimon/html-webpack-plugin/issues/1522); I believe they still need to do some work to properly support webpack v5.0.0, and in the meantime, I would consider using our plugin alongside it unsupported. I did not want that to block getting this new code out there for the Workbox v6 prereleases, though.


test/workbox-webpack-plugin/node/v4/generate-sw.js
```diff
@@ -6,24 +6,35 @@
   https://opensource.org/licenses/MIT.
 */
 
+// workbox-webpack-plugin needs to do require('webpack'), and in order to test
+// against multiple webpack versions, we need that to resolve to whatever the
+// correct webpack is for this test.
```
philipwalton, line 11: Maybe add a link somewhere to [your post about this](https://jeffy.info/2020/10/01/testing-multiple-webpack-versions.html), so anyone looking at this in the future can understand why this was done?

philipwalton (approved):
LGTM (though not being super familiar with webpack, I can't really provide much feedback on the implementation).

test/workbox-webpack-plugin/node/v4/generate-sw.js
jeffposnick: Sure.

jeffposnick:
I added the comment to the test and updated the `devDependency` to the webpack `v5.0.0` release.

PR #2645, opened 2020-10-07
Adjust PrecacheEntry type to allow null revisions

### LINKED ISSUE #2644: PrecacheEntry type does not allow setting revision to null
**Library Affected**:
*workbox-precaching*

**Browser & Platform**:
*all browsers*

When passing precaching entries with revision information in the URL itself, the revision property should be set to `null`.

The documentation describing this can be found here: https://developers.google.com/web/tools/workbox/modules/workbox-precaching#explanation_of_the_precache_list

The types included with `workbox-precaching` currently do not allow this. Instead, a revision of either `string` or `undefined` is allowed. The types should be adjusted to also allow `null`.

Manduro:
This allows the revision to be set to null, which is necessary when the revisioning information is in the URL itself.

R: @jeffposnick @philipwalton

Fixes #2644

### DIFF
--- a/packages/workbox-precaching/src/_types.ts
+++ b/packages/workbox-precaching/src/_types.ts
@@ -20,7 +20,7 @@ export interface CleanupResult {
 export interface PrecacheEntry {
   integrity?: string;
   url: string;
-  revision?: string;
+  revision?: string | null;
 }
 export interface PrecacheRouteOptions {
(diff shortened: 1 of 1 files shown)


jeffposnick:
Thanks! I remember finding that annoying once and then forgot to make the change 😄

PR #2656, opened 2020-10-23
Mark BackgroundSyncPlugin options param as optional

jeffposnick:
R: @philipwalton

I just bumped into this when using the TypeScript definition of the `BackgroundSyncPlugin` constructor. I am pretty sure that the `options` param is meant to be optional.


jeffposnick:
(Actually, there are a few more plugins that have the same problem.)

jeffposnick:
Okay, modified `BroadcastUpdatePlugin` as well.

philipwalton (approved):

PR #2664, opened 2020-11-03
Feature/workbox recipes

### LINKED ISSUE #2622: Make recipes easier to consume
There are two sets of recipes that Workbox has published, [common recipes](https://developers.google.com/web/tools/workbox/guides/common-recipes) and [advanced recipes](https://developers.google.com/web/tools/workbox/guides/advanced-recipes), and there are likely other patterns the Workbox community have come up with that are similar. I think it may be useful to provide easy-to-consume versions of these recipes to reduce onboarding friction to Workbox. I could envision something like the following:

**[Offline Page Only Recipe](https://developers.google.com/web/tools/workbox/guides/advanced-recipes#offline_page_only)**
```js
import { OfflinePage } from 'workbox-recipes';

new OfflinePage('/offline.html');
```

**[Google Fonts](https://developers.google.com/web/tools/workbox/guides/common-recipes#google_fonts) and [Caching Images](https://developers.google.com/web/tools/workbox/guides/common-recipes#caching_images) Recipes**
```js
import { GoogleFontsCache, ImageCache} from 'workbox-recipes';

// Cache the Google Fonts stylesheets with StaleWhileRevalidate and cache the underlying font files for 1 year, max 30 entries
new GoogleFontsCache(30, 60 * 60 * 24 * 365);

// Images would be similar, here max 60 entries for 30 days
new ImageCache(60, 30 * 24 * 60 * 60);
```

Would love some community feedback here, if this is something worth pursuing and if so, what recipes might we want to prioritize.

Snugug:
R: @jeffposnick @philipwalton

Fixes #2622

Adds an experimental `workbox-recipes` module that bundles the configuration for common recipes into config-optional function calls. Example of what default usage would look like:

```js
import {
  pageCache,
  imageCache,
  staticResourceCache,
  googleFontsCache,
  offlineFallback,
} from 'workbox-recipes';
import { precacheAndRoute } from 'workbox-precaching';

// Include offline.html in the manifest
precacheAndRoute(self.__WB_MANIFEST);

// request.mode === 'navigate', network first,   'pages' cachename, 3s timeout, [0, 200] statuses
pageCache();

// request.destination === 'style' || 'script' || 'worker', stale while revalidate first,   'static-resources' cachename, [0, 200] statuses
staticResourceCache();

// request.destination === 'image', cache first,   'images' cachename,, [0, 200] statuses, 30 day max age, 60 max entries
imageCache();

// Same as https://developers.google.com/web/tools/workbox/guides/common-recipes
googleFontsCache();

// Same as https://developers.google.com/web/tools/workbox/guides/advanced-recipes#comprehensive_fallbacks, but just setCacheHandler, and pulling from precache match
offlineFallback();
```


jeffposnick:
The test failure is from https://github.com/GoogleChrome/workbox/blob/d62c18598934e1d8f0e2bd2ff63eb0d7d3e6535b/test/workbox-sw/sw/controllers/test-WorkboxSW.mjs#L187-L198

I am personally okay with treating `workbox-recipes` as something separate from the other modules, and preventing it from being loading from the CDN via `workbox-sw`. (Similarly, I don't think we need to copy it over via `workbox-build`'s `copyLibraries` function.)

You can change that test to the return early when `namespace === 'workbox' || namespace === 'recipes'` and I think things will pass.

PR #2673, opened 2020-11-12
Make use of AssetInfo in webpack build

### LINKED ISSUE #2221: Use webpack v5's immutable asset metadata
https://github.com/webpack/webpack/pull/9687 added new metadata to webpack's asset output, letting plugins determine whether or not an asset URL is immutable (i.e. whether its filename contains a hash).

We could use that as a signal to supplement `dontCacheBustURLsMatching` in our `workbox-webpack-plugin` configuration once it makes it into a webpack v5 release.

### LINKED ISSUE #2672: Set minimized: true on the GenerateSW-created asset
See https://github.com/webpack-contrib/compression-webpack-plugin/issues/218#issuecomment-726196160

jeffposnick:
R: @Snugug @housseindjirdeh @philipwalton

Fixes #2221, fixes #2672

Some small changes to the `workbox-webpack-plugin` to take proper advantage of `webpack`'s `AssetInfo` capabilities.

When we create an asset via `GenerateSW` and we know that it's already been compressed, we now set `minimized: true`.

When we read assets from the build to generate our precache manifest, if any assets have `immutable: true`, that's a clear sign that we don't need to generate hashes for them, so we can return `null` for the `revision` parameter. The JSDoc for `dontCacheBustURLsMatching` has also been updated to reflect this.

### DIFF
--- a/packages/workbox-webpack-plugin/src/generate-sw.js
+++ b/packages/workbox-webpack-plugin/src/generate-sw.js
@@ -75,10 +75,10 @@ class GenerateSW {
    *
    * @param {RegExp} [config.dontCacheBustURLsMatching] Assets that match this will be
    * assumed to be uniquely versioned via their URL, and exempted from the normal
-   * HTTP cache-busting that's done when populating the precache. While not
-   * required, it's recommended that if your existing build process already
-   * inserts a `[hash]` value into each filename, you provide a RegExp that will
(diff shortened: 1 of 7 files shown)


Snugug (approved):
This LGTM from the comments from other webpack threads, but could probably use a LGTM from someone who knows webpack 6 better than I do.

PR #2675, opened 2020-11-13
Update the stage at which the webpack plugin runs

### LINKED ISSUE #2670: InjectManifest does not include index.html generated by html-webpack-plugin in 6.0.0-alpha.3
**Library Affected**:
workbox-webpack-plugin `6.0.0-alpha.3`

**Browser & Platform**:
Node JS

When upgrading to `@next`, I noticed that the `index.html` generated by html-webpack-plugin is no longer included in the manifest entries list. This issue disappeared when I rolled back to 5.1.4 :) 

I made a minimal repo demonstrating the problem here: https://github.com/dermoumi/workbox-inject-manifest

Hopefully this can help, keep up the good work guys!

jeffposnick:
R: @Snugug @philipwalton

Fixes #2670

This is a tactical fix to get compatibility back with `html-webpack-plugin` in `webpack` v5 until https://github.com/webpack/webpack/issues/11822 is resolved.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4168,7 +4168,6 @@
       "version": "7.2.4",
       "resolved": "https://registry.npmjs.org/@types/eslint/-/eslint-7.2.4.tgz",
       "integrity": "sha512-YCY4kzHMsHoyKspQH+nwSe+70Kep7Vjt2X+dZe5Vs2vkRudqtoFoUIv1RlJmZB8Hbp7McneupoZij4PadxsK5Q==",
-      "dev": true,
       "requires": {
         "@types/estree": "*",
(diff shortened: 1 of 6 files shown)


Snugug (approved):
Again, don't know much about Webpack/plugin internals, but from reading the related issue this LGTM from what I can tell.

Only question I have, and is likely because I don't know the test suite well, is if there needs to be a test for this written for `generate-sw` in addition to `inject-manifest`, but I won't block on an answer.

Snugug:
@jeffposnick told me that the test is just being moved to the correct `describe` block and everything else is already covered by tests in `html-webpack-plugin`, so 👍 from me

PR #2690, opened 2020-12-03
Adds gulp build to top-level version script

### LINKED ISSUE #2610: Regression in lerna versioning in the new gulp publish
Hey @philipwalton—hopefully you can remember how you fixed this in the v5 branch, but unfortunately the v6 releases with the new `gulp publish` pipeline aren't getting the `lerna` version bump prior to the `npm` publish and `git` tag.

Do you happen to recall? Or if you could event just point me to the PR I could try to use that as the basis for re-adding it to the `gulp publish` pipeline.

jeffposnick:
R: @Snugug @philipwalton

Fixes #2610

I *think* this should work, and will result in `gulp build` only being run once (as opposed to being run once per package, like our old release system did, which took forever).

This top-level `version` script is [automatically run](https://github.com/lerna/lerna/tree/main/commands/version#lifecycle-scripts) by `lerna` once, prior to commiting to `git`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "node": ">=10.0.0"
   },
   "scripts": {
-    "version": "git add -A packages",
+    "version": "gulp build && git add -A packages",
     "gulp": "gulp"
   },
   "nyc": {


package.json
```diff
@@ -4,7 +4,7 @@
     "node": ">=10.0.0"
   },
   "scripts": {
-    "version": "git add -A packages",
+    "version": "gulp build && git add -A packages",
```
Snugug, line 7:
```suggestion
    "preversion": "gulp build",
    "version": "git add -A packages",
```

By putting it into a `pre` [prefixed script](https://docs.npmjs.com/cli/v6/using-npm/scripts#pre--post-scripts), we don't need to rely on the Unix specific `&&` functionality but we get the net same result. IMO it's also more semantic, as it's a prerequisite.

Snugug (changes requested):
A small change, but something that I think makes Scripts easier to digest and makes this less Unix dependent.

Snugug:
(sorry, that was suppose to be a nit not a blocking request, if you'd prefer not to do it that's fine, just LMK and I'll approve)

package.json
jeffposnick: As per https://github.com/lerna/lerna/tree/main/commands/version#lifecycle-scripts;

```
// preversion:  Run BEFORE bumping the package version.
// version:     Run AFTER bumping the package version, but BEFORE commit.
// postversion: Run AFTER bumping the package version, and AFTER commit.
```

We need the `gulp build` to happen after the package versions have been bumped by `lerna`, but before the commit, so I think we need to chain it as a two-part `version` script.

jeffposnick: (Also, I don't use Windows regularly, but https://github.com/npm/npm/issues/4040#issuecomment-209034612 makes it sound like `&&` works on Windows?)

Snugug (approved):

package.json
Snugug: sgtm. Can leave as-is.

PR #2693, opened 2020-12-04
Fix skipWaiting not working

### LINKED ISSUE #2692: Service worker stuck in busy state when SSE request hits the cache
**Library Affected**:
workbox-routing

**Browser & Platform**:
all browsers

Bug report:

Consider the following service worker

```js
import { clientsClaim, skipWaiting } from 'workbox-core';
import { registerRoute } from **'workbox-routing';
import { StaleWhileRevalidate } from 'workbox-strategies';

skipWaiting();
clientsClaim();

registerRoute(
  ({ request, url }) => request.url.origin === self.location.origin && /(?<!service-worker)\.js$/.test(url.pathname),
  new StaleWhileRevalidate({
    cacheName: 'app',
  }),
);
```

When a new service worker exists and installation of the newer worker begins, it would be expected that the use of `skipWaiting` would activate the new service worker.

This isn't the case however because the _registerRoute_ function instantiates a fetch handler that never calls [waitUntil](https://developer.mozilla.org/en-US/docs/Web/API/ExtendableEvent/waitUntil).

The result of this is that the updated service worker will never activate (stuck in `installed` state) until the user's browser session is closed.

andyrichardson:
R: @jeffposnick @philipwalton

Fixes #2692 

Fixes issue where `skipWaiting` doesn't work as expected when using `registerRoute`.

### DIFF
--- a/packages/workbox-routing/src/Router.ts
+++ b/packages/workbox-routing/src/Router.ts
@@ -75,6 +75,7 @@ class Router {
       const {request} = event;
       const responsePromise = this.handleRequest({request, event});
       if (responsePromise) {
+        event.waitUntil(responsePromise);
         event.respondWith(responsePromise);
       }
     }) as EventListener);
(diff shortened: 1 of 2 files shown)


jeffposnick:
(As per https://github.com/GoogleChrome/workbox/issues/2692, I don't believe this PR is needed.)

PR #2699, opened 2020-12-11
Add setCatchHandler option for routes

### LINKED ISSUE #2318: New Route.setCatchHandler() method
**Library Affected**:
*workbox-routing*

I was thinking that we should add in a `Route`-level `setCatchHandler()` method, which would behave like the existing `Router.setCatchHandler()`, but instead of being a global to the `Router` instance, it would apply only when there's an exception throw by the `HandlerCallback` associated with the specific `Route`.

Something like:

```
const navigationRoute = new NavigationRoute(new NetworkOnly());

// Everything that extends Route would support setCatchHandler():
navigationRoute.setCatchHandler((routeHandlerCallbackParams) => {
  return caches.match('/offline.html');
  // Or do something using routeHandlerCallbackParams to create a Response
});

registerRoute(navigationRoute);
```

This would be post-v5, as it should be backwards-compatible.

Snugug:
Fixes #2318

Added `setCatchHandler` to `Route` base class. Will try the route's catch handler first, if it exists, and falls back to default one if it doesn't, or if the route's catch handler has an error.

### DIFF
--- a/packages/workbox-routing/src/Router.ts
+++ b/packages/workbox-routing/src/Router.ts
@@ -282,7 +282,7 @@ class Router {
           return this._catchHandler.handle({url, request, event});
         }
 
-        return Promise.reject(err);
+        throw err;
       });
     }
 
--- a/packages/workbox-routing/src/Router.ts
+++ b/packages/workbox-routing/src/Router.ts
@@ -244,17 +244,40 @@ class Router {
       responsePromise = Promise.reject(err);
     }
 
-    if (responsePromise instanceof Promise && this._catchHandler) {
-      responsePromise = responsePromise.catch((err) => {
+    // Get route's catch handler, if it exists
+    const catchHandler = route && route.catchHandler;
+
+    if (responsePromise instanceof Promise && (this._catchHandler || catchHandler)) {
+      responsePromise = responsePromise.catch(async (err) => {
+        // If there's a route catch handler, process that first
+        if (catchHandler) {
+          if (process.env.NODE_ENV !== 'production') {
+            // Still include URL here as it will be async from the console group
+            // and may not make sense without the URL
+            logger.groupCollapsed(`Error thrown when responding to: ` +
+              ` ${getFriendlyURL(url)}. Falling back to route's Catch Handler.`);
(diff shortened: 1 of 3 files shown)


Snugug:
Looks like the Mac suite failed post-run in the cache phase? Re-running to hopefully get that to not fail.

packages/workbox-routing/src/Router.ts
```diff
@@ -244,17 +244,40 @@ class Router {
...
+            logger.groupEnd();
+          }
+
+          try {
+            return await catchHandler.handle({url, request, event, params});
+          } catch (catchErr) {
+            err = catchErr;
+          }
+        }
+
```
jeffposnick, line 270: I don't think you can count on `this._catchHandler` being truthy anymore, since the `if()` statement that guards this is now uses `this._catchHandler || catchHandler` as the criteria.

So you'd need to do an explicit check for `this._catchHandler` before executing these statements .

(You should also be able to remove the `!` from `this._catchHandler!` on line 281 once you do that.)

jeffposnick (changes requested):

packages/workbox-routing/src/Router.ts
Snugug: Updated

packages/workbox-routing/src/Router.ts
```diff
@@ -244,18 +244,45 @@ class Router {
...
+              ` ${getFriendlyURL(url)}. Falling back to global Catch Handler.`);
+            logger.error(`Error thrown by:`, route);
+            logger.error(err);
+            logger.groupEnd();
+          }
+          
+          return this._catchHandler.handle({url, request, event});
+        }
+
+        return Promise.reject(err);
```
jeffposnick:
```suggestion
        throw err;
```

We're inside an `async` function, so I think it's clearer to explicitly `throw` instead of creating a rejected `Promise`.

(Unless there's a reason why this has to `Promise.reject()`, in which case, can you explain that in a comment?)

jeffposnick (changes requested):
Just the small suggestion.

PR #2718, opened 2021-01-08
Feature/recipes precache and normalize

Snugug:
R: @jeffposnick @philipwalton

* Adds a new Workbox Recipe: `warmStrategyCache`, which takes an array of paths and a Workbox strategy and warms that strategy's cache with those paths at service worker install time.
* Adds `warmCache` option to page, image, and static resource recipes to allow users to warm those caches
* Adds `plugins` option to page, image, and static resource recipes to allow users to pass additional plugins to those recipes, allowing a user, for instance, to add a URL normalization plugin to the page recipe
* Removes dependency on precaching from offlineFallback recipe. If precaching is used, it will still attempt to pull from there, but if not, it will pull from its own cache.


Snugug:
I'm not sure what the error in the Mac test suite is, and why the Windows one didn't fail. If someone can point me to what I need to fix I'd be more than happy to fix it.

packages/workbox-recipes/src/warmStrategyCache.ts
```diff
@@ -0,0 +1,30 @@
...
+export interface WarmStrategyCacheOptions {
+  paths: Array<string>;
+  strategy: Strategy;
+}
+
+// Give TypeScript the correct global.
+declare let self: ServiceWorkerGlobalScope;
+
+/**
+ *
```
jeffposnick:
```suggestion
 * @memberof module:workbox-recipes
 
```

This should fix the documentation-related error. (Without it, JSDocs thinks that the function is in the global namespace.)

packages/workbox-recipes/src/warmStrategyCache.ts
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * 
+ * @param {Object} options 
+ * @param {string[]} options.paths Paths to warm the strategy's cache with
+ * @param {Strategy} options.strategy Strategy to use
+ */
+function warmStrategyCache(options: WarmStrategyCacheOptions): void {
+  self.addEventListener('install', event => {
+    const warm = options.paths.map(path => options.strategy.handleAll({
```
jeffposnick: I think `done` (or `donePromises`) would be a more descriptive variable name than `warm`, especially since it's not entirely obvious from reading the code that the `[1]` index represents a `done` promise.

packages/workbox-recipes/src/warmStrategyCache.ts
```diff
@@ -0,0 +1,30 @@
+import { Strategy } from 'workbox-strategies/src/Strategy';
+
+import './_version.js';
+
+export interface WarmStrategyCacheOptions {
+  paths: Array<string>;
```
jeffposnick: Can you switch from calling this `paths` to `urls`?

(As far as I can tell, you should be able to pass in any absolute or relative URL inside this array, right?)

jeffposnick (changes requested):
Thanks!

I still need to give some of the modified files a more complete review, but here's some feedback on the new file.

Snugug:
@jeffposnick Looks like test are passing now. Here's the Workbox Recipes docs update to go along w/this
https://github.com/google/WebFundamentals/pull/9210

jeffposnick (approved):

PR #2735, opened 2021-01-26
chunks/excludeChunks can match compilation.chunkNames

### LINKED ISSUE #2717: Workbox throws warning but still caches it (was provided in your Workbox chunks config, but was not found in the compilation.)
**Library Affected**:
workbox-webpack-plugin: ^6.0.2

Currently, I'm upgrading from workbox-webpack-plugin: ^4.3.1 to ^6.0.2, but I noticed a warning that I'm not sure how I should solve it.

It says: `WARNING in The chunk '/js/manifest' was provided in your Workbox chunks config, but was not found in the compilation``

The weird part is that the `manifest.js` file is still cached by the SW (this same warning happens for `vendor.js` as well and is cached too)

image

Why does it still cache it when it says it couldn't find it?

Besides that, I'm struggling to find out what exactly my chunk name for manifest is. I know this is an issue from another plugin, however since using Laravel Mix v6.*, I cannot see the chunk names. It only outputs this in the terminal:

```
Laravel Mix v6.0.8   
                        

✔ Compiled Successfully in 38735ms
┌───────────────────────────────────┬────────────┐
│ File                              │ Size       │
├───────────────────────────────────┼────────────┤
│ /js/vendor.js                     │ 4.05 MiB   │
├───────────────────────────────────┼────────────┤
│ css/app.css                       │ 246 KiB    │
├───────────────────────────────────┼────────────┤
│ /js/app.js                        │ 104 KiB    │
├───────────────────────────────────┼────────────┤
│ images/vendor/intl-tel-input/bui… │ 69.2 KiB   │
├───────────────────────────────────┼────────────┤
│ service-worker.js                 │ 22.8 KiB   │
├───────────────────────────────────┼────────────┤
│ js/6231.9b6787592e9af06465aa.js   │ 6.46 KiB   │
├───────────────────────────────────┼────────────┤
│ js/1901.f3d670a3cfb776389054.js   │ 6.17 KiB   │
├───────────────────────────────────┼────────────┤
│ /js/manifest.js                   │ 6.15 KiB   │
```

But this clearly shows how `/js/manifest.js` is written.

In the `webpack.mix.js` file I wrote this:

 ```
       new InjectManifest({
            swSrc: './resources/js/sw.js',
            swDest: 'service-worker.js',
            // precaching only these.
            chunks: [
                'routes-settings', 'routes-sell-food', 'routes-checkout',
                '/js/app', '/js/moment', '/js/canvas', '/js/manifest',
                '/js/vendor',
            ],
            maximumFileSizeToCacheInBytes: 5000000,
        }),
```

Everything in the chunks array worked well besides the `/js/manifest` and `/js/vendor`. 

p.S. I hope there are no misunderstandings, I want to cache those two files, but I'm just curious why the warnings are still there

### PR DESCRIPTION
**Pull request by jeffposnick**
R: @developit @philipwalton @housseindjirdeh 

Fixes #2717

`workbox-webpack-plugin` provides a way of allow/denylisting the assets in its precache manifest based on chunk names.

When making changes to the webpack plugin for the last major release, we started looking at the `compilation.namedChunkGroups` property to check for matches in `chunks`/`excludeChunks`. This neglects the fact that developers might also expect the names they provide to match against `compilation.namedChunks` (i.e. the actual chunk names, not the group names).

The code in this PR will still give preference to a match in `compilation.namedChunkGroups` (so it should be backwards-compatible), but for situations in which there's no match, it will fall back to checking `compilation.namedChunks`.

### DIFF
--- a/packages/workbox-webpack-plugin/src/lib/get-manifest-entries-from-compilation.js
+++ b/packages/workbox-webpack-plugin/src/lib/get-manifest-entries-from-compilation.js
@@ -41,21 +41,54 @@ function checkConditions(asset, compilation, conditions = []) {
 }
 
 /**
- * Returns the names of all the assets in a chunk group.
+ * Returns the names of all the assets in all the chunks in a chunk group,
+ * if provided a chunk group name.
+ * Otherwise, if provided a chunk name, return all the assets in that chunk.
+ * Otherwise, if there isn't a chunk group or chunk with that name, return null.
  *
- * @param {ChunkGroup} chunkGroup
+ * @param {Compilation} compilation
+ * @param {string} chunkOrGroup
+ * @return {Array<Asset>|null}
+ * @private
+ */
+function getNamesOfAssetsInChunkOrGroup(compilation, chunkOrGroup) {
+  const chunkGroup = compilation.namedChunkGroups &&
+      compilation.namedChunkGroups.get(chunkOrGroup);
+  if (chunkGroup) {
+    const assetNames = [];
+    for (const chunk of chunkGroup.chunks) {
+      assetNames.push(...getNamesOfAssetsInChunk(chunk));
+    }
+    return assetNames;
+  } else {
+    const chunk = compilation.namedChunks &&
+        compilation.namedChunks.get(chunkOrGroup);
+    if (chunk) {
+      return getNamesOfAssetsInChunk(chunk);
+    }
+  }
+
+  // If we get here, there's no chunkGroup or chunk with that name.
+  return null;
+}
+
+/**
+ * Returns the names of all the assets in a chunk.
+ *
+ * @param {Chunk} chunk
  * @return {Array<Asset>}
  * @private
  */
-function getNamesOfAssetsInChunkGroup(chunkGroup) {
+function getNamesOfAssetsInChunk(chunk) {
   const assetNames = [];
-  for (const chunk of chunkGroup.chunks) {
-    assetNames.splice(0, 0, ...chunk.files);
-    // This only appears to be set in webpack v5.
-    if (chunk.auxiliaryFiles) {
-      assetNames.splice(0, 0, ...chunk.auxiliaryFiles);
-    }
+
+  assetNames.push(...chunk.files);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/workbox-webpack-plugin/src/lib/get-manifest-entries-from-compilation.js [hunk 1]
```diff
@@ -41,21 +41,54 @@ function checkConditions(asset, compilation, conditions = []) {
...
+function getNamesOfAssetsInChunk(chunk) {
   const assetNames = [];
-  for (const chunk of chunkGroup.chunks) {
-    assetNames.splice(0, 0, ...chunk.files);
-    // This only appears to be set in webpack v5.
-    if (chunk.auxiliaryFiles) {
-      assetNames.splice(0, 0, ...chunk.auxiliaryFiles);
-    }
+
+  assetNames.push(...chunk.files);
```
jeffposnick, line 85: I... honestly can't remember why this was written as a `splice(0, 0, ...)` previously. I don't believe that it matters whether the items are prepended or appended.

developit: If the ordering of these lists ends up determining the order of the URLs in the cache manifest, maybe this was done to prioritize resources? Prepending seems arbitrary though, I can't think of any reason to prioritize assets from chunk groups over assets from un-grouped chunks.

developit (approved):
> Full disclosure: I have barely used Webpack 5 yet.

This approach seems reasonable to me. It's easier to understand than having exclude/include match both chunks and groups even when there are collisions.

packages/workbox-webpack-plugin/src/lib/get-manifest-entries-from-compilation.js
jeffposnick: Nah, we do an explicit sort on the `url` property of each entry later on. I think it was just a mistaken optimization on my part.

jeffposnick:
Thanks for taking the time to look at this, @developit!

PR #2738, opened 2021-01-28
Ensure bad responses aren't cached

### LINKED ISSUE #2737: workbox-precache caches error responses
Hi

**Library Affected**:
workbox-precache

**Browser & Platform**:
Google Chrome latest

Workbox precache module caches 404 error response, but it shouldn't, See screenshot:

image

Using workbox 6.0.2

My sw.js

```
const precacheUrls = self.__WB_MANIFEST;
if (precacheUrls) {
  precacheAndRoute(precacheUrls);
}
```

jeffposnick:
R: @philipwalton

Fixes #2737

This was a regression introduced in Workbox v6's modified precaching logic. This PR ensures that a bad response isn't saved to the cache before installation is halted.

### DIFF
--- a/packages/workbox-precaching/src/PrecacheStrategy.ts
+++ b/packages/workbox-precaching/src/PrecacheStrategy.ts
@@ -176,7 +176,7 @@ class PrecacheStrategy extends Strategy {
    * 
    * At the time the strategy is used (i.e. during an `install` event), there
    * needs to be at least one plugin that implements `cacheWillUpdate` in the
-   * array, other the `copyRedirectedCacheableResponsesPlugin`.
+   * array, other than `copyRedirectedCacheableResponsesPlugin`.
    * 
    * - If this method is called and there are no suitable `cacheWillUpdate`
    * plugins, we need to add `defaultPrecacheCacheabilityPlugin`.
@@ -205,7 +205,7 @@ class PrecacheStrategy extends Strategy {
         continue;
       }
 
-      // Save the default plugin's index, in case it needs to be remove.
+      // Save the default plugin's index, in case it needs to be removed.
       if (plugin === PrecacheStrategy.defaultPrecacheCacheabilityPlugin) {
         defaultPluginIndex = index;
       }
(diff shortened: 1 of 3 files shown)


philipwalton (approved):

packages/workbox-precaching/src/PrecacheStrategy.ts
```diff
@@ -162,27 +165,63 @@ class PrecacheStrategy extends Strategy {
...
-   * with a `cacheWillUpdate` callback is passed, then the strategy should
-   * defer to that plugin's logic.
+   * This method is complex, as there a number of things to account for:
+   * 
+   * The `plugins` array can be set at construction, and/or it might be added to
+   * to at any time before the strategy is used.
+   * 
+   * At the time the strategy is used (i.e. during an `install` event), there
+   * needs to be at least one plugin that implements `cacheWillUpdate` in the
+   * array, other the `copyRedirectedCacheableResponsesPlugin`.
```
philipwalton: Other than?

packages/workbox-precaching/src/PrecacheStrategy.ts
```diff
@@ -162,27 +165,63 @@ class PrecacheStrategy extends Strategy {
...
+    let defaultPluginIndex: number | null = null;
+    let cacheWillUpdatePluginCount = 0;
+
+    for (const [index, plugin] of this.plugins.entries()) {
+      // Ignore the copy redirected plugin when determining what to do.
+      if (plugin === PrecacheStrategy.copyRedirectedCacheableResponsesPlugin) {
+        continue;
+      }
+
+      // Save the default plugin's index, in case it needs to be remove.
```
philipwalton: s/remove./removed./

philipwalton (approved):
This approach LGTM (just noticed a few spelling/grammar issues in the comments).

One thought though: instead of adding conditionally adding and removing `defaultPrecacheCacheabilityPlugin`, did you consider just always adding and removing it? E.g. before the install event happens, always add it if needed and then always remove it afterward.

Since it only applies to the install event, and since you have to make the check anyway, perhaps keeping it from running during fetch events is good in general.

jeffposnick:
After discussing offline, it's not possible for the `PrecacheStrategy` to run code once, at a point in time right before the `install` event happens. That's something that only the `PrecacheController` could do, and I want to ensure that `PrecacheStrategy` behaves properly even if it's not called by `PrecacheController`.

PR #2744, opened 2021-02-04
Fix NetworkFirst waitUntil

### LINKED ISSUE #2721: Service worker stuck in busy state with NetworkFirst requests
**Library Affected**:

workbox-strategies 6.0.2

**Browser & Platform**:

Tested on Google Chrome 87.0.4280.141 for Windows

My application is set up with the ["offer a page reload"](https://developers.google.com/web/tools/workbox/guides/advanced-recipes#offer_a_page_reload_for_users) recipe and has a route registered with NetworkFirst. However, I noticed that the application wasn't responding when I clicked the upgrade prompt. Going under DevTools -> Application -> Service Workers and clicking "skipWaiting" next to the new service worker instance also seemed to have no effect. If I clicked "stop" next to the old service worker instance to forcibly terminate it, the application reloaded as expected. If I waited 5 minutes, the application reloaded as expected.

As best as I can tell, the symptoms are identical to #2692, except that I'm not using server-side events.

If I replace my NetworkFirst strategy with StaleWhileRevalidate, everything works.

Similar to #2692, calling waitUntil seems to fix this. Specifically, I found NetworkFirst's Promise.race call at https://github.com/GoogleChrome/workbox/blob/025657cde5114bb301407d8e6bc546a21302c244/packages/workbox-strategies/src/NetworkFirst.ts#L120

and inserted a waitUntil call there:

```ts
        let responsePromise = Promise.race(promises);
        handler.waitUntil(responsePromise);
        let response = await responsePromise;
```

I'm happy to open a PR if that would help, although, after reading the discussion at #2692, I wasn't sure if that's the correct fix.

joshkel:
NetworkFirst uses two promises (one for the network request, and one for the timeout). If the network request succeeds, then the timeout promise's timer is canceled. However, it attempted to wait until both promises resolved before the handler resolved as done; this meant that, if the network request succeeded, then it never resolved its handlerDone / awaitComplete promise.

*Note*: There are two potential issues with this PR:

* I wasn't sure what to do about the "If Promise.race() resolved with null" case, when the NetworkFirst strategy falls back to waiting on the network request. Should the code ensure that `handler.waitUntil(networkRequest)` is called in the case of a network timeout + cache miss?
* I updated a test to try and cover this case, but I couldn't figure out how to run the test suite to confirm my changes. (I found the `test_server` gulp task and navigated to `http://localhost:3004/test/workbox-strategies/sw`, but I get console errors, so I must be doing something wrong.)

R: @jeffposnick @philipwalton

Fixes #2721

### DIFF
--- a/packages/workbox-strategies/src/NetworkFirst.ts
+++ b/packages/workbox-strategies/src/NetworkFirst.ts
@@ -99,7 +99,7 @@ class NetworkFirst extends Strategy {
       });
     }
 
-    const promises = [];
+    const promises: Promise<Response | undefined>[] = [];
     let timeoutId: number | undefined;
 
     if (this._networkTimeoutSeconds) {
@@ -113,18 +113,16 @@ class NetworkFirst extends Strategy {
 
     promises.push(networkPromise);
 
-    // Promise.race() will resolve as soon as the first promise resolves.
-    const combinedPromise = Promise.race(promises);
-    handler.waitUntil(combinedPromise);
-    let response = await combinedPromise;
-    // If Promise.race() resolved with null, it might be due to a network
-    // timeout + a cache miss. If that were to happen, we'd rather wait until
-    // the networkPromise resolves instead of returning null.
-    // Note that it's fine to await an already-resolved promise, so we don't
-    // have to check to see if it's still "in flight".
-    if (!response) {
-      response = await networkPromise;
-    }
+    const response = await handler.waitUntil((async () =>
+      // Promise.race() will resolve as soon as the first promise resolves.
+      await handler.waitUntil(Promise.race(promises))
+      // If Promise.race() resolved with null, it might be due to a network
+      // timeout + a cache miss. If that were to happen, we'd rather wait until
+      // the networkPromise resolves instead of returning null.
+      // Note that it's fine to await an already-resolved promise, so we don't
+      // have to check to see if it's still "in flight".
+      || await networkPromise
+    )());
 
     if (process.env.NODE_ENV !== 'production') {
       logger.groupCollapsed(
--- a/packages/workbox-strategies/src/NetworkFirst.ts
+++ b/packages/workbox-strategies/src/NetworkFirst.ts
@@ -112,12 +112,11 @@ class NetworkFirst extends Strategy {
         this._getNetworkPromise({timeoutId, request, logs, handler});
 
     promises.push(networkPromise);
-    for (const promise of promises) {
-      handler.waitUntil(promise);
-    }
 
     // Promise.race() will resolve as soon as the first promise resolves.
-    let response = await Promise.race(promises);
+    const combinedPromise = Promise.race(promises);
+    handler.waitUntil(combinedPromise);
+    let response = await combinedPromise;
     // If Promise.race() resolved with null, it might be due to a network
     // timeout + a cache miss. If that were to happen, we'd rather wait until
     // the networkPromise resolves instead of returning null.
--- a/packages/workbox-strategies/src/NetworkFirst.ts
+++ b/packages/workbox-strategies/src/NetworkFirst.ts
@@ -113,16 +113,16 @@ class NetworkFirst extends Strategy {
 
     promises.push(networkPromise);
 
-    const response = await handler.waitUntil((async () =>
+    const response = await handler.waitUntil((async () => {
       // Promise.race() will resolve as soon as the first promise resolves.
-      await handler.waitUntil(Promise.race(promises))
-      // If Promise.race() resolved with null, it might be due to a network
-      // timeout + a cache miss. If that were to happen, we'd rather wait until
-      // the networkPromise resolves instead of returning null.
-      // Note that it's fine to await an already-resolved promise, so we don't
-      // have to check to see if it's still "in flight".
-      || await networkPromise
-    )());
+      return await handler.waitUntil(Promise.race(promises)) ||
+          // If Promise.race() resolved with null, it might be due to a network
+          // timeout + a cache miss. If that were to happen, we'd rather wait until
+          // the networkPromise resolves instead of returning null.
+          // Note that it's fine to await an already-resolved promise, so we don't
+          // have to check to see if it's still "in flight".
+          await networkPromise;
+    })());
 
     if (process.env.NODE_ENV !== 'production') {
       logger.groupCollapsed(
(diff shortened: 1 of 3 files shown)


jeffposnick:
Thanks a ton for investigating this and proposing a fix, @joshkel!

I'll take a look, but @philipwalton touched this code more recently, so it would be great if he could give a review as well.

(Regarding manual test running, instead of using `http://localhost:3004/test/workbox-strategies/sw`, it should be `http://localhost:3004/test/workbox-strategies/sw/`, with a trailing `/`, or else the service worker won't register properly. This always annoys me too, and we should fix the server to automatically redirect.)

test/workbox-strategies/sw/test-NetworkFirst.mjs
```diff
@@ -141,12 +141,37 @@ describe(`NetworkFirst`, function() {
...
       await cache.put(request, injectedResponse.clone());
 
-      const handlePromise = networkFirst.handle({
+      const [handlePromise, donePromise] = networkFirst.handleAll({
+        request,
+        event,
+      });
+
+      await eventDoneWaiting(event);
+      await donePromise;
```
philipwalton, line 149: I don't you think you need both `eventDoneWaiting()` and `donePromise` (I expect them to resolve at the same time. Were you seeing a situations where that was not the case?

test/workbox-strategies/sw/test-NetworkFirst.mjs
```diff
@@ -141,12 +141,37 @@ describe(`NetworkFirst`, function() {
...
+
+      const networkFirst = new NetworkFirst({networkTimeoutSeconds});
+
+      const [handlePromise, donePromise] = networkFirst.handleAll({
         request,
         event,
       });
 
       await eventDoneWaiting(event);
+      await donePromise;
```
philipwalton, line 173: I think you also want to test that the handler completes before 1000ms passes, right? I.e. from my read of your changes this test would still pass with the old logic.

philipwalton (changes requested):
Thanks @joshkel, and nice catch! I left a few comments in the test, but overall the approach LGTM.

test/workbox-strategies/sw/test-NetworkFirst.mjs
joshkel: In the scenario I was seeing, if the network request succeeded, then it canceled the timeout, and so donePromise never resolved. However, the handlePromise still resolved, so the bug (the unresolved promise) went unnoticed until I tried to upgrade the service worker; at that point, the promise kept the old service worker in the "running" state.

For this scenario ("should return the cached response if the network request times out"), both promises resolved even before my changes. But, since the "successful completion before timeout" scenario involved a hard-to-notice bug with only one promise remaining, it seemed worthwhile to add an explicit test for the "timeout" scenario, too.

I'm happy to delete this if you don't think it's worth keeping.

test/workbox-strategies/sw/test-NetworkFirst.mjs
joshkel: Prior to my changes, donePromise never resolved, instead of resolving after 1000ms.

But verifying that it completes quickly is a good addition; thank you. Please see the latest commit.

(I also changed the networkTimeoutSeconds from 1000 to 10; I thought that might avoid potential confusion between 1000 seconds and 1000 ms, and 10 seconds is still plenty to make the test serve its purpose of failing if the promise doesn't resolve.)

jeffposnick (changes requested):
Thanks so much for digging into this, @joshkel!

To the point you made in your PR description: yes, I think that technically you should add to line 127:

```
handler.waitUntil(networkPromise); // add this
response = await networkPromise;
```

in the case where there's a cache miss and you need to fall back to the network.

joshkel:
Thanks, @jeffposnick. I don't think that adding 

```
handler.waitUntil(networkPromise);
```

is the best solution, because that partially undoes the timeout - if the network is slow, the initial response can still come through when the timeout expires, but the handler's `waitUntil` / `doneWaiting` won't resolve until the network request resolves.

I pushed an alternate approach; please let me know what you think. Thanks.

jeffposnick:
Thanks again for your continued work on getting this straight! This latest update is 👍 from me, but I would appreciate a final look from @philipwalton before we merge.

jeffposnick (approved):

packages/workbox-strategies/src/NetworkFirst.ts
```diff
@@ -112,20 +112,17 @@ class NetworkFirst extends Strategy {
...
-    let response = await Promise.race(promises);
-    // If Promise.race() resolved with null, it might be due to a network
-    // timeout + a cache miss. If that were to happen, we'd rather wait until
-    // the networkPromise resolves instead of returning null.
-    // Note that it's fine to await an already-resolved promise, so we don't
-    // have to check to see if it's still "in flight".
-    if (!response) {
-      response = await networkPromise;
-    }
+    const response = await handler.waitUntil((async () =>
```
philipwalton: Nit: can your reformat this to fit with Google JS style (explicit return statement, `||` at the end of the previous line, and 4-space indentation for line continuation)? Here's the correct formatting:

```js
const response = await handler.waitUntil((async () => {
  // Promise.race() will resolve as soon as the first promise resolves.
  return await handler.waitUntil(Promise.race(promises)) ||
      // If Promise.race() resolved with null, it might be due to a network
      // timeout + a cache miss. If that were to happen, we'd rather wait until
      // the networkPromise resolves instead of returning null.
      // Note that it's fine to await an already-resolved promise, so we don't
      // have to check to see if it's still "in flight".
      await networkPromise;
})());
```

test/workbox-strategies/sw/test-NetworkFirst.mjs
philipwalton: The logic for `donePromise` is virtually identical (actually it was copy/pasted from) the logic for `eventDoneWaiting()`, so you shouldn't need both. I'd prefer to just use `donePromise` since it uses library code rather than a test helper.

test/workbox-strategies/sw/test-NetworkFirst.mjs
philipwalton: Same comment as above, just remove the call to `eventDoneWaiting()`.

Let me know if for some reason the tests don't pass with that change (it shouldn't have any effect).

philipwalton (changes requested):

joshkel:
Done. Thanks for the reviews, @jeffposnick and @philipwalton.

philipwalton (approved):
LGTM, thanks again for catching this!

Issue #2150, opened 2019-07-29
Failing workbox-window integration test on Safari


jeffposnick:
After upgrading to `selenium-webdriver` v4 and running the Safari (production) test suite locally, I get consistent failures at

```
  1) [workbox-window] Workbox
       "after each" hook for "reports all events for an external SW registration":
     
  NoSuchWindowError
      at Object.throwDecodedError (node_modules/selenium-webdriver/lib/error.js:550:15)
      at parseHttpResponse (node_modules/selenium-webdriver/lib/http.js:560:13)
      at Executor.execute (node_modules/selenium-webdriver/lib/http.js:486:26)
      at process._tickCallback (internal/process/next_tick.js:68:7)
```

jeffposnick:
Unfortunately, that test is still flakey.

C.f. https://travis-ci.org/GoogleChrome/workbox/builds/574870617 in Safari Stable:

```sh
  1) [workbox-window] Workbox
       register
         reports all events for an external SW registration:
     
  TimeoutError
      at Object.throwDecodedError (node_modules/selenium-webdriver/lib/error.js:550:15)
      at parseHttpResponse (node_modules/selenium-webdriver/lib/http.js:560:13)
      at Executor.execute (node_modules/selenium-webdriver/lib/http.js:486:26)
      at <anonymous>
      at process._tickDomainCallback (internal/process/next_tick.js:229:7)
```

I'm assuming we see a slightly different error now due to catching the exception, but it's not enough to eliminate the flakiness.

@philipwalton, as the test's author, do you have any idea what might be causing the occasional failures?

philipwalton:
Since creating #2210, this failure has turned from flaky to consistent. (Not sure why, since that PR didn't make any `workbox-window` changes).

After spending a bunch of time trying to narrow down exactly what changed that's making this now happen more frequently, I couldn't identify anything. (It's also really hard because it's also flakey on master, so it's hard to determine if a particular change is making it _more_ flakey.)

At this point, this issue is blocking us from merging #2210, so unless we can figure out why it's failing, I propose we temporarily skip the failing test in Safari until a solution is found.

@jeffposnick what do you think?

philipwalton:
Not sure if this is related, but [this issue](https://github.com/SeleniumHQ/selenium/issues/3796#issuecomment-459928531) suggests the `NoSuchWindowError` error has been fixed in the latest Safari Tech Preview. However, I tried running the tests in STP and still saw the error, so our issue may be different.

[...]

PR #2751, opened 2021-02-12
Fix a flakey test, and remove retries

jeffposnick:
R: @philipwalton

I got to the bottom of #2321 and #2150, by removing the call to close a tab prior to switching to tabs, and that seems to make Safari happy in my local testing.

I also removed all retries from our tests, as I don't think anything should currently be flakey, and honestly, if it is, I'd rather know and fix it.

### DIFF
--- a/gulp-tasks/test-integration.js
+++ b/gulp-tasks/test-integration.js
@@ -25,7 +25,7 @@ function runFiles(filePaths) {
 
   return new Promise((resolve, reject) => {
     const mocha = new Mocha({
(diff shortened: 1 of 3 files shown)


jeffposnick:
I also discovered that we were... just not running the `workbox-cacheable-response` unit tests at all? And instead running the `workbox-broadcast-update` unit tests twice.

jeffposnick:
(Removing retries has effectively kicked the beehive wrt other flakiness that I'm investigating...)

PR #2752, opened 2021-02-12
Set wasWaitingBeforeRegister, if missing

### LINKED ISSUE #2526: `waiting` event is missing `wasWaitingBeforeRegister` property in Chrome
**Library Affected**:
workbox-window

**Browser & Platform**:
Chrome 83.0.4103.61, Linux

Event in `waiting` event handler (`wb.addEventListener('waiting', (event) => { ...`) in Google Chrome is missing `wasWaitingBeforeRegister` property but has `isUpdate`. In Firefox it has only `wasWaitingBeforeRegister`.

Also in case of Chrome if I reload the page without calling `skipWaiting` in sw then I see two `waiting` events. One has `wasWaitingBeforeRegister` property and *redundent* SW and other has `isUpdate` property and *installed* SW.

I am not sure if #2525 is related as I am encountering both issues even I followed instructions from https://developers.google.com/web/tools/workbox/guides/get-started.

Code:
```javascript
const wb = new Workbox('/sw.js');

wb.addEventListener('waiting', (event) => {
  console.log('waiting', event);

  if (event.isUpdate || event.wasWaitingBeforeRegister) { // I must use both conditions
    showSkipWaitingPrompt();
  }
});

jeffposnick:
R: @philipwalton

Fixes #2526, I think. Your feedback is appreciated, @philipwalton.


philipwalton:
To confirm, AFAICT these changes are just to make the `wasWaitingBeforeRegister` property always present always and always boolean, right?

I'm fine with that change, but I'm not sure if that will really address the linked bug because the logic in that code was just checking for truthiness.

I suspect the issue in that bug may be due to slight differences in SW lifecycle between Chrome and FF that are probably worth digging in to.

jeffposnick:
Okay, let's hold off on this then. I see that these properties are `(boolean or undefined)` in the docs, so I guess that's good enough.

PR #2756, opened 2021-02-15
Add docs on running the test suite

joshkel:
It took me a bit to figure out how to run the test suite, so I thought some docs might help.

Feedback / additions / corrections welcome.

R: @jeffposnick @philipwalton


jeffposnick (approved):
Thanks for these contributions, @joshkel!

PR #2775, opened 2021-03-10
Fix the PrecacheController JSDocs

### LINKED ISSUE #2774: Undocumented v6 breaking change: PrecacheController.install/active
Hi,

Not 100% sure but it seems like in the code and TS typedefs, the PrecacheController install/activate methods now take the event as param, while it wasn't the case before.

Not passing the event lead to console errors after an upgrade:

image

```
Uncaught TypeError: Cannot read property 'waitUntil' of undefined
``` 

FYI the doc seems wrong to me and this is also not mentioned on the v5-v6 breaking changes, and might be useful to add/fix ;)

https://developers.google.com/web/tools/workbox/modules/workbox-precaching

jeffposnick:
R: @philipwalton @tropicadri
CC: @slorber

Fixes #2774

To accompany this change, I'll also be updating the following separately:

- https://github.com/GoogleChrome/workbox/releases/tag/v6.0.0
- https://developers.google.com/web/tools/workbox/guides/migrations/migrate-from-v5

### DIFF
--- a/packages/workbox-precaching/src/PrecacheController.ts
+++ b/packages/workbox-precaching/src/PrecacheController.ts
@@ -177,8 +177,7 @@ class PrecacheController {
    * Note: this method calls `event.waitUntil()` for you, so you do not need
    * to call it yourself in your event handlers.
    *
-   * @param {Object} options
-   * @param {Event} options.event The install event.
+   * @param {ExtendableEvent} event
    * @return {Promise<module:workbox-precaching.InstallResult>}
    */
   install(event: ExtendableEvent) {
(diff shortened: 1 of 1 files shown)


jeffposnick:
Actually, this is [called out as a breaking change](https://github.com/GoogleChrome/workbox/releases/tag/v6.0.0#:~:text=The%20install()%20and%20activate()%20methods) in the release notes already, but not in the v5 to v6 upgrade guide. I'll get that guide updated.

tropicadri (approved):

slorber:
Thanks

don't forget to update this page: https://developers.google.com/web/tools/workbox/modules/workbox-precaching

(assuming it's supposed to show latest/v6 code, which is quite unclear imho)

image

jeffposnick:
https://github.com/google/WebFundamentals/pull/9303/commits/ede49ffe4ac3a8b37196c02e93a8bbb1ad6f03e5 covers those other docs.

Thanks!

PR #2787, opened 2021-03-19
Support for external controlling events

### LINKED ISSUE #2786: workbox-window should fire controlling events with isExternal: true
In Workbox v6, we modified the behavior of most of the previous `workbox-window` events so that instead of having "expected" and "external" flavors (`waiting`, `externalwaiting`, etc.) we would always fire events with the expected name (`waiting`, etc.) and set the `isExternal` boolean flag accordingly.

This change did not happen for the `controlling` event, however, and the only time `controlling` will be fired is if the service worker that has taken control of the page is the same as the service worker associated with `workbox-window`'s original registration.

We should update that so that the `controlling` event is also fired when it's a different service worker that's taken control, and set `isExternal: true` in that case. (And set `isExternal: false` otherwise, which also isn't being set.)

https://github.com/GoogleChrome/workbox/blob/27080cccfbc360b121083bb47e20429997051a6f/packages/workbox-window/src/Workbox.ts#L503-L517

It's arguable whether this is a bug fix or a breaking change—the documentation implies that the `controlling` event would be fired with `isExternal: true` in this scenario, so I'm erring on the side of bug fix.

CC: @philipwalton in case he has an opinion.

jeffposnick:
R: @philipwalton @tropicadri

Fixes #2786

The intention in v6 was to fire all synthetic events on all `Workbox` instances, regardless of whether they originated from the "current" or "external" service worker. Developers who need to distinguish between the two could check the `isExternal` boolean flag on the event.

This logic wasn't implemented for the `controlling` synthetic event, causing developer confusion. This PR adds that functionality.

As mentioned in the original issue, I consider this a bug fix as opposed to a breaking change, as we state in the [updated documentation](https://developers.google.com/web/tools/workbox/modules/workbox-window#when_an_unexpected_version_of_the_service_worker_is_found):

> As of Workbox v6 and later, these events are equivalent to the events documented above, with the addition of an isExternal: true property set on each event object. If your web application needs to implement specific logic to handle an "external" service worker, you can check for that property in your event handlers.

### DIFF
--- a/packages/workbox-window/src/Workbox.ts
+++ b/packages/workbox-window/src/Workbox.ts
@@ -502,13 +502,20 @@ class Workbox extends WorkboxEventTarget {
    */
   private readonly _onControllerChange = (originalEvent: Event) => {
     const sw = this._sw;
-    if (sw === navigator.serviceWorker.controller) {
-      this.dispatchEvent(new WorkboxEvent('controlling', {
-        sw,
-        originalEvent,
-        isUpdate: this._isUpdate,
-      }));
+    const isExternal = sw !== navigator.serviceWorker.controller;
 
+    // Unconditionally dispatch the controlling event, with isExternal set
+    // to distinguish between controller changes due to the initial registration
+    // vs. an update-check or other tab's registration.
+    // See https://github.com/GoogleChrome/workbox/issues/2786
+    this.dispatchEvent(new WorkboxEvent('controlling', {
+      isExternal,
+      originalEvent,
+      sw,
+      isUpdate: this._isUpdate,
+    }));
+
+    if (!isExternal) {
       if (process.env.NODE_ENV !== 'production') {
         logger.log('Registered service worker now controlling this page.');
       }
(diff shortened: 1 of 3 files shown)


packages/workbox-window/src/Workbox.ts
```diff
@@ -502,13 +502,20 @@ class Workbox extends WorkboxEventTarget {
...
+    // vs. an update-check or other tab's registration.
+    // See https://github.com/GoogleChrome/workbox/issues/2786
+    this.dispatchEvent(new WorkboxEvent('controlling', {
+      isExternal,
+      originalEvent,
+      sw,
+      isUpdate: this._isUpdate,
+    }));
+
+    if (!isExternal) {
```
jeffposnick, line 518: I decided that the `controlling` promise would _not_ resolve (if it hasn't already been resolved) if the new controller is an "external" service worker, as there is no means to distinguish between expected/external in that case.

tropicadri (approved):

PR #2799, opened 2021-04-01
Rollup update

tropicadri:
Updating rollup version and rollup-plugin-node-resolve. Verified build output is the same with new version.

Options modified to be compatible with the new version:

- https://rollupjs.org/guide/en/#outputesmodule
- https://rollupjs.org/guide/en/#treeshake moduleSideEffects 

Rollup part of #2479


jeffposnick (changes requested):
One small addition: can you remove this line?

https://github.com/GoogleChrome/workbox/blob/2426f72aef51ec78c0a750c79d6615ee0aee3fe3/.ncurc.js#L10

That way, `npx npm-check-updates` run from the top-level directory will be able to update `rollup` in the future.

Otherwise, the builds look identical. Thanks!

jeffposnick (approved):

Issue #2801, opened 2021-04-02
Deprecate DBWrapper


tropicadri:
**Library Affected**:
*workbox-background-sync, workbox-expiration, workbox-core*

**Browser & Platform**:
 "all browsers".*

**Issue or Feature Request Description**:
Deprecate the [DBWrapper](https://github.com/GoogleChrome/workbox/blob/v6/packages/workbox-core/src/_private/DBWrapper.ts) in favor of [idb](https://www.npmjs.com/package/idb), idb provides the functionality the wrapper gives and will help making the code more readable, and is a tested and widely used solution.

Issue #2602, opened 2020-08-14
Many TypeScript linter issues with latest @typescript-eslint


jeffposnick:
There are 225 TypeScript linter errors and 119 warnings in our codebase that I encountered after bumping the `@typescript-eslint` dependency versions locally. I'm going to roll that back for now, but we need to address these at some point.

For the record, here's the full list:

```
[10:09:43] Error: Command failed with exit code 1: eslint **/*.ts --config typescript.eslintrc.js --ignore-path .gitignore

/Users/jeff/git/workbox/packages/workbox-background-sync/src/Queue.ts
   37:14  error    Don't use `object` as a type. The `object` type is currently hard to use ([see this issue](https://github.com/microsoft/TypeScript/issues/21732)).
Consider using `Record<string, unknown>` instead, as it allows you to more easily inspect and use the keys  @typescript-eslint/ban-types
  121:3   warning  Missing return type on function                                                                                                                                                                                                                                @typescript-eslint/explicit-module-boundary-types
  141:3   warning  Missing return type on function                                                                                                                                                                                                                                @typescript-eslint/explicit-module-boundary-types
  176:3   warning  Missing return type on function                                                                                                                                                                                                                                @typescript-eslint/explicit-module-boundary-types
  202:3   warning  Missing return type on function                                                                                                                                                                                                                                @typescript-eslint/explicit-module-boundary-types
  213:3   warning  Missing return type on function                                                                                                                                                                                                                                @typescript-eslint/explicit-module-boundary-types
  223:3   warning  Missing return type on function                                                                                                                                                                                                                                @typescript-eslint/explicit-module-boundary-types
  253:3   warning  Missing return type on function                                                                                                                                                                                                                                @typescript-eslint/explicit-module-boundary-types
  319:3   warning  Missing return type on function                                                                                                                                                                                                                                @typescript-eslint/explicit-module-boundary-types
  348:3   warning  Missing return type on function                                                                                                                                                                                                                                @typescript-eslint/explicit-module-boundary-types
  386:15  error    Unsafe assignment of an any value                                                                                                                                                                                                                              @typescript-eslint/no-unsafe-assignment
  415:7   error    Promises must be handled appropriately or explicitly marked as ignored with the `void` operator                                                                                                                                                                @typescript-eslint/no-floating-promises
  427:3   warning  Missing return type on function                                                                                                                                                                                                                                @typescript-eslint/explicit-module-boundary-types

[...]

PR #2805, opened 2021-04-06
Adding the QueueDb and using it in background sync instead of DBWrapper

tropicadri:
This is the first part of #2801 
I created a new class to interact with IDB, to manage the queues in workbox-background-sync. This new class encapsulates details like the object store name and index, as well as the types the DB should manage.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "html-webpack-plugin-v4": "npm:html-webpack-plugin@^4.5.0",
     "html-webpack-plugin-v5": "npm:html-webpack-plugin@^5.0.0-alpha.14",
     "husky": "^4.2.5",
+    "idb": "^6.0.0",
     "jsdoc": "^3.6.5",
     "jsdoc-baseline": "^0.1.5",
     "lerna": "^3.22.1",
--- a/packages/workbox-background-sync/src/Queue.ts
+++ b/packages/workbox-background-sync/src/Queue.ts
@@ -10,10 +10,10 @@ import {WorkboxError} from 'workbox-core/_private/WorkboxError.js';
 import {logger} from 'workbox-core/_private/logger.js';
 import {assert} from 'workbox-core/_private/assert.js';
 import {getFriendlyURL} from 'workbox-core/_private/getFriendlyURL.js';
-import {UnidentifiedQueueStoreEntry, QueueStore} from './lib/QueueStore.js';
+import {QueueStore} from './lib/QueueStore.js';
+import {UnidentifiedQueueStoreEntry} from './lib/QueueDb';
 import {StorableRequest} from './lib/StorableRequest.js';
 import './_version.js';
-
 
 // Give TypeScript the correct global.
 declare let self: ServiceWorkerGlobalScope;
@@ -238,7 +238,6 @@ class Queue {
 
     return unexpiredEntries;
   }
-
 
   /**
    * Adds the entry to the QueueStore and registers for a sync event.
--- a/packages/workbox-background-sync/src/Queue.ts
+++ b/packages/workbox-background-sync/src/Queue.ts
@@ -11,7 +11,7 @@ import {logger} from 'workbox-core/_private/logger.js';
 import {assert} from 'workbox-core/_private/assert.js';
 import {getFriendlyURL} from 'workbox-core/_private/getFriendlyURL.js';
 import {QueueStore} from './lib/QueueStore.js';
-import {UnidentifiedQueueStoreEntry} from './lib/QueueDb';
+import {UnidentifiedQueueStoreEntry} from './lib/QueueDb.js';
 import {StorableRequest} from './lib/StorableRequest.js';
 import './_version.js';
 
--- a/packages/workbox-background-sync/src/lib/QueueDb.ts
+++ b/packages/workbox-background-sync/src/lib/QueueDb.ts
@@ -13,7 +13,7 @@ import '../_version.js';
 interface QueueDBSchema extends DBSchema {
   requests: {
     key: number;
-    value: UnidentifiedQueueStoreEntry;
+    value: QueueStoreEntry;
     indexes: {queueName: string};
   };
 }
@@ -53,7 +53,7 @@ export class QueueDb {
    */
   async addEntry(entry: UnidentifiedQueueStoreEntry) {
     const db = await this.getDb();
-    await db.add(REQUEST_OBJECT_STORE_NAME, entry);
+    await db.add(REQUEST_OBJECT_STORE_NAME, entry as QueueStoreEntry);
   }
 
   /**
@@ -77,7 +77,7 @@ export class QueueDb {
     queueName: string,
   ): Promise<QueueStoreEntry[]> {
     const db = await this.getDb();
-    const results = await db.getAllFromIndex(REQUEST_OBJECT_STORE_NAME, QUEUE_NAME_INDEX, IDBKeyRange.only(queueName)) as QueueStoreEntry[];
+    const results = await db.getAllFromIndex(REQUEST_OBJECT_STORE_NAME, QUEUE_NAME_INDEX, IDBKeyRange.only(queueName));
     return results ? results : new Array<QueueStoreEntry>();
   }
 
@@ -127,7 +127,7 @@ export class QueueDb {
     const cursor = await db.transaction(REQUEST_OBJECT_STORE_NAME)
       .store.index(QUEUE_NAME_INDEX)
       .openCursor(query, direction);
-    return cursor?.value as QueueStoreEntry;
+    return cursor?.value;
   }
 
   /**
@@ -141,7 +141,7 @@ export class QueueDb {
         upgrade: this._upgradeDb,
       });
     }
-    return this._db
+    return this._db;
   }
 
   /**
--- a/packages/workbox-background-sync/src/lib/QueueDb.ts
+++ b/packages/workbox-background-sync/src/lib/QueueDb.ts
@@ -1,0 +1,149 @@
+/*
+  Copyright 2018 Google LLC
+
+  Use of this source code is governed by an MIT-style
+  license that can be found in the LICENSE file or at
+  https://opensource.org/licenses/MIT.
+*/
+
+import {openDB, DBSchema, IDBPDatabase} from 'idb';
+import {RequestData} from './StorableRequest.js';
+
+interface QueueDBSchema extends DBSchema {
+  requests: {
+    key: number;
+    value: UnidentifiedQueueStoreEntry;
+    indexes: {queueName: string};
+  };
+}
+
+const DB_VERSION = 3;
+const DB_NAME = 'workbox-background-sync';
+const OBJECT_STORE_NAME = 'requests';
+const INDEXED_PROP = 'queueName';
+
+export interface UnidentifiedQueueStoreEntry {
+  requestData: RequestData;
+  timestamp: number;
+  id?: number;
+  queueName?: string;
+  metadata?: object;
+}
+
+export interface QueueStoreEntry extends UnidentifiedQueueStoreEntry {
+  id: number;
+}
+
+/**
+ * A class to interact directly with IndexedDB to store a retrieve queue requests.
+ *
+ * @private
+ */
+
+export class QueueDb {
+  private _db: IDBPDatabase<QueueDBSchema> | null = null;
+
+  /**
+   * Opens a connection to IDBDatabase.
+   *
+   * @private
+   */
+  private async open() {
+    if (!this._db) {
+      this._db = await openDB(DB_NAME, DB_VERSION, {
+        upgrade: this._upgradeDb,
+      });
+    }
+  }
+
+  /**
+   *
+   * @param {UnidentifiedQueueStoreEntry} entry
+   */
+  async addEntry(entry: UnidentifiedQueueStoreEntry) {
+    await this.open();
+    await this._db!.add(OBJECT_STORE_NAME, entry);
+  }
+
+  /**
+   * Returns the first entry in the ObjectStore
+   *
+   * @return {QueueStoreEntry}
+   */
+  async getFirstEntry(): Promise<QueueStoreEntry | any> {
+    await this.open();
+    const cursor = await this._db
+      ?.transaction(OBJECT_STORE_NAME)
+      .store.openCursor();
+
+    if (cursor) {
+      return cursor.value as QueueStoreEntry;
+    }
+  }
+
+  /**
+   *
+   * @param query
+   * @return {Promise<QueueStoreEntry[]>}
+   */
+  async getAllEntriesFromIndex(
+    query?: IDBKeyRange,
+  ): Promise<QueueStoreEntry[]> {
+    await this.open();
+    const results =
+      await this._db?.getAllFromIndex(OBJECT_STORE_NAME, INDEXED_PROP, query) as QueueStoreEntry[];
+    return results ? results : new Array<QueueStoreEntry>();
+  }
+
+  /**
+   *
+   * @param {number} id the id of the entry to be deleted
+   */
+  async deleteEntry(id: number) {
+    await this.open();
+    await this._db?.delete(OBJECT_STORE_NAME, id);
+  }
+
+  /**
+   * Returns either the first or the last entries, depending on direction. Uses the index.
+   *
+   * @param {IDBCursorDirection} direction
+   * @param {IDBKeyRange} query
+   * @return {Promise<QueueStoreEntry | any>}
+   */
+  async getEndEntryFromIndex(
+    {direction}: {direction?: IDBCursorDirection},
+    query?: IDBKeyRange,
+  ): Promise<QueueStoreEntry | any> {
+    await this.open();
+
+    const cursor = await this._db?.transaction(OBJECT_STORE_NAME)
+      .store.index(INDEXED_PROP)
+      .openCursor(query, direction);
+
+    if (cursor) {
+      return cursor.value as QueueStoreEntry;
+    }
+  }
+
+  /**
+   * Upgrades QueueDB
+   *
+   * @param {IDBPDatabase<QueueDBSchema>} db
+   * @param {number} oldVersion
+   * @private
+   */
+  private _upgradeDb(db: IDBPDatabase<QueueDBSchema>, oldVersion: number) {
+    if (oldVersion > 0 && oldVersion < DB_VERSION) {
+      if (db.objectStoreNames.contains(OBJECT_STORE_NAME)) {
+        db.deleteObjectStore(OBJECT_STORE_NAME);
+      }
+    }
+
+    const objStore = db.createObjectStore(OBJECT_STORE_NAME, {
+      autoIncrement: true,
+      keyPath: 'id',
+    });
+    objStore.createIndex(INDEXED_PROP, INDEXED_PROP, {unique: false});
+  }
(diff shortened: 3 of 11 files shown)


packages/workbox-background-sync/src/Queue.ts
```diff
@@ -10,11 +10,11 @@ import {WorkboxError} from 'workbox-core/_private/WorkboxError.js';
 import {logger} from 'workbox-core/_private/logger.js';
 import {assert} from 'workbox-core/_private/assert.js';
 import {getFriendlyURL} from 'workbox-core/_private/getFriendlyURL.js';
-import {UnidentifiedQueueStoreEntry, QueueStore} from './lib/QueueStore.js';
+import {QueueStore} from './lib/QueueStore.js';
+import {UnidentifiedQueueStoreEntry} from './lib/QueueDb';
```
jeffposnick: This needs to have a `.js` file extension or certain use cases break.

package.json
```diff
@@ -81,6 +81,7 @@
     "html-webpack-plugin-v4": "npm:html-webpack-plugin@^4.5.0",
     "html-webpack-plugin-v5": "npm:html-webpack-plugin@^5.0.0-alpha.14",
     "husky": "^4.2.5",
+    "idb": "^6.0.0",
```
jeffposnick, line 89: Just double-checking—what necessitates adding in the `idb` dependency at the top-level?Is it there for the test server, for creating our `build/` bundles, or some other purpose?

packages/workbox-background-sync/src/lib/QueueDb.ts
```diff
@@ -0,0 +1,149 @@
+/*
+  Copyright 2018 Google LLC
```
jeffposnick: You can switch this to 2021.

packages/workbox-background-sync/src/lib/QueueDb.ts
```diff
@@ -0,0 +1,149 @@
...
+const DB_NAME = 'workbox-background-sync';
+const OBJECT_STORE_NAME = 'requests';
+const INDEXED_PROP = 'queueName';
+
+export interface UnidentifiedQueueStoreEntry {
+  requestData: RequestData;
+  timestamp: number;
+  id?: number;
+  queueName?: string;
+  metadata?: object;
```
jeffposnick, line 31: I know that this is what the old TypeScript definition used, but maybe there's an opportunity to clean that up a bit to remove some of the ambiguity, since `object` is overly broad: https://2ality.com/2020/01/typing-objects-typescript.html#object-vs.-object-in-typescript

Something like `Record<string, any>` might make more sense, or something more strict if we know that the values of each property will only be a certain subset of types.

(I'm assuming we can making "breaking" changes like that for this definition since it's internal-only, and not part of the public API, but correct me if I'm wrong.)

packages/workbox-background-sync/src/lib/QueueDb.ts
```diff
@@ -0,0 +1,149 @@
...
+}
+
+export interface QueueStoreEntry extends UnidentifiedQueueStoreEntry {
+  id: number;
+}
+
+/**
+ * A class to interact directly with IndexedDB to store a retrieve queue requests.
```
jeffposnick: I think there's a typo in this description.

packages/workbox-background-sync/src/lib/QueueDb.ts
```diff
@@ -0,0 +1,149 @@
...
+  private async open() {
+    if (!this._db) {
+      this._db = await openDB(DB_NAME, DB_VERSION, {
+        upgrade: this._upgradeDb,
+      });
+    }
+  }
+
+  /**
+   *
```
jeffposnick: If this is meant to be public, could you add a JSDoc description of what it does? Otherwise, add `@private`.

(That applies to the other methods as well.)

packages/workbox-background-sync/src/lib/QueueDb.ts
```diff
@@ -0,0 +1,149 @@
...
+
+export class QueueDb {
+  private _db: IDBPDatabase<QueueDBSchema> | null = null;
+
+  /**
+   * Opens a connection to IDBDatabase.
+   *
+   * @private
+   */
+  private async open() {
```
jeffposnick: Usage might be cleaner if you have this method always return `this._db`, and name it `getDB()` or something like that.

That way you wouldn't have to put the `!` after `this._db!` all over the place.

packages/workbox-background-sync/src/lib/QueueDb.ts
```diff
@@ -0,0 +1,149 @@
...
+    await this.open();
+    await this._db!.add(OBJECT_STORE_NAME, entry);
+  }
+
+  /**
+   * Returns the first entry in the ObjectStore
+   *
+   * @return {QueueStoreEntry}
+   */
+  async getFirstEntry(): Promise<QueueStoreEntry | any> {
```
jeffposnick: `Promise<QueueStoreEntry | any>` is the same as `Promise<any>` (for the same reason that `someValue || true` is the same as `true` in logic).

I think you mentioned that this might be `Promise<QueueStoreEntry | undefined>`?

packages/workbox-background-sync/src/lib/QueueDb.ts
```diff
@@ -0,0 +1,149 @@
...
+    await this.open();
+    const cursor = await this._db
+      ?.transaction(OBJECT_STORE_NAME)
+      .store.openCursor();
+
+    if (cursor) {
+      return cursor.value as QueueStoreEntry;
```
jeffposnick: I'm kind of surprised that you have to cast `cursor.value` to `QueueStoreEntry` here—what is the type of `cursor.value` if you don't cast? One of the advantages of the IDB library, I thought, was that it "knew" about the types defined in `QueueDBSchema`.

packages/workbox-background-sync/src/lib/QueueDb.ts
```diff
@@ -0,0 +1,149 @@
...
+   *
+   * @param query
+   * @return {Promise<QueueStoreEntry[]>}
+   */
+  async getAllEntriesFromIndex(
+    query?: IDBKeyRange,
+  ): Promise<QueueStoreEntry[]> {
+    await this.open();
+    const results =
+      await this._db?.getAllFromIndex(OBJECT_STORE_NAME, INDEXED_PROP, query) as QueueStoreEntry[];
```
jeffposnick: As with previous comment, is casting to `QueueStoreEntry[]` necessary? If `getAllFromIndex()` is returning something really generic, like `any[]` by default, then maybe the IDB library's schema definition isn't being set properly?

packages/workbox-background-sync/src/lib/QueueDb.ts
```diff
@@ -0,0 +1,149 @@
...
+   * Returns either the first or the last entries, depending on direction. Uses the index.
+   *
+   * @param {IDBCursorDirection} direction
+   * @param {IDBKeyRange} query
+   * @return {Promise<QueueStoreEntry | any>}
+   */
+  async getEndEntryFromIndex(
+    {direction}: {direction?: IDBCursorDirection},
+    query?: IDBKeyRange,
+  ): Promise<QueueStoreEntry | any> {
```
jeffposnick: Same comment about `Promise<QueueStoreEntry | any>` being the same as `Promise<any>`.

packages/workbox-background-sync/src/lib/QueueDb.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  /**
+   * Returns either the first or the last entries, depending on direction. Uses the index.
+   *
+   * @param {IDBCursorDirection} direction
+   * @param {IDBKeyRange} query
+   * @return {Promise<QueueStoreEntry | any>}
+   */
+  async getEndEntryFromIndex(
+    {direction}: {direction?: IDBCursorDirection},
```
jeffposnick: It seems a little awkward to have one parameter as a destructured object, and the second parameter as just a normal object.

Since I think this is not exposed publicly, do you want to just change the method signature so that it's either a single parameter that gets destructured, or alternatively two parameters without any destructuring? So:

```
async getEndEntryFromIndex({direction, query}: {
  direction?: IDBCursorDirection,
  query?: IDBKeyRange,
})
```

or

```
async getEndEntryFromIndex(direction?: IDBCursorDirection, query?: IDBKeyRange)
```

Also, if you know that one or both of the parameter values will always be set, you don't need to add a `?` after them. It's a TypeScript best practice to be as strict as possible and only loosen up restrictions (via `?`) if there are times when you really don't know if the value will be `undefined`.

packages/workbox-background-sync/src/lib/QueueStore.ts
```diff
@@ -7,27 +7,8 @@
...
-  timestamp: number;
-  id?: number;
-  queueName?: string;
-  metadata?: object;
-}
-
-export interface QueueStoreEntry extends UnidentifiedQueueStoreEntry {
-  id: number;
-}
+import {UnidentifiedQueueStoreEntry, QueueStoreEntry, QueueDb} from './QueueDb';
```
jeffposnick: Please add the `.js` suffix.

packages/workbox-background-sync/src/lib/QueueStore.ts
```diff
@@ -153,11 +130,10 @@ export class QueueStore {
    * @return {Promise<Array<Object>>}
    * @private
    */
-  async getAll(): Promise<QueueStoreEntry[]> {
-    return await this._db.getAllMatching(OBJECT_STORE_NAME, {
-      index: INDEXED_PROP,
-      query: IDBKeyRange.only(this._queueName),
-    });
+  async getAll(): Promise<QueueStoreEntry[] | any> {
```
jeffposnick: Same comment about how `Promise<QueueStoreEntry[] | any>` really means `Promise<any>`.

packages/workbox-background-sync/src/lib/QueueStore.ts
```diff
@@ -183,38 +159,14 @@ export class QueueStore {
    * @private
    */
   async _removeEntry({direction}: {direction?: IDBCursorDirection}) {
-    const [entry] = await this._db.getAllMatching(OBJECT_STORE_NAME, {
-      direction,
-      index: INDEXED_PROP,
-      query: IDBKeyRange.only(this._queueName),
-      count: 1,
-    });
+    const entry = await this._queueDb.getEndEntryFromIndex(
```
jeffposnick: See my earlier comment about the method signature for `getEndEntryFromIndex()`, as passing in the first parameter wrapped in `{}` looks awkward.

jeffposnick (changes requested):
Thanks for working on this!

In addition to the inline comments, I had a question about `test-QueueStore.mjs`. There are a lot of test cases removed, but no equivalents added. Is this because the test cases cover usage of `idb` that is already tested by the library's own test suite?

packages/workbox-background-sync/src/Queue.ts
tropicadri: updated.

packages/workbox-background-sync/src/lib/QueueDb.ts
tropicadri: Agreed, fixed it.

packages/workbox-background-sync/src/lib/QueueDb.ts
tropicadri: Yes, copy pasted from QueueStore, restructured all this for better readability + clearer names.

packages/workbox-background-sync/src/lib/QueueStore.ts
tropicadri: We should find a way to add this to our lint :)

package.json
tropicadri: For the test server. Correct me if I'm wrong but to use it in the tests, it needs to be included at the top-level. That is why I added the 'node_modules' as a directory in sw-bundle.js. Not sure if there is another way to do it.

packages/workbox-background-sync/src/lib/QueueDb.ts
tropicadri: Yes, I fixed them.

packages/workbox-background-sync/src/lib/QueueDb.ts
tropicadri: There are two interfaces defined for an entry UnidentifiedQueueStoreEntry and QueueStoreEntry, the difference is that QueueStoreEntry has to have an id, for the UnidentifiedQueueStoreEntry the id is optional. 

To add a record to it has to be unidentified because IDB sets the for the record when it. So the type is specified in the schema as UnidentifiedQueueEntry, line 16. The methods return Promise<UnidentifiedQueryEntry> that is why the need for the casting. No sure if there is a better way to handle that.

packages/workbox-background-sync/src/lib/QueueDb.ts
tropicadri: See above.

packages/workbox-background-sync/src/lib/QueueStore.ts
tropicadri: Fixed.

packages/workbox-background-sync/src/lib/QueueDb.ts
tropicadri: We talked about this a bit, but we ultimately expose it as public as part of the API: https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-background-sync.Queue#pushRequest so won't change it for now.

tropicadri:
> Thanks for working on this!
> 
> In addition to the inline comments, I had a question about `test-QueueStore.mjs`. There are a lot of test cases removed, but no equivalents added. Is this because the test cases cover usage of `idb` that is already tested by the library's own test suite?

Not sure if you saw this file test-QueueDb.mjs: https://github.com/GoogleChrome/workbox/pull/2805/files#diff-0527463f9abf935dcbb6cd2edf286adef83a169bbccfa88aa1bc43e18518cd35 (github collapses the file because the diff is big)

The tests that I removed from test-QueueStore.mjs where moved there, because the functionality is now implemented there + added more tests for the class

package.json
jeffposnick: 👍

packages/workbox-background-sync/src/lib/QueueDb.ts
jeffposnick: 👍

jeffposnick:
I have been burned so many times by GitHub collapsing files that have too many changes. (**The files with a lot of changes are the ones I want to pay closest attention to, GitHub!**)

I'll take a look at the new tests.

jeffposnick (changes requested):
The tests that I originally overlooked look good!

The main unresolved issue has to do with the distinction between `UnidentifiedQueueStoreEntry` and `QueueStoreEntry`.

Not surprisingly, other folks have run into this same issue, it seems: https://github.com/jakearchibald/idb/issues/150

Until that underlying `idb` library issue is fixed, it sounds like we need to work around it in our code. I've got an inline comment about doing the type conversion inside of `addEntry()` instead of when IDB is read from—I think that would be cleaner overall.

packages/workbox-background-sync/src/lib/QueueDb.ts
```diff
@@ -0,0 +1,167 @@
...
+*/
+
+import {openDB, DBSchema, IDBPDatabase} from 'idb';
+import {RequestData} from './StorableRequest.js';
+import '../_version.js';
+
+interface QueueDBSchema extends DBSchema {
+  requests: {
+    key: number;
+    value: UnidentifiedQueueStoreEntry;
```
jeffposnick: Can this be switched to `QueueStoreEntry`?

packages/workbox-background-sync/src/lib/QueueDb.ts
```diff
@@ -0,0 +1,167 @@
...
+  private _db: IDBPDatabase<QueueDBSchema> | null = null;
+
+  /**
+   * Add QueueStoreEntry to underlying db.
+   * 
+   * @param {UnidentifiedQueueStoreEntry} entry
+   */
+  async addEntry(entry: UnidentifiedQueueStoreEntry) {
+    const db = await this.getDb();
+    await db.add(REQUEST_OBJECT_STORE_NAME, entry);
```
jeffposnick: Can this be switched to

```
await db.add(REQUEST_OBJECT_STORE_NAME, entry as QueueStoreEntry);
```

I think if we just "fake" the type info in that one place, and everywhere that we read from IDB we get back a `QueueStoreEntry`, that would be cleaner overall.

packages/workbox-background-sync/src/lib/QueueDb.ts
```diff
@@ -0,0 +1,167 @@
...
+   *
+   * @private
+   */
+  private async getDb() {
+    if (!this._db) {
+      this._db = await openDB(DB_NAME, DB_VERSION, {
+        upgrade: this._upgradeDb,
+      });
+    }
+    return this._db
```
jeffposnick: Missing trailing `;`.

(I don't know why our linting doesn't pick this up. Maybe once https://github.com/GoogleChrome/workbox/issues/2602 is resolved it might?)

packages/workbox-background-sync/src/lib/QueueDb.ts
```diff
@@ -0,0 +1,167 @@
...
+  async getEndEntryFromIndex(
+    query: IDBKeyRange,
+    direction: IDBCursorDirection,
+  ): Promise<QueueStoreEntry | undefined> {
+    const db = await this.getDb();
+
+    const cursor = await db.transaction(REQUEST_OBJECT_STORE_NAME)
+      .store.index(QUEUE_NAME_INDEX)
+      .openCursor(query, direction);
+    return cursor?.value as QueueStoreEntry;
```
jeffposnick: If you change the schema to define `value` as `QueueStoreEntry` you shouldn't have to cast here anymore.

packages/workbox-background-sync/src/lib/QueueDb.ts
```diff
@@ -0,0 +1,167 @@
...
+   * Get all the entries filtered by index
+   * 
+   * @param queueName
+   * @return {Promise<QueueStoreEntry[]>}
+   */
+  async getAllEntriesByQueueName(
+    queueName: string,
+  ): Promise<QueueStoreEntry[]> {
+    const db = await this.getDb();
+    const results = await db.getAllFromIndex(REQUEST_OBJECT_STORE_NAME, QUEUE_NAME_INDEX, IDBKeyRange.only(queueName)) as QueueStoreEntry[];
```
jeffposnick: You can remove this type conversion if you change the `value` field to be of type `QueueStoreEntry`.

jeffposnick (changes requested):
(Sorry, these are the comments that were meant to be attached to the previous review.)

The tests that I originally overlooked look good!

The main unresolved issue has to do with the distinction between `UnidentifiedQueueStoreEntry` and `QueueStoreEntry`.

Not surprisingly, other folks have run into this same issue, it seems: https://github.com/jakearchibald/idb/issues/150

Until that underlying `idb` library issue is fixed, it sounds like we need to work around it in our code. I've got an inline comment about doing the type conversion inside of `addEntry()` instead of when IDB is read from—I think that would be cleaner overall.

jeffposnick (approved):
Awesome! Looks good to me.

PR #2811, opened 2021-04-12
Adding .vscode to .gitignore to be able to customize workspace

tropicadri:
Adding .vscode to .gitignore to be able to customize workspace

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ docs
 local-builds/
 firebase-debug.log
 .firebase
+.vscode/
 
 # Generated TypeScript files and build data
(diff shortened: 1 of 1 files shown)


jeffposnick (commented):
It's not a huge deal, but what I tend to do for this sort of thing is keep ignore settings specific to my work setup in the `.gitignore` that's in my home directory, which I have [configured](https://tekin.co.uk/2020/03/maintain-a-global-git-ignore-file) to automatically apply to my local git environment:

```sh
$ cat ~/.gitignore
.idea/
*.iml
.vscode/
.DS_Store
```

Would you be okay with that in lieu of this PR?

PR #2820, opened 2021-04-15
Removing githead attribute from package.json files from lerna error

### LINKED ISSUE #2819: Clear gitHead attribute added by lerna when published failed
**Library Affected**:
*all packages*

During the last deploy lerna added an attribute to each package.json, apparently this can happen if the publish doesn't finish correctly. We need to clean that attribute from all the our package.json. See more details here: https://github.com/lerna/lerna/issues/1880

tropicadri:
Fixes #2819

### DIFF
--- a/packages/workbox-background-sync/package.json
+++ b/packages/workbox-background-sync/package.json
@@ -25,6 +25,5 @@
   "types": "index.d.ts",
   "dependencies": {
     "workbox-core": "^6.1.5"
-  },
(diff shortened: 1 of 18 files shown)


jeffposnick (approved):
Just to confirm, this wasn't causing any issues for users of our packages, right? It's just internal cleanup?

tropicadri:
> Just to confirm, this wasn't causing any issues for users of our packages, right? It's just internal cleanup?

Yes, everything was working as intended just noticed it when I was doing idb changes :)

PR #2823, opened 2021-04-22
fetch() stubs in Google Analytics tests

jeffposnick:
R: @tropicadri

This kind of papers over the issue we were seeing with the `workbox-google-analytics` tests in Chrome, but at the same time, making real requests to Google Analytics URLs in order to test routing isn't necessary.

### DIFF
--- a/test/workbox-google-analytics/sw/test-initialize.mjs
+++ b/test/workbox-google-analytics/sw/test-initialize.mjs
@@ -149,7 +149,6 @@ describe(`initialize`, function() {
   it(`should register GET/POST routes for collect endpoints`, async function() {
     // Stub out fetch(), since this test is about routing, and doesn't need to
     // contact the production Google Analytics endpoints.
-    // (This also works around some unresolved issues in Chrome 90+.)
     sandbox.stub(self, 'fetch').callsFake(() => new Response('ignored'));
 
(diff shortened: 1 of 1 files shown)


test/workbox-google-analytics/sw/test-initialize.mjs
```diff
@@ -147,6 +147,11 @@ describe(`initialize`, function() {
   });
 
   it(`should register GET/POST routes for collect endpoints`, async function() {
+    // Stub out fetch(), since this test is about routing, and doesn't need to
+    // contact the production Google Analytics endpoints.
+    // (This also works around some unresolved issues in Chrome 90+.)
```
tropicadri: Do we want to specifically call out the issue in Chrome 90? The stubbing is the right thing to do, it would be sort of a legacy comment that no one other than you and me would understand and that in a couple of years of months even us won't remember what it was about 

Same for bellow.

tropicadri (changes requested):
Should we also stub fetch here for consistency?  line 198 `it(`should not alter successful hit payloads`, async function() {`

test/workbox-google-analytics/sw/test-initialize.mjs
jeffposnick: Sure, dropping the comment.

tropicadri (approved):

PR #2838, opened 2021-05-06
DbWrapper deprecation

### LINKED ISSUE #2801: Deprecate DBWrapper

tropicadri:
Fixes #2801

Changes in workbox-expiration to use idb and deleting DBWrapper.


packages/workbox-expiration/src/models/CacheTimestampsModel.ts
```diff
@@ -76,7 +78,9 @@ class CacheTimestampsModel {
 
     // Previous versions of `workbox-expiration` used `this._cacheName`
     // as the IDBDatabase name.
-    deleteDatabase(this._cacheName);
+    if (this?._cacheName) {
```
jeffposnick: I'm curious as to why it's `this?` instead of `this`. When would `this` not be defined?

jeffposnick (commented):
Thanks! This all looks a lot cleaner, and I'm glad we were able to drop so much old code.

I just had the one small comment inline.

packages/workbox-expiration/src/models/CacheTimestampsModel.ts
tropicadri: Glad you asked, I meant to ask you about that, we call the method to upgradeDB in tests, for example: https://github.com/GoogleChrome/workbox/blob/v6/test/workbox-expiration/sw/test-CacheTimestampsModel.mjs#L16, and that is when `this` would not be defined. I am not sure if this is still the correct way to do this tests, because the test would be the one creating the IDB because it is opened before each test, maybe we should refactor the tests to open the DB only in the test cases that needs it and only by passing the name?

jeffposnick: Hmmm, that's interesting.

I don't think most of the tests end up ever triggering the IDB upgrade flow, do they? Can we just leave out `onupgradeneeded` from when we call `openDB()` in our tests?

I think would be cleaner than having a situation in which we need to use `this?` in the non-test code.

jeffposnick (changes requested):

jeffposnick:
@tropicadri pointed out that `onupgradeneeded` always runs when going from the initial empty IDB state to an initial IDB instance (not just when going from an initial IDB instance to a new one), so forget what I mentioned about leaving out `onupgradeneeded`.

As discussed, moving the call to `openDB()` to run later in the test suite makes the most sense.

packages/workbox-expiration/src/models/CacheTimestampsModel.ts
tropicadri: Turns out my approach also wouldn't work, so I sort of separated the dbupgrade and the deleting of old dbs and that seems to work, let me know what you think. Thank you!

jeffposnick (approved):

PR #2882, opened 2021-07-04
feat: migrate workbox-webpack-plugin to typescript

### LINKED ISSUE #2478: Migrate workbox-webpack-plugin to TypeScript
Placeholder to track the migration.

roikoren755:
- ensure that `gulp build && gulp lint test` passes locally.

R: @jeffposnick @tropicadri

Fixes #2478

This is a work in progress.
Couldn't get tests to run locally, and there are some `@ts-ignore` comments littered about in places that import things from `workbox-build` or `workbox-google-analytics` that I couldn't quite figure out.

I think getting this package to use the new TS version of `workbox-build` would get it almost ready to merge 🤞.

Feel free to push changes, or point me in the right direction to fix things :)


tropicadri:
@roikoren755 Thank you for this, sorry for the late reply we've been doing some upgrading of TS lint and as you saw workbox-build migration to TS, I'll take a look at the PR and help where I can :)

tropicadri:
This is far from ready, but I wanted to share the updates. 

- I merged with v6 branch that has the latest eslint version, we don't use @ts-ignore when we need to disable eslint we use // eslint-disable...
- I reused the types from workbox-build instead of creating new ones
- I setup the build to use compile with Typescript
- Still fixing tests.

Let me know if you have any questions, thank you for your patience with this one.

jeffposnick:
Thanks! I'll take a look at this soon.

package.json
```diff
@@ -22,6 +22,7 @@
     "@types/fs-extra": "^9.0.5",
     "@types/lodash": "^4.14.165",
     "@types/stringify-object": "^3.3.0",
+    "@types/webpack": "^5.28.0",
```
jeffposnick: Should this be at the top-level package.json `devDependencies`? I would have thought it would be in `devDependencies` of `packages/workbox-webpack-plugin/package.json`.

packages/workbox-webpack-plugin/src/generate-sw.ts
```diff
@@ -6,26 +6,27 @@
...
 // https://github.com/webpack/webpack/issues/11425#issuecomment-686607633
 const {RawSource} = webpack.sources || require('webpack-sources');
 
 // Used to keep track of swDest files written by *any* instance of this plugin.
 // See https://github.com/GoogleChrome/workbox/issues/2181
-const _generatedAssetNames = new Set();
+const _generatedAssetNames = new Set<string>();
+
+interface GenerateSWConfig extends WebpackGenerateSWOptions {
+  manifestEntries?: Array<ManifestEntry>;
```
jeffposnick, line 28: I'm not sure what's going on here with the optional `manifestEntries`. Could you explain?

packages/workbox-webpack-plugin/src/generate-sw.ts
```diff
@@ -79,7 +82,7 @@ class GenerateSW {
    * [asset's metadata](https://github.com/webpack/webpack/issues/9038) is used
    * to determine whether it's immutable or not.)
    *
-   * @param {Array<string|RegExp|Function>} [config.exclude=[/\.map$/, /^manifest.*\.js$]]
+   * @param {Array<string|RegExp>} [config.exclude=[/\.map$/, /^manifest.*\.js$]]
```
jeffposnick, line 85: Is there a reason that `Function` is dropped here? I see that the code is still in place to support functions as include/exclude criteria.

(In the future, this JSDoc won't be a source of truth, so it doesn't matter too much.)

packages/workbox-webpack-plugin/src/inject-manifest.ts
```diff
@@ -6,24 +6,23 @@
...
-const relativeToOutputPath = require('./lib/relative-to-output-path');
+import {escapeRegExp} from 'workbox-build/build/lib/escape-regexp';
+import {replaceAndUpdateSourceMap} from 'workbox-build/build/lib/replace-and-update-source-map';
+import {validateWebpackInjectManifestOptions} from 'workbox-build/build/lib/validate-options';
+import prettyBytes from 'pretty-bytes';
+import stringify from 'fast-json-stable-stringify';
+import upath from 'upath';
+import webpack from 'webpack';
+
+//import {CommonConfig} from './types';
```
jeffposnick: You can remove this comment.

packages/workbox-webpack-plugin/src/inject-manifest.ts
```diff
@@ -205,25 +222,27 @@ class InjectManifest {
 
     if (Array.isArray(this.config.webpackCompilationPlugins)) {
       for (const plugin of this.config.webpackCompilationPlugins) {
-        plugin.apply(childCompiler);
+        plugin.apply(childCompiler); //eslint-disable-line
```
jeffposnick: What's the underlying issue here?

packages/workbox-webpack-plugin/src/inject-manifest.ts
```diff
@@ -249,15 +273,23 @@ class InjectManifest {
...
+  ): Promise<void> {
     try {
       this.config = validateWebpackInjectManifestOptions(this.config);
     } catch (error) {
-      throw new Error(`Please check your ${this.constructor.name} plugin ` +
-        `configuration:\n${error.message}`);
+      if (error instanceof Error) {
+        throw new Error(
+          `Please check your ${this.constructor.name} plugin ` +
+            // eslint-disable-next-line
```
jeffposnick: Is this needed?

packages/workbox-webpack-plugin/src/inject-manifest.ts
```diff
@@ -300,72 +345,87 @@ class InjectManifest {
...
+      config.swDest!,
+    );
 
     if (sourcemapAssetName) {
       _generatedAssetNames.add(sourcemapAssetName);
       const sourcemapAsset = compilation.getAsset(sourcemapAssetName);
       const {source, map} = await replaceAndUpdateSourceMap({
-        jsFilename: config.swDest,
-        originalMap: JSON.parse(sourcemapAsset.source.source()),
+        // eslint-disable-next-line
```
jeffposnick: What's the error here? If possible, could you use the form of the comment that disables a specific rule rather than all rules, so that it's clearer what's being worked around?

https://eslint.org/docs/user-guide/configuring/rules#using-configuration-comments

packages/workbox-webpack-plugin/src/lib/get-manifest-entries-from-compilation.ts
```diff
@@ -6,12 +6,18 @@
...
 */
 
-const {matchPart} = require('webpack').ModuleFilenameHelpers;
-const {transformManifest} =
-    require('workbox-build/build/lib/transform-manifest');
-
-const getAssetHash = require('./get-asset-hash');
-const resolveWebpackURL = require('./resolve-webpack-url');
+import {Asset, Chunk, Compilation, ModuleFilenameHelpers} from 'webpack';
+import {WebpackError} from 'webpack';
```
jeffposnick: You can include this is the previous line's `import ...`

packages/workbox-webpack-plugin/src/lib/get-manifest-entries-from-compilation.ts
```diff
@@ -24,14 +30,15 @@ const resolveWebpackURL = require('./resolve-webpack-url');
...
+function checkConditions(
+  asset: Asset,
+  conditions: Array<string | RegExp | ((arg0: string) => boolean)> = [],
+): boolean {
   for (const condition of conditions) {
     if (typeof condition === 'function') {
-      if (condition({asset, compilation})) {
-        return true;
-      }
+      return condition(asset.name);
```
jeffposnick: I wonder if this is something that changed between `webpack` v4 and v5. I do see some v5 types that imply that conditions take a single string (and https://webpack.js.org/configuration/module/#condition sort of implies that too), but at one point I thought we needed to pass in `{asset, compilation}`... I am having trouble figuring out why that is, though.

But, since we aim to make this a backwards—compatible release, I feel like we need to keep the old method call here, and move to the "correct" signature of `(value: string) => boolean` as a breaking change in Workbox v7.

packages/workbox-webpack-plugin/src/lib/get-manifest-entries-from-compilation.ts
```diff
@@ -159,14 +181,14 @@ function filterAssets(compilation, config) {
     }
 
     // Next, check asset-level checks via includes/excludes:
-    const isExcluded = checkConditions(asset, compilation, config.exclude);
+    const isExcluded = checkConditions(asset, config.exclude);
```
jeffposnick: See previous comment—we might need to keep the old signature for backwards compatibiltiy.

packages/workbox-webpack-plugin/src/lib/get-manifest-entries-from-compilation.ts
```diff
@@ -178,17 +200,20 @@ function filterAssets(compilation, config) {
...
+  config: WebpackGenerateSWOptions | WebpackInjectManifestOptions,
+): Promise<{size: number; sortedEntries: ManifestEntry[]}> {
   const filteredAssets = filterAssets(compilation, config);
 
   const {publicPath} = compilation.options.output;
 
   const fileDetails = Array.from(filteredAssets).map((asset) => {
     return {
-      file: resolveWebpackURL(publicPath, asset.name),
+      file: resolveWebpackURL(publicPath as string, asset.name),
```
jeffposnick, line 225: I had no idea `publicPath` could be `((pathData: PathData, assetInfo?: AssetInfo | undefined) => string)` and I don't see any valid values other than a `string` mentioned in the docs at https://webpack.js.org/guides/public-path/

We can probably just ignore that possibility and unconditionally cast to `string` like you're doing, but it's weird...

jeffposnick (changes requested):
Thanks, @tropicadri! I have a few comments to go over once you're back in the office, but generally speaking, this looks good!

package.json
tropicadri: You are 100% right

packages/workbox-webpack-plugin/src/inject-manifest.ts
tropicadri: Done

packages/workbox-webpack-plugin/src/lib/get-manifest-entries-from-compilation.ts
tropicadri: Done

packages/workbox-webpack-plugin/src/generate-sw.ts
tropicadri: I am not sure why it was dropped, but I put it back to keep it consistent with the webpack options.

packages/workbox-webpack-plugin/src/inject-manifest.ts
tropicadri: it doesn't know what type plugin is, I'll check with you tomorrow, if they are specific type (I'm using EntryPlugin) or if we should do the disabling of the linting or something else.

packages/workbox-webpack-plugin/src/inject-manifest.ts
tropicadri: I fixed the ones that didn't have the issue

packages/workbox-webpack-plugin/src/lib/get-manifest-entries-from-compilation.ts
tropicadri: I had to do some "anying" to keep it the way it was before, it doesn't seem to have changed a lot: https://v4.webpack.js.org/configuration/module/#condition, I am not sure why we are passing `{asset, compilation}` it seems that would invoke the case where compilation receives an object and all the properties should match, but I have no idea why it is implemented this way...so we'll keep it and change in v7

packages/workbox-webpack-plugin/src/lib/get-manifest-entries-from-compilation.ts
tropicadri: same as above

packages/workbox-webpack-plugin/src/lib/get-manifest-entries-from-compilation.ts
tropicadri: keeping the casting and if we run into troubles will change it.

packages/workbox-webpack-plugin/src/inject-manifest.ts
tropicadri: I had to disable the check for unsafe type and for unsafe access, I think there are different types for v4 and v5 so we have plugins be 'any' so I added comments about all that.

packages/workbox-webpack-plugin/src/generate-sw.ts
tropicadri: since we are initializing the config with {} it complains if the manifestEntries are not optional...

packages/workbox-webpack-plugin/src/inject-manifest.ts
jeffposnick: Thanks!

jeffposnick (approved):
🥳 

Thanks so much to @roikoren755 for kicking this off, and @tropicadri for finishing it up!

PR #2916, opened 2021-08-13
Debug logging when caching Vary: headers

### LINKED ISSUE #2818: Better dev logging for Vary: headers

**Library Affected**:
workbox-expiration

**Browser & Platform**:
Chrome 89.0.4389.114 (Official Build) (64-bit) on Linux

I observed a problem where [my cache kept adding entries and never deleting them, even though I was using ExpirationPlugin with maxEntries of 1](https://stackoverflow.com/questions/67085513/workbox-expirationplugin-with-maxentries-of-1-seems-to-not-enforce-any-limit-on).

Through some console.log debugging, I narrowed it down to [this piece of code in Workbox](https://github.com/GoogleChrome/workbox/blob/ae29e1748bb976351db8e0c64a9ea9fc8ba2583f/packages/workbox-expiration/src/CacheExpiration.ts#L115-L117). I found that `urlsExpired` looks correct and `cache.delete` is being called with no error, but the cache entry is not actually being deleted. Checking the return value of `cache.delete` and I see it resolves to false.

[MDN says:](https://developer.mozilla.org/en-US/docs/Web/API/Cache/delete)

> resolves to `true` if the cache entry is deleted, or `false` otherwise

[This article says:](https://developers.google.com/web/ilt/pwa/caching-files-with-service-worker)

> If it doesn't find the item, it resolves to false.

So I guess that implies it's not finding the item? But look. Here's how I added my console.logs:

```js
    console.log('urlsExpired', urlsExpired);
    console.log('keys', await cache.keys());
    for (const url of urlsExpired) {
      console.log('delete', url, this._matchOptions);
      const deleted = await cache.delete(url, this._matchOptions);
      console.log('after delete', url, deleted);
    }
    console.log('keys2', await cache.keys());
```

And here's the console output I see:

image

[MDN says the first argument can be a Request object or a URL](https://developer.mozilla.org/en-US/docs/Web/API/Cache/delete), so I don't know why it wouldn't find the entry in the cache that is clearly there at the first "keys" log.

So that's why I'm submitting this issue in addition to making that StackOverflow post... I'm concerned this is either a bug in Workbox, or a bug in Chrome that Workbox needs to work around. My apologies if it's just me doing something stupid!

jeffposnick:
R: @tropicadri

Fixes #2818

This attempts to improve a common developer pain point via some extra debug-level logging.


packages/workbox-strategies/src/StrategyHandler.ts
```diff
@@ -319,6 +319,16 @@ class StrategyHandler {
           method: effectiveRequest.method,
         });
       }
+
+      // See https://github.com/GoogleChrome/workbox/issues/2818
+      if (response.headers.has('Vary')) {
+        logger.debug(
+          `The response for ${getFriendlyURL(effectiveRequest.url)} ` +
+            `has a 'Vary: ${response.headers.get('Vary')}' header. ` +
```
tropicadri: please fix: "Invalid type "string | null" of template literal expression", apparently Typescript doesn't understand that you already checked with the .has?  ¯\_(ツ)_/¯

tropicadri (approved):
It is a very straight forward change, I'll approve it pending the fix for the ts-lint complaint.

PR #2918, opened 2021-08-17
Use Application Default Credentials instead of Service Account

jeffposnick:
R: @tropicadri

This switches from using service account info to cached Application Default Credentials when interacting with the Google Cloud Storage API.

Folks who have permissions to publish to Google Cloud Storage will need to run `gcloud auth application-default login` prior to using the script in order to generate local credentials.

I'm not 100% if this will work as-is, so we may need to tweak it before we do our next CDN publish.

### DIFF
--- a/gulp-tasks/utils/cdn-helper.js
+++ b/gulp-tasks/utils/cdn-helper.js
@@ -6,8 +6,6 @@
   https://opensource.org/licenses/MIT.
 */
 
-const {oneLine} = require('common-tags');
-const fs = require('fs-extra');
 const glob = require('glob');
 const path = require('path');
 const {Storage} = require('@google-cloud/storage');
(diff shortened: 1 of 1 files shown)


gulp-tasks/utils/cdn-helper.js
```diff
@@ -30,23 +26,10 @@ class CDNHelper {
...
-  to the CDN.
-  Please get the service account file and put it at the root of the
-  workbox repo.
-
-  File Path Tested: '${SERVICE_ACCOUNT_PATH}'.
-`;
-        throw new Error(errorMessage);
-      }
-
+      // Run `gcloud auth application-default login` if needed.
```
tropicadri, line 29: Should we log this instructions some where if the script fails?

jeffposnick: Maybe, but I'm not actually sure _what_ will fail. Like, I don't think the `Storage` constructor will throw (but maybe it will)? We should try it out and adjust as needed during the next release.

We're the only two people who will be running this, so I'm not as worried.

tropicadri: SGTM

tropicadri (approved):

PR #2921, opened 2021-08-18
Allow precaching "repair" when using subresource integrity

### LINKED ISSUE #2858: Allow on-demand precaching for cache recovery
**Library Affected**:
workbox-precaching

**Browser & Platform**:
All

In Workbox v6, PrecacheController changed such that precaching can only be performed during certain SW lifecycle events. This unfortunately breaks a use case that was supported in v5 and before -- cache recovery. If SW cache entries are removed by a browser while the SW stays active, offline mode will not work even though the SW seems healthy. We were doing cache validation based on the embedded manifest during app startup, and triggering a precache run with `controller.install()` if any entries from the manifest were missing from the cache.

A workaround for this case is to unregister the service worker such that reinstallation will trigger precaching, but critically, this won't happen until the next application load, which is problematic for long-lived app sessions.

A top-level API like `validateAndRecoverPrecache` next to `precacheAndRoute` would be ideal, but barring that, restoring ad-hoc usage of the PrecacheController's installer would also work.

jeffposnick:
R: @tropicadri

Fixes #2858

Although unexpected, there are edge cases where the precache might not be in an inconsistent state, most likely due to a developer manually deleting something in DevTools.

When this happens, `workbox-precaching` defaults to falling-back to using a network response (assuming the device is online) when there's a precaching miss. Up until now, `workbox-precaching` hasn't attempting to use this network response to repopulate the precache, because there are no guarantees that the network response corresponds to the version of the asset specified in the precache manifest.

However, if the precache entry includes an [`integrity` property](https://developers.google.com/web/tools/workbox/modules/workbox-precaching#using_subresource_integrity), then [subresource integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) guarantees that the response _does_ correspond to the same version of the asset in the manifest. So it should be safe to "repair" the cache with that response.

(The `revision` field can't be used for this purpose, because while it's traditionally set to the hash of an asset's content, it could theoretically be set to anything, and it would be much too expensive to calculate the hash of the response body via JavaScript at runtime.)


jeffposnick:
(Actually, hold off on the review. I noticed something that needs to be changed when testing locally.)

jeffposnick:
...I realize now that the code in this PR will work if there's an `integrity` property on the actual request being made, but the `integrity` information from the precache manifest is being ignored. I need to find a clean way to pass information about the precache manifest contents (which is stored in `PrecacheController`) to the code that handles requests (which is in `PrecacheStrategy`).

jeffposnick:
I've got a more robust approach, but it's blocked on #2914 (we need to be able to check the `integrity` property of the incoming `request` in the `PrecacheStrategy`, but without that PR, it's lost).

test/workbox-precaching/sw/test-PrecacheStrategy.mjs
```diff
@@ -54,6 +54,44 @@ describe(`PrecacheStrategy()`, function() {
...
+
+      const response3 = await ps.handle(createFetchEvent('/three'));
+      expect(await response3.text()).to.equal('Fetched Response');
+      expect(self.fetch.callCount).to.equal(2);
+
+      // /two should be there, since integrity is used. /three shouldn't.
+      const cachedUrls = (await cache.keys()).map((request) => request.url);
+      expect(cachedUrls).to.eql([
+        `${location.origin}/one`,
+        `${location.origin}/two`,
```
tropicadri, line 120: This is failing because the integrity in the request is 'some-hash' but there is integrity in manifest is ""

tropicadri (changes requested):

test/workbox-precaching/sw/test-PrecacheStrategy.mjs
jeffposnick: I fixed the test to account for the current implementation.

tropicadri (approved):

PR #2923, opened 2021-08-21
Run a single, top-level tsc --build

jeffposnick:
R: @tropicadri 

TypeScript has support for [project references](https://www.typescriptlang.org/docs/handbook/project-references.html), which work well for monorepos, like our own. We've been configuring our `tsconfig.json` files for use with project references, but we haven't actually been taking advantage of them from a compilation perspective. Instead, our build process would start a new `tsc --build` for each sub-package. This is partly because it was easier to migrate to TypeScript if we kept our old per-package build process in place, but now that we're almost all TypeScript, it can be improved.

This PR removes all of the per-package `tsc --build` invocations in favor of a single `tsc --build` run with our top-level `tsconfig.json`, which includes all of the project references.

This PR also stops cleaning up the TypeScript build artifacts from the previous build prior to starting a new build, as the [`composite` setting](https://www.typescriptlang.org/tsconfig/#composite) means that they can be used to speed up subsequent builds. This is useful when running `gulp build` a few times in row while testing out some changes. I've moved the clean-up process to the `gulp publish` task flow, to ensure that nothing out of date ends up  inadvertently published. (I haven't tested this flow yet, so this is something to keep an eye on the next time we cut a release.)

With these changes, I've seen `gulp build` drop from around **75 seconds to 30 seconds** on my laptop.


gulp-tasks/transpile-typescript.js
```diff
@@ -68,79 +59,36 @@ function needsTranspile(packageName) {
...
 
 /**
- * Transpiles a package iff it has TypeScript source files.
- *
- * @param {string} packagePath
- * @param {Object} [options]
+ * Transpiles all packages listed in the root tsconfig.json's references section
+ * into .js and .d.ts files. Creates stub .mjs files that re-export the contents
+ * of the .js files.
+ *
```
tropicadri:
```suggestion
 *
```

tropicadri (approved):

PR #2924, opened 2021-08-21
Use prettier against our codebase

jeffposnick:
R: @tropicadri

This PR enables `prettier` as a pre-commit hook via `husky`, and includes the output of running `prettier --write .` against our current codebase.

We should probably hold off on merging this until the other open PRs are merged, but I figured I'd get the PR filed.


jeffposnick:
Closing in favor of #2929.

PR #2934, opened 2021-09-03
changing durability to relaxed

### LINKED ISSUE #2854: Consider `durability: relaxed` for indexeddb writes
**Library Affected**:
*workbox-expiration*

**Browser & Platform**:
Chrome

Hi, I'm a former chrome engineer working on a web app now. We've studied some chrome traces of our app where the expiration cache timestamps writes get backed up and cause significant contention. This bottleneck can be alleviated by setting `durability: relaxed` on any `readwrite` indexeddb transactions. This prevents blocking on disk flushes. It's the default behavior in Firefox and probably will (or should) be the default in Chrome at some point. That's because it's actually a very significant perf gain for a very minor durability sacrifice -- catastrophic events like power failure, OS crash, disk full may cause data loss, but things like browser crash or tab close do not.

There's more detail in the chrome status:
https://www.chromestatus.com/feature/5730701489995776

If this behavior makes sense, the fix would entail changing `put`s and similar here:
https://github.com/GoogleChrome/workbox/blob/bbaa0bb5bc7830543be4f170d108842dd1f19e01/packages/workbox-expiration/src/models/CacheTimestampsModel.ts#L114

Slow (current):
```ts
await db.put(CACHE_OBJECT_STORE, entry);
```

Fast (untested in this context):
```ts
const tx = db.transaction(CACHE_OBJECT_STORE, 'readwrite', { durability: 'relaxed' });
await tx.store.put(CACHE_OBJECT_STORE, entry);
await tx.done;
```

tropicadri:
After reading #2854 and this [article]( https://nolanlawson.com/2021/08/22/speeding-up-indexeddb-reads-and-writes/), seeing it is the default behavior in FF that seems to improve the performance, that is supported in Chrome and seems to be soon supported in Safari, I think it is a good idea to change the durability.

Fixes #2854

### DIFF
--- a/packages/workbox-background-sync/src/lib/QueueDb.ts
+++ b/packages/workbox-background-sync/src/lib/QueueDb.ts
@@ -56,7 +56,9 @@ export class QueueDb {
    */
   async addEntry(entry: UnidentifiedQueueStoreEntry): Promise<void> {
     const db = await this.getDb();
-    await db.add(REQUEST_OBJECT_STORE_NAME, entry as QueueStoreEntry);
+    const tx = db.transaction(REQUEST_OBJECT_STORE_NAME, 'readwrite', { durability: 'relaxed' });
+    await tx.store.add(entry as QueueStoreEntry);
+    await tx.done;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


jeffposnick (changes requested):
We also write to IndexedDB (via `add()` instead of `put()`) at: https://github.com/GoogleChrome/workbox/blob/63ea2ca52e7278411a71520f48bd9cf700694edd/packages/workbox-background-sync/src/lib/QueueDb.ts#L57-L60

Should that be moved inside a transaction with relaxed durability as well?

jeffposnick (approved):

PR #2941, opened 2021-09-15
Add size() to expose amount of requests in queue

### LINKED ISSUE #2940: expose amount of requests in queue
**Library Affected**:
`workbox-background-sync`

**Browser & Platform**:
N/a

I've got a custom `onSync` function for `workbox-background-sync` as I need to modify requests as they are sent over the networ... anyway, this means my app has a small synchronisation icon in the app bar which displays the number of requests that are still queued up when the customer has no (or a very slow) network.

I currently do this using `(await queue.getAll()).length` and decrementing the number every time I shift a request from the queue in my `onSync` function.

This works, except for the case where the user is causing more requests to be added to the queue while `onSync` is running, in which case the counter will go into the negative.

I can solve this by doing `(await queue.getAll()).length` before every single request, but that feels like it would be terribly inefficient.

I noticed that `getAll` (a few levels deep) is essentially:
```
  async getAllEntriesByQueueName(
    queueName: string,
  ): Promise<QueueStoreEntry[]> {
    const db = await this.getDb();
    const results = await db.getAllFromIndex(REQUEST_OBJECT_STORE_NAME, QUEUE_NAME_INDEX, IDBKeyRange.only(queueName));
    return results ? results : new Array<QueueStoreEntry>();
  }
  ```
  Seeing `idb` is used under the hood, we can create a new function (almost of clone of the one above) that uses `countFromIndex` instead. With that, something like `queue.size()` could be exposed (or `.getCount`, whatever).
  
  Would this be something that could be accepted as a feature? I can implement it myself and provide a pull request :-)

### PR DESCRIPTION
**Pull request by StephanBijzitter**
R: @jeffposnick @tropicadri

Fixes  GoogleChrome/workbox#2940

In short, this allows a developer to perform a quick count of what's left in the queue, so that a number could be displayed in a status indicator... or whatever else the developer wants to do with the number.

It's not as heavy as `(await queue.getAll()).length` and doesn't modify the queue

>- ensure that `gulp build && gulp lint test` passes locally.

Well, I tried... but Windows doesn't very much like the npm scripts and as such I had to bypass the git hooks as well, it just wouldn't run.. Hopefully I managed to keep everything consistent with the surrounding code :-)

### DIFF
--- a/test/workbox-background-sync/sw/lib/test-QueueDb.mjs
+++ b/test/workbox-background-sync/sw/lib/test-QueueDb.mjs
@@ -350,6 +350,26 @@ describe(`QueueDb`, () => {
     });
   });
 
+  describe('getEntryCountByQueueName', () => {
+    it(`should return the number of entries in IDB filtered by index`, async () => {
+      const queueDb = new QueueDb();
+
+      await queueDb.addEntry(entry1);
+      await queueDb.addEntry(entry2);
+      await queueDb.addEntry(entry3);
+      await queueDb.addEntry(entry4);
+      await queueDb.addEntry(entry5);
+
+      expect(await queueDb.getEntryCountByQueueName('a')).to.equal(3);
+      expect(await queueDb.getEntryCountByQueueName('b')).to.equal(2);
+
+      await db.clear('requests');
+
+      expect(await queueDb.getEntryCountByQueueName('a')).to.equal(0);
+      expect(await queueDb.getEntryCountByQueueName('b')).to.equal(0);
+    });
+  });
+
   describe('deleteEntry', () => {
     it(`should delete an entry for the given ID`, async () => {
       const queueDb = new QueueDb();
--- a/packages/workbox-background-sync/src/Queue.ts
+++ b/packages/workbox-background-sync/src/Queue.ts
@@ -243,6 +243,16 @@ class Queue {
   }
 
   /**
+   * Returns the number of entries present in the queue.
+   * Note that expired entries (per `maxRetentionTime`) are also included in this count.
+   *
+   * @return {Promise<number>}
+   */
+  async size(): Promise<number> {
+    return await this._queueStore.size();
+  }
+
+  /**
    * Adds the entry to the QueueStore and registers for a sync event.
    *
    * @param {Object} entry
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by jeffposnick** [comment]
Thanks for submitting this!

We have a _lot_ of `workbox-background-sync` [feature requests open](https://github.com/GoogleChrome/workbox/issues?q=is%3Aissue+is%3Aopen+label%3Aworkbox-background-sync) at the moment, and I have to admit that it's one of the areas of the Workbox codebase that I am personally less familiar with. I'm going to take some time to look at this change and see whether it also address some of the other open issues.

(Regarding running the `gulp` tasks, apologies about that. I'll take a look at our new setup and figure out the Windows compatibility story.)

File: test/workbox-background-sync/sw/lib/test-QueueDb.mjs [hunk 1]
```diff
@@ -350,6 +350,26 @@ describe(`QueueDb`, () => {
...
+    it(`should return the number of entries in IDB filtered by index`, async () => {
+      const queueDb = new QueueDb();
+
+      await queueDb.addEntry(entry1);
+      await queueDb.addEntry(entry2);
+      await queueDb.addEntry(entry3);
+      await queueDb.addEntry(entry4);
+      await queueDb.addEntry(entry5);
+
+      expect(await queueDb.getAllEntriesByQueueName('a')).to.equal(3);
```
jeffposnick: I think you mean to call `getEntryCountByQueueName()` here and later on on lines 368 and 369, but instead you're  calling `getAllEntriesByQueueName()`.

jeffposnick (changes requested):
I'd like @tropicadri to get a chance to review this, as she's touched the IndexedDB code most recently, but generally speaking, this looks like it would be useful as implemented.

test/workbox-background-sync/sw/lib/test-QueueDb.mjs
StephanBijzitter: You're right, must've missed this as I wrote this on Windows and (judging the CI) it only runs a subset of the tests.
Will fix this.

tropicadri (approved):
LGTM Thank you!

tropicadri:
> I'd like @tropicadri to get a chance to review this, as she's touched the IndexedDB code most recently, but generally speaking, this looks like it would be useful as implemented.

I checked it out, I agree with you :)

jeffposnick (approved):

PR #2947, opened 2021-09-29
Use lint-staged to run prettier

### LINKED ISSUE #2944: Fix git hooks/gulp on Windows
See #2941

jeffposnick:
Fixes #2944, hopefully. It seems to work as expected when I tested on a Windows tablet using Power Shell, with `node` and `npm` manually installed.

This switches to using `lint-staged` to orchestrate that process instead of `git-format-staged`.

Also, when investigating this issue, I realized that the glob patterns in our `.prettierignore` were not configured properly, and there were a bunch of TypeScript source files that should have been formatted, but weren't. So the cosmetic fixes to all those files got picked up in this PR as well.

CC: @StephanBijzitter


tropicadri (approved):

tropicadri (commented):
TS lint is complaining of a couple of issues other than that it lgtm

jeffposnick:
Some of the linting issues were introduced by `prettier` reformatting the code, so that our comments that disable `eslint` on the following line no longer applied to the line we expected...

That also made me realize that I neglected to add back the `pre-push` hook that runs our linting, since it's really silly to push to a PR only to find out that it fails CI. I've re-added that.

I also will update our guide to contributing code in a separate PR to clearly spell out what needs to be done to get up and running locally, including on Windows.

PR #2955, opened 2021-10-08
Export more classes/types publicly

### LINKED ISSUE #1824: Make StorableRequest publicly available
**Library Affected**:
*workbox-background-sync.*

Let user create Sotrable Request with something like :  
`new workbox. backgroundSync.StorableRequest(request)`
And the developper would be responsible using it, store it and recreate it using the existing functions `toRequest`, `fromRequest`, and `toObject`.

In some complicated places, we cannot use the background-sync as it is. 
In my case, I have to implement a complex control of the request queue in order to make sure some request are not send before others.
I don't use all the background-sync element at all.
The only thing I'm really using is the `StorableRequest` which is internally defined [here](https://github.com/GoogleChrome/workbox/blob/master/packages/workbox-background-sync/models/StorableRequest.mjs).

Doing so will simply expose a new class to the developper, to more feature requested here ;)

### LINKED ISSUE #2952: Why is PrecacheEntry interface not exported?
**Library Affected**:
workbox-precaching

**Browser & Platform**:
all browsers

The "precache" function takes this type as a parameter: `entries: Array<PrecacheEntry | string>`. However the PrecacheEntry interface [doesn't seem to be exported from the workbox-precaching package](https://github.com/GoogleChrome/workbox/blob/v6/packages/workbox-precaching/src/index.ts).

When I try to run

`import { PrecacheEntry } from "workbox-precaching";`

It tells me:

> Module '"workbox-precaching"' has no exported member 'PrecacheEntry'.

jeffposnick:
R: @tropicadri

Fixes #1824, fixes #2952

The additional exports of TypeScript types that might be useful to some developers via the top-level `index.ts` is hopefully not controversial.

There are also two previously private, internal classes in the `workbox-background-sync` module that this PR makes public. Based on the initial issue request, there's a use for developers accessing these directly, but I've warned in the JSDoc that most developers shouldn't. You've touched the `workbox-background-sync` code most recently, @tropicadri, so I'll defer to whether you anticipate any problems with making those classes public.

### DIFF
--- a/packages/workbox-background-sync/src/QueueStore.ts
+++ b/packages/workbox-background-sync/src/QueueStore.ts
@@ -1,0 +1,14 @@
+/*
+  Copyright 2018 Google LLC
+
+  Use of this source code is governed by an MIT-style
+  license that can be found in the LICENSE file or at
+  https://opensource.org/licenses/MIT.
+*/
+
+import './_version.js';
+
+// This is a temporary workaround to expose something from ./lib/ via our
+// top-level public API.
+// TODO: In Workbox v7, move the actual code from ./lib/ to this file.
+export {QueueStore} from './lib/QueueStore';
--- a/packages/workbox-background-sync/src/QueueStore.ts
+++ b/packages/workbox-background-sync/src/QueueStore.ts
@@ -1,5 +1,5 @@
 /*
-  Copyright 2018 Google LLC
+  Copyright 2021 Google LLC
 
   Use of this source code is governed by an MIT-style
   license that can be found in the LICENSE file or at
--- a/packages/workbox-background-sync/src/StorableRequest.ts
+++ b/packages/workbox-background-sync/src/StorableRequest.ts
@@ -1,0 +1,14 @@
+/*
+  Copyright 2018 Google LLC
+
+  Use of this source code is governed by an MIT-style
+  license that can be found in the LICENSE file or at
+  https://opensource.org/licenses/MIT.
+*/
+
+import './_version.js';
+
+// This is a temporary workaround to expose something from ./lib/ via our
+// top-level public API.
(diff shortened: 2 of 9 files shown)


tropicadri (changes requested):
The test check for the exported types and right now they are not matching:

  1) [all] Window and SW packages
       should have top a level module for every export in index.mjs (and vise-versa):

      AssertionError: expected [ Array(4) ] to deeply equal [ 'BackgroundSyncPlugin', 'Queue' ]
      + expected - actual

       [
         "BackgroundSyncPlugin"
         "Queue"
      -  "QueueStore"
      -  "StorableRequest"
       ]

jeffposnick:
Hmm... I remember when we added that restriction, and I guess it makes sense. I'll create a top-level module for each of those that re-exports the code in `./lib/`. Technically the `./lib/` code isn't part of our public API, so I could probably just move the modules out of `./lib/`, but I have a feeling there might be some people who are using them at their current location, so we'll have to save that until a future major release of Workbox.

packages/workbox-background-sync/src/QueueStore.ts
```diff
@@ -0,0 +1,14 @@
+/*
+  Copyright 2018 Google LLC
```
tropicadri:
```suggestion
  Copyright 2021 Google LLC
```

packages/workbox-background-sync/src/StorableRequest.ts
```diff
@@ -0,0 +1,14 @@
+/*
+  Copyright 2018 Google LLC
```
tropicadri:
```suggestion
  Copyright 2021 Google LLC
```

tropicadri (approved):

PR #2962, opened 2021-10-18
upgrade to @surma/rollup-plugin-off-main-thread dependencies from 1.4.1 to 2.2.2 (#2927)

### LINKED ISSUE #2927: workbox-webpack-plugin has indirect dependency on vulnerable ejs 2.7.4 package - needs update

**Library Affected**:
*workbox-webpack-plugin, workbox-build*

workbox-webpack-plugin has indirect dependency on ejs 2.7.4 with Arbitrary Code Injection vulnerability caused by filename which isn't sanitized for display. The issue is fixed in ejs version 3.1.6. (https://github.com/mde/ejs/commit/abaee2be937236b1b8da9a1f55096c17dda905fd)

Are there plans to upgrade the dependencies to address this?

Indirect dependency chain:
workbox-webpack-plugin
 --> workbox-build
          --> @surma/rollup-plugin-off-main-thread
                   --> ejs

The fix should really be to update @surma/rollup-plugin-off-main-thread to use the new ejs version, and then the dependencies, above it, must be updated as well.

fuzail-ahmed:
upgrade to @surma/rollup-plugin-off-main-thread dependencies from 1.4.1 to 2.2.2 (#2927)

Fixes #2927 
updated @surma/rollup-plugin-off-main-thread to use the new ejs version 3.1.6

### DIFF
--- a/infra/testing/validator/service-worker-runtime.js
+++ b/infra/testing/validator/service-worker-runtime.js
@@ -136,7 +136,7 @@ function setupSpiesAndContextForGenerateSW() {
   const context = Object.assign(
     {
       importScripts,
-      define: (_, scripts, callback) => {
+      define: (scripts, callback) => {
         importScripts(...scripts);
         callback(workboxContext);
       },
--- a/packages/workbox-background-sync/package-lock.json
+++ b/packages/workbox-background-sync/package-lock.json
@@ -8,11 +8,6 @@
       "version": "6.1.4",
       "resolved": "https://registry.npmjs.org/idb/-/idb-6.1.4.tgz",
       "integrity": "sha512-DshI5yxIB3NYc47cPpfipYX8MSIgQPqVR+WoaGI9EDq6cnLGgGYR1fp6z8/Bq9vMS8Jq1bS3eWUgXpFO5+ypSA=="
-    },
-    "workbox-core": {
-      "version": "6.3.0",
-      "resolved": "https://registry.npmjs.org/workbox-core/-/workbox-core-6.3.0.tgz",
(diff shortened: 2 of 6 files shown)


fuzail-ahmed:
@ViPai82, @jeffposnick Please review

jeffposnick:
Thanks for the PR; I was unaware that there was already a version of OMT out that had the updated deps.

However, i think there might be some breaking changes in OMT that make this more of an involved update beyond just changing `package.json`, as our test suite is now failing:

https://github.com/GoogleChrome/workbox/pull/2962/checks?check_run_id=3928423296#step:7:331

Let me know if you feel comfortable investigating that further; otherwise, I'm happy to take this over.

jeffposnick:
I'd like to get this merged so that we can include it in our upcoming v6.4.0 release. I'm going to assume you won't mind if I force-push a fix to your branch.

jeffposnick:
The PR is now failing on Node v10 (which we still support in Workbox v6) because of the use of `matchAll()` in `@surma/rollup-plugin-off-main-thread dependencies`.

[That module](https://github.com/surma/rollup-plugin-off-main-thread/blob/master/package.json) doesn't advertise a minimum required version of Node, and I'm not sure whether this was an intentional breaking change or not. I'll see if @surma is open to swapping out that method for something compatible with Node v10.

jeffposnick:
See https://github.com/surma/rollup-plugin-off-main-thread/issues/49

jeffposnick (approved):
Okay, the dependencies are in good shape now. Thanks!

PR #3007, opened 2021-12-24
Revert "JSDoc link cleanup"

swansoffie001:
Reverts GoogleChrome/workbox#2999


jeffposnick:
Closing, as I'm not sure what the context is, and I don't think we want to revert this.

PR #3009, opened 2022-01-05
refactor(build): flatten output of `workbox-build`

MichaelSolati:
**Prior to creating a pull request, please follow all the steps in the [contributing guide](https://github.com/GoogleChrome/workbox/blob/v6/CONTRIBUTING.md).**

R: @jeffposnick @tropicadri

Fixes #_issue number_

_Description of what's changed/fixed._

In the process of adding `workbox-build` doc generation to developer.chrome.com GoogleChrome/developer.chrome.com#1903 I realized that the transpiled code for the `workbox-build` library are put into a `build` directory instead of being top level. This is unlike a slew of the other libraries and causes the workbox docs on DCC to look like this...

image

The `workbox-build` library is referenced as `workbox-build/build`. This PR flattens the output of the `workbox-build` build to match the other libraries and make it play nice with the doc generation for DCC.


jeffposnick:
Thanks, @MichaelSolati!

There's unfortunately still some places in the codebase that still assume the `/build/` path (as evidenced by this [CI failure](https://github.com/GoogleChrome/workbox/runs/4731479671?check_suite_focus=true#step:7:327)).

In the back of my head, I'm worried that there are developers out there who rely on specific pieces of `workbox-build` that are not technically part of the exported public interface, but which are accessible via direct package paths. This is a gray area when it comes to semver, but our general philosophy has been to defer these sorts of changes until the next major semver release comes around (unless there is an overwhelming argument in the other direction).

Can we sync up to see whether we could address this via changes to how d.c.c. ingests the TypeScript definitions?

PR #3012, opened 2022-01-08
removing dynamic method calling to prevent minification error

### LINKED ISSUE #3006: Dynamic method calls in Queue.ts could fail with minification
There are some dynamic method calls in `Queue.ts` along the lines of:

https://github.com/GoogleChrome/workbox/blob/48f53f790ce7f0aa1a8431bdc605d589b648e329/packages/workbox-background-sync/src/Queue.ts#L279

(and a few more similar ones).

These method calls can fail if minification changes the names of the methods on the underlying `QueueStore` object.

I think it might be possible (via `declare` in TypeScript?) to give hints to various minifiers that they shouldn't do renaming, but honestly it might be easier just to use `if()`/`else()` conditionals to find the right method name to call, since there are usually only a couple to choose from for each code path.

tropicadri:
Fixes #3006 

Because I'm not sure how to reliably test the `declare` approach, I went with calling the methods explicitly.


jeffposnick (approved):
Thanks for taking care of this!

PR #3013, opened 2022-01-10
Improved TSDocs

### LINKED ISSUE #2436: Better discoverability for generateSW's runtime caching documentation
**Library Affected**:
*workbox-build*

We don't have any prominently featured "real world" examples of using `generateSW` with `runtimeCaching` to produce a service worker that is focused on runtime caching.

The documentation for the feature, at https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-build#.RuntimeCachingEntry, is rather buried.

This is related to #2423, in that it has to do with our runtime caching documentation being harder to find vs. our precaching documentation.

CC: @slightlyoff

### LINKED ISSUE #3011: TSDoc cleanup for workbox-build
There are a few things we need to do to https://github.com/GoogleChrome/workbox/blob/v6/packages/workbox-build/src/types.ts to clean up the TypeScript doc so that it could be properly rendered at developer.chrome.com

- Add the [parameter-level inline documentation](https://github.com/GoogleChrome/workbox/blob/48f53f790ce7f0aa1a8431bdc605d589b648e329/packages/workbox-build/src/generate-sw.ts#L18-L31) that existed in the JSDoc to the TSDoc.
- Ensure that all subtypes that are referenced via the top-level public composite types are `export`ed, or else the subtype docs won't be generated.

jeffposnick:
R: @tropicadri
CC: @MichaelSolati 

Fixes #3011, fixes #2436

### DIFF
--- a/packages/workbox-build/src/types.ts
+++ b/packages/workbox-build/src/types.ts
@@ -15,7 +15,7 @@ export interface ManifestEntry {
   url: string;
 }
 
-type StrategyName =
+export type StrategyName =
   | 'CacheFirst'
   | 'CacheOnly'
   | 'NetworkFirst'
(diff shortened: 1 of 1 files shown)


jeffposnick:
(Merging as this is docs-only.)

PR #3037, opened 2022-02-25
Don't use unknown for workbox-build method config

jeffposnick:
Prior to this, the `getManifest()`, `generateSW()`, and `injectManifest()` methods of `workbox-build` were marked as taking in a configuration object with type `unknown`. The idea was that prior to validating the configuration (which happens immediately), we should treat the object as potentially anything.

Unfortunately, this leads to a bad experience in our [TSDocs](https://developer.chrome.com/docs/workbox/reference/workbox-build/#method-generateSW), and likely for users who rely on TypeScript code completion.

Additionally, if you actually attempt to pass in anything other than the support options, the validation step should immediately throw, so passing in just anything didn't actually "work." Because of this, I consider adding in explicit types for the config to be a bug fix rather than a breaking change, as we never should have had `unknown` to begin with.

### DIFF
--- a/packages/workbox-build/src/generate-sw.ts
+++ b/packages/workbox-build/src/generate-sw.ts
@@ -8,7 +8,7 @@
 
 import upath from 'upath';
 
-import {BuildResult, GetManifestOptions} from './types';
+import {BuildResult, GetManifestOptions, GenerateSWOptions} from './types';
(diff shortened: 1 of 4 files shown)


tropicadri (approved):
Yay, I like this change.

PR #3046, opened 2022-03-21
check bodyUsed property for post request

### LINKED ISSUE #2552: [Workbox Google Analytics] [Offline] No payload sent to GTM when network resumed.
**Library Affected**:
*workbox-google-analytics*

**Browser & Platform**:
*Google Chrome Version 83.0.4103.116 (Official Build) (64-bit).*

**Issue Description**:
*I used GTM in my site with heavily using `dataLayer` to sent data to GTM and GA.*

The problem is GTM request is using `POST` method with passing data via long query parameters append in the url, no body used.

currently,
when offline, workbox stores the un-reachable GTM requests to indexedDB and used  them later when network resumed. after network resumed, Workbox retry all the requests but without body sent.

here is my debugging console with showing values of variables in the scope of `onSync` method from google-analytics in compiled version (v5.1.2) (https://github.com/GoogleChrome/workbox/blob/v6/packages/workbox-google-analytics/src/initialize.ts#L49)

Screen Shot 2563-06-25 at 13 43 24

As you can see ( I debugged in generated version), I did print `t` in console, which is params. It was only return `qt` (queue time) which should return many parameters according to the request context in the right side panel ( `s.url` variable )

My purpose solution is at this line (https://github.com/GoogleChrome/workbox/blob/v6/packages/workbox-google-analytics/src/initialize.ts#L59) 
we cloud do check `bodyUsed` also.

jamesyoon11:
R: @philipwalton @jeffposnick @tropicadri

Fixes #2552

 
It seems `analytics.js` used to send GET request so it didn't have any problem. However, it has been changed to POST request only. As a result, it sends an empty payload.
 
Solution: Check the bodyUsed property and if it is false, make it to use `url.searchParams`.
 
Screen_Shot_2022-03-21_at_12_00_41_PM
 
Screen_Shot_2022-03-21_at_12_01_14_PM

### DIFF
--- a/packages/workbox-google-analytics/src/initialize.ts
+++ b/packages/workbox-google-analytics/src/initialize.ts
@@ -54,7 +54,7 @@ const createOnSyncCallback = (config: GoogleAnalyticsInitializeOptions) => {
 
       try {
         // Measurement protocol requests can set their payload parameters in
-        // either the URL query string (for GET requests) or the POST body if bodyUsed is true.
+        // either the URL query string (for GET requests) or the POST body if bodyUsed value is true.
         const params =
           request.method === 'POST' && request.bodyUsed
             ? new URLSearchParams(await request.clone().text())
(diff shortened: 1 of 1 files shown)


kayal28 (changes requested):
packages/workbox-google-analytics/src/initialize.ts

tomayac:
Hi there,

The Workbox team

PR #3056, opened 2022-04-04
Fix extending webpack plugin

### LINKED ISSUE #3039: Make `config` property of `InjectManifest` class `protected` instead of `private``
**Library Affected**:
`workbox-webpack-plugin`

So, I am making a plugin that allows me to use workbox with angular, and I use typescript. Even my webpack config uses typescript.
As part of this effort, I use a class that inherits from `InjectManifest`. 
I need to add `index.html` and `manifest.webmanifest` to `additionalManifestEntries`. This addition needs to be done not just during the initial webpack config setup. Rather, it must be done every time compilation runs (which is every time I make a change and save), so that I can add the compilation hash as the revision number for the two entries. This lets the existing service worker know that some file has changed somewhere, without having to add any lazy-loaded modules to the manifest entries.

For this, I need access to the `config` property of the `InjectManifest` class which I inherit from, but it is marked as `private`, so I cannot access it.
As illustrated by my example above, this is poor for extensibility. If possible, I recommend making it `protected` instead of `private`. That way, it remains private. But those who want to extend the behavior of the plugin can do so easily.

Currently, the workaround for this is to go into the `node_modules` folder and manually edit the `.d.ts`. This has enabled me to successfully finish my plugin, but I will have to keep returning to this whenever a new version of workbox releases.

DibyodyutiMondal:
R: @jeffposnick @tropicadri

Fixes #3039 

Made the `config` property of the webpack plugins `protected`.
This will allow child classes to access and modify the config property if necessary.

PS:
Ran into an issue while running the commands mentioned in `contrbuting.md`. The commands ran fine and modified a bunch of files (mostly `package.json`s and `content not modified`s), but when I tried to commit these extra changes, it gave me an error about husky. After that, the changes disappeared as well. I am not familiar with husky so I could not figure out what to do even after reading the docs. Re-running the commands did not make the changes come back either. So I made the changes on github.com in the browser itself and fetched the commits back. Given the nature of the changes I made, I assume that it does not really matter? Still worth mentioning, just in case.

### DIFF
--- a/packages/workbox-webpack-plugin/src/generate-sw.ts
+++ b/packages/workbox-webpack-plugin/src/generate-sw.ts
@@ -69,7 +69,7 @@ export interface GenerateSWConfig extends WebpackGenerateSWOptions {
  * @memberof module:workbox-webpack-plugin
  */
 class GenerateSW {
-  private config: GenerateSWConfig;
+  protected config: GenerateSWConfig;
   private alreadyCalled: boolean;
 
   /**
--- a/packages/workbox-webpack-plugin/src/inject-manifest.ts
+++ b/packages/workbox-webpack-plugin/src/inject-manifest.ts
@@ -54,7 +54,7 @@ const {RawSource} = webpack.sources || require('webpack-sources');
  * @memberof module:workbox-webpack-plugin
  */
 class InjectManifest {
-  private config: WebpackInjectManifestOptions;
+  protected config: WebpackInjectManifestOptions;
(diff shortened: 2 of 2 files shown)


DibyodyutiMondal:
@jeffposnick any update on this?

jeffposnick:
Sorry for the delay, and thanks for the contribution!

jeffposnick (approved):

Issue #3077, opened 2022-05-05
Service worker responds slowly (10 seconds) when many query parameters are present


westonruter:
**Library Affected**:
workbox-strategies

**Browser & Platform**:
Google Chrome v101.0.4951.41

**Issue or Feature Request Description**:

On a site using the PWA plugin for WordPress, which includes Workbox, the service worker is taking 10 seconds for the NetworkFirst strategy to respond to requests for URLs that have many query parameters.

Steps to reproduce:

1. Go to https://www.sensorsone.com/ and let the service worker install
2. Try reloading the page and see it only takes a couple seconds.
3. Now try going to the [same URL with many query parameters](https://www.sensorsone.com/?ab=4&v0=56&u0=degrees%20Celsius%20%28%C2%B0C%29&f0=C&v1=60&u1=degrees%20Celsius%20%28%C2%B0C%29&f1=C&v2=0.0005&u2=per%20degree%20Celsius%20%28%2F%C2%B0C%29&f2=CdT&v3=100&u3=centimetres%20%28cm%29&f3=cm&u4=centimetres%20%28cm%29&f4=cm&v5=100&u5=sq%20centimetres%20%28cm%C2%B2%29&f5=sqcm&v6=100.4&u6=sq%20centimetres%20%28cm%C2%B2%29&f6=sqcm&v7=100&u7=cu%20centimetres%20%28cc%2C%20cm%C2%B3%29&f7=cucm&v8=100.6&u8=cu%20centimetres%20%28cc%2C%20cm%C2%B3%29&f8=cucm), and notice that the response from the service worker takes over 10 seconds.
4. Now in the Application tab of DevTools, enable "Bypass for Network" and reload the same URL with many query parameters. Now the response is returned relatively quickly as in the first step.

Originally [reported](https://wordpress.org/support/topic/long-url-with-many-query-parameters-slowing-loading/) on the WordPress plugin support forum

westonruter:
I just saw that the user was running PWA plugin v0.6 which included Workbox v5. I'm asking them to update to v0.7 which includes the current Workbox version to see if that fixes the issue.

westonruter:
Actually, I can reproduce the issue on my own site with the latest versions installed. 

* Responds quickly after 115ms: https://weston.ruter.net/
[...]

PR #3078, opened 2022-05-09
Tweak TSDocs for navigation RegExps and modifyURLPrefix

jeffposnick:
R: @tropicadri @malchata 

This adds some more documentation that will hopefully help folks avoid the situation that @westonruter found himself in with #3077

It also adds an inline example for the `modifyURLPrefix` build configuration option, because I realize that what ends up rendered in the final HTML docs is not particularly useful.


malchata (approved):
Definitely not a super duper expert on this specific thing, but it reads well to me. So, LGTM!

PR #3101, opened 2022-07-25
[WIP] Playwright integration tests

jeffposnick:
This is a placeholder PR to document some of the WIP code for migrating our integration test suite off of `selenium-assistant` and onto Playwright.

Much of this is based on guidance from https://playwright.dev/, so that's the best place to start familiarizing yourself.

I'll include some inline comments with more context when needed.

This PR includes a rough translation of the integration tests for the following modules:
- `workbox-broadcast-update`
- `workbox-cacheable-response`
- `workbox-core`
- `workbox-expiration`
- `workbox-navigation-preload`
- `workbox-routing`

The tests in this PR pass for everything except for the navigation-related tests in `firefox`. I'm not sure what's up with those failures.

You can run the entire test suite against all browsers via `npm run test_playwright`. If you want to run a specific subset of tests against, e.g., a single browser, you can use `npx playwright test --project=chromium --config=playwright-tests/playwright.config.ts playwright-tests/workbox-cacheable-response`. You may also want to turn off headless mode so that you can examine the console logs when developing new tests. There's a lot more info in the [Playwright docs](https://playwright.dev/docs/running-tests).


.github/workflows/playwright.yml
```diff
@@ -0,0 +1,26 @@
+name: Playwright Tests
```
jeffposnick, line 1: I think I just copied this template for GH Actions integration from the Playwright starter guide. It will probably need to be customized.

infra/testing/server/index.js
```diff
@@ -31,6 +31,13 @@ function initApp() {
   // Exposed the `.body` property on requests for application/json.
   app.use(bodyParser.json());
 
+  app.use((req, _, next) => {
```
jeffposnick, line 34: One of the big changes in this PR not related to Playwright itself is refactoring the server code under `infra/testing/server/` to make it easier to run multiple tests in parallel. This means doing things like supporting unique URL prefixes that each test could set so that its test state would not interfere with the test state from another test running at the same time.

playwright-tests/lib/fetchHelper.ts
```diff
@@ -0,0 +1,21 @@
+import {Frame, Page} from '@playwright/test';
```
jeffposnick, line 1: The files under `playwright-tests/lib/` are roughly equivalent to the helpers under https://github.com/GoogleChrome/workbox/tree/v6/infra/testing

playwright-tests/playwright.config.ts
```diff
@@ -0,0 +1,81 @@
+import {PlaywrightTestConfig, devices} from '@playwright/test';
```
jeffposnick, line 1: Feel free to tweak this config as needed. A lot of it is copied over from the default examples in the docs.

https://playwright.dev/docs/test-configuration

playwright-tests/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
```
jeffposnick, line 1: I forget exactly why, but we need a custom TS config for the test suite.

playwright-tests/workbox-broadcast-update/spec.ts
```diff
@@ -0,0 +1,146 @@
...
+  const apiURL = `${baseURL}/__WORKBOX/uniqueETag`;
+  const url = generateIntegrationURL(baseURL, __dirname);
+  await page.goto(url);
+  const scope = await registerAndControl(page);
+
+  const firstResponse = await fetchAsString(page, apiURL);
+
+  expect(firstResponse).toMatch(/ETag is \d+\./);
+
+  const [messageData, secondResponse] = await Promise.all([
```
jeffposnick, line 28: This is a common pattern in the tests that I've moved over that seems to have led to more reliability. It registers the `message` handler and fires off the fetch request as part of the same async sequence, and waits until both promises are resolved before moving on to the next step in the test.

playwright-tests/workbox-broadcast-update/sw.js
```diff
@@ -0,0 +1,43 @@
+/*
```
jeffposnick, line 1: Generally speaking, most of the `sw.js` files from the old integration tests can be moved over to the new test suite as-is.

The small changes that I've made are usually around setting explicit `cacheName` properties using `self.registration.scope`, which are guaranteed to be unique for each test execution, thanks to the `generateIntegrationURL()` helper funciton.

That way, each cache ends up being assigned a unique name and that helps avoid conflicts when running multiple tests in parallel.

playwright-tests/workbox-broadcast-update/spec.ts
Kazy0004: ماذا يوجد لدي هنا على الموقع

Kazy0004:
لااعلم ماذا يحدث معي اريد منك المساعده ياعزيزي

tomayac:
Hi there,

The Workbox team

PR #3128, opened 2022-10-31
feat(build): support esbuild bundling

### LINKED ISSUE #3002: Move off of rollup and on to esbuild/swc/??? for generateSW
We currently use [Rollup](https://rollupjs.org/) in our build tools' "generate" modes to take a templated service worker source file + config, and produce either a single output service worker (if `inlineWorkbox` is `true`) or a service worker + custom Workbox runtime bundle (if `inlineWorkbox` is `false`; this mode uses [`rollup-plugin-off-main-thread`](https://github.com/surma/rollup-plugin-off-main-thread).

Rollup is a fantastic tool and we'll likely continue to use it to generate, e.g., our pre-bundled libraries that are released on the CDN. For the more basic use case within the "generate" modes, it can be overkill, and its runtime performance is generally slower than Rust- or Go-based bundlers.

We should reevaluate our options for Workbox v7.

yckbilly1929:
Fixes https://github.com/GoogleChrome/workbox/issues/3002 (Support using esbuild as bundling when generateSW)

This PR adds support for using esbuild as bundler by adding [esbuild](https://www.npmjs.com/package/esbuild) and [@esbuild-plugins/node-resolve](https://www.npmjs.com/package/@esbuild-plugins/node-resolve) to workbox-build. Feature is enabled only when esbuildOptions is passed as true or esbuild.BuildOptions to generateSW().

Differences:
- only supports inline runtime, as there is no native support to output as amd format (but magic solution exists though)
- default browser target changed from >= chrome56 to [chrome60, safari11, edge18]
- service-worker.js output size increases (from 21 kb to 23 kb in my case)


jessemoe (approved):

tomayac:
Hi there,

The Workbox team

PR #3205, opened 2023-05-05
feat(workbox-precaching) Support for concurrent prefetching

### LINKED ISSUE #2880: Option to precache multiple requests simultaneously
If I always have to wait for the full installation of the service worker before the page is displayed, why not to tell the service worker that assets can be loaded not sequentially, but as quickly as possible-in parallel?

On a slow connection, we have to wait a very long time for a sequential download, even on http2

My case (slow 3g connection): 
13 page assets are loaded in parallel in 5 seconds and then 22 assets are loaded with service worker sequentially for about a minute over http2 (13 of them are the same ones that loaded in 5 seconds for the page and the main size of all assets falls on them)

C-Hess:
Fixes #2880

Added the ability to precache requests concurrently. It looks like this used to be a feature in #2528, but was removed because it had no limits to the number of concurrent connections, resulting in a lack of resources for large sites with hundreds of assets.

This feature is _opt-in_, and should be non-breaking. Though it looks like this ticket is part of the V7 milestone. It may be possible to just bump a minor version for this.

Let me know if there's anything I missed that you'd like to see. One thing that might be interesting in the future is a dynamic concurrent request feature, where the lack of HTTP/2+ support or browser resource errors results in the throttling of the batch sizes.


packages/workbox-precaching/src/precacheAndRoute.ts
```diff
@@ -27,10 +28,11 @@ import './_version.js';
  */
 function precacheAndRoute(
   entries: Array<PrecacheEntry | string>,
-  options?: PrecacheRouteOptions,
+  routeOptions?: PrecacheRouteOptions,
+  controllerOptions?: PrecacheControllerOptions,
```
C-Hess, line 32: Not a big fan of the two option arguments. Combining them into a single options argument in a non-breaking manner didn't sit well with me either.

dohomi:
any chance that this open PR gets more attention? Its still an issue on larger SPAs where the reload takes minutes of time until reloading is possible.

dohomi:
@westonruter I would love to hear if there is any chance to get this released otherwise because it is still a major issue for users waiting for the files being pre-cached

packages/workbox-precaching/src/precacheAndRoute.ts
westonruter: I'm curious why the additional arguments couldn't be just added to `PrecacheRouteOptions`? Or rather, could just `concurrentRequests` be added to `PrecacheRouteOptions`? Then `getOrCreatePrecacheController` could be updated to just be passed the one object with `concurrentRequests` argument extracted from the options. This would avoid the need for a second options object and it wouldn't have any back-compat breakage as far as I can tell.

westonruter:
@dohomi Sorry, so far I've been focused on getting outdated dependencies updated. I haven't had time to work on new features.

dohomi:
No problem I understand you guys have other priorities. I migrated to https://serwist.pages.dev/ just in case other people also struggle with long delays of precaching. Thanks for the heads up

PR #3215, opened 2023-05-26
Dependency updates to remove critical warnings and fix release process

tropicadri:
Dependency updates to remove critical warnings and fix release process


malchata (approved):

rulgf:
Hi
Question, why is removing support for node <= 16 considered a patch release?

Thanks!

dpobel:
if I understand correctly, node 16 was defined as the minimum required version in 6.6.0 in a "soft" way and then the 6.6.1 enforced it in packages.json. This is breaking the build of our application :(

IMHO this is a BC break should have happened in a 7.0.0

rulgf:
I agree, this should not be considered a minor/patch release.

@tropicadri could it be possible to move this to 7.0.0 and revert 6.6.1 ?

PR #3234, opened 2023-07-28
fix(workbox-build): make source map filename consistent with injected source file

### LINKED ISSUE #3233: Source map support for injectManifest makes incorrect assumptions
**Library Affected**:
workbox-build

**Browser & Platform**:
n/a

It looks like source map support was added for the Webpack plugin in #2239 when injecting the manifest.  However, there is no such support for the `injectManifest` method in the _workbox-build_ package.  It would be helpful if the same code could be applied there, i.e. if a source map exists for `swSrc` then write an adjusted one for `swDest`.

steverep:
Fixes #3233 per comments on #2239 

- Name the output source map simply `swDest` with `.map` appended
- Replace the `sourceMappingURL` comment with the new filename
- Adjust tests to use different filenames for input and output
- Add simple test to check URL replacement


steverep:
@jeffposnick would it be possible to get your review here?  I just don't want to build a pipeline using this patch if you feel it's a no go.

jeffposnick:
CC: @tropicadri and @petele for guidance.

steverep:
At this point I'm wondering if `workbox` is even being maintained anymore?  It hasn't been touched in 6 months even for just responding to issues and dependency bumps.

steverep:
@tomayac any chance this could get looked at soon? It's not exactly a very taxing review.

tomayac:
> @tomayac any chance this could get looked at soon? It's not exactly a very taxing review.

At first sight this LGTM, but the tests fail. Could you look if it's related to this change? Thanks!

steverep:
> At first sight this LGTM, but the tests fail. Could you look if it's related to this change? Thanks!

I know they passed locally when I first created the PR but I'll take a quick look.

steverep:
The same 8 tests fail on the v7 branch, so they have nothing to do with these changes.

tomayac:
> The same 8 tests fail on the v7 branch, so they have nothing to do with these changes.

Thanks for checking! @tropicadri, over to you for a final look. I LGTM.

tomayac (approved):
LGTM, approval.

Tests were manually verified to fail for unrelated reasons.

steverep:
@tropicadri are you able to review and release this change?  I'd like to avoid creating a patch if I can.

Issue #3180, opened 2023-03-01
GenerateSW has been called multiple times, perhaps due to running webpack in --watch mode.


Janani-BlueOcean:
Welcome! Please use this template for reporting bugs or requesting features. For questions about using Workbox, the best place to ask is Stack Overflow, tagged with `[workbox]`: https://stackoverflow.com/questions/ask?tags=workbox

**Library Affected**:
_workbox-sw, workbox-build, etc._

**Browser & Platform**:
_E.g. Google Chrome v51.0.1 for Android, or "all browsers"._

**Issue or Feature Request Description**:
_Your request here._

_When reporting bugs, please include relevant JavaScript Console logs and links to public URLs at which the issue could be reproduced._

prprabhu-ms:
This is a known issue that only affects local devserver runs.

See https://github.com/GoogleChrome/workbox/issues/1790 for prior context.

jon-ressio:
Coming from a nextjs app, it would be really nice if this warning only got written one time per call to `next dev`.  For me I get it every time a set of server components are compiled.  It ends up being a significant console spam.

Any thoughts on something like below?  If I understand the code and how webpack is working, it would mean I'd see 1 warning for inject-manifest and another for generate-sw, but then nothing for subsequent calls to each?

I'll spin up a quick PR so we can discuss there...

![image](https://github.com/GoogleChrome/workbox/assets/114938161/c1678c59-4d34-4b90-bbb3-228890005320)

tomayac:
Hi there,

Workbox is [moving to a new engineering team within Google](https://github.com/GoogleChrome/workbox?tab=readme-ov-file#maintenance-update:~:text=Maintenance%20update,owners%20of%20Workbox.). As part of this move, we're declaring a partial bug bankruptcy to allow the new team to start fresh. We realize this isn't optimal, but realistically, this is the only way we see it working. For transparency, here're the criteria we applied:

- For PRs, we closed [everything that doesn't target the current `v7` branch](https://github.com/GoogleChrome/workbox/pulls?q=is%3Apr+is%3Aclosed+-base%3Av7).
- For Issues, we closed [everything that was created before 2024](https://github.com/GoogleChrome/workbox/issues?q=is%3Aissue+is%3Aclosed+created%3A%3C2024-01-01+). 

Thanks, and we hope for your understanding!
The Workbox team

PR #3246, opened 2023-09-14
Single warning per `webpack --watch`

jon-ressio:
- Only warn one time per execution of `webpack --watch`

Improves #1790, #3180 

I'd like to only see this warning printed one time per execution of `next dev`.  Right now I see it every time a different bundle/batch of server components is compiled or when any code is recompiled and a page reloaded, it's very noisy.

I just built and tested this PR locally and I see the GenerateSW warning 4 times (I think `next dev` manages multiple `webpack --watch` processes behind the scenes for the different server and client packages, etc...) but after that initial warning it was quiet.

If maintainers don't want to accept this PR as default behavior, I can also change this PR to check for an environment variable like `__WB_ONLY_WARN_ONCE` or something like that.  Just let me know

### DIFF
--- a/packages/workbox-webpack-plugin/src/generate-sw.ts
+++ b/packages/workbox-webpack-plugin/src/generate-sw.ts
@@ -71,6 +71,7 @@ export interface GenerateSWConfig extends WebpackGenerateSWOptions {
 class GenerateSW {
   protected config: GenerateSWConfig;
   private alreadyCalled: boolean;
+  private alreadyWarned: boolean;
 
   /**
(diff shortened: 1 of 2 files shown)


jon-ressio:
Just a note that this was an issue with the original next-pwa package which seems to be unmaintained for a while.  https://github.com/DuCanhGH/next-pwa doesn't cause the same log spam.

So I no longer need this PR merged but I'll leave it in case others think it makes sense.  Feel free to accept or reject

jessemoe (approved):
LGTM

PR #3172, opened 2023-02-01
StrategyHandler.doneWaiting unhandled rejection

joshkel:
Fixes #3171

`StrategyHandler.doneWaiting` throws on the first rejected promise that it encounters.  This means that subsequent rejected promises may result in unhandled rejection errors.

Promise.allSettled is a natural solution. It [does not appear to be supported](https://caniuse.com/mdn-javascript_builtins_promise_allsettled) on all browsers that workbox uses. I tried installing [promise.allsettled](https://www.npmjs.com/package/promise.allsettled) from NPM but had trouble getting the default import from that module to work correctly with Rollup, and I noticed that the workbox service worker libraries tend to avoid external dependencies in general, so I created a local version instead. I put it under workbox-core's `_private` directory, following the example of other utility code such as Deferred. If I should handle this differently, please let me know.

PR #3320, opened 2024-04-27
Strategy unhandled rejection

### LINKED ISSUE #3171: StrategyHandler.doneWaiting may result in unhandled promise rejections
**Library Affected**:
workbox-strategies

**Browser & Platform**:
all browsers

While investigating some unhandled rejection errors in our application, I found that `StrategyHandler.doneWaiting` throws on the first rejected promise that it encounters. This means that subsequent rejected promises may result in unhandled rejection errors.

joshkel:
Fixes https://github.com/GoogleChrome/workbox/issues/3171

StrategyHandler.doneWaiting throws on the first rejected promise that it encounters. This means that subsequent rejected promises may result in unhandled rejection errors.

Promise.allSettled is a natural solution. It [does not appear to be supported](https://caniuse.com/mdn-javascript_builtins_promise_allsettled) on all browsers that workbox uses. (However, I'm having trouble finding a clear statement of what browsers workbox supports, so I could be wrong.) I tried installing [promise.allsettled](https://www.npmjs.com/package/promise.allsettled) from NPM but had trouble getting the default import from that module to work correctly with Rollup, and I noticed that the workbox service worker libraries tend to avoid external dependencies in general, so I created a local version instead. I put it under workbox-core's _private directory, following the example of other utility code such as Deferred. If I should handle this differently, please let me know.

This is a rebase of #3172 against the v7 branch.


tomayac:
Fully approve of the intent of the PR. However, I don't think `Promise.allSettled()` needs polyfilling looking at [browser support](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled#browser_compatibility). Looking at devices that didn't make the Safari 13 cut, this are iPhone 5S, iPhone 6 and 6 Plus, the original iPad Air, the iPad mini 2 and iPad mini 3, and the 6th-generation iPod Touch. For Chromebooks, we'd be locking out Chromebooks stuck on Chrome <76. @tropicadri, what do you think?

packages/workbox-strategies/src/StrategyHandler.ts
```diff
@@ -556,13 +560,17 @@ class StrategyHandler {
...
+   * `waitUntil()` method), otherwise the service worker thread may be killed
    * prior to your work completing.
    */
   async doneWaiting(): Promise<void> {
-    let promise;
-    while ((promise = this._extendLifetimePromises.shift())) {
-      await promise;
+    while (this._extendLifetimePromises.length) {
+      const promises = this._extendLifetimePromises.splice(0);
+      const result = await allSettled(promises);
```
westonruter: Should this reuse `Promise.allSettled()` if it is available?

```suggestion
      const result = Promise.allSettled ? await Promise.allSettled(promises) : await allSettled(promises);
```

packages/workbox-core/src/_private/allSettled.ts
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Promise.allSettled polyfill based on
+ *
+ * https://github.com/es-shims/Promise.allSettled/blob/main/implementation.js
+ *
+ * which is (c) 2019 Jordan Harband and used under the terms of the MIT license.
+ *
+ * @private
+ */
+function allSettled<T>(
```
westonruter: Alternatively to my suggestion below, this could do something like:

```suggestion
const allSettled = Promise.allSettled ? Promise.allSettled.bind( Promise ) : function allSettledImpl<T>(
```

westonruter:
> Fully approve of the intent of the PR. However, I don't think `Promise.allSettled()` needs polyfilling looking at [browser support](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled#browser_compatibility).

I'll note that `Promise.allSettled()` is now **Baseline Widely available**. So yeah, maybe no polyfill is needed?

joshkel:
Thanks for the feedback, @westonruter (and @tomayac). Since `Promise.allSettled` is widely available, it seems like no polyfill is needed. However, making TypeScript happy with `Promise.allSettled` will require changing [tsconfig.json's](https://github.com/GoogleChrome/workbox/blob/v7/tsconfig.json) `lib` from `es2017` to `es2020`. Is it okay to do that as part of this PR?

westonruter:
@joshkel Yes, that makes sense to me, since `Promise.allSettled()` was introduced in `es2020`.

joshkel:
@westonruter I've switched to the native version. Thanks again.

packages/workbox-strategies/src/StrategyHandler.ts
```diff
@@ -556,13 +556,17 @@ class StrategyHandler {
...
    * passed to an event's `waitUntil()` method (not this handler's
-   * `waitUntil()` method), otherwise the service worker thread my be killed
+   * `waitUntil()` method), otherwise the service worker thread may be killed
    * prior to your work completing.
    */
   async doneWaiting(): Promise<void> {
-    let promise;
-    while ((promise = this._extendLifetimePromises.shift())) {
-      await promise;
+    while (this._extendLifetimePromises.length) {
```
westonruter, line 563: I suppose the `while` loop needs to be retained in case additional promises are added to the array while awaiting the ones that were removed during the first iteration?

joshkel: Yes. (That was my rationale, at least.)

westonruter (approved):

westonruter:
On a related note, I just recalled that the [Browser Support](https://github.com/GoogleChrome/workbox/wiki/Browser-Support) page indicates that only the current version minus 2 are to be supported. This is now clearly the case for `Promise.allSettled()`!

image

PR #3387, opened 2025-01-09
Update dependencies

westonruter:
* Add `.nvmrc` populated with `18` to correspond with the oldest version of node currently in LTS maintenance. 
* Bump `engines.node` from `>=16.0.0` to `>=18.0.0`.
* Run `npm audit fix`.

PR #3392, opened 2025-02-04
Update dependencies in nested packages

westonruter:
This is a follow-up to #3387 which did the same for the main `package.json`, except this time it is done for each of the `packages/workbox-*` directories:

* Update the old `package-lock.json`. 
* Run `npm audit fix`.


adamsilverstein (approved):
Great

PR #3394, opened 2025-02-05
Update engines.node in packages and node-version in pull-request workflow

westonruter:
This is a follow-up to https://github.com/GoogleChrome/workbox/pull/3387 and https://github.com/GoogleChrome/workbox/pull/3392. I neglected to update `engines.node` in the latter as I had done in the former. This version of node corresponds with the oldest version of node currently in LTS maintenance. 

This also updates the `node-version` in the `pull-request` workflow.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -11,7 +11,7 @@ jobs:
 
       - uses: actions/setup-node@v1
         with:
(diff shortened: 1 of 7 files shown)


adamsilverstein (commented):
Nice!

adamsilverstein (approved):

PR #3402, opened 2025-03-04
Add composite to compilerOptions in the TypeScript config for packages that reference other packages

### LINKED ISSUE #3401: tsc: Referenced project must have setting "composite": true.
**Library Affected**:
All workbox libraries referencing other workbox libraries

**Typescript Version**:
5.7.3

I use workbox-precaching for my project. However I've got these messages from VS Code:

> Referenced project '/path/to/node_modules/workbox-core' must have setting "composite": true.

> Referenced project '/path/to/node_modules/workbox-routing' must have setting "composite": true.

> Referenced project '/path/to/node_modules/workbox-strategies' must have setting "composite": true.

Image

According to the [TypeScript Handbook](https://www.typescriptlang.org/docs/handbook/project-references.html#composite), `referenced projects must have the new composite setting enabled`.

It seems the libraries below referenced by other libraries:

1. workbox-background-sync
1. workbox-broadcast-update
1. workbox-build
1. workbox-cacheable-response
1. workbox-core
1. workbox-expiration
1. workbox-google-analytics
1. workbox-routing
1. workbox-strategies

Adding `"composite": true` to each `tsconfig.json` would fix these warnings.

westonruter:
Fixes #3401

See https://www.typescriptlang.org/docs/handbook/project-references.html#composite

### DIFF
--- a/packages/workbox-background-sync/tsconfig.json
+++ b/packages/workbox-background-sync/tsconfig.json
@@ -1,6 +1,7 @@
 {
   "extends": "../../tsconfig",
   "compilerOptions": {
+    "composite": true,
     "outDir": "./",
     "rootDir": "./src",
     "tsBuildInfoFile": "./tsconfig.tsbuildinfo"
--- a/packages/workbox-broadcast-update/tsconfig.json
+++ b/packages/workbox-broadcast-update/tsconfig.json
@@ -1,6 +1,7 @@
 {
   "extends": "../../tsconfig",
   "compilerOptions": {
+    "composite": true,
     "outDir": "./",
     "rootDir": "./src",
     "tsBuildInfoFile": "./tsconfig.tsbuildinfo"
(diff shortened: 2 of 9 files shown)


westonruter:
@swissspidy Are you familiar with the `composite` compiler option?

swissspidy:
Not familiar per se, but in the Web Stories repo with a similar structure it is enabled too. But I'm surprised that here it needs to be applied to every single package 🤔  They all extend the base TSConfig (`"extends": "../../tsconfig",`), which already sets it:

https://github.com/GoogleChrome/workbox/blob/77dc1bd4fce7836fa47c9085aa0cf522bee5ae3d/tsconfig.json#L3

So I thought it would already be enabled 🤷 

The main difference in the Web Stories repo is that there are two files in the root: `tsconfig.json` (which just references all the individual packages) and `tsconfig.shared.json` (holds `compilerOptions` etc. and is extended by every single package).

But if the reporter says this fixes the issue then I suppose this way works too.

swissspidy (approved):

PR #3440, opened 2025-11-19
Update GitHub Actions to latest versions

swissspidy:
Fixes the broken Scorecards workflow which currently fails with this:

> This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v3.1.2`. Learn more: https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/


tomayac (approved):
LGTM, thank you

PR #3483, opened 2026-03-27
Fix/precache cleanupresult compat

### LINKED ISSUE #3400: Bug: Incorrect type definition of CleanupResult
According to the type definition, CleanupResult should have a property “deletedCacheRequests”. The PrecacheController returns a CleanupResult in the activate function, but instead of the “deletedCacheRequests” property you get “deletedURLs”

Type Definition CleanupResult:
https://github.com/GoogleChrome/workbox/blob/v7/packages/workbox-precaching/src/_types.ts#L17

Implementation of activate Function:
https://github.com/GoogleChrome/workbox/blob/v7/packages/workbox-precaching/src/PrecacheController.ts#L267

Ujjwaljain16:
Fixes #3400

### Summary

This PR resolves a mismatch between the declared `CleanupResult` type and the actual runtime behavior of `PrecacheController.activate()`.

Currently:

* The type defines `deletedCacheRequests`
* The implementation returns `deletedURLs`
* Tests and runtime behavior already rely on `deletedURLs`

This leads to an inconsistent public API where TypeScript users are guided to use a property that does not exist at runtime.

---

### Changes

* **Align runtime and types**

  * `deletedURLs` is now treated as the canonical field
* **Add backward compatibility**

  * `deletedCacheRequests` is returned as an alias pointing to the same array
* **Update type definition**

  * Include both fields in `CleanupResult`
  * Mark `deletedCacheRequests` as deprecated
* **Update JSDoc**

  * Reflect `deletedURLs` as the primary field
  * Document deprecation of `deletedCacheRequests`
* **Add test coverage**

  * Verify both `deletedURLs` and `deletedCacheRequests` return consistent values

---

### Rationale

* Runtime behavior and existing tests already establish `deletedURLs` as the de facto API
* Renaming runtime to match types would risk breaking existing JavaScript consumers
* Removing `deletedCacheRequests` from types would break existing TypeScript users

This approach:

* aligns types with actual behavior
* avoids breaking changes
* provides a clear migration path

---

### Notes

* `deletedCacheRequests` is now deprecated and can be removed in a future major release
* `deletedURLs` is more semantically accurate (array of URL strings rather than Request objects)

---

### Testing

* Existing tests pass unchanged
* Added assertions to ensure both properties return identical values

---

### Risk

Low , this is an additive, backward-compatible change with no impact on runtime logic or precaching behavior

### DIFF
--- a/packages/workbox-precaching/src/PrecacheController.ts
+++ b/packages/workbox-precaching/src/PrecacheController.ts
@@ -264,7 +264,10 @@ class PrecacheController {
         printCleanupDetails(deletedURLs);
       }
 
-      return {deletedURLs};
+      return {
+        deletedURLs,
+        deletedCacheRequests: deletedURLs,
+      };
     });
   }
 
--- a/packages/workbox-precaching/src/_types.ts
+++ b/packages/workbox-precaching/src/_types.ts
@@ -14,7 +14,12 @@ export interface InstallResult {
 }
 
 export interface CleanupResult {
-  deletedCacheRequests: string[];
+  deletedURLs: string[];
+  /**
+   * @deprecated Use `deletedURLs` instead.
+   * TODO: Remove in the next major release.
+   */
+  deletedCacheRequests?: string[];
 }
 
 export declare interface PrecacheEntry {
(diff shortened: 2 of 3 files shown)


Ujjwaljain16:
waiting for the review

Ujjwaljain16:
@swissspidy  can i get the review?

Ujjwaljain16:
CI test failing due to version mismatch unrelated to my changes

swissspidy:
Thanks for the PR. I don't have much capacity right now, but that unrelated test failure would definitely need fixing first in another PR.

PR #3489, opened 2026-04-01
Replace lodash with eta

rtritto:
`lodash` [v4.18.0](https://github.com/lodash/lodash/releases/tag/4.18.0) introduced the issue https://github.com/lodash/lodash/issues/6167 with `lodash.template`.

E.g.:
```console
Error: Unable to generate service worker from template. 'assignWith is not defined'
    at populateSWTemplate (C:\Users\<USER>\AppData\Local\Yarn\Berry\cache\workbox-build-npm-7.4.0-c84561662c-10c0.zip\node_modules\workbox-build\build\lib\populate-sw-template.js:76:15)
    at writeSWUsingDefaultTemplate (C:\Users\<USER>\AppData\Local\Yarn\Berry\cache\workbox-build-npm-7.4.0-c84561662c-10c0.zip\node_modules\workbox-build\build\lib\write-sw-using-default-template.js:28:73)
    at generateSW (C:\Users\<USER>\AppData\Local\Yarn\Berry\cache\workbox-build-npm-7.4.0-c84561662c-10c0.zip\node_modules\workbox-build\build\generate-sw.js:95:23)
```

`lodash.template` is also deprecated (soruce [lodash.template](https://www.npmjs.com/package/lodash.template)):

> This package is deprecated. Use https://socket.dev/npm/package/eta instead.

`lodash` is used only in the `workbox-build` package (it uses `lodash.template`) and it can be replaced with `eta` (https://www.npmjs.com/package/eta).
The `@types/lodash` dependency at root `package.json` can also be removed.


rtritto:
@google-cla I signed the CLA

rtritto:
FYI @swissspidy

rtritto:
@swissspidy all test passed, please can you review?

swissspidy (approved):
Thanks for your PR

rtritto:
@swissspidy thanks!

PS: next time you merge a PR, I reccomend you to use `squash` instead of `merge`: by this way you'll only have 1 commit in the history

jayaddison:
NB: this appears to have been due to a CLI mapping that was not updated during the release of various `lodash` packages at v4.18.0 -- a fixup release v4.18.1 has recently been published that should repair those: https://github.com/lodash/lodash/releases/tag/4.18.1

jayaddison:
A question: is the `eta` package used here from the source repository at: https://github.com/bgub/eta/

packages/workbox-build/src/lib/populate-sw-template.ts
```diff
@@ -15,6 +15,11 @@ import {runtimeCachingConverter} from './runtime-caching-converter';
 import {stringifyWithoutComments} from './stringify-without-comments';
 import {swTemplate} from '../templates/sw-template';
 
+const eta = new Eta({
+  useWith: true,
```
jayaddison, line 19: NB: I think this relates to the [JavaScript `with` function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/with) -- something that can be used by both `lodash.template` and also `eta` -- but is deprecated.

rtritto: @jayaddison how should we change the code?

rtritto:
> A question: is the `eta` package used here from the source repository at: https://github.com/bgub/eta/

Yes

jayaddison:
> > A question: is the `eta` package used here from the source repository at: https://github.com/bgub/eta/
> 
> Yes

Ok, thanks @rtritto

rtritto:
@jayaddison should we remove `useWith: true` setting?

packages/workbox-build/src/lib/populate-sw-template.ts
jayaddison: @rtritto Personally I'm a bit skeptical about changing the dependency - `lodash.template` v4.18.0 may have introduced this error (shortly-after resolved in v4.18.1), but aside from that, the dependency itself has been stable for a long time and is used in many places, which to me is usually a sign that it's stable and reliable.  Code that _doesn't_ change often can be a good thing.

If there are significant benefits - e.g. much smaller dependency surface and codebase, better stability, auditing/assurance, etc, then perhaps `eta` would be worth migrating to - but I'm not aware of any significant improvements/drawbacks to back that up.

Neither `eta` nor `lodash.template` is a "logic-less templating" engine -- so I'd guess that they're relatively similar in terms of security and functionality.

However: the use of the `with` statement in JavaScript, highlighted by this `eta` migration, does make me think that if we move back to `lodash.template`, then we could explore using the `variable` option, which means that the evaluated code _does not_ use the `with` statement.

NB: there was a security bug about that a few years ago, that was subsequently resolved:

- https://github.com/advisories/GHSA-35jh-r3h4-6jhm
- https://github.com/lodash/lodash/commit/3469357cff396a26c363f8c1b5a91dde28ba4b1c

...so some caution could be worthwhile.

jayaddison:
@rtritto I don't have an opinion about that yet, I haven't read much about the functionality or source code of `eta`.

jayaddison:
Ah, also, I forgot: I'm a bit wary about this bugreport in `eta` - to me it indicates that the template parsing it does may be fragile: https://github.com/bgub/eta/issues/337

packages/workbox-build/src/lib/populate-sw-template.ts
rtritto: `lodash` is a generic library with drawbacks of performance and weight where there is an official deprecation of `lodash.template`.
So it's reasonable to replace it with a library with the specific purpose.
As officially suggested, an alternative is `eta`.
The `useWith: true` can be removed if it's not needed. Probably it should be removed `with () {} in JavaScript slows down execution and can cause confusing bugs` (see https://dev.to/bgub/i-built-a-js-template-engine-3x-faster-than-ejs-lj8).

rtritto:
> Ah, also, I forgot: I'm a bit wary about this bugreport in `eta` - to me it indicates that the template parsing it does may be fragile: [bgub/eta#337](https://github.com/bgub/eta/issues/337)

There is an opened PR

jayaddison:
@jayaddison 
> > Ah, also, I forgot: I'm a bit wary about this bugreport in `eta` - to me it indicates that the template parsing it does may be fragile: [bgub/eta#337](https://github.com/bgub/eta/issues/337)

@rtritto 
> There is an opened PR

That PR did exist when you mentioned it @rtritto -- but I can't find it when I view the `eta` pull requests any more, so I am wondering if it has been closed and/or deleted.

Although the usage of templates in `workbox` is fairly minimal, I personally feel that staying with `lodash` as the templating engine -- although with the `variable` option enabled, so that JavaScript `with` statements are not required (something that I wouldn't have learned about if not for the migration to `eta`) -- is probably safer.

That's based on my sense that slow-moving, widely-used dependencies tend to be mature, and that faster-moving ones can be risky.

@swissspidy @rtritto what do you think?

swissspidy:
Would you mind opening a new issue with your proposal for visibility? Discussions on closed PRs are easily missed.

jayaddison:
Sensible, thanks @swissspidy; I've opened a separate thread to consider migrating back to `lodash` at: #3529

PR #3513, opened 2026-05-08
refactor(workbox-build): remove useWith from Eta template configuration

rtritto:
This PR refactors the Eta template rendering configuration within `workbox-build` by removing the `useWith: true` option.

Using `with () { ... }` blocks (which `useWith` relies on) is generally considered bad practice in modern JavaScript due to performance implications, potential scoping issues, and incompatibility with Strict Mode. 

To accommodate this change, the service worker template (`sw-template.ts`) has been updated to explicitly reference injected variables using the default `it.` object property, ensuring compatibility and cleaner variable resolution during template compilation.

### Changes
* **`packages/workbox-build/src/lib/populate-sw-template.ts`**: Removed `useWith: true` from the Eta initialization options.
* **`packages/workbox-build/src/templates/sw-template.ts`**: Updated all template variable interpolations (e.g., `<%= importScripts %>`, `<%= use(...) %>`) to use the explicit `it.` prefix (e.g., `<%= it.importScripts %>`, `<%= it.use(...) %>`).

### Reference
`with () {} in JavaScript slows down execution and can cause confusing bugs` from https://dev.to/bgub/i-built-a-js-template-engine-3x-faster-than-ejs-lj8

### DIFF
--- a/packages/workbox-build/src/lib/populate-sw-template.ts
+++ b/packages/workbox-build/src/lib/populate-sw-template.ts
@@ -16,7 +16,6 @@ import {stringifyWithoutComments} from './stringify-without-comments';
 import {swTemplate} from '../templates/sw-template';
 
 const eta = new Eta({
-  useWith: true,
   autoEscape: false,
 })
 
--- a/packages/workbox-build/src/templates/sw-template.ts
+++ b/packages/workbox-build/src/templates/sw-template.ts
@@ -18,17 +18,17 @@ export const swTemplate = `/**
  * See https://goo.gl/2aRDsh
  */
 
-<% if (importScripts) { %>
+<% if (it.importScripts) { %>
 importScripts(
-  <%= importScripts.map(JSON.stringify).join(',\\n  ') %>
+  <%= it.importScripts.map(JSON.stringify).join(',\\n  ') %>
 );
 <% } %>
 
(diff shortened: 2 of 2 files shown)


rtritto:
FYI @jayaddison

jayaddison:
Thanks @rtritto.  Could we enable strict mode for the template evaluation?

rtritto:
Nice catch, I did the commit

rtritto:
Removed the commit.

> The reason `useStrict` is not a valid configuration option in modern versions of Eta is that simply removing `useWith: true` already solves the core issue.
> 
> Without `useWith: true`, Eta no longer wraps the template data in a `with(it) { ... }` block. Instead, it compiles the template into a standard, clean JavaScript function.
> 
> Since modern Node.js and TypeScript project environments (especially those using ES Modules or transpilers) run in Strict > Mode by default, the generated template function will naturally inherit this Strict Mode. This automatically provides the expected performance optimizations and security benefits.
>
>If you absolutely need to explicitly enforce the `"use strict";` directive inside the compiled function body, the supported way to do this in Eta is to add it directly at the very top of your template file `sw-template.ts`:
>```js
> <% "use strict"; %>
> /**
>  * Welcome to your Workbox-powered service worker!
>  ...
>```
> However, in most modern codebases, just removing `useWith: true` and explicitly referencing variables with it. is the correct and sufficient approach.

